PR #418, opened 2024-06-26
Add signing, entitlements, and notarization

DavisVaughan:


DavisVaughan:
Likely going to have to go a different way, so no need to keep this open

PR #426, opened 2024-07-06
Update time and ahash

yutannihilation:
Fix https://github.com/posit-dev/positron/issues/3911.

Update the time crate and ahash crate to solve compilation errors. Honestly I don't understand the details, but it seems the Rust compiler will be more strict about type ambiguity.


lionel-:
Thanks!

PR #427, opened 2024-07-06
Run CI on nightly

yutannihilation:
Even if you aren't very eager to update the toolchain, probably it's a good idea to run test on nightly as well so that you can detect the future breakage like https://github.com/posit-dev/positron/issues/3911 early.


lionel-:
seems reasonable, thanks!

PR #428, opened 2024-07-06
Remove unnecessary mut from OnceCell and OnceLock

yutannihilation:
Fix https://github.com/posit-dev/positron/issues/3912

`OnceCell::set()` doesn't require mutability.

https://doc.rust-lang.org/std/cell/struct.OnceCell.html#method.set


lionel-:
Thanks!

PR #429, opened 2024-07-07
Add a workaround for an argument matching error

yutannihilation:
A workaround for https://github.com/posit-dev/positron/issues/3467.

I read https://github.com/posit-dev/positron/issues/3467#issuecomment-2158872927, and I don't think this is the proper fix to the problem, but it might be good as a temporary workaround if it will take some time to overhaul the signature-help method.

### DIFF
--- a/crates/ark/src/lsp/completions/sources/unique/custom.rs
+++ b/crates/ark/src/lsp/completions/sources/unique/custom.rs
@@ -18,6 +18,7 @@ use stdext::unwrap;
 use stdext::IntoResult;
 use tower_lsp::lsp_types::CompletionItem;
 
+use crate::lsp;
 use crate::lsp::completions::completion_item::completion_item;
 use crate::lsp::completions::completion_item::completion_item_from_dataset;
 use crate::lsp::completions::completion_item::completion_item_from_package;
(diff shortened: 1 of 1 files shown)


crates/ark/src/lsp/completions/sources/unique/custom.rs
```diff
@@ -82,8 +82,16 @@ pub fn completions_from_custom_source_impl(
...
-    let index = signature.active_parameter.into_result()?;
-    let parameter = parameters.get(index as usize).into_result()?;
+    let index = signature.active_parameter.into_result()? as usize;
+    // TODO: Currently, argument matching is not very accurate. This is just a
+    // workaround to supresses the error rather than showing a cryptic error
+    // message to users, but there should be some better option.
+    //
+    // cf. https://github.com/posit-dev/positron/issues/3467
+    if index >= parameters.len() {
+        return Ok(None);
```
lionel-, line 92: Can you add some ERROR logging with `lsp::log_error!` please?

lionel- (commented):
It's a good idea to silently return until the underlying bug is fixed.

I verified that it fixes the `textDocument/completion` notifications but not the `completionItem/resolve` ones in functions taking dots like `bar <- function(...) {}`. Would you like to do the same for these? Could also be a different PR or we could do it.

yutannihilation:
Thanks for the review! I added `lsp::log_error!`.

> but not the `completionItem/resolve` ones in functions taking dots like `bar <- function(...) {}`.

It seems this occurs here (I'll paste the traceback in the original issue). I'm not yet sure if this is the same category of error that needs to be handled in this pull request, so I prefer a separate request at the moment.

https://github.com/posit-dev/ark/blob/8d44a5bc325312ede6e86d28ca5a57d82394b6e0/crates/ark/src/lsp/help.rs#L140-L144

lionel-:
Thanks! These notifications are annoying.

PR #432, opened 2024-07-08
Always display NA as unquoted in the Variables pane

yutannihilation:
This pull request addresses https://github.com/posit-dev/positron/issues/3810

Since `format()` and `format_one()` belongs to `Vector`, I think `format_with_string_options()` should also belong to `Vector` and be called there. But, I'm not sure if it's a good idea to include `FormattedVectorCharacterOptions` in the signature of all types of `Vector`. Considering https://github.com/posit-dev/positron/issues/2860 will introduce a facility to handle this kind of distinction easier, I'm not sure if this is worth, but I share this code just in case this helps.

### DIFF
--- a/crates/harp/src/vector/mod.rs
+++ b/crates/harp/src/vector/mod.rs
@@ -115,12 +115,4 @@ pub trait Vector {
             None => String::from("NA"),
         }
     }
-
-    fn format_with_options(&self, value: String, options: &FormatOptions) -> String {
-        if options.quote {
-            format!("\"{}\"", value.replace("\"", "\\\""))
-        } else {
-            value
-        }
-    }
 }
--- a/crates/harp/src/vector/mod.rs
+++ b/crates/harp/src/vector/mod.rs
@@ -12,7 +12,7 @@ use libr::SEXP;
 use crate::error::Result;
 use crate::utils::r_assert_capacity;
 use crate::utils::r_assert_type;
-use crate::vector::formatted_vector::FormattedVectorCharacterOptions;
+use crate::vector::formatted_vector::FormatOptions;
 
 pub mod character_vector;
 pub use character_vector::CharacterVector;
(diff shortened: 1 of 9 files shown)


lionel- (commented):
I don't mind having it in the signature, the added verbosity has the advantage that callers get reminded of formatting options. @dfalbel Since you worked on formatting, could you review this PR please?

crates/harp/src/vector/mod.rs
```diff
@@ -100,12 +101,32 @@ pub trait Vector {
...
+    fn format_one(
+        &self,
+        x: Self::Type,
+        options: Option<&FormattedVectorCharacterOptions>,
+    ) -> String;
+
+    fn format_elt_unchecked(
+        &self,
+        index: isize,
+        options: Option<&FormattedVectorCharacterOptions>,
```
dfalbel, line 113: I think we should move the definition of `FormattedVectorCharacterOptions` into this file. 
We probably also want to change it's name to something like `FormatOptions` as it's no longer specific to character vectors.

crates/harp/src/vector/mod.rs
```diff
@@ -100,12 +101,32 @@ pub trait Vector {
...
+    ) -> String {
         match self.get_unchecked(index) {
-            Some(x) => self.format_one(x),
+            Some(x) => self.format_one(x, options),
             None => String::from("NA"),
         }
     }
+
+    fn format_with_string_options(
```
dfalbel, line 121: I think this should be renamed to `format_with_options`.

dfalbel (commented):
These options were introduced in order to be able to control the formatting for the data explorer, but we no longer use the `FormattedVector` interface (see https://github.com/posit-dev/ark/pull/382) because this interface wasn't flexible enough to support both use cases. AFAICT `FormattedVector` is currently only used by the Variables pane, so we can make it specific for it.

Maybe a better approach is to simply remove those and we can add it back if we need it at some point. 

We should also add a test for the `NA` formatting here: https://github.com/posit-dev/ark/blob/0071f7a90ce035066883cfdab45248c4dd6fbdf9/crates/harp/src/vector/formatted_vector.rs#L242

yutannihilation:
Thanks for the review! Renamed the struct and add a test.

Also, thanks for the context. So, maybe it needs some more cleanup? I want to keep this pull request small for now (mainly because I don't know well about the implementations), but I can address in a separate pull request.

crates/harp/src/vector/mod.rs
```diff
@@ -100,12 +107,20 @@ pub trait Vector {
...
-    fn format_elt_unchecked(&self, index: isize) -> String {
+    fn format_elt_unchecked(&self, index: isize, options: Option<&FormatOptions>) -> String {
         match self.get_unchecked(index) {
-            Some(x) => self.format_one(x),
+            Some(x) => self.format_one(x, options),
             None => String::from("NA"),
         }
     }
+
+    fn format_with_options(&self, value: String, options: &FormatOptions) -> String {
```
dfalbel, line 119: If this only affects the character vectors, this should be moved to the `character_vector.rs` file.

crates/harp/src/vector/mod.rs
```diff
@@ -37,6 +37,13 @@ pub use raw_vector::RawVector;
 pub mod formatted_vector;
 pub mod names;
 
+// Formatting options for character vectors
+pub struct FormatOptions {
+    // Wether to quote the strings or not (defaults to `true`)
+    // If `true`, elements will be quoted during format so, eg: c("a", "b") becomes ("\"a\"", "\"b\"") in Rust
```
dfalbel, line 43: We should document that the `quote` option only affects character vectors.

dfalbel (approved):
Thanks @yutannihilation for the PR! I have just added a few more comments, but this looks great!

yutannihilation:
Thanks! I inlined `format_with_options()` and added a note.

PR #439, opened 2024-07-16
Add default argument values to signature labels

DavisVaughan:
Addresses https://github.com/posit-dev/positron/issues/2779

It is best to read this PR 1 commit at a time. Each commit is nicely self contained.

Adds support for showing default argument values in the function signature label. This makes the signature label wayyyy more useful for R code, as it often shows you the set of possible enum values. It's also just typically useful to know the default arguments of the function you are calling.

Screenshot 2024-07-16 at 5 20 18 PM

Screenshot 2024-07-16 at 5 20 28 PM

Positions are updated so they correctly highlight as you work through the function call

Screenshot 2024-07-16 at 5 20 42 PM

You'll note in the code that I have `vec_label()`, which supports both the scalar and vector case. The scalar case is pretty common, because `fn <- function(x = 1)` inlines the `1` as a double vector of length 1. The vector case is not so common. `fn <- function(x = c("a", "b"))` represents `c("a", "b")` _as a call_, so it goes through that path instead. That said, in some cases we do support inlining more complex objects, like `vctrs:::fn_inline_formals()`, so I went ahead and added support for vectors of length >1 in `vec_label()` too.

---

Related to https://github.com/posit-dev/positron/issues/1612. I think this PR is actually an easier solution, because when we look at the `formals()` of, say, `stringr::boundary()`:

```r
> formals(stringr::boundary)
$type
c("character", "line_break", "sentence", "word")

$skip_word_none
[1] NA

$...
```

we actually see `c("character", "line_break", "sentence", "word")` expressed _as a call_. So to implement the feature described there we'd have to evaluate that call, which can be tricky as many times it won't be able to be evaluated. This PR at least makes it so you can see the values without evaluating the call.


crates/ark/src/modules/positron/environment.R
```diff
@@ -14,15 +14,3 @@
 
     readLines(tf)
 }
-
-#' @export
-.ps.environment.describeCall <- function(expr, width.cutoff = 500L, nlines = -1L) {
```
DavisVaughan, line 19: Moved from ark to harp, and renamed to `expr_deparse()`

(Needed to call it from harp tests directly)

crates/harp/src/pretty.rs
DavisVaughan, line 1: Extracted from the DAP variables implementation. Seemed pretty useful here too.

crates/harp/src/pretty.rs
```diff
@@ -0,0 +1,191 @@
...
+    } else {
+        let mut out = String::from("\"");
+        let elt = r_str_to_owned_utf8(x).unwrap_or(String::from("???"));
+        out.push_str(&elt);
+        out.push_str("\"");
+        out
+    }
+}
+
+pub fn r_s3_pretty_class(x: SEXP) -> harp::Result<String> {
```
DavisVaughan, line 90: A shared way to get `"<first_class>"` in a reliable and failsafe way

kevinushey:
Throwing it out there: do we want to do anything to neatly truncate the signature tooltip in case of functions with a large number of arguments, or very large default arguments?

How would it feel if we only displayed the default argument values for the current / active formal argument, if any?

Some examples of functions with large formals:

- `str.default`
- `read.table`
- `install.packages`

DavisVaughan:
> How would it feel if we only displayed the default argument values for the current / active formal argument, if any?

I originally thought that wasn't possible because I thought we only got called 1 time up front, and then the frontend managed the "active parameter" stuff based on where the user's cursor it. But it seems like we are actually getting called after each `,` and `=` keypress, so we could do your suggestion if we wanted to.

DavisVaughan:
> do we want to do anything to neatly truncate the signature tooltip in case of functions with a large number of arguments

We'd have to be very careful with this because we also add parameter offsets to each parameter (which are offsets into the `label`). And `r_signature_help()` is utilized in places besides just the `SignatureHelp` handler (like in generating custom completions for `Sys.setenv()` for example) so we'd have to be careful there too.

Like, if we add `<...>` into the `label` or something, do parameter offsets map to that position when they are too long?

DavisVaughan:
Coming back to:

> do we want to do anything to neatly truncate the signature tooltip in case of functions with a large number of arguments, or very large default arguments?

I don't currently feel like truncation is necessary, even with long signatures like `install.packages()`. It just doesn't feel _that_ bad (see images above), and the alternative would mean that sometimes you'd have to fall back to the help documentation page, and that we'd have to additionally manage mapping the user's cursor position into the truncated range for highlighting the "current parameter" and that could be messy.

> How would it feel if we only displayed the default argument values for the current / active formal argument, if any?

My gut feeling is that it is more useful to see them all at once. Like, with `fill()` here even before I've put in the `data` I feel like it is useful to see the possible values for the other arguments. It's kind of hard to put into words why I feel this is more useful than only showing the values once you get to the argument itself.

Screenshot 2024-08-28 at 3 49 31 PM

kevinushey:
On the balance, I think you're right -- put another way, we probably shouldn't pessimize the design for the relatively more rare cases of functions with overly large signatures. We might still want to consider some sort of smart truncation for those cases, but even then doing nothing is probably fine too.

lionel-:
> I don't currently feel like truncation is necessary, even with long signatures like install.packages(). It just doesn't feel that bad (see images above), and the alternative would mean that sometimes you'd have to fall back to the help documentation page

This is my feeling as well.

crates/harp/src/pretty.rs
```diff
@@ -0,0 +1,191 @@
...
+            assert_eq!(r_lgl_to_pretty_string(1), String::from("TRUE"));
+            assert_eq!(r_lgl_to_pretty_string(0), String::from("FALSE"));
+            assert_eq!(r_lgl_to_pretty_string(r_lgl_na()), String::from("NA"));
+
+            assert_eq!(r_int_to_pretty_string(1), String::from("1L"));
+            assert_eq!(r_int_to_pretty_string(0), String::from("0L"));
+            assert_eq!(r_int_to_pretty_string(-1), String::from("-1L"));
+            assert_eq!(r_int_to_pretty_string(r_int_na()), String::from("NA"));
+
+            assert_eq!(r_dbl_to_pretty_string(1.5), String::from("1.5"));
```
lionel-, line 144: Needs a test with fractional values.

crates/harp/src/pretty.rs
```diff
@@ -0,0 +1,191 @@
+//
+// pretty.rs
```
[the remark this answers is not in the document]
lionel-: oh or the original `harp::lgl_to_string()`

crates/ark/src/lsp/signature_help.rs
```diff
@@ -336,13 +352,155 @@ fn is_within_call_parentheses(x: &Point, node: &Node) -> bool {
...
+fn null_label() -> String {
+    r_null_to_pretty_string()
+}
+
+fn list_label() -> String {
+    // TODO: Should we try and do better? I doubt an inlined list is that common.
+    String::from("<list>")
+}
+
+fn vec_label(x: SEXP, elt_to_pretty_string: fn(SEXP, isize) -> String) -> String {
```
lionel-, line 423: Do we need to be defensive against large inlined objects? E.g. ellipsise with `...` after 10 elements.

crates/ark/src/lsp/signature_help.rs
```diff
@@ -336,13 +352,155 @@ fn is_within_call_parentheses(x: &Point, node: &Node) -> bool {
...
+        },
+    }
+}
+
+fn null_label() -> String {
+    r_null_to_pretty_string()
+}
+
+fn list_label() -> String {
+    // TODO: Should we try and do better? I doubt an inlined list is that common.
```
lionel-, line 419: Probably best not to, especially if we need to consider whether someone inlined a data frame or model fit in the formals of their function 😱

crates/ark/src/lsp/signature_help.rs
lionel-: Or maybe we should be defensive downstream in `argument_label()` and truncate any label that exceeds a threshold?

crates/ark/src/lsp/signature_help.rs
lionel-: The most likely way this could happen is with code that generates functions from data.

lionel- (approved):
Looks good! A lot of the awkward precautions taken in this PR to deal with edge cases will no longer be necessary once we use source as input rather than live values.

crates/harp/src/pretty.rs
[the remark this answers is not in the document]
DavisVaughan: The idea was to imply there is some kind of opinionated nature about how it converts to string, and that it is mainly useful for showing in the IDE in some place, rather than as some "definitive" way to convert this object to a string. But I can go back to `lgl_to_string()` and we can tweak in the future as needed

crates/harp/src/pretty.rs
```diff
@@ -0,0 +1,191 @@
...
+    } else if r_dbl_is_nan(x) {
+        String::from("NaN")
+    } else if !r_dbl_is_finite(x) {
+        if x.is_sign_positive() {
+            String::from("Inf")
+        } else {
+            String::from("-Inf")
+        }
+    } else {
+        x.to_string()
```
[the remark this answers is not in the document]
DavisVaughan: The `to_string()` method for `f64` has some bonkers code for using the "shortest" possible string and it seems to work quite well, so im inclined to leave it as is. I did add a few more tests though https://doc.rust-lang.org/1.80.1/src/core/fmt/float.rs.html#192

We can set a precision but it seems to always apply, like setting a precision of 2 forces `1.` to be `1.00`. We could try and work around that by only applying a precision when there is no fractional value, but then you also have to decide on what precision to use when there are fractional values...and at that point I just gave up vs letting the Rust algorithm work its magic

crates/ark/src/lsp/signature_help.rs
DavisVaughan: Within argument truncation seems pretty nice for vectors

Screenshot 2024-08-29 at 1 29 55 PM

DavisVaughan: Also added less elegant forced truncation on any individual argument. Keeping individual arguments from being too longer is nicer than trying to truncate the overall signature because we don't have to add any extra logic to track the position of the "current" argument.

https://github.com/user-attachments/assets/5388ba3c-d41f-473a-8dd9-c17b0e5d9037

PR #443, opened 2024-07-17
Add `try_eval_with_x()` POC

DavisVaughan:
POC, not done, needs discussion with @lionel- about what this should look like. Part of a larger plan to stop inlining the heck out of everything when we do an R function call - see https://github.com/posit-dev/ark/issues/695

This technically addresses https://github.com/posit-dev/positron/issues/4008 and is motivated by it

The problem in that issue is that when `r_format(x)` calls the R level `harp_format(x)`, it _inlines_ `x` into the call. This means that when `harp_format()` errors and the R traceback is captured, the inlined version of `x` is captured in the R traceback, which we then print in the Output pane.

In this case `x` was an absolutely massive brmsfit object with elements in it that we don't know how to format, causing an error to get thrown from `harp_format()`, causing it to get inlined.

There are really two things to address here:

- First off, we actually expect `harp_format()` to error sometimes, so when we propagate its `Result` upward, do we really want the full traceback report to be provided whenever that `Error` is eventually logged? It feels to me like this is an "expected" error in some cases, and seeing the full traceback is noisy and misleading. I really just want to see the `message`, I think.

- Secondly, I think we need to invest in some infrastructure that allows us to replicate rlang's `eval_with_x()` idea. i.e. create a child env, define variables in it, and evaluate in that child env. This avoids the inline problem and results in cleaner tracebacks. I'd really like for this to be hooked into `RCall` and `RFunction` somehow, but I think we need to talk about the idea some more.

This PR starts by adding `try_eval_with_x()`, but it feels like the wrong level of abstraction

---

Reprex:

Run

```r
library(marginaleffects)
library(brms)

mod_miss <- insight::download_model("brms_miss_1")
```

Then go to the Variables pane and try and expand `mod_miss`.

Before:

Note that the `R backtrace` contains the entire inlined object.

Screenshot 2024-07-17 at 4 39 35 PM

After (note that we still noisily show the R thread traceback here, but at least we can tell what is going on):

Screenshot 2024-07-17 at 4 36 17 PM

### DIFF
--- a/crates/ark/src/dap/dap_variables.rs
+++ b/crates/ark/src/dap/dap_variables.rs
@@ -444,7 +444,7 @@ mod tests {
     use harp::exec::RFunctionExt;
     use harp::object::*;
     use harp::r_char;
-    use harp::utils::r_envir_set;
+    use harp::utils::r_env_poke;
     use libr::*;
 
     use crate::dap::dap_variables::env_binding_variable;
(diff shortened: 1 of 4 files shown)


DavisVaughan:
Will revive as needed

PR #444, opened 2024-07-19
Only emit dynamic plots for console sessions

jmcphers:
This change fixes a frequently reported bug in which Jupyter R notebooks in Positron don't show plots in the right location.

The problem is that dynamic plots (which always go to the Plots pane) are always emitted if Positron is connected to Ark. 

The fix is pretty simple; we just _also_ need to be sure we're in console mode before emitting a dynamic plot. 

Addresses https://github.com/posit-dev/positron/issues/3846

### DIFF
--- a/crates/ark/src/interface.rs
+++ b/crates/ark/src/interface.rs
@@ -124,6 +124,7 @@ use crate::startup;
 use crate::sys::console::console_to_utf8;
 
 /// An enum representing the different modes in which the R session can run.
+#[derive(PartialEq, Clone)]
 pub enum SessionMode {
     /// A session with an interactive console (REPL), such as in Positron.
     Console,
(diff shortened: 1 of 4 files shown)


juliasilge (approved):
This is now working great!

awkward

Note that the plot which is created is _extremely_ awkward in its default aspect ratio. I observe this for both base and ggplot2 plots, in both untitled and saved notebooks. 

How does this get controlled? I don't suppose this is an easy fix with the current PR?

jmcphers:
It's currently hardcoded to be that size! I feel weird changing it -- surely these values were chosen for a reason -- but you're absolutely right that it looks awkward. 

Updated to use a traditional 4:3 landscape orientation in https://github.com/posit-dev/ark/pull/444/commits/2f63cbcf69d92bdbbb85c3403c2c1a693cd2ba3b.

PR #446, opened 2024-07-23
Data Explorer: Add support for truncating outputs at specified length

dfalbel:
Makes the data explorer safer by avoiding transfering a lot of data over RPC if single cells contain a lot of data.
Address: https://github.com/posit-dev/positron/issues/3823

### DIFF
--- a/crates/ark/src/data_explorer/format.rs
+++ b/crates/ark/src/data_explorer/format.rs
@@ -49,7 +49,33 @@ pub fn format_string(x: SEXP, format_options: &FormatOptions) -> Vec<String> {
 }
 
 fn format(x: SEXP, format_options: &FormatOptions) -> Vec<FormattedValue> {
-    format_values(x, format_options).unwrap_or(unknown_format(x))
+    let mut formatted = format_values(x, format_options).unwrap_or(unknown_format(x));
+
+    // Truncate the values if they are too long
+    formatted.iter_mut().for_each(|v| {
+        if let FormattedValue::Value(x) = v {
+            truncate_inplace(x, format_options.max_value_length);
+        }
+    });
+
+    formatted
+}
+
+// Truncating strings in Rust is more complicated that one would imagine.
+// If you index using eg s[..6] that would take 6 bytes, which is not necessarily 6 characters.
+// We need to iterate over the characters and truncate the string at the right character.
+// See also https://doc.rust-lang.org/book/ch08-02-strings.html
+fn truncate_inplace(s: &mut String, max_len: i64) {
+    // Returns the indice of the nth character in the string if it exists
+    let index = s.char_indices().nth(max_len as usize);
+
+    // If an index is found, truncate the string to that index
+    match index {
+        // truncate is an in-place operation that takes the new max index as input
+        // but looking the string as a sequence of bytes, not chars.
+        Some((i, _)) => s.truncate(i),
+        None => (),
+    }
 }
 
 fn unknown_format(x: SEXP) -> Vec<FormattedValue> {
@@ -425,6 +451,7 @@ mod tests {
             small_num_digits: 4,
             max_integral_digits: 7,
             thousands_sep: Some(",".to_string()),
+            max_value_length: 100,
         }
     }
 
@@ -514,6 +541,7 @@ mod tests {
                 small_num_digits: 4,
                 max_integral_digits: 7,
                 thousands_sep: None,
+                max_value_length: 100,
             };
             let expected = vec![
                 "0.00",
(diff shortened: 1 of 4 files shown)


crates/ark/src/data_explorer/format.rs
```diff
@@ -49,7 +49,33 @@ pub fn format_string(x: SEXP, format_options: &FormatOptions) -> Vec<String> {
...
+// See also https://doc.rust-lang.org/book/ch08-02-strings.html
+fn truncate_inplace(s: &mut String, max_len: i64) {
+    // Returns the indice of the nth character in the string if it exists
+    let index = s.char_indices().nth(max_len as usize);
+
+    // If an index is found, truncate the string to that index
+    match index {
+        // truncate is an in-place operation that takes the new max index as input
+        // but looking the string as a sequence of bytes, not chars.
+        Some((i, _)) => s.truncate(i),
```
DavisVaughan, line 76: Is this right? Doesn't `i` correspond to the _i-th character_? But `truncate()` is looking for _the i-th byte_?

DavisVaughan: Ah according to this, the `i` there is the _byte index_ of that n-th character, so I think you are right
https://stackoverflow.com/questions/38461429/how-can-i-truncate-a-string-to-have-at-most-n-characters

crates/ark/src/data_explorer/format.rs
```diff
@@ -49,7 +49,33 @@ pub fn format_string(x: SEXP, format_options: &FormatOptions) -> Vec<String> {
...
 fn format(x: SEXP, format_options: &FormatOptions) -> Vec<FormattedValue> {
-    format_values(x, format_options).unwrap_or(unknown_format(x))
+    let mut formatted = format_values(x, format_options).unwrap_or(unknown_format(x));
+
+    // Truncate the values if they are too long
+    formatted.iter_mut().for_each(|v| {
+        if let FormattedValue::Value(x) = v {
+            truncate_inplace(x, format_options.max_value_length);
+        }
+    });
```
DavisVaughan, line 59: I think I would vote for calling `truncate_inplace()` _inside_ of `format_values()`. That seems like it would make more sense to me?

crates/ark/src/data_explorer/format.rs
```diff
@@ -709,4 +738,33 @@ mod tests {
...
+        r_test(|| {
+            let mut options = default_options();
+            options.max_value_length = 3;
+
+            let data = r_parse_eval0(r#"c("aaaaa", "aaaaaaaa", "aa")"#, R_ENVS.global).unwrap();
+            let formatted = format_column(data.sexp, &options);
+            assert_eq!(formatted, vec![
+                ColumnValue::FormattedValue("aaa".to_string()),
+                ColumnValue::FormattedValue("aaa".to_string()),
+                ColumnValue::FormattedValue("aa".to_string()),
```
DavisVaughan, line 753: On the frontend, does the user get any indication of the fact that the value was truncated? It might be nice for them to have some visual indication of this, like a trailing `...` or something.

I assume that would mean the backend would have to send over some details in `ColumnValue::FormattedValue` like `truncated: true` to tell the frontend that it was truncated. May be worth it, not sure. I can just imagine some confusion coming from the fact that it could _look_ like you have all the values, but in reality some truncation happened

DavisVaughan (approved):

crates/ark/src/data_explorer/format.rs
dfalbel: Yep, it's the byte index for the specific character: https://doc.rust-lang.org/std/primitive.str.html#method.char_indices

crates/ark/src/data_explorer/format.rs
dfalbel: The reason for adding it here, is mostly to avoid adding truncation in multiple places as we have early returns in `format_values`. Since `format_values` is not public interface, seems fine to leave it here?

crates/ark/src/data_explorer/format.rs
dfalbel: Currently with the default option to `1000` (set by the frot-end) it is actually impossible for a user to see the end of the string, so the truncation is only there as a safety feature to avoid sending a huge amount of data to the front-end in case there's a single cell with a huge text inside.

crates/ark/src/data_explorer/format.rs
DavisVaughan: 👍  i was looking for information about what the first element in the tuple corresponded to over on the `CharIndices` iterator page, but couldn't find it! Didn't think to look at the str method page
https://doc.rust-lang.org/std/str/struct.CharIndices.html

PR #448, opened 2024-07-25
Send HTML widgets as Show File events to Positron

jmcphers:
This change is the R side of https://github.com/posit-dev/positron/pull/4151; see that PR for a detailed explanation of the changes.


crates/ark/src/modules/positron/html_widgets.R
```diff
@@ -9,28 +9,49 @@
...
+    tmp_file <- htmltools::html_print(rendered, viewer = NULL)
+
+    # Derive the height of the viewer pane from the sizing policy of the widget.
+    is_plot <- x$sizingPolicy$knitr$figure
+    height <- x$sizingPolicy$viewer$paneHeight
+    if (identical(height, 'maximize')) {
+        height <- -1
+    } else if (is.null(height)) {
+        height <- 0
+    }
```
DavisVaughan, line 23: For both `is_plot` and `height`, maybe add additional checks that they are bool-like and integer-like respectively? If they aren't, may result in a rust panic. Could have default values if we detect weirdness

crates/ark/src/modules/positron/html_widgets.R
```diff
@@ -9,28 +9,49 @@
...
     add_s3_override("print.htmlwidget", .ps.view_html_widget)
+    add_s3_override("print.shiny.tag", .ps.view_html_widget)
+    add_s3_override("print.shiny.tag.list", .ps.view_html_widget)
 }
 
 #' @export
 .ps.viewer.removeOverrides <- function() {
     remove_s3_override("print.htmlwidget")
+    remove_s3_override("print.shiny.tag", .ps.view_html_widget)
+    remove_s3_override("print.shiny.tag.list", .ps.view_html_widget)
```
DavisVaughan, line 54:
```suggestion
    remove_s3_override("print.shiny.tag")
    remove_s3_override("print.shiny.tag.list")
```

I imagine the fact that this slipped through suggests that we haven't actually exercised this code yet, otherwise we would have seen an error already?

crates/ark/src/modules/positron/viewer.R
```diff
@@ -6,11 +6,28 @@
 #
 
 options("viewer" = function(url, height = NULL, ...) {
+    # Normalize paths for comparison. This is necessary because on e.g. macOS,
+    # the `tempdir()` may contain `//` or other non-standard path separators.
+    normalizedPath <- normalizePath(url, mustWork = FALSE)
```
DavisVaughan, line 11:
```suggestion
    # Normalize paths for comparison. This is necessary because on e.g. macOS,
    # the `tempdir()` may contain `//` or other non-standard path separators.
    # See https://github.com/posit-dev/positron/issues/3018.
    normalizedPath <- normalizePath(url, mustWork = FALSE)
```

This fixes https://github.com/posit-dev/positron/issues/3018 right? You could include that in your list of issues that your Positron side PR addresses.

And https://github.com/posit-dev/positron/issues/3269 since it was probably the same underlying issue

crates/ark/src/modules/positron/viewer.R
```diff
@@ -6,11 +6,28 @@
...
+    normalizedTempdir <- normalizePath(tempdir(), mustWork = FALSE)
+
     # Is the URL a temporary file?
-    if (startsWith(url, tempdir())) {
+    if (startsWith(normalizedPath, normalizedTempdir)) {
+        # Use the filename as the label, unless it's an index file, in which
+        # case use the directory name.
+        fname <- tolower(basename(url))
```
DavisVaughan, line 18: Should you just go ahead and use the normalized `url` everywhere? i.e. when getting the label and passing through to Rust? Seems right to me to do that

crates/ark/src/modules/positron/viewer.R
```diff
@@ -6,11 +6,28 @@
...
+        # R HTML widgets get printed to temporary files starting with the name
+        # "viewhtml". This makes an ugly label, so we give it a nicer one.
+        if (startsWith(fname, "viewhtml")) {
+            fname <- "R HTML widget"
+        }
+
         # If so, open it in the HTML viewer.
-        .ps.Call("ps_html_viewer", url)
-        # TODO: handle `height` for HTML viewer
+        .ps.Call("ps_html_viewer", url, fname, height, FALSE)
```
DavisVaughan, line 30: Should probably do some R side validation of `height` here?

Like, in other places i saw this

```
if (is.null(height)) {
        height <- 0
    }
```

crates/ark/src/modules/rstudio/stubs.R
```diff
@@ -36,3 +36,27 @@
...
+    if (!is.character(url) || (length(url) != 1))
+        stop("url must be a single element character vector.")
+    if (identical(height, "maximize"))
+        height <- -1
+    if (!is.null(height) && (!is.numeric(height) || (length(height) !=
+        1)))
+        stop("height must be a single element numeric vector or 'maximize'.")
+    if (is.null(height)) {
+        height <- 0
+    }
```
DavisVaughan, line 51: This is what i was looking for in the other two places `height` is used! Could be a little helper like `height <- check_viewer_height(height)`

crates/ark/src/modules/rstudio/stubs.R
```diff
@@ -36,3 +36,27 @@
...
+    fname <- tolower(basename(url))
+    if (identical(fname, "index.html") || identical(fname, "index.htm")) {
+        fname <- basename(dirname(url))
+    }
```
DavisVaughan, line 55: You also have some `"viewhtml"` handling in `options("viewer" =`. Include that here?

Maybe a `viewer_file_name(url)` helper that can be used in both?

crates/ark/src/viewer.rs
```diff
@@ -41,17 +51,51 @@ fn emit_html_output(iopub_tx: Sender<IOPubMessage>, path: String) -> Result<()>
...
+            // Emit HTML output
             let main = RMain::get();
             let iopub_tx = main.get_iopub_tx().clone();
-            if let Err(err) = emit_html_output(iopub_tx, path) {
-                log::error!("Failed to emit HTML output: {:?}", err);
+            if main.session_mode == SessionMode::Notebook {
+                // In notebook mode, send the output as a Jupyter display_data message
+                if let Err(err) = emit_html_output_jupyter(iopub_tx, path, label) {
+                    log::error!("Failed to emit HTML output: {:?}", err);
+                }
```
DavisVaughan, line 76: What "mode" is ark going to run in for Jupyter Console?

crates/ark/src/viewer.rs
```diff
@@ -41,17 +51,51 @@ fn emit_html_output(iopub_tx: Sender<IOPubMessage>, path: String) -> Result<()>
...
+    };
     match path {
         Ok(path) => {
-            // Emit the HTML output
+            // Emit HTML output
             let main = RMain::get();
             let iopub_tx = main.get_iopub_tx().clone();
-            if let Err(err) = emit_html_output(iopub_tx, path) {
-                log::error!("Failed to emit HTML output: {:?}", err);
+            if main.session_mode == SessionMode::Notebook {
```
DavisVaughan, line 72: If its ok with you, I have a slight preference for a `match` call here to force us to be explicit about our intended behavior for each possible `SessionMode` (it will force us to come back here and adjust the code if we add a new SessionMode variant). Looks like both `Console` and `Background` go through the else right now?

crates/ark/src/modules/positron/viewer.R
DavisVaughan: In fact, isn't a `NULL` height a problem here right now? Can that even be converted to an i32? If so I'd be surprised

crates/ark/src/viewer.rs
```diff
@@ -41,17 +51,51 @@ fn emit_html_output(iopub_tx: Sender<IOPubMessage>, path: String) -> Result<()>
...
+                // In notebook mode, send the output as a Jupyter display_data message
+                if let Err(err) = emit_html_output_jupyter(iopub_tx, path, label) {
+                    log::error!("Failed to emit HTML output: {:?}", err);
+                }
+            } else {
+                let is_plot = RObject::view(is_plot).to::<bool>();
+                let height = RObject::view(height).to::<i32>();
```
DavisVaughan, line 79:
```suggestion
                let height = RObject::view(height).to::<i64>();
```

Does this exist so you don't need the `into`?

crates/ark/src/modules/positron/viewer.R
DavisVaughan: Oh I see

```r
                    height: match height {
                        Ok(height) => height.into(),
                        Err(_) => 0,
                    },
```

That is probably catching the _error_ resulting from a `NULL` height and it just happens to get mapped to `0`. That definitely seems like a conversion to handle on the R side then

crates/ark/src/viewer.rs
```diff
@@ -41,17 +51,51 @@ fn emit_html_output(iopub_tx: Sender<IOPubMessage>, path: String) -> Result<()>
...
+                let params = ShowHtmlFileParams {
+                    path,
+                    title: label.clone(),
+                    height: match height {
+                        Ok(height) => height.into(),
+                        Err(_) => 0,
+                    },
```
DavisVaughan, line 86: Would you mind lifting this error checking out into its own line with a `log::warn()` in the failure path? 

Something like

```rust
let height = match height {
  Ok(height) => height.into(),
  Err(err) => {
    log::warn!("Can't convert `height` into an i32, using `0` as a fallback: {err:?}");
    0
  },
}
```

Same for `is_plot`.

That possibly could have helped catch the `NULL` `height` i mentioned earlier

crates/ark/src/viewer.rs
```diff
@@ -41,17 +51,51 @@ fn emit_html_output(iopub_tx: Sender<IOPubMessage>, path: String) -> Result<()>
...
+                    is_plot: match is_plot {
+                        Ok(plot) => plot,
+                        Err(_) => false,
+                    },
+                };
+                main.send_frontend_event(UiFrontendEvent::ShowHtmlFile(params));
+
+                let message = IOPubMessage::Stream(StreamOutput {
+                    name: Stream::Stdout,
+                    text: format!("<{}>", label),
```
DavisVaughan, line 96:
```suggestion
                    text: format!("<{label}>"),
```

crates/amalthea/src/comm/ui_comm.rs
```diff
@@ -251,6 +251,25 @@ pub struct ShowUrlParams {
...
 
+/// Parameters for the ShowHtmlFile method.
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+pub struct ShowHtmlFileParams {
+	/// The fully qualified filesystem path to the HTML file to display
+	pub path: String,
+
+	/// A title to be displayed in the viewer. May be empty, and can be
+	/// superseded by the title in the HTML file.
+	pub title: String,
```
DavisVaughan, line 262: Could be `Option<String>` based on your docs here, but I guess you're looking for an empty string on the frontend side?

DavisVaughan (approved):

DavisVaughan:
reprex is correctly going to the Viewer pane for me, but I prefer it without the IOPub Stream message. Otherwise it conflicts with reprex's actual output

Screenshot 2024-07-26 at 2 13 02 PM

DavisVaughan:
`profvis::profvis({Sys.sleep(2)})` experience still seems a bit wonky

https://github.com/user-attachments/assets/2e728b24-7b5c-4305-baad-c428caa5b819

jmcphers:
Profvis isn't showing up in the Viewer because it is explicitly asking not to, via `viewer.suppress` (which creates an object with `suppress_viewer` class instead of the `htmlwidget` class;) 

https://github.com/r-lib/profvis/blob/b8233d3f5295be64aa6b5719db457a94cfd5ebf6/R/profvis.R#L224-L227

```r
    sizingPolicy = htmlwidgets::sizingPolicy(
      padding = 0,
      browser.fill = TRUE,
      viewer.suppress = TRUE,
...
```

crates/ark/src/modules/positron/viewer.R
jmcphers: It does fix 3018! I'll resolve that one with the PR.

it doesn't fix 3269. :-(

crates/ark/src/modules/positron/html_widgets.R
jmcphers: Sure, done!

crates/ark/src/modules/positron/html_widgets.R
jmcphers: Yes, this code was adapted from RStudio, where we do this S3 override enable/disable constantly (as a preamble and then cleanup every time you run code in notebook/inline output mode). In Positron we flip the switch on but don't currently turn it off. Felt weird to not have the off switch but maybe I should remove it since we aren't using it...

crates/ark/src/modules/positron/viewer.R
jmcphers: Done!

crates/ark/src/modules/positron/viewer.R
jmcphers: I'll add handling on the R side too! But I'm going the leave the Rust bit in as a layer of defense (anyone can call this function if they try hard enough and I don't think a type error should cause a panic here)

crates/ark/src/modules/rstudio/stubs.R
jmcphers: done!

crates/ark/src/modules/rstudio/stubs.R
jmcphers: Sure!

crates/ark/src/viewer.rs
jmcphers: A good question ... I think it defaults to console mode, and we should switch it to notebook mode (also possible to just add session mode to the `argv` when registering the kernel).

crates/amalthea/src/comm/ui_comm.rs
jmcphers: A limitation of the code generator -- it supports optional properties on object, but not optional parameters on methods (which is where this structure is derived from).

crates/ark/src/viewer.rs
jmcphers: No, it doesn't. :-/

crates/ark/src/viewer.rs
jmcphers: can do!

crates/ark/src/viewer.rs
jmcphers: I just took this out as discussed above. I like having _something_ printed to the console as an acknowledgment that we did something (and at one point considered having it print out a link that opened the content in the Viewer so you could scroll back and replay it), but agree it's hard to get this right in all cases.

crates/ark/src/viewer.rs
jmcphers: Yep, I'll make that explicit.

PR #449, opened 2024-07-26
Add new `.ps.ui.evaluateWhenClause` OpenRPC method

juliasilge:
I was thinking about how to address https://github.com/posit-dev/positron/issues/2697 and have a proposed approach in this PR plus one I will open next in Positron. I looked at how various extensions (like the git extension, etc) check for whether we are in a git repo in the workspace and all that checking is typically based on context keys. In an extension specifically, you can check those keys with a [when clause](https://code.visualstudio.com/api/references/when-clause-contexts) and I realized that approach (i.e. a string) could work well as an OpenRPC method and would be flexible for other context keys we need to check from runtimes.

This PR does not yet actually check for git repos for #2697 but I wanted to get input on the approach before going much further. To see the behavior, you can check out code such as:

```r
.ps.ui.evaluateWhenClause("isLinux || isWindows")
.ps.ui.evaluateWhenClause("isMac")
.ps.ui.evaluateWhenClause("gitOpenRepositoryCount >= 1")
```

### DIFF
--- a/crates/ark/src/modules/rstudio/commands.R
+++ b/crates/ark/src/modules/rstudio/commands.R
@@ -8,7 +8,7 @@
         "saveAllSourceDocs" = "workbench.action.files.saveAll",
         # This command is a silent no-op in RStudio when there is no git repo:
         "vcsRefresh" = {
-            if (.ps.ui.evaluateWhenClause("gitOpenRepositoryCount >= 1")) {
+            if (.ps.ui.evaluateWhenClause("gitOpenRepositoryCount > 0")) {
                 "git.refresh"
             } else {
                 return()
--- a/crates/ark/src/modules/rstudio/commands.R
+++ b/crates/ark/src/modules/rstudio/commands.R
@@ -9,7 +9,7 @@
         # https://github.com/posit-dev/positron/issues/2697
         # This command is a silent no-op in RStudio when there is no git repo:
         "vcsRefresh" = {
-            if (.ps.ui.evaluateWhenClause("gitOpenRepositoryCount > 0")) {
+            if (.ps.ui.evaluateWhenClause("config.git.enabled && gitOpenRepositoryCount > 0")) {
                 "git.refresh"
             } else {
                 return()
--- a/crates/ark/src/modules/rstudio/commands.R
+++ b/crates/ark/src/modules/rstudio/commands.R
@@ -6,7 +6,14 @@
         "activateTerminal" = "workbench.action.terminal.focus",
         # This command includes untitled files in RStudio:
         "saveAllSourceDocs" = "workbench.action.files.saveAll",
-        "vcsRefresh" = "git.refresh",
+        # This command is a silent no-op in RStudio when there is no git repo:
+        "vcsRefresh" = {
+            if (.ps.ui.evaluateWhenClause("gitOpenRepositoryCount >= 1")) {
+                "git.refresh"
+            } else {
+                return()
+            }
+        },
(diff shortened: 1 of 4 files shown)


juliasilge:
Now with 6914a61258b9edbef44ab841cddecfa5b71eaa3d, the behavior for `.rs.api.executeCommand("vcsRefresh")` has changed to address #2697:

- If there is no repo, it is a silent no-op
- If there is a repo, the UI will refresh for the git repo state

crates/ark/src/modules/rstudio/commands.R
```diff
@@ -6,7 +6,14 @@
         "activateTerminal" = "workbench.action.terminal.focus",
         # This command includes untitled files in RStudio:
         "saveAllSourceDocs" = "workbench.action.files.saveAll",
-        "vcsRefresh" = "git.refresh",
+        # This command is a silent no-op in RStudio when there is no git repo:
+        "vcsRefresh" = {
+            if (.ps.ui.evaluateWhenClause("gitOpenRepositoryCount > 0")) {
```
juliasilge, line 11: There is a lot of things you can check about the git situation (for example, [see what is checked to show the "The folder currently open doesn't have a Git repository. You can initialize..." screen](https://github.com/posit-dev/positron/blob/693987748002d8aed55c2913dbe58290a249cb10/extensions/git/package.json#L3357)) but I am pretty sure `gitOpenRepositoryCount > 0` is necessary + sufficient for the refreshing.

crates/ark/src/modules/rstudio/commands.R
```diff
@@ -6,7 +6,14 @@
         "activateTerminal" = "workbench.action.terminal.focus",
         # This command includes untitled files in RStudio:
         "saveAllSourceDocs" = "workbench.action.files.saveAll",
-        "vcsRefresh" = "git.refresh",
+        # This command is a silent no-op in RStudio when there is no git repo:
```
jennybc, line 9: I think it *might* be worth noting the issue, in case anyone is tempted to mess with this:

```suggestion
        # https://github.com/posit-dev/positron/issues/2697
        # This command is a silent no-op in RStudio when there is no git repo:
```

DavisVaughan:
A note that an easy way to test this with a git repo is to set the workspace setting:

```
"git.enabled": false
```

Then running `devtools::test()` gives me the expected popup mentioned in https://github.com/posit-dev/positron/issues/2697.

---

Another note that I want to put somewhere is that I _used_ to have an issue with this in tree-sitter-r, which had `.git` at the top level, but had an R package in a subdirectory at `bindings/r/`. I typically like to open just `bindings/r/` in its own Positron instance, but that would trigger the above issues.

I since learned that when you open a subdirectory like `bindings/r/`, you can use the Source Control panel to _point git at a parent repository_ that the panel can track instead, and then everything magically works. I can't reproduce it now, but the Source Control panel had already found the parent git folder for me, I just had to click a box to say "yes please use that" and then everything worked nicely even without this fix.

crates/ark/src/modules/rstudio/commands.R
DavisVaughan: I would not be against doing `config.git.enabled && gitOpenRepositoryCount > 0`, but I'll leave that up to you

DavisVaughan (approved):
Works nicely with `devtools::test()` in a repo where I've turned off git support with `git.enabled: false`, which previously threw up that "no available repository" note.

Nice work! Super cool idea

PR #451, opened 2024-07-29
Data Explorer: Re-apply filters after Data Update

dfalbel:
Addresses https://github.com/posit-dev/positron/issues/4170 by making sure filters are also reapplied after data update events.

### DIFF
--- a/crates/ark/src/data_explorer/r_data_explorer.rs
+++ b/crates/ark/src/data_explorer/r_data_explorer.rs
@@ -381,6 +381,11 @@ impl RDataExplorer {
             if self.sort_keys.len() > 0 {
                 self.sorted_indices = Some(r_task(|| self.r_sort_rows())?);
             }
+
+            // This is a no-op if there are no filters to be applied.
+            let (indices, _) = self.row_filters_compute()?;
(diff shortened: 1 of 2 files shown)


crates/ark/src/data_explorer/r_data_explorer.rs
```diff
@@ -379,6 +379,11 @@ impl RDataExplorer {
             if self.sort_keys.len() > 0 {
                 self.sorted_indices = Some(r_task(|| self.r_sort_rows())?);
             }
+
+            // This is a no-op if there are no filters to be applied.
+            let (indices, _) = self.row_filters_compute()?;
+            self.filtered_indices = indices;
+
             self.apply_sorts_and_filters();
```
DavisVaughan, line 387:
```suggestion
            // Recompute and apply filters and sorts.
            let (indices, _) = self.row_filters_compute()?;
            self.filtered_indices = indices;
            self.apply_sorts_and_filters();
```

Maybe just copy/paste from the `if` branch path? Might help us keep it in sync. (I think the no-op can be implied)

DavisVaughan (approved):

PR #453, opened 2024-07-30
Switch GetSchema from column range to column indices

softwarenerd:
This PR switches `GetSchema` from accepting a column range (`start_index` / `num_columns`) to accepting an array of column indices (`column_indices`). This addresses a nascent bug because the Data Explorer was calculating column indices but then passing them as a column range.

This change also addresses the `i32` limitation of the previous `GetSchema` implementation. So this comment no longer applies:

```
// TODO: Support for data frames with over 2B rows. Note that neither base R nor
// tidyverse support long vectors in data frames, but data.table does.
```

It is now possible to load column indices greater than `2_147_483_647`.
It also attempts to bump the ark version to `0.1.121`.
It also adds `.DS_Store` to the `.gitignore` file.

Please do not merge this PR without also merging https://github.com/posit-dev/positron/pull/4180.

### DIFF
--- a/crates/ark/Cargo.toml
+++ b/crates/ark/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "ark"
-version = "0.1.121"
+version = "0.1.120"
 description = """
 Ark, an R Kernel.
 """
--- a/crates/ark/Cargo.toml
+++ b/crates/ark/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "ark"
-version = "0.1.120"
+version = "0.1.121"
 description = """
 Ark, an R Kernel.
 """
--- a/crates/ark/tests/data_explorer.rs
+++ b/crates/ark/tests/data_explorer.rs
@@ -142,7 +142,7 @@ fn default_format_options() -> FormatOptions {
 
 fn test_mtcars_sort(socket: CommSocket, has_row_names: bool, display_name: String) {
     // Get the schema for the test data set.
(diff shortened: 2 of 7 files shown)


crates/ark/tests/data_explorer.rs
```diff
@@ -423,8 +421,11 @@ fn test_matrix_support() {
 
         // Get the schema for the test data set.
         let req = DataExplorerBackendRequest::GetSchema(GetSchemaParams {
-            num_columns: 61,
-            start_index: 0,
+            column_indices: vec![
```
dfalbel, line 424: This could be something like `(0..61).collect_vec()`

crates/ark/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "ark"
-version = "0.1.120"
+version = "0.1.121"
```
dfalbel, line 3: By convention we bump ark versions in a separate PR (that doesn't need a review), because it's possible that the ark version is bumped by someoneelse while this PR is still in review.

dfalbel (commented):
Looks good

crates/ark/Cargo.toml
softwarenerd: Roger that.

crates/ark/tests/data_explorer.rs
softwarenerd: Indeed. I will do a PR for this.

PR #458, opened 2024-08-02
Data explorer: Add support for histograms and frequency tables

dfalbel:
This implements the OpenRPC protocol for support computing histograms and frequency tables. This is the backend for computing sparklines in the data explorer. 

Addresses https://github.com/posit-dev/positron/issues/4061

TODO:

- [x] Add support for frequency tables (consider factors, and character vectors)
- [x] Add tests for quantile computations.
- [x] Add implementation/tests for the Sturges algorithm
- [x] Plumb with the infra to produce results on the filtered data (if any filters are applied)

### DIFF
--- a/crates/ark/src/data_explorer/histogram.rs
+++ b/crates/ark/src/data_explorer/histogram.rs
@@ -2,6 +2,7 @@ use std::collections::HashMap;
 
 use amalthea::comm::data_explorer_comm::ColumnHistogram;
 use amalthea::comm::data_explorer_comm::ColumnHistogramParams;
+use amalthea::comm::data_explorer_comm::ColumnHistogramParamsMethod;
 use amalthea::comm::data_explorer_comm::ColumnQuantileValue;
 use amalthea::comm::data_explorer_comm::FormatOptions;
 use anyhow::anyhow;
@@ -45,9 +46,20 @@ pub fn profile_histogram(
         },
     }
 
+    let num_bins: RObject = match params.num_bins {
+        Some(v) => (v as i32).into(),
+        None => r_null().into(),
+    };
+
+    let method: RObject = match params.method {
+        ColumnHistogramParamsMethod::Fixed => "fixed".into(),
+        ColumnHistogramParamsMethod::Sturges => "sturges".into(),
+    };
+
     let results: HashMap<String, RObject> = RFunction::from("profile_histogram")
         .add(column)
-        .add(params.num_bins as i32)
+        .add(method)
+        .add(num_bins)
         .add(quantiles)
         .call_in(ARK_ENVS.positron_ns)?
         .try_into()?;
@@ -128,7 +140,8 @@ mod tests {
         let hist = profile_histogram(
             column.sexp,
             &ColumnHistogramParams {
-                num_bins,
+                method: ColumnHistogramParamsMethod::Fixed,
+                num_bins: Some(num_bins),
                 quantiles: None,
             },
             &default_options(),
@@ -151,7 +164,8 @@ mod tests {
         let hist = profile_histogram(
             column.sexp,
             &ColumnHistogramParams {
-                num_bins: 100,
+                method: ColumnHistogramParamsMethod::Fixed,
+                num_bins: Some(100),
                 quantiles: Some(quantiles),
             },
             &default_options(),
--- a/crates/ark/src/data_explorer/histogram.rs
+++ b/crates/ark/src/data_explorer/histogram.rs
@@ -155,6 +155,27 @@ mod tests {
         })
     }
 
+    fn test_histogram_sturges(code: &str, bin_edges: Vec<&str>, bin_counts: Vec<i64>) {
+        let column = r_parse_eval0(code, R_ENVS.global).unwrap();
+
+        let hist = profile_histogram(
+            column.sexp,
+            &ColumnHistogramParams {
+                method: ColumnHistogramParamsMethod::Sturges,
+                num_bins: None,
+                quantiles: None,
+            },
+            &default_options(),
+        )
+        .unwrap();
+
+        assert_eq!(hist, ColumnHistogram {
+            bin_edges: bin_edges.into_iter().map(|v| v.to_string()).collect(),
+            bin_counts,
+            quantiles: vec![]
+        })
+    }
+
     fn test_quantiles<T>(code: &str, quantiles: Vec<f64>, expected: T)
     where
         RObject: From<T>,
@@ -188,8 +209,8 @@ mod tests {
             test_histogram(
                 "0:10",
                 4,
-                vec!["0.00", "2.50", "5.00", "7.50", "10.00"],
-                vec![3, 3, 2, 3],
+                vec!["0.00", "2.00", "4.00", "6.00", "8.00", "10.00"],
+                vec![3, 2, 2, 2, 2],
             );
         })
     }
@@ -207,7 +228,7 @@ mod tests {
                     "2000-01-08 12:00:00",
                     "2000-01-11 00:00:00",
                 ],
-                vec![3, 2, 3, 3],
+                vec![3, 3, 2, 3],
             );
 
             test_histogram(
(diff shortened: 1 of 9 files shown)


crates/ark/src/modules/positron/r_data_explorer.R
```diff
@@ -337,3 +333,113 @@ write_html <- function(x, include_header) {
...
+    }
+
+    # The pretty bins algorithm doesn't really make sense for dates,
+    # so we generate our own bin_edges for them.
+    min_value <- min(x)
+    max_value <- max(x)
+    range <- max_value - min_value
+
+    if (inherits(x, "POSIXct") && as.integer(range) < num_bins) {
+     num_bins <- as.integer(range) + 1L
```
lionel-, line 379:
```suggestion
      num_bins <- as.integer(range) + 1L
```

crates/ark/src/modules/positron/r_data_explorer.R
```diff
@@ -337,3 +333,113 @@ write_html <- function(x, include_header) {
...
+    quantiles <- c() # we otherwise return an empty quantiles vector
+  }
+
+  method <- match.arg(method)
+  # For dates, `hist` does not compute the number of breaks automatically
+  # using default methods.
+  # We force something considering the integer representation.
+  if (inherits(x, "POSIXct")) {
+    if (method == "sturges") {
+        num_bins <- grDevices::nclass.Sturges(x)
```
lionel-, line 369:
```suggestion
      num_bins <- grDevices::nclass.Sturges(x)
```

crates/ark/src/modules/positron/r_data_explorer.R
```diff
@@ -337,3 +333,113 @@ write_html <- function(x, include_header) {
...
+    }
+
+    # The pretty bins algorithm doesn't really make sense for dates,
+    # so we generate our own bin_edges for them.
+    min_value <- min(x)
+    max_value <- max(x)
+    range <- max_value - min_value
```
lionel-, line 376: width rather than range?

crates/ark/src/modules/positron/r_data_explorer.R
```diff
@@ -337,3 +333,113 @@ write_html <- function(x, include_header) {
...
+    breaks <- seq(min_value, max_value, length.out = num_bins + 1)
+  } else if (method == "sturges") {
+    breaks <- "Sturges"
+  } else if (method == "fixed") {
+    # Note that with fixed, the number of bins can be slightly different in some cases
+    # due to R building pretty bin_edges. From `hist` docs:
+    # > In the last three cases the number is a suggestion only; as the breakpoints will
+    # > be set to `pretty` values.
+    breaks <- num_bins
+    stopifnot(is.integer(breaks) && length(breaks) == 1)
```
lionel-, line 391:
```suggestion
    stopifnot(is.integer(breaks), length(breaks) == 1)
```

crates/ark/src/modules/positron/r_data_explorer.R
```diff
@@ -337,3 +333,113 @@ write_html <- function(x, include_header) {
...
+  )
+}
+
+profile_frequency_table <- function(x, limit) {
+    x <- x[!is.na(x)]
+
+    if (length(x) == 0) {
+        return(list(
+            values = c(),
+            counts = c(),
```
lionel-, line 420: It seems clearer to explicitly use the `NULL` value? Esp. important to be clear that we return nullable values since these go through interop boundaries.

```suggestion
            values = NULL,
            counts = NULL,
```

crates/ark/src/data_explorer/histogram.rs
```diff
@@ -0,0 +1,579 @@
...
+            }
+        },
+        None => match r_typeof(column) {
+            INTSXP | REALSXP => {},
+            _ => return Err(anyhow!("Type not supported {:?}", r_typeof(column))),
+        },
+    }
+
+    let num_bins: RObject = match params.num_bins {
+        Some(v) => (v as i32).into(),
```
lionel-, line 52: Should we do these conversions with a tool that checks for boundaries? Not that it matters here but it seems like it'd be safer to get into a habit of checking bounds before truncating.

crates/ark/src/data_explorer/histogram.rs
```diff
@@ -0,0 +1,579 @@
...
+            bin_counts,
+            quantiles: vec![]
+        })
+    }
+
+    fn test_quantiles<T>(code: &str, quantiles: Vec<f64>, expected: T)
+    where
+        RObject: From<T>,
+    {
+        let column = r_parse_eval0(code, R_ENVS.global).unwrap();
```
lionel-, line 225: Might be worth adding `harp::parse_eval_global()`

lionel-: Done in https://github.com/posit-dev/ark/pull/480

By the way, if you don't need the global environments (to assign into or do search path lookups), better evaluate in the base env. This way readers don't have to wonder what the global env is used for and this eliminates a range of possible side effects that could contaminate other tests. I've also added `harp::parse_eval_base()` for this.

crates/ark/src/data_explorer/histogram.rs
```diff
@@ -0,0 +1,579 @@
...
+                "0:10",
+                4,
+                vec!["0.00", "2.00", "4.00", "6.00", "8.00", "10.00"],
+                vec![3, 2, 2, 2, 2],
+            );
+        })
+    }
+
+    #[test]
+    fn test_date_histogram() {
```
lionel-, line 285: Nice test util and great set of tests!

lionel- (approved):
Looks great!

crates/harp/src/object.rs
```diff
@@ -603,6 +604,18 @@ impl From<&Vec<i64>> for RObject {
     }
 }
 
+impl From<&Vec<f64>> for RObject {
+    fn from(values: &Vec<f64>) -> Self {
+        unsafe {
+            let vector = RObject::from(Rf_allocVector(REALSXP, values.len() as isize));
+            for idx in 0..values.len() {
+                SET_REAL_ELT(vector.sexp, idx as isize, values[idx] as c_double);
```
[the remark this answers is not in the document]
DavisVaughan: But I guess then we'd have to move from `from(values: &Vec<f64>)` to `try_from(values: &Vec<f64>)` so maybe we shouldn't worry about that right now.

(i do think maybe we should move to `try_from` in general, since allocations can always be a point of failure...)

DavisVaughan (approved):

crates/ark/src/modules/positron/r_data_explorer.R
```diff
@@ -346,3 +342,113 @@ write_html <- function(x, include_header) {
...
+  # use it, so it can be safely ignored.
+  suppressWarnings(h <- graphics::hist(x, breaks = breaks, plot = FALSE))
+
+  bin_edges <- h$breaks
+  bin_counts <- h$counts
+
+  # For dates, we convert back the breaks to the date representation.
+  if (inherits(x, "POSIXct")) {
+    bin_edges <- as.POSIXct(h$breaks, tz = attr(x, "tzone"))
+  }
```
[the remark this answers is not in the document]
dfalbel: Under the hood we'll call `format()` and dependning on the bin edges, it will add the seconds precision. But if bin_edges are round dates, it won't eg:

```
format(as.POSIXct(c("2010-01-01 00:00:00", "2010-01-02 00:00:00")))
```

But it's likely that it will include the seconds precision in mos cases.

crates/ark/src/modules/positron/r_data_explorer.R
```diff
@@ -346,3 +342,113 @@ write_html <- function(x, include_header) {
...
+  # using default methods.
+  # We force something considering the integer representation.
+  if (inherits(x, "POSIXct")) {
+    if (method == "sturges") {
+      num_bins <- grDevices::nclass.Sturges(x)
+    }
+
+    # The pretty bins algorithm doesn't really make sense for dates,
+    # so we generate our own bin_edges for them.
+    min_value <- min(x)
```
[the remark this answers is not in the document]
dfalbel: I was unaware of that, from my testing it seemed weird but I didn't go far investigating. I assumed they were simply ignoring pretty for dates.

crates/ark/src/modules/positron/r_data_explorer.R
```diff
@@ -346,3 +342,113 @@ write_html <- function(x, include_header) {
...
+
+  list(
+      bin_edges = bin_edges,
+      bin_counts = bin_counts,
+      quantiles = quantiles
+  )
+}
+
+profile_frequency_table <- function(x, limit) {
+    x <- x[!is.na(x)]
```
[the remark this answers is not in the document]
dfalbel: I think in this case we can show the non-finite values as elements in the freq table

dfalbel:
I have addressed most comments. Some pending discussions are: 

- Usage of `as i32` to cast from `i64` into an `RObject`. Since the specific usage in this PR is not problematic I didn't change it, but we can revisit in a future PR.
- Usage of `r_parse_eval0` in tests. Opened https://github.com/posit-dev/positron/issues/4497 to track and will do it once #480 is merged.

PR #467, opened 2024-08-08
Data Explorer: Make sure we update the data shape after a Data Update

dfalbel:
Addresses https://github.com/posit-dev/positron/issues/4286

When we verify if there was a data change in R, we check if we should send a SchemaChange or a DataUpdate event. 
When the schema changes, ie, types of columns or number of columns, we would correctly update the data shape.
For data updates we would skip this update, however it's necessary because even though the schema didn't change, it's possible that the number of rows has changed and we need to make sure the front-end knows about that.


juliasilge (approved):
I built ark from this PR and can observe the fix for changing the number of rows as outlined in the issue. This is a really simple fix so I feel confident approving. Thanks! 🙌

PR #468, opened 2024-08-09
Fix little bug in `.rs.api.viewer()`

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/4304

### DIFF
--- a/crates/ark/src/modules/rstudio/stubs.R
+++ b/crates/ark/src/modules/rstudio/stubs.R
@@ -45,11 +45,12 @@
     height <- .ps.validate.viewer.height(height)
 
     # Derive a title for the viewer from the path.
+    normalizedPath <- normalizePath(url, mustWork = FALSE)
     title <- .ps.viewer.title(normalizedPath)
 
     invisible(.Call("ps_html_viewer",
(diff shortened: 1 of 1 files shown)


crates/ark/src/modules/rstudio/stubs.R
```diff
@@ -45,11 +45,12 @@
     height <- .ps.validate.viewer.height(height)
 
     # Derive a title for the viewer from the path.
+    normalizedPath <- normalizePath(url, mustWork = FALSE)
     title <- .ps.viewer.title(normalizedPath)
 
     invisible(.Call("ps_html_viewer",
         url,     # The URL of the file to view
```
DavisVaughan, line 52:
```suggestion
    url <- normalizePath(url, mustWork = FALSE)
    
    # Derive a title for the viewer from the path.
    title <- .ps.viewer.title(url)

    invisible(.Call("ps_html_viewer",
        url,     # The URL of the file to view
```

The main advantage of my tweak is that we now also pass the normalized `url` to Rust as well, which feels right, and is what we've done elsewhere too

DavisVaughan (approved):

PR #471, opened 2024-08-14
Data Explorer: Avoid warnings when summarizing columns

dfalbel:
Addresses: https://github.com/posit-dev/positron/issues/4353

It fixes two separate issues:

- We actually want to return `None` when there are no statistics to be computed for numeric values. (Related to https://github.com/posit-dev/positron/issues/4352). This also avoids the warnings that appear in https://github.com/posit-dev/positron/issues/4353

- For datetimes containing invalid timezones such as `x <- as.POSIXct(c("2010-01-01 00:00:00"), tz = "+01:00")`, computing `min` or `max` will raise a warning, so we suppress those.

### DIFF
--- a/crates/ark/src/modules/positron/r_data_explorer.R
+++ b/crates/ark/src/modules/positron/r_data_explorer.R
@@ -45,6 +45,14 @@
 }
 
 summary_stats_number <- function(col) {
+
+    col <- col[!is.na(col)]
+
+    # Don't compute stats if the column is all NA's or empty.
+    if (length(col) == 0) {
+        return(numeric(0))
+    }
+
     c(
         min_value = min(col, na.rm = TRUE),
         max_value = max(col, na.rm = TRUE),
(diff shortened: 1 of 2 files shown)


crates/ark/src/modules/positron/r_data_explorer.R
```diff
@@ -63,13 +71,19 @@ summary_stats_boolean <- function(col) {
...
-    list(
-        min_date = as.character(min(col, na.rm = TRUE)),
-        mean_date = as.character(mean(col, na.rm = TRUE)),
-        median_date = as.character(stats::median(col, na.rm = TRUE)),
-        max_date = as.character(max(col, na.rm = TRUE)),
-        num_unique = length(unique(col))
-    )
+    # We have to suppress warnings here because malformed datetimes, eg:
+    # x <- as.POSIXct(c("2010-01-01 00:00:00"), tz = "+01:00")
+    # When calling `min` on `x` woudl raise a warnings.
```
softwarenerd, line 76: woudl -> would

softwarenerd (approved):
LG!

PR #473, opened 2024-08-16
Async callbacks for processing GetColumnProfile RPC requests

dfalbel:
This PR pairs with https://github.com/posit-dev/positron/pull/4326
And should be merged after #458

This makes `GetColumnProfile` requests asycronous giving the data explorer execution thread oportunities to compute other RPC methods, such as GetDataValues which should make the data explorer UX a little smoother.


lionel-:
@dfalbel I changed the base to `feature/histograms-frequency-tables` but I still get unrelated diffs. I'll review nonetheless using this commit as a guideline since it seems that's the only relevant one here: [Use async callbacks allowing other RPC methods to execute between get…](https://github.com/posit-dev/ark/pull/473/commits/2fb74ba277448a98ea347436b7495f0b393c6a5b)

dfalbel:
@lionel- I rebased on the histograms branch and It's now correct. 
But yes, https://github.com/posit-dev/ark/pull/473/commits/985a7d98cd5cb3ece103275ea0ef40a5d8d30c4d is the only relevant commit, the other commit is just fixing tests.

lionel- (commented):
Nice first stab!

I think we could make it simpler and safer by using the new idle task mechanism `r_task::spawn_idle()`:

- These tasks run on the R thread when R is idle instead of at interrupt time, which is the direction we'd like to take. It's also consistent with the data explorer being an idle-only feature (you can't use it when the kernel is busy on a shiny app or computing an expensive model).

- These tasks take rust async functions, which make it easy to yield for a tick, fail on timeout etc.

- They could own the `callback_id` and send the result back to the data explorer thread in the form of a response event. This way we won't need to spawn any threads or create any temporary channels. We just need a new permanent channel to transmit events to the data explorer thread, the receiving side of which would be handled by the data explorer event loop.

- The tasks may get delayed by an incoming execute request. This means there's also a possibility of sending outdated summaries if the data was updated, but I guess the frontend should handle that case. There should be no possibility of piling up idle tasks while R is busy since the shell socket is busy as well and can't take any new data explorer requests.

crates/ark/src/data_explorer/column_profile.rs
```diff
@@ -0,0 +1,289 @@
...
+    pub request: GetColumnProfilesParams,
+}
+
+pub async fn handle_columns_profiles_requests(
+    params: ProcessColumnsProfilesParams,
+    comm: CommSocket,
+) -> anyhow::Result<()> {
+    let event = process_columns_profiles_requests(params).await?;
+    let json_event = serde_json::to_value(event)?;
+    comm.outgoing_tx.send(CommMsg::Data(json_event))?;
```
lionel-, line 49: I'm still confused that we don't reply in case of errors, since there is a callback ID I would expect the frontend is keeping track of it. But I guess this is cleaned up when it no longer needs the answer because the data has changed? (I haven't looked into the frontend code yet)

crates/ark/src/data_explorer/r_data_explorer.rs
```diff
@@ -336,28 +333,37 @@ impl RDataExplorer {
...
-            if new == old {
-                None
+            let old = self.table.get();
+            let old = unwrap!(old, Err(_) => {
+                // This is AFAICT impossible because the table is only deleted when the data explorer instance is
+                // deleted and this method belongs to that data explorer instance.
+                log::error!("Old table has been deleted? This is unexpected, but we'll update the data explorer table.");
+                // It's `unsafe` because we RObject::new calls protect, and it shouldn't
+                // be called outside of the R main thread.
+                self.table.set(unsafe { RObject::new(new) });
```
lionel-, line 343: Can we change `set()` to take a thread safe object so it's clear what's going on?

crates/ark/src/data_explorer/table.rs
```diff
@@ -0,0 +1,61 @@
...
+use once_cell::sync::Lazy;
+
+use crate::thread::RThreadSafe;
+
+// Stores the table R objects objects that contain the data for each
+// data explorer instance.
+// This allows for background threads to easilly access the
+// instance related data without having to rely on the lifetime
+// of data explorer execution thread.
+// Since this is a DashMap, it's safe to access it's underlying data from
```
lionel-, line 20: Instead of a dashmap stored in a global, could we use an `Arc<Mutex<HashMap>>`? This should solve the lifetime concerns?

dfalbel:
Updated to use `spawn_idle`:

The most challending thing was that the task in `spawn_idle` can live longer than the data.frame we are currently visualizing, so the Rust compiler won't allows us to move that object to a different thread. It's also not safe to clone the table, in the rpc handler thread because it requires calling the R API. So similarly to the `VDOCS` table, we now store references to the currently visualized table in a global `DATA_EXPLORER_TABLES` `DashMap` and we created a wrapper `Table` object that abstracts away how the access to the RObject. It keep the requirements from `RThreadSafe<>` and can only be accessed from the main R thread.

Moved all profile handling functions to a separate file as they can no longer depend on `self`, since they are called from a thread that might outlive `self`.

I also had to change tests, because since there's no spawned thread in the tests, the tasks are now running synchronously and thus the messages get out of order.

crates/ark/src/data_explorer/column_profile.rs
```diff
@@ -0,0 +1,289 @@
...
+) -> anyhow::Result<DataExplorerFrontendEvent> {
+    let GetColumnProfilesParams {
+        callback_id,
+        profiles,
+        format_options,
+    } = params.request;
+    let span = tracing::trace_span!("get_profile", ns = callback_id);
+
+    // This is an R thread, so we can actually get the data frame.
+    // If it fails we quickly return an empty result set and end the task.
```
lionel-, line 64: Needs a discussion of how this can happen, i.e. when the data explorer instance has been dropped. This is a way of cancelling the task in that case.

crates/ark/src/data_explorer/table.rs
lionel-: Or passing around `Arc<Mutex<Option<RThreadSafeObject<RObject>>>>` as we discussed on Slack, so that an instance can set it to `None` when it's dropped, and getting a `None` would be a cancellation point for the task.

lionel- (commented):
Good progress! Some more changes needed as this is tricky.

crates/ark/src/r_task.rs
```diff
@@ -253,7 +253,7 @@ where
     Fut: Future<Output = ()> + 'static,
 {
     // Idle tasks are always run from the read-console loop
-    if !only_idle && unsafe { R_TASK_BYPASS } {
+    if unsafe { R_TASK_BYPASS } {
```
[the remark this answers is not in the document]
dfalbel: :P

I forgot I had to change that to make tests pass. In theory this only affects tests because of `R_TASK_BYPASS` and all tests seem to be passing normally.

If I don't add this, then the tests get deadlock as there's no R main thread to actually watch the task queue and execute them.

crates/ark/src/data_explorer/column_profile.rs
dfalbel: Yeah, currently the RPC contract does not implement any error handling, so if it fails and we never respond, all the front-end will see is a timeout. 

I added some handling here, so basically, when it fails we will return an empty result set:

https://github.com/posit-dev/ark/blob/56726b7ae1ceda0224efdda20435c2bc70adac25/crates/ark/src/data_explorer/column_profile.rs#L57-L66

We already return empty profiles if we fail to compute profiles, thus it seems reasonable to also do that if everything fails. 

We can add error handling too, we would need make `ReturnColumnsProfile` an enum so we would return either the actual profiles or an error message, but I'd rather add this in another PR because it will also require changes in the python backend.

crates/ark/src/data_explorer/table.rs
dfalbel: This worked nice! So got rid of the global dashmap, and lifetime is now managed with `Arc`. Failing to `get()` from the table means that the table has been deleted and so it can be used to cancel tasks in case the data explorer is closed, although we don't really use it for now.

We would probably also want to implement a cancelling mechanism based on the `callback_id`, so the front-end can cancel tasks it will no longer need. But we can leave this for another PR.

crates/ark/src/data_explorer/column_profile.rs
dfalbel: I created https://github.com/posit-dev/positron/issues/4559 to track this issue.

crates/ark/src/data_explorer/table.rs
lionel-: oh I meant cancelling the idle task, which does happen already via the propagated `get()` error, e.g. your comment

```rust
        // In case something goes wrong whhile computing the profiles, we send
        // an empty response. Ideally, we would have a way to comunicate an that
        // an error happened but it's not implemented yet.
```

crates/ark/src/data_explorer/column_profile.rs
```diff
@@ -0,0 +1,307 @@
...
+            profile_column(
+                data.clone(),
+                indices.clone(),
+                profile,
+                &format_options,
+                kind,
+            )
+            .await,
+        );
+        // Yield to the R event loop
```
lionel-, line 104:
```suggestion
        // Yield to the idle event loop
```

By R event loop we generally mean the one that runs at interrupt time.

crates/ark/src/data_explorer/column_profile.rs
```diff
@@ -0,0 +1,307 @@
...
+        tokio::task::yield_now().await;
+    }
+
+    Ok(results)
+}
+
+// This function does not return a Result because it must handle still handle other profile types
+// if one of them fails. Thus it needs to gracifully handle the errors that might have resulted
```
lionel-, line 112:
```suggestion
// if one of them fails. Thus it needs to gracefully handle the errors that might have resulted
```

crates/ark/src/data_explorer/column_profile.rs
```diff
@@ -0,0 +1,307 @@
...
+                match profile_req.profile_type {
+                    ColumnProfileType::SmallFrequencyTable => {
+                        output.small_frequency_table = frequency_table;
+                    },
+                    ColumnProfileType::LargeFrequencyTable => {
+                        output.large_frequency_table = frequency_table;
+                    },
+                    _ => {
+                        // This is technically unreachable!(), but not worth panicking if
+                        // this happens.
```
lionel-, line 208: Warn?

crates/ark/src/data_explorer/column_profile.rs
lionel-: Oh you referred to "R console loop" below which is good too.

crates/ark/src/data_explorer/column_profile.rs
```diff
@@ -0,0 +1,307 @@
...
+    let profiles = process_columns_profiles_requests(
+        params.table,
+        params.indices,
+        params.kind,
+        params.request.profiles,
+        params.request.format_options,
+    )
+    .await
+    .unwrap_or_else(|e| {
+        // In case something goes wrong whhile computing the profiles, we send
```
lionel-, line 59:
```suggestion
        // In case something goes wrong while computing the profiles, we send
```

lionel- (approved):
Looks good!

PR #474, opened 2024-08-16
Switch summary stats to supported

softwarenerd:
This PR switches `ColumnProfileType::SummaryStats` from `SupportStatus::Experimental` to `SupportStatus::Supported` so that users will see the column summary stats again.


dfalbel (approved):
LGTM!

PR #477, opened 2024-08-20
Avoid warning with invalid regex filter in Data Explorer

dfalbel:
Addresses: https://github.com/posit-dev/positron/issues/4392
The error was caused because `grepl` in R will show a warning and an error when an invalid regex appears.
We already catch the error, but the warning was leaking to the console.

QA: Following the issue description should not show any console warning any longer.


juliasilge (approved):
I built ark from this PR and observe the fix that I no longer get extraneous warnings in the console from a bad regex. This is a straightforward fix so I feel confident approving. Thanks! 🚀

PR #478, opened 2024-08-20
Data Explorer: Add support for factors summary stats

dfalbel:
Addresses https://github.com/posit-dev/positron/issues/2161

Without this PR we fail to compute stats for factors. They are treated as strings by the front-ent, thus they get into the codepath of calling `summary_stats_string()` and `nzchar` requires a character vector.

### DIFF
--- a/crates/ark/src/data_explorer/summary_stats.rs
+++ b/crates/ark/src/data_explorer/summary_stats.rs
@@ -248,6 +248,20 @@ mod tests {
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


juliasilge (approved):
I built ark from this PR and observe that I no longer see "N/A" empty and unique for factors:

Screenshot 2024-08-20 at 2 12 37 PM

This is a straightforward fix so I feel confident approving it. Thanks!

PR #482, opened 2024-08-22
Add `SrcRef` type and create srcrefs with `parse_status()`

lionel-:
Branched from #481.

Add `SrcRef` type and enable parsing with srcrefs in `parse_status()`. We now use this in `parse_exprs_with_srcrefs()` instead of calling back into the R-level `parse()`. Also adds a bunch of tests to this previously untested area.

Will be useful to write tools that detect expression boundaries in R code. Progress towards posit-dev/positron#1326

Edit: `srcref` is now a submodule of `parser`. I'll move `parse.rs` and `source.rs` to that submodule outside of this PR.


crates/harp/src/srcref.rs
```diff
@@ -0,0 +1,152 @@
...
+
+            assert_eq!(foo.line, Range { start: 0, end: 0 });
+            assert_eq!(foo.line_parsed, Range { start: 0, end: 0 });
+            assert_eq!(foo.column, Range { start: 0, end: 2 });
+            assert_eq!(foo.column_byte, Range { start: 0, end: 2 });
+
+            // `column_byte` is different because the character takes up two bytes
+            assert_eq!(utf8.line, Range { start: 2, end: 2 });
+            assert_eq!(utf8.line_parsed, Range { start: 2, end: 2 });
+            assert_eq!(utf8.column, Range { start: 0, end: 0 });
```
lionel-, line 125: I wonder if we should bump the column-end fields by 1. These empty ranges don't make much sense to me.

crates/harp/src/srcref.rs
```diff
@@ -0,0 +1,152 @@
...
+
+        let column = std::ops::Range {
+            start: (value.get_value(4)? - 1) as usize,
+            end: value.get_value(5)? as usize,
+        };
+
+        let column_byte = std::ops::Range {
+            start: (value.get_value(1)? - 1) as usize,
+            end: value.get_value(3)? as usize,
+        };
```
DavisVaughan, line 81: Does `end: value.get_value(5)?` and `end: value.get_value(3)?` need a `-1`? 

If not, I think we need some comments in here about the adjustments made to the structure of the ranges, and what we expect the result to look like, i.e. is the resulting range `[ ]` or `[ )` ?

crates/harp/src/parse.rs
```diff
@@ -43,60 +50,57 @@ pub fn parse(code: &str) -> crate::Result<RObject> {
...
-        // Because `parse(text =)` doesn't allow `\r\n` even on Windows
-        let code = convert_line_endings(code, LineEnding::Posix);
-        let code = r_string!(code, protect);
+    parse_exprs_ext(code, RParseOptions { srcref: true })
+}
 
-        RFunction::new("base", "parse")
-            .param("text", code)
-            .param("keep.source", true)
-            .call()
```
DavisVaughan, line 71: I also went through and confirmed that it looks like `parse(text = code, keep.source = TRUE)` is doing pretty much the exact same thing as our lower level code, i.e.:
- Calling `srcfilecopy()` to get the srcfile https://github.com/wch/r-source/blob/3052a2df21c794ead804f3960d10618ec10e90e1/src/library/base/R/parse.R#L27-L29
- Handing off to C level `R_ParseVector()` without much extra really https://github.com/wch/r-source/blob/3052a2df21c794ead804f3960d10618ec10e90e1/src/main/source.c#L286

The only difference I see between this and calling `parse()` directly is that `parse()` also fiddles with these global variables around encoding, which I think affect how `R_ParseVector()` works, but hopefully that doesn't really matter
https://github.com/wch/r-source/blob/3052a2df21c794ead804f3960d10618ec10e90e1/src/main/source.c#L240-L255

DavisVaughan (approved):

crates/harp/src/srcref.rs
lionel-: The resulting range is the conventional `[ )`

lionel-:
@DavisVaughan Can you take another quick look please?

* I've moved the new files to a `parser` submodule. I plan to move `parse.rs` and `source.rs` there outside of this PR.

* The srcfile is no longer a parse option but a variant of input. I structured things this way so that srcfile is owned by the caller, which allows it to retrieve parse data.

* I renamed `line` to `line_virtual` and `line_parsed` to `line`.

* The line fields are now `[ )` ranges like the column fields. It's important not only for internal consistency but also because it's expected by methods of `std::ops::Range`.

PR #486, opened 2024-08-28
Compile portably for x86_64 Linux

lionel-:
Addresses https://github.com/posit-dev/positron/issues/4506
Addresses the ark side of https://github.com/posit-dev/positron/issues/3854

Solves linking issues for:

- Open SSL by statically linking to libssl. See https://github.com/posit-dev/positron/issues/3854#issuecomment-2226860016 for an example of linking issue. Static linking is supported by the `openssl-sys` crate, though it's slightly tricky as it expects us to provide the location of the static lib and headers in a single folder. The system installation of libssl on Ubuntu puts headers in two folders, so we merge them as a workaround.

  Another approach supported by `openssl-sys` is to compile Open SSL as part of the build ("vendored" crate feature), but I had trouble getting it to work. Might be worth looking into further for arm64 support though.

- GNU libc by using the zig linker with https://github.com/rust-cross/cargo-zigbuild. This linker supports linking to specific versions of glibc: https://github.com/ziglang/glibc-abi-tool. We use this to target glibc 2.26, which is sufficiently old to run on RHEL8 or OpenSUSE 15.2.

Using these approaches we now get a pretty portable build of ark.

I thought that we'd also have to solve linking issues with the C++ runtime but that doesn't seem to be the case. In fact I no longer see libstdc++ in the output of `ldd`, which I find very surprising:

```
ldd ark
	linux-vdso.so.1 (0x00007fffd3d71000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f6f95b19000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f6f973a1000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f6f95800000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f6f9739c000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f6f973b3000)
```

I verified that the ark binary produced by https://github.com/posit-dev/ark/actions/runs/10591447650 works on the following fuzzbucket platforms:

```sh
fuzzbucket-client create ubuntu20-ide-prereqs-ci
fuzzbucket-client create ubuntu22-ide-prereqs
fuzzbucket-client create ubuntu24-ide-prereqs-ci
fuzzbucket-client create rhel8-ide-prereqs
fuzzbucket-client create rhel9-ide-prereqs
fuzzbucket-client create opensuse15.2
fuzzbucket-client create opensuse15.5-ide-prereqs-ci
```

Testing method: Connecting with jupyter-console and running `rnorm(1)`.

I removed the debug build from our matrix because statically linking to openssl with debug symbols causes the size of the build to explode to over 230mb. If debug symbols are needed, getting a local build of ark going is a simple matter of `git clone` + `rustup` + `cargo build`, which seems reasonably easy?

### DIFF
--- a/.github/workflows/release-linux.yml
+++ b/.github/workflows/release-linux.yml
@@ -25,7 +25,7 @@ jobs:
         strategy:
             matrix:
                 arch: [x64]
-                flavor: [release]
+                flavor: [debug, release]
 
         steps:
             - name: Checkout sources
@@ -41,24 +41,9 @@ jobs:
                   sudo apt install python3-pip
                   pip3 install ziglang
 
-                  # Install static OpenSSL
-                  sudo apt-get install -y libssl-dev
-
-                  # The openssl-sys crate expects to see all headers in one place but
-                  # Ubuntu puts configured headers in an arch-specific folder. We
-                  # merge these folders as a workaround.
-                  sudo cp /usr/include/x86_64-linux-gnu/openssl/* /usr/include/openssl
-
             - name: Compile ARK (${{ matrix.flavor }})
               run: |
                   cargo clean
-
-                  # Instruct the openssl-sys crate to link to libssl statically.
-                  # This is important for portability because distributions have
-                  # various requirements for their dynamic libssl libraries.
-                  export OPENSSL_STATIC=yes
-                  export OPENSSL_LIB_DIR=/usr/lib/x86_64-linux-gnu
-                  export OPENSSL_INCLUDE_DIR=/usr/include
 
                   # Use the zig linker. This allows linking to a specific version of glibc.
                   # We use a sufficiently old version that is available on all platforms we target.
--- a/.github/workflows/release-linux.yml
+++ b/.github/workflows/release-linux.yml
@@ -19,11 +19,13 @@ jobs:
         env:
             GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
             DEBUG_FLAG: ${{ matrix.flavor == 'debug' && '-debug' || '' }}
+            TARGET_FLAG: ${{ matrix.flavor == 'release' && '--release' || '' }}
+            GLIBC_MAX_VERSION: '2.26'  # Sufficiently old for all our target platforms
 
         strategy:
             matrix:
                 arch: [x64]
-                flavor: [debug, release]
+                flavor: [release]
 
         steps:
             - name: Checkout sources
@@ -34,16 +36,40 @@ jobs:
                   sudo apt-get update
                   sudo apt-get install -y cargo
 
+                  # We're linking with zig to select the libc version
+                  cargo install --locked cargo-zigbuild
+                  sudo apt install python3-pip
+                  pip3 install ziglang
+
+                  # Install static OpenSSL
+                  sudo apt-get install -y libssl-dev
+
+                  # The openssl-sys crate expects to see all headers in one place but
+                  # Ubuntu puts configured headers in an arch-specific folder. We
+                  # merge these folders as a workaround.
+                  sudo cp /usr/include/x86_64-linux-gnu/openssl/* /usr/include/openssl
+
             - name: Compile ARK (${{ matrix.flavor }})
               run: |
                   cargo clean
-                  cargo build ${{ matrix.flavor == 'release' && '--release' || '' }}
+
+                  # Instruct the openssl-sys crate to link to libssl statically.
+                  # This is important for portability because distributions have
+                  # various requirements for their dynamic libssl libraries.
+                  export OPENSSL_STATIC=yes
+                  export OPENSSL_LIB_DIR=/usr/lib/x86_64-linux-gnu
+                  export OPENSSL_INCLUDE_DIR=/usr/include
+
+                  # Use the zig linker. This allows linking to a specific version of glibc.
+                  # We use a sufficiently old version that is available on all platforms we target.
+                  # See https://github.com/ziglang/glibc-abi-tool
+                  cargo zigbuild --target x86_64-unknown-linux-gnu.$GLIBC_MAX_VERSION $TARGET_FLAG
 
             # Compress kernel to a zip file
             - name: Create archive
               run: |
                   # Enter the build directory
-                  pushd target/${{ matrix.flavor }}
+                  pushd target/x86_64-unknown-linux-gnu/${{ matrix.flavor }}
 
                   # Compress the kernel to an archive
                   ARCHIVE="$GITHUB_WORKSPACE/ark-${{ inputs.version }}-${{ matrix.flavor }}-linux-x64.zip"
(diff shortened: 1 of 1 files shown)


.github/workflows/release-linux.yml
```diff
@@ -19,11 +19,13 @@ jobs:
         env:
             GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
             DEBUG_FLAG: ${{ matrix.flavor == 'debug' && '-debug' || '' }}
```
DavisVaughan, line 21: Remove `DEBUG_FLAG`? Seems out of date

kevinushey:
IMHO static linking to OpenSSL and libc is a bad idea, unless we're certain it's trivial to update ark (e.g. if a user of Positron / ark needs to update in response to security vulnerabilities discovered in OpenSSL).

jmcphers:
This is super cool, but unfortunately I agree with @kevinushey. A portable build would be really convenient, but becoming an OpenSSL distributor (by statically linking it) would be a nightmare. https://openssl-library.org/news/vulnerabilities/index.html

Here are four other routes we could take:

- Look up OpenSSL symbols dynamically like we do for R. May or may not be possible; I'm aware that most OpenSSL usage may be in crates we do not own. We've discussed doing this for RStudio for years.
- Attempt to remove OpenSSL as a dependency entirely. Ark doesn't have a lot of crypto or use any TLS/SSH/etc. features. Maybe we could use alternatives for the few crypto-related features we have that don't require OpenSSL.
- Produce separate builds of Ark that link against the appropriate OpenSSL for each Linux distribution (minimally by ABI compatibility
- [Run away from home and live in the woods](https://www.wikihow.com/Run-Away-from-Home-and-Live-in-the-Woods)

Producing separate builds is probably the easiest way forward and it's what we have always done to fix this problem. But I think it would also be worth investigating where the dependency is coming from and seeing if there are alternatives.

DavisVaughan:
@jmcphers re bullet 2, like this? https://github.com/posit-dev/ark/pull/487

lionel-:
oh I had assumed that ssl was used somewhere in our transport stack...

I just found out it's very easy to analyse dependency chains:

```
% cargo tree --target all --no-dedupe -i openssl-sys
openssl-sys v0.9.93
├── native-tls v0.2.11
│   ├── hyper-tls v0.5.0
│   │   └── reqwest v0.11.22
│   │       └── ark v0.1.126 (/Users/lionel/Sync/Projects/Positron/ark/crates/ark)
│   ├── reqwest v0.11.22
│   │   └── ark v0.1.126 (/Users/lionel/Sync/Projects/Positron/ark/crates/ark)
│   └── tokio-native-tls v0.3.1
│       ├── hyper-tls v0.5.0
│       │   └── reqwest v0.11.22
│       │       └── ark v0.1.126 (/Users/lionel/Sync/Projects/Positron/ark/crates/ark)
│       └── reqwest v0.11.22
│           └── ark v0.1.126 (/Users/lionel/Sync/Projects/Positron/ark/crates/ark)
└── openssl v0.10.57
    └── native-tls v0.2.11
        ├── hyper-tls v0.5.0
        │   └── reqwest v0.11.22
        │       └── ark v0.1.126 (/Users/lionel/Sync/Projects/Positron/ark/crates/ark)
        ├── reqwest v0.11.22
        │   └── ark v0.1.126 (/Users/lionel/Sync/Projects/Positron/ark/crates/ark)
        └── tokio-native-tls v0.3.1
            ├── hyper-tls v0.5.0
            │   └── reqwest v0.11.22
            │       └── ark v0.1.126 (/Users/lionel/Sync/Projects/Positron/ark/crates/ark)
            └── reqwest v0.11.22
                └── ark v0.1.126 (/Users/lionel/Sync/Projects/Positron/ark/crates/ark)
```

lionel-:
I had just assumed that ark, by way of running R code, is a security gap in itself.

@jmcphers @kevinushey Do the security concerns apply equally to libc? What about all the Rust dependencies statically linked to ark?

My main question is whether we can use the zigbuild trick to link to old GNU libc versions?

lionel-:
- Now branched from @DavisVaughan's branch. We no longer depend on libssl at all.

- With that out of the way it was easy to cross-compile for arm, adresses posit-dev/positron#4507.

This still links to an old GNU libc. If that's an issue, we could produce a bunch of builds for the various libc we still support.

jmcphers (approved):
Built locally on macOS and Linux, nice!

I'm not too familiar with zigbuild and the libc implications. Does this wind up statically linking libc? (don't want to do that if we can avoid it, libc also regularly needs patching)

lionel-:
@jmcphers It's dynamically linked but against our target version of libc. So it does pretty much what we need it to AFAICT.

jmcphers:
great!

lionel-:
If you're curious how it works, see the glibc section in https://andrewkelley.me/post/zig-cc-powerful-drop-in-replacement-gcc-clang.html (the whole post is worth a read).

---

Now I was wondering why libstdc++ disappeared from the `ldd` output. It turns out that zig is not only used for linking together compiled object files, it's also used to compile C and C++ files. This means zeromq is now compiled by zig via llvm.

While glibc is linked dynamically for C code, by default C++ code gets statically linked with llvm-libcxx to avoid ABI issues. There is a way to dynamically link to libstdc++ instead, but:

- This might be tricky to set up via `cargo zigbuild`.
- We might get into ABI compat issues.
- We already statically link to the Rust runtime and all our Rust dependencies, so surely we're way more exposed via the Rust stack. But there might be process/political reasons why libstdc++ is considered more critical?

@jmcphers What do you think? Do we need to find another way?

lionel-:
A search for libcxx vulnerabilities shows very few hits compared to glibc, so it seems much safer to distribute as part of ark (not that we could statically link to glibc if we wanted to, just using it as a point of comparison).

> Regarding the licence, it's apache 2.0 with restrictions, but it seems fine:
https://llvm.org/docs/DeveloperPolicy.html#new-llvm-project-license-framework
In particular:
Binaries that include LLVM must reproduce the copyright notice (e.g. in an included README file or in an “About” box), unless the LLVM code was added as a by-product of compilation. For example, if an LLVM runtime library like compiler_rt or libc++ was automatically included into your application by the compiler, you do not need to attribute it.

See also MS comments in https://devblogs.microsoft.com/cppblog/open-sourcing-msvcs-stl:

> As a customer of MSVC’s STL, you might be wondering whether this new license creates new obligations for you. Microsoft’s position is that the text of the Apache License v2.0 with LLVM Exceptions (specifically, the wording of the LLVM Exceptions) clearly states that when you compile your own source code using MSVC’s STL to produce object code or similar output (e.g. static libraries, dynamic libraries, and executables), you aren’t required to provide attribution when shipping your compiled products to your end users. This is another reason we’ve chosen this license: to avoid disrupting our customers’ businesses.

In light of this, we've decided to go with the zig approach as that allows us to compile portably for all supported distributions on both x86_64 and arm64.

PR #489, opened 2024-08-28
Add retries to Help proxy request

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/3753

Folks in that issue ⤴️ are experiencing problems seeing R Help pages in the Help pane. Some things we can notice:

- None of us internally have been able to reproduce the problem.
- Some folks _do_ occasionally see the Help page they are querying for! Seems like an intermittent problem.
- We see errors in logs like "operation was canceled: connection closed before message completed".
- Some folks have their problems totally solved after installing a new version of Positron.

Given all that, seems like we may just need to retry the request, or at the least that is a first possible step as outlined by @DavisVaughan in https://github.com/posit-dev/positron/issues/3753#issuecomment-2245984690.

### DIFF
--- a/crates/ark/Cargo.toml
+++ b/crates/ark/Cargo.toml
@@ -41,7 +41,9 @@ notify = "6.0.0"
 once_cell = "1.17.1"
 parking_lot = "0.12.1"
 regex = "1.10.0"
-reqwest = { version = "0.11.20", features = ["json"] }
+reqwest = { version = "0.12.5", features = ["json"] }
+reqwest-retry = "0.6.1"
+reqwest-middleware = "0.3.3"
 ropey = "1.6.0"
 rust-embed = "8.0.0"
 scraper = "0.15.0"
--- a/crates/ark/src/help_proxy.rs
+++ b/crates/ark/src/help_proxy.rs
@@ -17,6 +17,10 @@ use actix_web::HttpServer;
 use harp::exec::RFunction;
 use harp::exec::RFunctionExt;
 use mime_guess::from_path;
+use reqwest::Client;
+use reqwest_middleware::ClientBuilder;
+use reqwest_retry::policies::ExponentialBackoff;
+use reqwest_retry::RetryTransientMiddleware;
 use rust_embed::RustEmbed;
 use serde::Deserialize;
 use stdext::spawn;
@@ -128,8 +132,14 @@ async fn proxy_request(req: HttpRequest, app_state: web::Data<AppState>) -> Http
     // Add query from original request back to URL.
     target_url.set_query(Some(query));
 
+    // Set up reqwest client with 3 retries.
+    let retry_policy = ExponentialBackoff::builder().build_with_max_retries(3);
+    let client = ClientBuilder::new(Client::new())
+        .with(RetryTransientMiddleware::new_with_policy(retry_policy))
+        .build();
+
     // Get the target URL.
-    match reqwest::get(target_url.clone()).await {
+    match client.get(target_url.clone()).send().await {
         // OK.
         Ok(response) => {
             // We only handle OK. Everything else is unexpected.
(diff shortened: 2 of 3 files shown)


crates/ark/src/help_proxy.rs
```diff
@@ -242,3 +253,22 @@ async fn preview_img(params: web::Query<PreviewRdParams>) -> HttpResponse {
...
+    let out = actix_web::http::header::HeaderValue::from_bytes(x.as_bytes());
+
+    // We've checked these are the same underlying structure, so assert that this works.
+    let mut out = out.unwrap();
+
+    // Set the one other field that defaults to `false`, just in case.
+    out.set_sensitive(x.is_sensitive());
+
+    out
+}
```
juliasilge, line 274: Joint work with @DavisVaughan who discovered that:

- actix-web re-exports http 0.2.7's `HeaderValue`
- reqwest re-exports http 1.0.0's `HeaderValue`

Here is the commit where reqwest changed this:
https://github.com/seanmonstar/reqwest/commit/eb94f26919e881177ad3d9cd172f47d1a8263799

crates/ark/src/help_proxy.rs
```diff
@@ -128,8 +132,14 @@ async fn proxy_request(req: HttpRequest, app_state: web::Data<AppState>) -> Http
     // Add query from original request back to URL.
     target_url.set_query(Some(query));
 
+    // Set up reqwest client with 3 retries.
+    let retry_policy = ExponentialBackoff::builder().build_with_max_retries(3);
+    let client = ClientBuilder::new(Client::new())
+        .with(RetryTransientMiddleware::new_with_policy(retry_policy))
+        .build();
+
```
juliasilge, line 140: Shall we start with 3 retries?

crates/ark/src/help_proxy.rs
```diff
@@ -17,6 +17,10 @@ use actix_web::HttpServer;
 use harp::exec::RFunction;
 use harp::exec::RFunctionExt;
 use mime_guess::from_path;
+use reqwest::Client;
+use reqwest_middleware::ClientBuilder;
+use reqwest_retry::policies::ExponentialBackoff;
+use reqwest_retry::RetryTransientMiddleware;
```
juliasilge, line 23: The reqwest crate doesn't have any retry mechanism, but we can use these related crates to set up middleware and retries.

crates/ark/Cargo.toml
```diff
@@ -41,7 +41,9 @@ notify = "6.0.0"
 once_cell = "1.17.1"
 parking_lot = "0.12.1"
 regex = "1.10.0"
-reqwest = { version = "0.11.20", features = ["json"] }
+reqwest = { version = "0.12.5", features = ["json"] }
+reqwest-retry = "0.6.1"
+reqwest-middleware = "0.3.3"
```
DavisVaughan, line 46: Good news is that both reqwest-retry and reqwest-middleware are smart enough to set `default-features = false` for their reqwest dependency
https://github.com/TrueLayer/reqwest-middleware/blob/6cd4eb93f8567bece4526f050cd6f7c2bde7371e/reqwest-middleware/Cargo.toml#L24
https://github.com/TrueLayer/reqwest-middleware/blob/6cd4eb93f8567bece4526f050cd6f7c2bde7371e/reqwest-retry/Cargo.toml#L19

Making #487 still viable even after this

crates/ark/src/help_proxy.rs
```diff
@@ -128,8 +132,14 @@ async fn proxy_request(req: HttpRequest, app_state: web::Data<AppState>) -> Http
     // Add query from original request back to URL.
     target_url.set_query(Some(query));
 
+    // Set up reqwest client with 3 retries.
```
DavisVaughan, line 135:
```suggestion
    // Set up reqwest client with 3 retries (posit-dev/positron#3753).
```

crates/ark/src/help_proxy.rs
```diff
@@ -128,8 +132,14 @@ async fn proxy_request(req: HttpRequest, app_state: web::Data<AppState>) -> Http
...
+    let retry_policy = ExponentialBackoff::builder().build_with_max_retries(3);
```
DavisVaughan, line 136: For the curious, here are the other default options that get set for the backoff:

https://docs.rs/retry-policies/0.4.0/src/retry_policies/policies/exponential_backoff.rs.html#148-151

Definitions
https://docs.rs/reqwest-retry/latest/reqwest_retry/policies/struct.ExponentialBackoff.html#fields

DavisVaughan (approved):
Seems right to me

PR #490, opened 2024-08-28
Transfer combined `path_and_query()` all at once

DavisVaughan:
CC @juliasilge, sorry I didn't get to look at your PR earlier. Turns out you can do this path and query dance all at once


juliasilge (approved):
Looks good! 🙌 

I built ark from this PR and help requests both like `?lm` and like `??lm` continue to work as expected.

PR #491, opened 2024-08-29
Upkeep for release builds

lionel-:
- Move release builds to own files so they can be triggered independently during development and without causing side effects (assets, slack notifications).

- Update artifacts action from deprecated v3 to v4.


jmcphers (approved):
LGTM!

PR #492, opened 2024-08-29
Use github runner for macOS builds

lionel-:
Branched from #491.

Addresses https://github.com/posit-dev/positron/issues/4499.

### DIFF
--- a/.github/workflows/release-macos.yml
+++ b/.github/workflows/release-macos.yml
@@ -14,7 +14,7 @@ jobs:
     # Build ARK for macOS. Both arm64 (Apple Silicon) and x64 (Intel) hosts.
     build_macos:
         name: Build macOS
-        runs-on: [self-hosted-production, macos, arm64]
+        runs-on: macos-latest
         timeout-minutes: 40
 
         env:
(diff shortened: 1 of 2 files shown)


.github/workflows/release-macos.yml
```diff
@@ -41,12 +41,19 @@ jobs:
...
                 rustup default stable
 
+            - name: Setup for x86 cross-compiling
+              if: matrix.arch == 'x64'
+              run: |
+                # Install x86 homebrew
+                arch -x86_64 /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
+
+                # Install x86 Rust target
+                rustup target add x86_64-apple-darwin
```
DavisVaughan, line 51:
```suggestion
            - name: Install brew and Rust target
              run: |
                arch -${{matrix.arch_terminal}} /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"

                rustup target add ${{ matrix.rust_target_prefix }}-apple-darwin
```

I think I would like the idea of making this fully general and splitting it into two steps of `Install brew` and `Install Rust target`, but they would look like what I have inserted as a suggestion here.

I realize it would probably be a no-op for the arm case, but we've already got those nice env vars set up...so it seems nice to use them in this way?

lionel-: I think arm64 brew is already installed, and I wouldn't expect the install script to detect existing installs and decide to do a no-op? I could see it fail with an error. So I'd rather avoid calling it for arm64.

The rustup command is a no-op though.

I think the conditional `if:` step is clearer. And when we stop supporting intel, it can be removed without any other changes.

PR #500, opened 2024-09-03
Recognize non-function pages in `RHtmlHelp` and split it into topic vs function

DavisVaughan:
Closes https://github.com/posit-dev/ark/pull/431 - alternate approach

Addresses https://github.com/posit-dev/positron/issues/3467#issuecomment-2213963654 (meaning we can close the issue after this)

Example from https://github.com/posit-dev/ark/pull/431:

```r
bar <- function(...) {}

bar()
#   ^
#   get errors when moving a cursor to here
```

When you get to `bar(<here>)`, we try to look up help documentation for `?bar`. That actually gets mapped to `?plotmath`, a pure topic page (i.e. no functions on there). This was causing our `RHtmlHelp::parameters()` function to freak out, because it was only to be used on function pages.

`RHtmlHelp` is a generic help engine, so it should be able to look up non-function arbitrary topics. And in fact we do this for package pages, i.e. `"dplyr-package"` in `resolve_package_completion_item()`, so we can't just throw out any help page that doesn't look like function help (which was the #431 approach).

Instead, I've split `RHtmlHelp` into:
- `from_topic()`, which sets `function = false` internally
- `from_function()`, which sets `function = true` internally

`from_function()` also returns `None` if a new `is_function()` internal helper returns `false` on the found HTML. It uses the approach from #431 which was to detect a `Usage` section, which should only apply to functions.

We now use `from_function()` everywhere except `resolve_package_completion_item()`, but I see us potentially using `from_topic()` more in the future too.

The new internal `function: bool` flag is useful because it allows us to error immediately in `RHtmlHelp::parameter()` and `RHtmlHelp::parameters()` if we don't give it function related help. We could try to encode this into the type system but that seemed like overkill for now.

---

I've also introduced snapshot testing through the insta crate in this PR. It was useful for catching an expected error message, and I think we are going to want to use it elsewhere too.

### DIFF
--- a/crates/ark/src/lsp/completions/sources/composite/search_path.rs
+++ b/crates/ark/src/lsp/completions/sources/composite/search_path.rs
@@ -9,9 +9,10 @@ use anyhow::Result;
 use harp::exec::RFunction;
 use harp::exec::RFunctionExt;
 use harp::utils::r_env_is_pkg_env;
-use harp::utils::r_envir_name;
+use harp::utils::r_pkg_env_name;
 use harp::vector::CharacterVector;
 use harp::vector::Vector;
+use harp::RObject;
 use libr::R_EmptyEnv;
 use libr::R_GlobalEnv;
 use libr::R_lsInternal;
@@ -41,13 +42,23 @@ pub(super) fn completions_from_search_path(
         let mut envir = R_GlobalEnv;
 
         while envir != R_EmptyEnv {
-            // Get environment name
-            let name = r_envir_name(envir)?;
+            let is_pkg_env = r_env_is_pkg_env(envir);
+
+            // Get package environment name, if there is one
+            let name = if is_pkg_env {
+                let name = RObject::from(r_pkg_env_name(envir));
+                let name = String::try_from(name)?;
+                Some(name)
+            } else {
+                None
+            };
+
+            let name = name.as_deref();
 
             // If this is a package environment, we will need to force promises to give meaningful completions,
             // particularly with functions because we add a `CompletionItem::command()` that adds trailing `()` onto
             // the completion and triggers parameter completions.
-            let promise_strategy = if r_env_is_pkg_env(envir) {
+            let promise_strategy = if is_pkg_env {
                 PromiseStrategy::Force
             } else {
                 PromiseStrategy::Simple
@@ -71,12 +82,9 @@ pub(super) fn completions_from_search_path(
                 }
 
                 // Add the completion item.
-                let Some(item) = completion_item_from_symbol(
-                    symbol,
-                    envir,
-                    Some(name.as_str()),
-                    promise_strategy.clone(),
-                ) else {
+                let Some(item) =
+                    completion_item_from_symbol(symbol, envir, name, promise_strategy.clone())
+                else {
                     log::error!("Completion symbol '{symbol}' was unexpectedly not found.");
                     continue;
                 };
--- a/crates/ark/src/lsp/help.rs
+++ b/crates/ark/src/lsp/help.rs
@@ -365,3 +365,64 @@ fn for_each_section(doc: &Html, mut callback: impl FnMut(ElementRef, Vec<Element
         callback(header, elements);
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use crate::lsp::help::RHtmlHelp;
+    use crate::lsp::help::Status;
+    use crate::test::r_test;
+
+    #[test]
+    fn test_help_from_function() {
+        r_test(|| {
+            let help = unsafe { RHtmlHelp::from_function("match", None) };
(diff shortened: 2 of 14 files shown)


crates/ark/src/lsp/completions/sources/composite/search_path.rs
```diff
@@ -41,13 +42,23 @@ pub(super) fn completions_from_search_path(
...
+            let is_pkg_env = r_env_is_pkg_env(envir);
+
+            // Get package environment name, if there is one
+            let name = if is_pkg_env {
+                let name = RObject::from(r_pkg_env_name(envir));
+                let name = String::try_from(name)?;
+                Some(name)
+            } else {
+                None
+            };
```
DavisVaughan, line 54: There was also a bug here where `r_envir_name()` was returning things like `"0x123456"` for non-package environments, which was getting fed all the way through to `.ps.help.getHtmlHelpContents()` as the package name to use.

I've changed this to restrict it to only get a `name` for a package environment

crates/ark/src/lsp/help.rs
```diff
@@ -57,7 +81,22 @@ impl RHtmlHelp {
...
+    /// Uses a heuristic of looking for a `Usage` section to determine if this looks like
+    /// function help or not. Can't look for `Arguments`, as some functions don't have
+    /// any!
+    fn is_function(x: &Html) -> bool {
+        // Find all h3 headers in the document
+        let selector = Selector::parse("h3").unwrap();
+        let mut headers = x.select(&selector);
+
+        // Do any have a usage section?
+        headers.any(|header| header.html() == "<h3>Usage</h3>")
```
DavisVaughan, line 99: The `Usage` trick

crates/ark/src/lsp/help.rs
```diff
@@ -121,22 +160,32 @@ impl RHtmlHelp {
...
-        // </td></tr>
-        //
-        // Note that parameters might be parsed as part of different, multiple tables;
-        // we need to iterate over all tables after the Arguments header.
+    ) -> anyhow::Result<()> {
+        if !self.function {
+            return Err(anyhow::anyhow!(
+                "Called `parameters()` on a topic that isn't a function."
+            ));
+        }
```
DavisVaughan, line 187: Note the slightly safer error here

crates/ark/src/lsp/help.rs
```diff
@@ -306,3 +365,64 @@ fn for_each_section(doc: &Html, mut callback: impl FnMut(ElementRef, Vec<Element
...
+    }
+
+    #[test]
+    fn test_parameter_on_non_functions() {
+        r_test(|| {
+            let help = unsafe { RHtmlHelp::from_topic("plotmath", None) };
+            let help = help.unwrap().unwrap();
+            // Errors immediately
+            insta::assert_snapshot!(help.parameter("foo").unwrap_err());
+        });
```
DavisVaughan, line 426: This is what a snapshot test looks like

crates/ark/src/lsp/completions/sources/composite/search_path.rs
lionel-: No need to do now, but it would be useful to have an `Environment::kind()` method. It'd return an enum with variants for packages, namespaces, imports, etc, and the variant would contain the name.

crates/ark/src/lsp/help.rs
lionel-: I guess the html structure is stable enough? Can we analyse the Rd instead, perhaps eventually? It feels like we should be able to do better (though it's fine to do it this way now as a workaround, I'm just discovering this problem space and thinking aloud).

crates/ark/src/lsp/help.rs
```diff
@@ -32,7 +34,31 @@ pub enum Status {
 }
 
 impl RHtmlHelp {
-    pub unsafe fn new(topic: &str, package: Option<&str>) -> Result<Option<Self>> {
+    pub unsafe fn from_topic(topic: &str, package: Option<&str>) -> anyhow::Result<Option<Self>> {
```
lionel-, line 37: Why is this unsafe? Because it calls R? Can we use some other signal like `r_` (if specific methods) or `r::` (if the whole struct)?

lionel- (approved):

crates/ark/src/lsp/help.rs
DavisVaughan: > Can we analyse the Rd instead

Possibly yea, we'd have to return the `Rd` path (or contents) from `.ps.help.getHtmlHelpContents()` as well as the HTML contents (and maybe change the function name), and then store the contents in `RHtmlHelp` alongside the HTML. It seems like both installed help and dev help have access to a real `Rd` file on disk, so that's good.

And then add our own Rd parser on top of that to analyze it

PR #501, opened 2024-09-03
Data explorer: Allow for optional summary stats for dates

dfalbel:
Addresses https://github.com/posit-dev/positron/issues/4356
Pairs with: https://github.com/posit-dev/positron/pull/4564

This makes the output more consistent with the numbers summary statistics which are all optional and are not displayed when an error happens during the computation.

### DIFF
--- a/crates/amalthea/src/comm/data_explorer_comm.rs
+++ b/crates/amalthea/src/comm/data_explorer_comm.rs
@@ -383,38 +383,38 @@ pub struct SummaryStatsString {
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
 pub struct SummaryStatsDate {
 	/// The exact number of distinct values
-	pub num_unique: i64,
+	pub num_unique: Option<i64>,
 
 	/// Minimum date value as string
-	pub min_date: String,
+	pub min_date: Option<String>,
 
 	/// Average date value as string
-	pub mean_date: String,
+	pub mean_date: Option<String>,
 
 	/// Sample median (50% value) date value as string
-	pub median_date: String,
+	pub median_date: Option<String>,
 
 	/// Maximum date value as string
-	pub max_date: String
+	pub max_date: Option<String>
 }
 
 /// SummaryStatsDatetime in Schemas
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
 pub struct SummaryStatsDatetime {
 	/// The exact number of distinct values
-	pub num_unique: i64,
+	pub num_unique: Option<i64>,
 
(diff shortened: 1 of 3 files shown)


crates/amalthea/src/comm/data_explorer_comm.rs
```diff
@@ -383,38 +383,38 @@ pub struct SummaryStatsString {
...
-	pub num_unique: i64,
+	pub num_unique: Option<i64>,
 
 	/// Minimum date value as string
-	pub min_date: String,
+	pub min_date: Option<String>,
 
 	/// Average date value as string
-	pub mean_date: String,
+	pub mean_date: Option<String>,
```
DavisVaughan, line 392: Are we really showing mean date? That doesn't mean anything practical

crates/ark/src/data_explorer/summary_stats.rs
```diff
@@ -155,11 +161,19 @@ fn empty_column_summary_stats() -> data_explorer_comm::ColumnSummaryStats {
...
+where
+    Return: TryFrom<T>,
+{
     let value = stats.get(name);
 
     match value {
-        Some(value) => Ok(value.clone()),
+        Some(value) => {
+            let value: Return = unwrap!(value.clone().try_into(), Err(_) => {
+                return Err(anyhow!("Can't cast to return type."))
```
DavisVaughan, line 173:
```suggestion
            let value: Return = unwrap!(value.clone().try_into(), Err(_) => {
                return Err(anyhow!("Can't cast to return type: {err:?}"))
```

We don't care about the reason here?

crates/ark/src/modules/positron/r_data_explorer.R
```diff
@@ -78,16 +78,31 @@ summary_stats_date <- function(col) {
     # When calling `min` on `x` would raise a warning.
     # Turns out, some Parquet files might generate malformed timezones too.
     suppressWarnings({
+        # When all values in the column are NA's, there min and max return -Inf and +Inf,
+        # mean returns NaN and median returns NA. We make everything return `NULL` so we
```
DavisVaughan, line 82: Yea, really hate the R inconsistencies here 😢

DavisVaughan (approved):

crates/amalthea/src/comm/data_explorer_comm.rs
dfalbel: We're not showing it in the UI, even though we compute it. I had the same feeling here: https://github.com/posit-dev/positron/pull/4345#pullrequestreview-2238371237

PR #502, opened 2024-09-03
Add attribute to `repos`, for renv users

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/4255

You can check out the approach RStudio takes here: 
https://github.com/rstudio/rstudio/blob/078e21116b0e34aff92addf961699017adb62fc5/src/cpp/r/R/Tools.R#L698

I saw [your note here](https://github.com/rstudio/renv/issues/1963#issuecomment-2272152739) @DavisVaughan. Given that we are unlikely to be able to change this attribute in RStudio, I don't tend to think we will gain much by using `IDE` or `default`. If you have a strong opinion otherwise, happy to change this, though!

## QA Notes

After this PR is merged and ark is bumped, you will see (edited after code review to change name of attribute):

``` r
getOption("repos")
#>                        CRAN 
#> "https://cran.rstudio.com/" 
#> attr(,"IDE")
#> [1] TRUE
```

<sup>Created on 2024-09-04 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

Notice the new attribute.

Once we get this into a release build, we should update at https://github.com/rstudio/renv/issues/1963.

### DIFF
--- a/crates/ark/src/modules/positron/options.R
+++ b/crates/ark/src/modules/positron/options.R
@@ -27,8 +27,27 @@ options(device = function() {
 })
 
 # Set cran mirror
-repos <- getOption("repos")
-rstudio_cran <- "https://cran.rstudio.com/"
+set_cran_mirror <- function(repos) {
+    rstudio_cran <- "https://cran.rstudio.com/"
+    if (is.null(repos) || !is.character(repos)) {
+        return(c(CRAN = rstudio_cran))
+    } else {
+        if ("CRAN" %in% names(repos)) {
+            if (identical(repos[["CRAN"]], "@CRAN@")) {
+                repos[["CRAN"]] <- rstudio_cran
+                return(repos)
+            }
+        } else {
+            return(c(CRAN = rstudio_cran, repos))
+        }
+    }
+}
+
+original_repos <- getOption("repos")
+repos <- set_cran_mirror(original_repos)
+attr(repos, "Positron") <- TRUE
+options(repos = repos)
+
 
 if (is.null(repos) || !is.character(repos)) {
     options(repos = c(CRAN = rstudio_cran))
--- a/crates/ark/src/modules/positron/options.R
+++ b/crates/ark/src/modules/positron/options.R
@@ -27,41 +27,21 @@ options(device = function() {
 })
 
 # Set cran mirror
-set_cran_mirror <- function(repos) {
-    rstudio_cran <- "https://cran.rstudio.com/"
-    if (is.null(repos) || !is.character(repos)) {
-        return(c(CRAN = rstudio_cran))
-    } else {
-        if ("CRAN" %in% names(repos)) {
-            if (identical(repos[["CRAN"]], "@CRAN@")) {
-                repos[["CRAN"]] <- rstudio_cran
-                return(repos)
-            }
-        } else {
-            return(c(CRAN = rstudio_cran, repos))
-        }
-    }
-}
-
-original_repos <- getOption("repos")
-repos <- set_cran_mirror(original_repos)
-attr(repos, "Positron") <- TRUE
-options(repos = repos)
-
-
+repos <- getOption("repos")
+rstudio_cran <- "https://cran.rstudio.com/"
 if (is.null(repos) || !is.character(repos)) {
-    options(repos = c(CRAN = rstudio_cran))
+    repos <- c(CRAN = rstudio_cran)
 } else {
     if ("CRAN" %in% names(repos)) {
         if (identical(repos[["CRAN"]], "@CRAN@")) {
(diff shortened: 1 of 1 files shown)


crates/ark/src/modules/positron/options.R
```diff
@@ -29,20 +29,19 @@ options(device = function() {
...
         if (identical(repos[["CRAN"]], "@CRAN@")) {
             repos[["CRAN"]] <- rstudio_cran
-            options(repos = repos)
         }
     } else {
         repos <- c(CRAN = rstudio_cran, repos)
-        options(repos = repos)
     }
 }
+attr(repos, "Positron") <- TRUE
```
DavisVaughan, line 43: We should not unconditionally set this. IIUC, we should only set this if we detect the "default" repos behavior, i.e. if we see `"@CRAN@"` as the `repos[["CRAN"]]` value

That is what RStudio does

https://github.com/rstudio/rstudio/blob/078e21116b0e34aff92addf961699017adb62fc5/src/cpp/r/R/Tools.R#L682-L712

I _think_ that means you only set the attribute in the `if (identical(repos[["CRAN"]], "@CRAN@")) {` branch

---

In theory this allows a user to set their own `repos` in their `.Rprofile`, and in that case we are not supposed to mark that with a `"Positron"` attribute, as that did not come from us. In practice that might not work, because I _think_ this code might get run before we source a user's `.Rprofile` on their behalf, but that would be another bug to fix separately, I think.

crates/ark/src/modules/positron/options.R
```diff
@@ -29,20 +29,19 @@ options(device = function() {
...
+options(repos = repos)
 
 # Show Plumber apps in the viewer
```
DavisVaughan, line 46: I have one other larger worry about this in general.

None of this is going to kick in if the user is using a version of R installed with Rig. That's because Rig sets `repos` for you in the special `/library/base/R/Rprofile` that R forcibly loads on every startup (like, rig straight up tweaks this file that you get when you download R itself)

That means that every time you load up a rig installed version of R, it will _always_ have `repos` set and it will never be `"@CRAN@"`.

https://github.com/r-lib/rig/blob/140115c9b565167670cfc6f303e6c968c563db98/src/windows.rs#L342
https://github.com/r-lib/rig/blob/140115c9b565167670cfc6f303e6c968c563db98/src/macos.rs#L788
https://github.com/r-lib/rig/blob/140115c9b565167670cfc6f303e6c968c563db98/src/linux.rs#L579-L613

That means we would _never_ add the `"Positron"` attribute, and RStudio never adds the `"RStudio"` attribute.

I can confirm this by loading up my non-rig R-devel version in a plain terminal:

```
> getOption("repos")
    CRAN
"@CRAN@"
```

vs a rig installed R 4.4

```
> getOption("repos")
                         CRAN
"https://cloud.r-project.org"
```

Now, this _may_ be ok for Positron because on Linux you can see in the above link that rig sets up PPM as the `repos` option, so that should mean that users get binaries when using renv + rig on Linux, even though renv isn't the one setting the `repos` option.

The bigger issue would be for renv itself, because it means that renv will never detect a "default" case and can't set the `repos` option to the user's chosen configuration option for `ppm.url()`, which defaults to normal PPM but can be customized by the user:
https://github.com/rstudio/renv/blob/fcfe748225a88bf56b1964868604353c2e73bf7c/R/init.R#L341

crates/ark/src/modules/positron/options.R
DavisVaughan, line 43: I would still be in favor of `IDE` as the attribute name

It's the same idea as when we changed clickable hyperlinks from `rstudio:run` to `x-r-run`, its just more generic and allows other IDEs/editors to opt in without weirdly having to set `"RStudio"` or `"Positron"` as the attribute name.

In other words, to me it is less about getting RStudio to switch to this and more about making it slightly more generic for any IDE/editor to use.

crates/ark/src/modules/positron/options.R
DavisVaughan: I do not think we should let this stop us from moving on this PR, but @kevinushey may want to take a closer look at this weird rig + renv intersection issue, especially the inability to set a user customized `config$ppm.url()`

DavisVaughan (approved):

crates/ark/src/modules/positron/options.R
juliasilge: Ah, I misunderstood before how this was being used! Thanks 👍

crates/ark/src/modules/positron/options.R
juliasilge: I am not a rig user 😆 so I did not notice this.

I will add a note for renv when I say this new attribute is available.

PR #503, opened 2024-09-04
Add `try_*_get()` for each vector type, and fail more elegantly in `Vector` element extraction

DavisVaughan:
Closes https://github.com/posit-dev/ark/pull/499 (alternative approach)
Addresses https://github.com/posit-dev/positron/issues/4537

If we try and extract an element with `get_unchecked()` that causes an error in an ALTREP `Elt` method (including but not limited to an OOM error), we now log an error and return `error_elt()` for that type, which is typically the `NA` value for that type, but for raw vectors it is `0` as an `Rbyte`.

I considered propagating the error up through the return value of `get_unchecked()` but that was _painful_. We use it in so many places that don't expect it to fail that propagating an error through would be a huge ordeal that doesn't seem worth it for this somewhat rare case. I think logging an error + showing `NA` or something like it is possibly the best we can do.

I confirmed that

```r
x <- 1:1e10
names(x) <- x
```

doesn't crash immediately anymore, but expanding `x` in the Variables pane hangs and causes a 5 second timeout. Something is definitely not right here because even `x <- 1:1e5` hangs for a long time on `main` when you try and expand `x` in the Variables pane.

### DIFF
--- a/crates/ark/src/variables/variable.rs
+++ b/crates/ark/src/variables/variable.rs
@@ -1264,3 +1264,86 @@ pub fn plain_binding_force_with_rollback(binding: &Binding) -> anyhow::Result<RO
         _ => Err(anyhow!("Unexpected binding type")),
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use regex::Regex;
+
+    use crate::test::r_test;
+    use crate::variables::variable::WorkspaceVariableDisplayValue;
+
+    fn get_display_value(code: &str) -> String {
+        WorkspaceVariableDisplayValue::from(harp::parse_eval_base(code).unwrap().sexp).display_value
+    }
+
+    fn expect_display_value(code: &str, expected: &str) {
+        let display = get_display_value(code);
+        assert_eq!(display, expected.to_string());
+    }
+
+    #[test]
+    fn test_simple_display_values() {
+        r_test(|| {
+            expect_display_value("1", "1");
+            expect_display_value("1L", "1");
+            expect_display_value("'a'", "\"a\"");
+            expect_display_value("NULL", "NULL");
+            expect_display_value("TRUE", "TRUE");
+            expect_display_value("FALSE", "FALSE");
+            expect_display_value("1i", "0+1i");
+        })
+    }
+
+    #[test]
+    fn test_data_frame_display_value() {
+        r_test(|| {
+            expect_display_value("datasets::mtcars", "[32 rows x 11 columns] <data.frame>");
+            expect_display_value("matrix(1:4, ncol=2)", "[[1 2], [3 4]]");
+        })
+    }
+
+    #[test]
+    fn test_list_display_value() {
+        r_test(|| {
+            expect_display_value("list(x=1:4)", "[x = 1 2 3 4]");
+            expect_display_value("list(1:4)", "[1 2 3 4]");
+        })
+    }
+
+    #[test]
+    fn test_functions_display_value() {
+        r_test(|| {
+            expect_display_value("function() NULL", "function () ");
+            expect_display_value("function(a) NULL", "function (a) ");
+            expect_display_value("function(a, b) NULL", "function (a, b) ");
+            expect_display_value("function(a = 1, b) NULL", "function (a = 1, b) ");
+        })
+    }
+
+    #[test]
+    fn test_altrep_is_not_materialized() {
+        r_test(|| {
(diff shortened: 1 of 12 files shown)


DavisVaughan:
In general things like this should no longer immediately crash the R session. Note that printing `z` in the console will immediately trigger the R error.

Also note that using `STRING_ELT()` on `z` results in the allocation error, which is why you see `NA` values in the Variables pane. Using `STRING_ELT()` is the best we can do in the _general_ case for ALTREP objects (because it gives them a chance to do low memory operations if they can) but it just happens to trigger allocations for the deferred string ALTREP class in use here.

```r
x <- 1:1e10
names(x) <- x
z <- names(x)
```

Screenshot 2024-09-04 at 11 33 37 AM

dfalbel (approved):
LGTM! Is it worth copying the tests from #499 ?

crates/ark/src/variables/variable.rs
```diff
@@ -1264,3 +1264,86 @@ pub fn plain_binding_force_with_rollback(binding: &Binding) -> anyhow::Result<RO
         _ => Err(anyhow!("Unexpected binding type")),
     }
 }
+
+#[cfg(test)]
+mod tests {
```
lionel-, line 1269: Nice!

crates/ark/src/variables/variable.rs
```diff
@@ -1264,3 +1264,86 @@ pub fn plain_binding_force_with_rollback(binding: &Binding) -> anyhow::Result<RO
...
+            // of the STRSXP vector inside the ALTREP `Elt` method for deferred strings,
+            // which throws an OOM error when trying to look at elements of `names(x)`.
+            // We catch this, log an error, and return `NA` as the element value since
+            // we can't determine what it is.
+            let success = Regex::new(r#"^"1" "2" "3""#).unwrap();
+            let failure = Regex::new(r#"^NA NA NA"#).unwrap();
+
+            // Small, should always pass with `success`
+            let display = get_display_value("local({x = 1:1e3; names(x) = x; names(x)})");
+            assert!(success.is_match(display.as_str()) || failure.is_match(display.as_str()));
```
lionel-, line 1342:
```suggestion
            assert!(success.is_match(display.as_str()));
```

lionel- (approved):
LGTM!

I'm slightly worried about the performance overhead. This could be improved by:

- Detecting whether a vector is altrep or not on construction.
- If not altrep, get full vector slice and use that to index an element
- If altrep, use protected get, and this could be improved later on to retrieve regions so the protection would only happen once per region.

lionel-:
> and this could be improved later on to retrieve regions so the protection would only happen once per region.

I'm not familiar with altrep but would this behave better than the `_ELT()` methods? I.e. would this avoid a full materialisation? It seems like that's the purpose of retrieving a region slice?

lionel-:
Do we need to be worried about the variable and data viewer panes being confusing with altrep objects? Would it be possible for the formatting paths to use the `Result` getter and display, say, `<error>` instead of `NA`, so that the output is at least googlable?

DavisVaughan:
Note to self to wait for https://github.com/posit-dev/ark/pull/508 and rework this on top of that after it is merged

DavisVaughan:
> would this avoid a full materialisation

It is completely up to the ALTREP class. The ALTREP author could decide that `INTEGER_GET_REGION()` can be implemented without materializing (by providing an `INTEGER_OR_NULL()` method), but `INTEGER_ELT()` can't. The important thing for us is that we don't have any control over that, and just have to try and use "the friendliest thing we can".

---

Using `INTEGER_GET_REGION()` combined with having a small `buffer` vector of the underlying C type is not unreasonable. cpp11 does exactly this actually. It _may_ avoid repeated `INTEGER_ELT()` calls _if_ the ALTREP type can provide a const DATAPTR without materializing:
https://github.com/wch/r-source/blob/64873e17ff8be3c305b0a7487e01b266bc4660c9/src/main/altrep.c#L404-L417

Otherwise it falls back to this:
https://github.com/wch/r-source/blob/64873e17ff8be3c305b0a7487e01b266bc4660c9/src/main/altrep.c#L755-L763

And FWIW, deferred strings can't provide a dataptr before it has fully materialized, so it would always fall back to the `ELT` loop fallback approach
https://github.com/wch/r-source/blob/64873e17ff8be3c305b0a7487e01b266bc4660c9/src/main/altclasses.c#L764-L768

DavisVaughan:
Will revive as needed, not worth keeping open for right now I think

PR #506, opened 2024-09-06
Reticulate support

dfalbel:
Add support for reticulate sessions. See https://github.com/posit-dev/positron/pull/4603

### DIFF
--- a/crates/ark/src/modules/positron/reticulate.R
+++ b/crates/ark/src/modules/positron/reticulate.R
@@ -1,0 +1,6 @@
+
+#' @export
+.ps.rpc.reticulate_interpreter_path <- function() {
+    if (!.ps.is_installed("reticulate")) return("")
+    reticulate::py_discover_config()$python
+}
--- a/crates/ark/src/modules/positron/reticulate.R
+++ b/crates/ark/src/modules/positron/reticulate.R
@@ -4,3 +4,13 @@
     if (!.ps.is_installed("reticulate")) return("")
     reticulate::py_discover_config()$python
 }
+
+#' @export
+.ps.reticulate_open <- function(id) {
+    .ps.Call("ps_reticulate_open")
+}
+
+#' @export
+.ps.reticulate_focus <- function(id) {
+    .ps.Call("ps_reticulate_focus", id)
+}
--- a/crates/ark/src/modules/positron/reticulate.R
+++ b/crates/ark/src/modules/positron/reticulate.R
@@ -9,8 +9,3 @@
 .ps.reticulate_open <- function(id) {
     .ps.Call("ps_reticulate_open")
 }
-
-#' @export
-.ps.reticulate_focus <- function(id) {
-    .ps.Call("ps_reticulate_focus", id)
-}
--- a/crates/ark/src/modules/positron/reticulate.R
+++ b/crates/ark/src/modules/positron/reticulate.R
@@ -1,16 +1,10 @@
-
-#' @export
-.ps.rpc.reticulate_interpreter_path <- function() {
-    if (!.ps.is_installed("reticulate")) return("")
-    reticulate::py_discover_config()$python
-}
-
 #' @export
 .ps.reticulate_open <- function(input="") {
     .ps.Call("ps_reticulate_open", input)
 }
 
 #' Used by the front-end to install reticulate
+#' A modal asking to install is shown before calling this.
 #' @export
 .ps.rpc.install_reticulate <- function() {
     tryCatch({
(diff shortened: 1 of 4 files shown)


crates/ark/src/reticulate.rs
```diff
@@ -0,0 +1,145 @@
...
+            String::from("positron.reticulate"),
+        );
+
+        let service = Self {
+            comm,
+            comm_manager_tx,
+        };
+
+        let event = CommManagerEvent::Opened(service.comm.clone(), serde_json::Value::Null);
+        unwrap!(service.comm_manager_tx.send(event), Err(e) => {
```
lionel-, line 43: If return value is not used, better use the standard `if let Err(err)`

crates/ark/src/reticulate.rs
```diff
@@ -0,0 +1,145 @@
...
+
+    // If there's an id already registered, we just need to send the focus event
+    let mut comm_id_guard = unwrap!(
+        RETICULATE_COMM_ID.try_lock(),
+        Err(e) => {
+            return Err(anyhow!("Could not access comm_id. Error {}", e));
+        }
+    );
+
+    if let Some(id) = (*comm_id_guard).clone() {
```
lionel-, line 120: Nit: More idiomatic and transparent to clone at the call site that needs to own the object (i.e. innermost scope).

crates/ark/src/reticulate.rs
```diff
@@ -0,0 +1,145 @@
...
+// the comm_id that is returned by this function.
+#[harp::register]
+pub unsafe extern "C" fn ps_reticulate_open(input: SEXP) -> Result<SEXP, anyhow::Error> {
+    let main = RMain::get();
+
+    if !RMain::initialized() {
+        return Ok(R_NilValue);
+    }
+
+    let input = RObject::try_from(input)?;
```
lionel-, line 109: We're converging towards this style which makes for clearer code thanks to the type annotation:

```suggestion
    let input: RObject = input.try_into()?;
```

lionel-: Also for consistency with the line below.

crates/ark/src/reticulate.rs
```diff
@@ -0,0 +1,145 @@
...
+            }
+        }
+
+        // before finalizing the thread we make sure to send a close message to the front end
+        if let Err(err) = self.comm.outgoing_tx.send(CommMsg::Close) {
+            log::error!("Reticulate: Error while sending comm_close to front end: {err:?}");
+        }
+
+        let mut comm_id_guard = unwrap!(
+            RETICULATE_COMM_ID.try_lock(),
```
lionel-, line 81: `try_lock()` is very low level. Note that it's allowed to fail spuriously. Should use `.lock().unwrap()` here, as we are not worried about contention or deadlocks.

crates/ark/src/modules/positron/reticulate.R
```diff
@@ -0,0 +1,99 @@
+#' @export
+.ps.reticulate_open <- function(input="") {
+    .ps.Call("ps_reticulate_open", input)
+}
+
+#' Used by the front-end to install reticulate
+#' A modal asking to install is shown before calling this.
+#' @export
+.ps.rpc.install_reticulate <- function() {
```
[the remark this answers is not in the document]
lionel-: Actually shouldn't the frontend use the existing way of installing packages, i.e. `r.packageInstall()`?

crates/ark/src/modules/positron/reticulate.R
```diff
@@ -0,0 +1,99 @@
...
+    )
+}
+
+#' @export
+.ps.rpc.reticulate_start_kernel <- function(kernelPath, connectionFile, logFile, logLevel) {
+    tryCatch({
+        reticulate:::py_run_file_on_thread(
+            file = kernelPath,
+            args = c(
+                "-f", connectionFile,
```
lionel-, line 88: I assume this is for running ipykernel in the Python thread with the connection information generated by Positron? I think that could use some comments about what is going on.

crates/ark/src/reticulate.rs
```diff
@@ -0,0 +1,145 @@
...
+                log::error!("Reticulate: Error while receiving message from frontend: {err:?}");
+                break;
+            });
+
+            if let CommMsg::Close = msg {
+                self.comm.outgoing_tx.send(CommMsg::Close).unwrap();
+                break;
+            }
+
+            // Forward data msgs to the frontend
```
lionel-, line 68: We're just resending the data messages back? Could you add comments about that please?

crates/ark/src/reticulate.rs
```diff
@@ -0,0 +1,145 @@
...
+// We should aim at having at most **1** client per R session.
+// Further actions that reticulate can ask the front-end can be requested through
+// the comm_id that is returned by this function.
+#[harp::register]
+pub unsafe extern "C" fn ps_reticulate_open(input: SEXP) -> Result<SEXP, anyhow::Error> {
+    let main = RMain::get();
+
+    if !RMain::initialized() {
+        return Ok(R_NilValue);
+    }
```
lionel-, line 107: This should probably surface an error.

But once https://github.com/posit-dev/ark/pull/542 is merged, you can use `RMain::wait_initialized()`. If you need R to be ready too, you can also use `RMain::wait_r_initialized()`.

That said this method is called from R so actually this probably should be removed altogether.

lionel-:
My main comment is about `try_lock()` which should be replaced by `lock()` to avoid rarely occurring surprising behaviour.

crates/ark/src/modules/positron/reticulate.R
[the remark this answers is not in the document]
dfalbel: `r.packageInstall` is essentially just `devtools::install()` it doesn't allow installing arbitrary packages.
I can add `.ps.rpc.install_package(pkg)` if we want a more general approach. 

yes, the front-end can set verbosity

crates/ark/src/reticulate.rs
dfalbel: We don't need at, I copied from another comm by mistake.

crates/ark/src/reticulate.rs
dfalbel: This was required for testing the connections comm, because loading a package could call comm_open, but it's not required here and I accidentally copied it over.

crates/ark/src/modules/positron/reticulate.R
dfalbel: Added some comments! It would be nice to keep this close to the call site: https://github.com/posit-dev/positron/blob/5f3a1fa6c9d9e0eeae5b63483024e852fc5acc83/extensions/positron-reticulate/src/extension.ts#L383-L390 but `RuntimeSession.execute` does not allow awaiting /and or handling errors.

crates/ark/src/reticulate.rs
dfalbel: Used `.or_log_error()` instead

dfalbel:
I addressed most points in https://github.com/posit-dev/ark/pull/506/commits/286a24aa0000ba736694dd133e9de117ba9d7d20

Also moved `.ps.rpc.install_reticulate()` into a more general `.ps.rpc.install_packages`.

lionel- (approved):
LG!

PR #507, opened 2024-09-06
Expose environment variables from R process

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/2723

With this new function (callable via `call_method()` in Positron) we can get the values for a set of environment variables starting with a given regex from the R process:

``` r
.ps.rpc.get_env_vars("POSITRON")
#> $POSITRON
#> [1] "1"
#> 
#> $POSITRON_VERSION
#> [1] "2024.09.0"
```

If you need more than one set, you can do something like `.ps.rpc.get_env_vars("POSITRON|VSCODE")`.

### DIFF
--- a/crates/ark/src/modules/positron/system.R
+++ b/crates/ark/src/modules/positron/system.R
@@ -21,3 +21,11 @@ has_x11 <- function() {
     out <- as.list(vapply(cats, Sys.getlocale, "string", USE.NAMES = TRUE))
     c(LANG = Sys.getenv("LANG"), out)
 }
+
+#' Reports a vector of environment variables for the R process.
+#' @returns Named character vector of env vars that start with the given regex.
+#' @export
+.ps.rpc.get_env_vars <- function(pattern) {
+    matches <- grep(pattern, names(Sys.getenv()), value = TRUE)
+    as.list(vapply(matches, Sys.getenv, "string"))
+}
--- a/crates/ark/src/modules/positron/system.R
+++ b/crates/ark/src/modules/positron/system.R
@@ -28,5 +28,5 @@ has_x11 <- function() {
 #' @returns Values of the environment variables as a list.
 #' @export
 .ps.rpc.get_env_vars <- function(x = NULL) {
-    as.list(Sys.getenv(x))
+    as.list(Sys.getenv(x, names = TRUE))
 }
--- a/crates/ark/src/modules/positron/system.R
+++ b/crates/ark/src/modules/positron/system.R
@@ -22,11 +22,12 @@ has_x11 <- function() {
     c(LANG = Sys.getenv("LANG"), out)
 }
 
-#' Reports a vector of environment variables for the R process.
-#' @returns Named character vector of env vars that start with the given regex.
+#' Reports a list of environment variables for the R process.
+#' @returns List of all env vars, or optionally the subset that start with the
+#' given regex.
 #' @export
 .ps.rpc.get_env_vars <- function(pattern = NULL) {
-  ev <- as.list(Sys.getenv())
-  m <- if (is.null(pattern)) TRUE else grep(pattern, names(ev))
-  ev[m]
-}
+    ev <- as.list(Sys.getenv())
+    m <- if (is.null(pattern)) TRUE else grep(pattern, names(ev))
+    ev[m]
+  }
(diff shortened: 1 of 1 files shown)


crates/ark/src/modules/positron/system.R
```diff
@@ -21,3 +21,11 @@ has_x11 <- function() {
...
     c(LANG = Sys.getenv("LANG"), out)
 }
+
+#' Reports a vector of environment variables for the R process.
+#' @returns Named character vector of env vars that start with the given regex.
+#' @export
+.ps.rpc.get_env_vars <- function(pattern) {
+    matches <- grep(pattern, names(Sys.getenv()), value = TRUE)
+    as.list(vapply(matches, Sys.getenv, "string"))
+}
```
jennybc, line 31: I wonder if we will want some default behaviour that just returns all the env vars. Maybe something like:

```suggestion
.ps.rpc.get_env_vars <- function(pattern = NULL) {
  ev <- as.list(Sys.getenv())
  m <- if (is.null(pattern)) TRUE else grep(pattern, names(ev))
  ev[m]
}
```

crates/ark/src/modules/positron/system.R
```diff
@@ -21,3 +21,11 @@ has_x11 <- function() {
     out <- as.list(vapply(cats, Sys.getlocale, "string", USE.NAMES = TRUE))
     c(LANG = Sys.getenv("LANG"), out)
 }
+
+#' Reports a vector of environment variables for the R process.
+#' @returns Named character vector of env vars that start with the given regex.
+#' @export
+.ps.rpc.get_env_vars <- function(pattern) {
+    matches <- grep(pattern, names(Sys.getenv()), value = TRUE)
```
lionel-, line 29: Will the return value be too unstable to safely work with, since the grep pattern might match unexpected variables? Alternatively the string could be split by a separator like `;`, and each of the components retrieved in order. This way there is an invariant relationship between inputs and outputs.

juliasilge: Can you give me an example of the kind of unexpected variable you are thinking of?

DavisVaughan: I would much prefer explicit extraction of env vars rather than a regex approach

juliasilge: So for the [motivating testthat example](https://github.com/posit-dev/positron/issues/2723), you would want to pass over every `TESTTHAT*` env var explicitly (and need to update the list if testthat adds one) rather than choose them via their prefix?

lionel-: On second thought, since the goal is to reproduce the environment for specific packages, matching a prefix seems like a reasonable way of doing that.

But I wonder if these variables should be pushed from ark to the frontend via the UI comm after each top-level execution to ensure the frontend is up-to-date without getting into concurrency issues of a pull approach (will R be busy when an update is needed, etc).

juliasilge: The motivating example here is to use these environment variables in _tasks_ so I think what we have here is a good option, given those constraints. We have to pass the env var into the task object, similar to how we get the pandoc path here: 
https://github.com/posit-dev/positron/blob/348f25f96745f00beed473b6d000bf6b7e6889c8/extensions/positron-r/src/tasks.ts#L65

These env vars need to get passed through via the session manager, much like the locale info that gets used here:
https://github.com/posit-dev/positron/blob/348f25f96745f00beed473b6d000bf6b7e6889c8/extensions/positron-r/src/session.ts#L358

crates/ark/src/modules/positron/system.R
juliasilge: I tested this out and it works well. 👍

juliasilge:
@jmcphers also thought that a regex was a bad idea and we should do the work of passing specific env vars by name. This tiny new helper now works like this:

``` r
.ps.rpc.get_env_vars(c("USER", "POSITRON", "POTATO"))
$USER
[1] "juliasilge"

$POSITRON
[1] "1"

$POTATO
[1] ""
```

So basically an unfancy and tiny wrapper around `Sys.getenv()`.

DavisVaughan (approved):
I'm cool with how this works now if you can get approval on the Positron side too

Eventually supporting a push based model does sounds nice though

juliasilge:
I opened this issue to track a push-based model when the need becomes more urgent: https://github.com/posit-dev/positron/issues/4641

PR #508, opened 2024-09-09
Add `ParseData` type

lionel-:
This new type wraps the data frame of parse information generated by the R parser when:

- A source file environment is provided to the parser (which is the case at R level when `getOption("keep.source")` is `TRUE`).
- `getOption("keep.parse.data")` is `TRUE` (the default).

This parse data is interesting for the task of determining expression boundaries in a string because it is written by the parser even in the case of an incomplete or invalid input. Unfortunately there is no information about parse failures that we can use, but knowing the boundaries of valid expressions will prove very useful as it allows the following strategy:

- Use parse data to figure out boundaries of complete expressions.
- If there was an error, do a binary search with `R_ParseVector()` to figure out the boundaries of incomplete and invalid inputs in the last expression.

Progress towards posit-dev/positron#1326.

This work is supported by the following infrastructure improvements that are also part of this PR:

- New `DataFrame` type that validates its input. Once constructed, the following guarantees are provided:
    - List storage and `"data.frame"` class
    - Dimensions and valid names
    - All columns are vectors and the same size as the data frame

- There is a new `Error::MissingColumnError` returned from `DataFrame::col()` when an expected column is missing.

- New `harp::assert_class()` returns a new `Error::UnexpectedClass`.

- New `harp::unreachable!()` macro returns an `Error::AnyhowError`.

- New `.inherits()`, `.class()`, and `.dim()` methods for `RObject`.
 
- New coercion methods from `&RObject` to `Vec<T>` used to validate column types. These use a generic helper. I tried to integrate this helper in the `Vector` trait but couldn't solve compiler errors. Alternatively we could generate them with `vector_macros` but unless it saves a of work or substantially reduces the chance of undry typos we're probably better off with explicit code.

- The `iter()` method is no longer generated by `vector_macros`, it's now part of the `Vector` trait. This allows generic implementations to create vector iterators.

- The `TryFrom` methods for `SEXP -> impl Vector` now create empty vectors if passed `NULL`. This in turn is used in `&RObject -> Vec<T>` (this was necessary because the existing conversion method for `RObject -> Vec<i32>` now uses the new helper for `&RObject` and it supports `NULL` inputs, which seems reasonable enough for a conversion method).

### DIFF
--- a/crates/harp/src/data_frame.rs
+++ b/crates/harp/src/data_frame.rs
@@ -18,11 +18,13 @@ impl DataFrame {
         harp::assert_class(sexp, "data.frame")?;
 
         let Some(dim) = list.obj.dim()? else {
-            return Err(harp::anyhow!("Data frame doesn't have dimensions"));
+            return Err(harp::unexpected_structure!(
+                "Data frame doesn't have dimensions"
+            ));
         };
 
         if dim.len() != 2 {
-            return Err(harp::anyhow!(
+            return Err(harp::unexpected_structure!(
                 "Data frame must have 2 dimensions, instead it has {}",
                 dim.len()
             ));
@@ -34,10 +36,10 @@ impl DataFrame {
         let obj = RObject::view(sexp);
 
         let Some(names) = obj.names() else {
-            return Err(harp::anyhow!("Data frame doesn't have names"));
+            return Err(harp::unexpected_structure!("Data frame doesn't have names"));
         };
         let Ok(names) = harp::assert_non_optional(names) else {
-            return Err(harp::anyhow!("Data frame has missing names"));
+            return Err(harp::unexpected_structure!("Data frame has missing names"));
         };
 
         Ok(Self {
--- a/crates/harp/src/data_frame.rs
+++ b/crates/harp/src/data_frame.rs
@@ -2,6 +2,12 @@ use crate::vector::Vector;
 use crate::List;
 use crate::RObject;
 
+/// Typed data frame
+///
+/// Type guarantees:
+/// - Storage type is list.
+/// - Class is `"data.frame"`.
+/// - All columns are vectors and the same size as the data frame.
 #[derive(Debug)]
 pub struct DataFrame {
     pub list: List,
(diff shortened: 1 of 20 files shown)


crates/harp/src/object.rs
```diff
@@ -121,18 +122,20 @@ impl AsRef<SEXP> for RObject {
 }
 
 pub trait RObjectExt<T> {
-    unsafe fn elt(&self, index: T) -> crate::error::Result<RObject>;
+    fn elt(&self, index: T) -> crate::error::Result<RObject>;
 }
 
 impl PartialEq for RObject {
+    // FIXME: At call sites, not obvious that this is doing identity comparisons.
+    // Can we require explicit `FOO.sexp == BAR.sexp`?
```
DavisVaughan, line 130: I'd like that

crates/harp/src/object.rs
```diff
@@ -474,6 +477,37 @@ impl RObject {
...
+            ));
+        }
+
+        Ok(Some(class.try_into()?))
+    }
+
+    pub fn dim(&self) -> harp::Result<Option<Vec<usize>>> {
+        let dim: RObject = r_dim(self.sexp).into();
+
+        if dim.sexp == RObject::null().sexp {
```
DavisVaughan, line 501:
```suggestion
        if dim.sexp == harp::r_null() {
```

crates/harp/src/vector/list.rs
```diff
@@ -75,8 +78,8 @@ impl super::Vector for List {
         let mut data = data.into_iter();
 
         let size = data.len();
-        let inner = crate::alloc_list(size).unwrap();
-        let inner: RObject = inner.into();
+        let sexp = crate::alloc_list(size).unwrap();
```
DavisVaughan, line 81: Could reword this a little to use `sexp` directly in the loop and rename `inner` to `obj` here

DavisVaughan (approved):

crates/harp/src/data_frame.rs
```diff
@@ -0,0 +1,178 @@
...
+        harp::assert_class(sexp, "data.frame")?;
+
+        // SAFETY: Protected by `list`
+        let obj = RObject::view(sexp);
+
+        let dim = unsafe { harp::df_dim(obj.sexp) }?;
+        let nrow = dim.num_rows as usize;
+        let ncol = list.obj.length() as usize;
+
+        let Some(names) = obj.names() else {
```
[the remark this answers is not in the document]
lionel-: The safety note is about why we store an `RObject` view in `DataFrame.obj`.

PR #512, opened 2024-09-09
Bump tree-sitter-r 2 - Add support for `StringContent` and `EscapeSequence` nodes

DavisVaughan:
Bumps us one more commit in tree-sitter-r to support the new `string_content` and `escape_sequence` children of a `string` node https://github.com/r-lib/tree-sitter-r/compare/bc8919d3c38b816652e1e2d1a1be037cf74364cb...9d1a68f8f239bc3749a481ac85e2163e24f6362c.

These are nice tree-sitter features but are a bit annoying for us. It means that `DocumentContext` will drill allllllll the way down to one of the following, rather than finding and stopping at just a `NodeType::String` like before:
- `Anonymous("'")` or `Anonymous("\"")`
- `StringContent`
- `EscapeSequence`

This means we have to walk back up the tree a bit when we need to work with or detect the `String` node. I think I've found all the places we do this.

I'm hoping this is the most annoying tree-sitter related tweak I'll have to do.

### DIFF
--- a/crates/ark/src/lsp/completions/sources/composite/workspace.rs
+++ b/crates/ark/src/lsp/completions/sources/composite/workspace.rs
@@ -20,6 +20,7 @@ use crate::lsp::indexer;
 use crate::lsp::state::WorldState;
 use crate::lsp::traits::rope::RopeExt;
 use crate::lsp::traits::string::StringExt;
+use crate::treesitter::node_find_string;
 use crate::treesitter::NodeTypeExt;
 
 pub(super) fn completions_from_workspace(
@@ -41,8 +42,8 @@ pub(super) fn completions_from_workspace(
         }
     }
 
-    if node.is_string() {
-        log::error!("Should have already been handled by file path completions source");
+    if node_find_string(&node).is_some() {
+        log::error!("Should have already been handled by string completions source");
         return Ok(None);
     }
 
--- a/crates/ark/src/lsp/completions/sources/composite/workspace.rs
+++ b/crates/ark/src/lsp/completions/sources/composite/workspace.rs
@@ -20,7 +20,7 @@ use crate::lsp::indexer;
 use crate::lsp::state::WorldState;
 use crate::lsp::traits::rope::RopeExt;
 use crate::lsp::traits::string::StringExt;
-use crate::treesitter::node_find_string;
+use crate::treesitter::node_in_string;
 use crate::treesitter::NodeTypeExt;
 
 pub(super) fn completions_from_workspace(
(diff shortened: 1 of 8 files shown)


crates/ark/src/lsp/completions/sources/unique/file_path.rs
```diff
@@ -33,11 +35,7 @@ pub(super) fn completions_from_string_file_path(
     // NOTE: This includes the quotation characters on the string, and so
     // also includes any internal escapes! We need to decode the R string
     // before searching the path entries.
-    let token = context
-        .document
-        .contents
-        .node_slice(&context.node)?
-        .to_string();
+    let token = context.document.contents.node_slice(&node)?.to_string();
```
DavisVaughan, line 38: Can't use node directly from `context` anymore, it is likely a child of the `String` node we care about

crates/ark/src/treesitter.rs
```diff
@@ -27,6 +27,8 @@ pub enum NodeType {
     Complex,
     Float,
     String,
+    StringContent,
+    EscapeSequence,
```
DavisVaughan, line 31: New types

crates/ark/src/treesitter.rs
```diff
@@ -404,12 +410,37 @@ impl NodeTypeExt for Node<'_> {
...
+
+        while let Some(parent) = node.parent() {
+            if callback(parent) {
+                return Some(parent);
+            }
+            node = parent;
+        }
+
+        None
+    }
```
DavisVaughan, line 425: Questionably efficient but it works for this use case

crates/ark/src/lsp/completions/sources/composite/workspace.rs
```diff
@@ -41,8 +42,8 @@ pub(super) fn completions_from_workspace(
         }
     }
 
-    if node.is_string() {
-        log::error!("Should have already been handled by file path completions source");
+    if node_find_string(&node).is_some() {
```
lionel-, line 45: May be worth wrapping in `node_in_string()`?

crates/ark/src/treesitter.rs
lionel-: hmm I _think_ we don't need this method? Instead we can use the `ancestors()` method I added with the `find()` iterator method?

Or at least implement in terms of these.

lionel- (approved):

crates/ark/src/treesitter.rs
DavisVaughan: Nice I forgot about that

PR #517, opened 2024-09-10
Bump tree-sitter-r 3 - New `"open"`/`"close"` fields, can use `node.is_missing()`, no more `..i` diagnostic

DavisVaughan:
Addresses https://github.com/posit-dev/positron/issues/3632

Includes these 11 commits from tree-sitter-r https://github.com/r-lib/tree-sitter-r/compare/9d1a68f8f239bc3749a481ac85e2163e24f6362c...63ee9b10de3b1e4dfaf40e36b45e9ae3c9ed8a4f

Three notable ones:
- https://github.com/r-lib/tree-sitter-r/pull/110 (greatly simplifies some existing code around this)
- https://github.com/r-lib/tree-sitter-r/pull/112 (allows us to actually use `node.is_missing()`, fixing one of our diagnostics)
- https://github.com/r-lib/tree-sitter-r/pull/118 (fixes `..i` diagnostic)

### DIFF
--- a/crates/ark/src/lsp/completions/sources/common/subset.rs
+++ b/crates/ark/src/lsp/completions/sources/common/subset.rs
@@ -9,53 +9,21 @@ use tree_sitter::Node;
 use tree_sitter::Point;
 
 use crate::lsp::traits::point::PointExt;
-use crate::treesitter::NodeType;
-use crate::treesitter::NodeTypeExt;
 
-pub(crate) fn is_within_subset_delimiters(
-    x: &Point,
-    subset_node: &Node,
-    subset_type: &NodeType,
-) -> bool {
-    let (open, close) = match subset_type {
-        NodeType::Subset => ("[", "]"),
-        NodeType::Subset2 => ("[[", "]]"),
-        _ => std::unreachable!(),
-    };
-
+pub(crate) fn is_within_subset_delimiters(x: &Point, subset_node: &Node) -> bool {
     let Some(arguments) = subset_node.child_by_field_name("arguments") else {
         return false;
     };
 
-    let n_children = arguments.child_count();
-
-    if n_children < 2 {
-        return false;
-    }
-
-    let Some(open_node) = arguments.child(1 - 1) else {
+    let Some(open) = arguments.child_by_field_name("open") else {
         return false;
     };
-    let Some(close_node) = arguments.child(n_children - 1) else {
+    let Some(close) = arguments.child_by_field_name("close") else {
         return false;
     };
 
-    // Ensure open and closing nodes are the right type
-    if !matches!(
-        open_node.node_type(),
-        NodeType::Anonymous(kind) if kind == open
-    ) {
-        return false;
-    }
-    if !matches!(
-        close_node.node_type(),
-        NodeType::Anonymous(kind) if kind == close
-    ) {
-        return false;
-    }
-
-    let contains_start = x.is_after_or_equal(open_node.end_position());
-    let contains_end = x.is_before_or_equal(close_node.start_position());
+    let contains_start = x.is_after_or_equal(open.end_position());
+    let contains_end = x.is_before_or_equal(close.start_position());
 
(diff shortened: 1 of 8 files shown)


crates/ark/src/lsp/diagnostics.rs
```diff
@@ -863,38 +863,52 @@ fn check_unclosed_arguments(
...
+        return Ok(false);
     };
 
-    let arguments = unwrap!(node.child_by_field_name("arguments"), None => {
-        return false.ok();
-    });
+    let token = match node.node_type() {
+        NodeType::Call => "(",
+        NodeType::Subset => "[",
+        NodeType::Subset2 => "[[",
```
DavisVaughan, line 873: I'm pretty sure this diagnostic was never triggering before!

crates/ark/src/lsp/diagnostics.rs
```diff
@@ -1074,6 +1115,18 @@ mod tests {
...
+    fn test_no_diagnostic_for_dot_dot_i() {
+        r_test(|| {
+            let text = "..1 + ..2 + 3";
+            let document = Document::new(text, None);
+
+            let diagnostics = generate_diagnostics(document, DEFAULT_STATE.clone());
+
+            assert!(diagnostics.is_empty());
+        })
+    }
```
DavisVaughan, line 1128: Yay

crates/ark/src/lsp/completions/sources/common/subset.rs
```diff
@@ -9,53 +9,21 @@ use tree_sitter::Node;
...
-    subset_node: &Node,
-    subset_type: &NodeType,
-) -> bool {
-    let (open, close) = match subset_type {
-        NodeType::Subset => ("[", "]"),
-        NodeType::Subset2 => ("[[", "]]"),
-        _ => std::unreachable!(),
-    };
 
+pub(crate) fn is_within_subset_delimiters(x: &Point, subset_node: &Node) -> bool {
```
lionel-, line 13: Nice simplification

crates/ark/src/lsp/diagnostics.rs
```diff
@@ -863,38 +863,52 @@ fn check_unclosed_arguments(
...
+        NodeType::Subset2 => "[[",
+        _ => return Ok(false),
+    };
 
-    let n = arguments.child_count();
-    if n == 0 {
-        return false.ok();
+    let range = open.range();
+    let range = convert_tree_sitter_range_to_lsp_range(context.contents, range);
+    let message = format!("unmatched opening token '{token}'");
```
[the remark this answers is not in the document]
DavisVaughan: This gets reworked later in #523

crates/ark/src/lsp/signature_help.rs
```diff
@@ -329,21 +329,12 @@ fn is_within_call_parentheses(x: &Point, node: &Node) -> bool {
...
-
-    let open = arguments.child(1 - 1).unwrap();
-    let close = arguments.child(n_children - 1).unwrap();
-
-    if open.node_type() != NodeType::Anonymous(String::from("(")) {
-        return false;
-    }
-    if close.node_type() != NodeType::Anonymous(String::from(")")) {
+    };
+    let Some(close) = arguments.child_by_field_name("close") else {
```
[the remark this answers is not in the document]
DavisVaughan: I _think_ I'm leaning towards trying to avoid too much usage of `is_missing()` in the semantic paths for now. Hopefully Rowan makes this a non issue, and I don't think it is really going to hurt much as is

lionel-:
I forgot these notes:

> If a syntax error spans > 20 lines, it is now truncated to only show a squiggle on the start_point of the range, and it states that this is a Syntax error. Starts here and ends on row {row}

I like it!

> we only consider running semantic diagnostics down a section of the tree if we know that section of the tree does not contain any syntax errors.

Seems reasonable, and solves the problem of false positives semantic diagnostics due to syntax errors.

PR #519, opened 2024-09-10
Data Explorer: Add basic implementation of "not contains" search type

wesm:
Ark side of https://github.com/posit-dev/positron/pull/4633. The unit tests here for these filters are a bit janky and don't seem to test correctness so I didn't add a test right way. It might make sense to give them a complete working over.


dfalbel (approved):
LGTM!

PR #520, opened 2024-09-11
Use standard style in `handle_dotty_assignment()`

lionel-:
What do you think about this more standard style @kevinushey?

- Only create errors when something goes wrong (@DavisVaughan mentions this avoids performance issues with backtrace creation in some cases).

- Use standard constructors like `Ok()` and `Some()`.

-  If we're not interested in a value except for exiting, use explicit exits instead of `?`. I feel like the latter is the equivalent of doing `predicate() || return()` instead of the clearer `if (predicate()) return()`

### DIFF
--- a/crates/ark/src/lsp/diagnostics.rs
+++ b/crates/ark/src/lsp/diagnostics.rs
@@ -171,34 +171,44 @@ fn is_within_magrittr_pipe(node: Node, context: &mut DiagnosticContext) -> bool 
     }
 }
 
-fn handle_dotty_assignment(lhs: Node, context: &mut DiagnosticContext) -> Result<()> {
+fn handle_dotty_assignment(lhs: Node, context: &mut DiagnosticContext) -> Option<()> {
     // Check that the lhs is a subset call of the form `.[]`.
-    lhs.is_subset().into_result()?;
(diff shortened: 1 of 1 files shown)


crates/ark/src/lsp/diagnostics.rs
```diff
@@ -171,42 +171,52 @@ fn is_within_magrittr_pipe(node: Node, context: &mut DiagnosticContext) -> bool
...
+    if !lhs.is_subset() {
+        return None;
+    };
+
+    let function = lhs.child_by_field_name("function")?;
+    if !function.is_identifier_or_string() {
+        return None;
+    };
+
+    let value = context.contents.node_slice(&function).ok()?.to_string();
```
DavisVaughan, line 185: I have a slight preference for actually logging the potential error here and manually doing `return None`, like in an `Err` match arm

DavisVaughan: Ah I see how often you use it below though, hmm.

DavisVaughan (approved):
I generally find this much easier to read

kevinushey (approved):
LGTM! I wasn't aware (or had forgotten?) about the potential performance costs of backtrace creation as welll.

PR #522, opened 2024-09-13
Detect boundaries of parse inputs

lionel-:
Progress towards https://github.com/posit-dev/positron/issues/1326.

`parse_boundaries()` uses the R parser to detect the line boundaries of:

- Complete inputs (zero or more)
- Incomplete inputs (optional)
- Invalid inputs (optional)

The boundaries are for _lines of inputs_ rather than _expressions_. For instance, `foo; bar` has one input whereas `foo\nbar` has two inputs.

Invariants:
- Empty lines and lines containing only whitespace or comments are complete inputs.
- There is always at least one range since the empty string is a complete input.
- The ranges are sorted and non-overlapping.
- Inputs are classified in complete, incomplete, and error sections. The
  sections are sorted in this order (there cannot be complete inputs after
  an incomplete or error one, there cannot be an incomplete input after
  an error one, and error inputs are always trailing).
- There is only one incomplete and one error input in a set of inputs.

Approach:

I originally thought I'd use the parse data artifacts created by side effects to detect boundaries of complete expressions in case of incomplete and invalid inputs (see infrastructure implemented in #508). The goal was to avoid non-linear performance as the number of lines increases. I didn't end up doing that because the parse data is actually unusable for more complex cases than what I tried during my exploration.

Instead, I had to resort to parsing line by line. I start with the entire set of lines and back up one line at a time until the parse fully completes. Along the way I keep track of the error and incomplete sections of the input. In the most common cases (valid inputs, short incomplete input at the end), this should only require one or a few iterations. The boundaries of complete expressions are retrieved from the source references of the parsed expressions (using infrastructure from #482) and then transformed to complete inputs.

Supporting infrastructure:

- New `CharacterVector::slice()` method that returns a `&[SEXP]` and a corresponding `CharacterVector::TryFrom<&[SEXP]>` method. (Eventually `List` should gain those as well.) These methods make it easy to slice character vectors from the Rust side. I use this to shorten the vector of lines without having to start from a `&str` and reallocate `CHARSXP`s everytime.

- `SrcFile` can now be constructed from a `CharacterVector` with `TryFrom`.

- `Vector::create()` is no longer `unsafe`.

- `ArkRange::TryFrom<SrcRef>` method, although I didn't end up using it.

### DIFF
--- a/crates/ark/src/analysis/parse_boundaries.rs
+++ b/crates/ark/src/analysis/parse_boundaries.rs
@@ -10,6 +10,7 @@ use harp::vector::Vector;
 use harp::ParseResult;
 use harp::RObject;
 
+/// Boundaries are ranges over lines of text.
 #[derive(Debug)]
 pub struct ParseBoundaries {
     pub complete: Vec<std::ops::Range<usize>>,
@@ -17,8 +18,28 @@ pub struct ParseBoundaries {
     pub error: Option<std::ops::Range<usize>>,
 }
 
+/// Parse boundaries of R inputs
+///
+/// Takes a string of R code and detects which lines parse as complete,
+/// incomplete, and error inputs.
+///
+/// Invariants:
+/// - There is always at least one range as the empty string is a complete input.
+/// - The ranges are sorted and non-overlapping.
+/// - Inputs are classified in complete, incomplete, and error sections. The
+///   sections are sorted in this order (there cannot be complete inputs after
+///   an incomplete or error one, there cannot be an incomplete input after
+///   an error one, and error inputs are always trailing).
+/// - There is only one incomplete and one error input in a set of inputs.
 pub fn parse_boundaries(text: &str) -> anyhow::Result<ParseBoundaries> {
-    let lines: Vec<&str> = text.lines().collect();
+    let mut lines: Vec<&str> = text.lines().collect();
+
+    // Rectify for `lines()` ignoring trailing empty lines
+    match text.chars().last() {
+        Some(last) if last == '\n' => lines.push(""),
+        None => lines.push(""),
+        _ => {},
+    }
 
     // Create a duplicate vector of lines on the R side too so we don't have to
     // reallocate memory each time we parse a new subset of lines
@@ -55,11 +76,8 @@ pub fn parse_boundaries(text: &str) -> anyhow::Result<ParseBoundaries> {
     for current_line in (0..n_lines).rev() {
         let mut record_complete = |exprs: RObject| -> anyhow::Result<()> {
             let srcrefs = exprs.srcrefs()?;
-            let ranges: Vec<std::ops::Range<usize>> =
+            let mut ranges: Vec<std::ops::Range<usize>> =
                 srcrefs.into_iter().map(|srcref| srcref.line).collect();
-
-            // Merge expressions separated by semicolons
-            let mut ranges = merge_overlapping(ranges);
 
             complete.append(&mut ranges);
             Ok(())
@@ -104,11 +122,19 @@ pub fn parse_boundaries(text: &str) -> anyhow::Result<ParseBoundaries> {
     record_incomplete(0, &incomplete_end);
     record_error(0, &error_end);
 
-    Ok(ParseBoundaries {
+    // Merge expressions separated by semicolons
+    let complete = merge_overlapping(complete);
+
+    let boundaries = ParseBoundaries {
         complete,
         incomplete,
         error,
-    })
+    };
+
+    // Fill any gaps with one-liner complete expressions
+    let boundaries = fill_gaps(boundaries, n_lines);
+
+    Ok(boundaries)
 }
 
 fn merge_overlapping<T>(ranges: Vec<std::ops::Range<T>>) -> Vec<std::ops::Range<T>>
@@ -136,6 +162,66 @@ where
     ranges.into_iter().fold(vec![], merge)
 }
 
+// Fill any gaps with one-liner complete expressions. This applies
+// to empty lines, with or without spaces and tabs, and possibly
+// with a trailing comment.
+fn fill_gaps(mut boundaries: ParseBoundaries, n_lines: usize) -> ParseBoundaries {
+    let mut filled = vec![];
+    let ranges = boundaries.complete;
+
+    let mut last_line = 0;
+
+    let range_from = |start| std::ops::Range {
+        start,
+        end: start + 1,
+    };
+
+    // Fill leading whitespace with empty input ranges
+    if let Some(first) = ranges
+        .get(0)
+        .or(boundaries.incomplete.as_ref())
+        .or(boundaries.error.as_ref())
+    {
+        for start in 0..first.start {
+            let range = range_from(start);
+            last_line = range.end;
+            filled.push(range)
+        }
+    }
+
+    // Fill gaps between complete expressions
+    {
+        let iter = ranges.into_iter();
+        for range in iter {
+            // We found a gap, fill ranges for lines in that gap
+            if !range.contains(&last_line) {
+                for start in last_line..range.start {
+                    filled.push(range_from(start))
+                }
+            }
+
+            last_line = range.end;
+            filled.push(range);
+        }
+    }
+
+    // Fill trailing whitespace (between complete and incomplete|error\eof)
+    let last_boundary = filled.last().map(|r| r.end).unwrap_or(0);
+    let next_boundary = boundaries
+        .incomplete
+        .as_ref()
+        .or(boundaries.error.as_ref())
+        .map(|r| r.start)
+        .unwrap_or(n_lines);
+
+    for start in last_boundary..next_boundary {
+        filled.push(range_from(start))
+    }
+
+    boundaries.complete = filled;
+    boundaries
+}
+
 #[cfg(test)]
 mod tests {
     use harp::assert_match;
@@ -146,30 +232,53 @@ mod tests {
     #[test]
     fn test_parse_boundaries_complete() {
         r_test(|| {
-            let boundaries = parse_boundaries("").unwrap();
-            let expected_complete: Vec<std::ops::Range<usize>> = vec![];
-            assert_eq!(boundaries.complete, expected_complete);
+            let boundaries = parse_boundaries("foo").unwrap();
+            #[rustfmt::skip]
+            assert_eq!(boundaries.complete, vec![
+                std::ops::Range { start: 0, end: 1 },
+            ]);
             assert_match!(boundaries.incomplete, None);
             assert_match!(boundaries.error, None);
 
-            let boundaries = parse_boundaries("\n\n  ").unwrap();
-            let expected_complete: Vec<std::ops::Range<usize>> = vec![];
-            assert_eq!(boundaries.complete, expected_complete);
+            let boundaries = parse_boundaries("foo\nbarbaz  ").unwrap();
+            assert_eq!(boundaries.complete, vec![
+                std::ops::Range { start: 0, end: 1 },
+                std::ops::Range { start: 1, end: 2 }
+            ]);
+            assert_match!(boundaries.incomplete, None);
+            assert_match!(boundaries.error, None);
+        })
+    }
+
+    #[test]
+    fn test_parse_boundaries_whitespace() {
+        r_test(|| {
+            let boundaries = parse_boundaries("").unwrap();
+            #[rustfmt::skip]
+            assert_eq!(boundaries.complete, vec![
+                std::ops::Range { start: 0, end: 1 },
+            ]);
+            assert_match!(boundaries.incomplete, None);
+            assert_match!(boundaries.error, None);
+
+            let boundaries = parse_boundaries("\n\n  \n").unwrap();
+            assert_eq!(boundaries.complete, vec![
+                std::ops::Range { start: 0, end: 1 },
+                std::ops::Range { start: 1, end: 2 },
+                std::ops::Range { start: 2, end: 3 },
+                std::ops::Range { start: 3, end: 4 },
+            ]);
             assert_match!(boundaries.incomplete, None);
             assert_match!(boundaries.error, None);
 
             let boundaries = parse_boundaries("\n  foo\n  \n\n").unwrap();
-            let expected_complete = vec![std::ops::Range { start: 1, end: 2 }];
-            assert_eq!(boundaries.complete, expected_complete);
-            assert_match!(boundaries.incomplete, None);
-            assert_match!(boundaries.error, None);
-
-            let boundaries = parse_boundaries("foo\nbarbaz  \n").unwrap();
-            let expected_complete = vec![std::ops::Range { start: 0, end: 1 }, std::ops::Range {
-                start: 1,
-                end: 2,
-            }];
-            assert_eq!(boundaries.complete, expected_complete);
+            assert_eq!(boundaries.complete, vec![
+                std::ops::Range { start: 0, end: 1 },
+                std::ops::Range { start: 1, end: 2 },
+                std::ops::Range { start: 2, end: 3 },
+                std::ops::Range { start: 3, end: 4 },
+                std::ops::Range { start: 4, end: 5 },
+            ]);
             assert_match!(boundaries.incomplete, None);
             assert_match!(boundaries.error, None);
         })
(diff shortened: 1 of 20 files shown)


lionel-:
TODO: 

Error section should contain line number and error message

Whitespace/comment inputs should be tagged with a boolean

crates/harp/src/object.rs
```diff
@@ -248,6 +248,10 @@ pub fn r_dbl_begin(x: SEXP) -> *mut f64 {
     unsafe { REAL(x) }
 }
 
+pub unsafe fn chr_cbegin(x: SEXP) -> *const SEXP {
+    libr::DATAPTR_RO(x) as *const SEXP
+}
```
DavisVaughan, line 253: Everything else here is safe. Do you want to eventually make these `unsafe` again? If so I'd propose keeping it safe for consistency for now and then we move away from it (but to me this feels like it should be safe)

crates/harp/src/parse.rs
```diff
@@ -98,7 +116,7 @@ pub fn parse_status<'a>(input: &ParseInput<'a>) -> crate::Result<ParseResult> {
         match status {
             libr::ParseStatus_PARSE_OK => Ok(ParseResult::Complete(result)),
             libr::ParseStatus_PARSE_INCOMPLETE => Ok(ParseResult::Incomplete),
-            libr::ParseStatus_PARSE_ERROR => Err(crate::Error::ParseSyntaxError {
+            libr::ParseStatus_PARSE_ERROR => Ok(ParseResult::SyntaxError {
```
DavisVaughan, line 119: Oh that's a great idea

crates/harp/src/parser/srcref.rs
```diff
@@ -118,11 +118,10 @@ impl TryFrom<RObject> for SrcRef {
 /// Creates the same sort of srcfile object as with `parse(text = )`.
 /// Takes code as an R string containing newlines, or as a R vector of lines.
 impl SrcFile {
-    pub fn new_virtual(text: &str) -> harp::Result<Self> {
-        let input = crate::as_parse_text(text);
+    pub fn new_virtual(text: RObject) -> harp::Result<Self> {
```
DavisVaughan, line 121:
```suggestion
    fn new_virtual(text: RObject) -> harp::Result<Self> {
```

It feels like you'll want to make this with `try_from()` now? Even if you are coming from an `RObject` (which we don't have a `try_from()` method for, but we could.

crates/harp/src/parse.rs
```diff
@@ -76,9 +79,24 @@ pub fn parse_exprs_ext<'a>(input: &ParseInput<'a>) -> crate::Result<RObject> {
...
+pub fn parse_with_parse_data(text: &str) -> crate::Result<(ParseResult, ParseData)> {
+    let srcfile = srcref::SrcFile::try_from(text)?;
+
+    // Fill parse data in `srcfile` by side effect
+    let status = parse_status(&ParseInput::SrcFile(&srcfile))?;
+
+    let parse_data = ParseData::from_srcfile(&srcfile)?;
+
+    Ok((status, parse_data))
+}
```
DavisVaughan, line 98: - This method does not seem to be used
- The leading comment about `Return type would be clearer` seems outdated because syntax error is integrated into `ParseResult` now I think

crates/harp/src/parser/parse_data.rs
```diff
@@ -108,11 +113,23 @@ impl ParseData {
...
 }
 
+impl ParseDataNode {
+    pub fn as_point_range(&self) -> std::ops::Range<(usize, usize)> {
+        std::ops::Range {
+            start: (self.line.start, self.column.start),
+            end: (self.line.end, self.column.end),
+        }
+    }
+}
```
DavisVaughan, line 131: Not used?

Also feels like it should return a typed `ParseDataRange` to be more self documenting

(I've noticed that the ruff team uses custom types for almost _everything_, and I really think it makes the code easier to read)

crates/harp/src/vector/character_vector.rs
```diff
@@ -17,13 +17,24 @@ use libr::STRING_ELT;
...
 pub struct CharacterVector {
-    object: RObject,
+    pub object: RObject,
+}
+
+impl CharacterVector {
+    pub fn slice(&self) -> &[SEXP] {
+        unsafe {
+            let data = harp::chr_cbegin(self.object.sexp);
+            std::slice::from_raw_parts(data, self.len())
```
DavisVaughan, line 35: Cool that this does `len * mem::size_of::<T>()` internally automatically

crates/harp/src/vector/character_vector.rs
```diff
@@ -94,6 +107,35 @@ impl Vector for CharacterVector {
...
+            let vec = Self::with_length(value.len());
+            let sexp = vec.object.sexp;
+            let mut iter = value.into_iter();
+
+            for i in 0..value.len() {
+                let value = iter.next().unwrap_unchecked();
+                r_assert_type(*value, &[libr::CHARSXP])?;
+
+                r_chr_poke(sexp, i as R_xlen_t, *value);
+            }
```
DavisVaughan, line 124:
```suggestion

            for (i, elt) in value.into_iter().enumerate() {
                r_assert_type(*elt, &[libr::CHARSXP])?;
                r_chr_poke(sexp, i as R_xlen_t, *elt);
            }
```

DavisVaughan (approved):

crates/ark/src/analysis/parse_boundaries.rs
```diff
@@ -0,0 +1,375 @@
...
+            let srcrefs = exprs.srcrefs()?;
+            let mut ranges: Vec<std::ops::Range<usize>> =
+                srcrefs.into_iter().map(|srcref| srcref.line).collect();
+
+            complete.append(&mut ranges);
+            Ok(())
+        };
+
+        // Grab all code up to current line
+        let subset = &lines_r.slice()[..current_line + 1];
```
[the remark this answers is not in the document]
DavisVaughan: Oh oops, maybe I'm wrong. I now realize that `(0..n_lines)` in the for loop line would expand to `[0, ..., n_lines - 1]` and then everything should work fine

crates/ark/src/analysis/parse_boundaries.rs
```diff
@@ -0,0 +1,375 @@
...
+        let subset = CharacterVector::try_from(subset)?;
```
DavisVaughan, line 88: I dislike that you have to do a (potentially large) redundant allocation on the 1st iteration, which most of the time is all you need to do (i.e. no incomplete and no error)

DavisVaughan: Since you only reference `&subset` below, it does seem like you could add a special case for the first iteration that remapped `lines_r` directly as `subset` so you don't need that first allocation

I know that is uglier but it might be worth it

DavisVaughan: How about this, with a `let mut first = true;` outside the loop?

```rust
        // Parse within source file to get source references.
        // Avoid allocation on the first iteration, since often there are no issues.
        let srcfile = if first {
            first = false;
            harp::srcref::SrcFile::try_from(&lines_r)?
        } else {
            // Grab all code up to current line
            let subset = &lines_r.slice()[..current_line + 1];
            let subset = CharacterVector::try_from(subset)?;
            harp::srcref::SrcFile::try_from(&subset)?
        };
```

crates/ark/src/analysis/parse_boundaries.rs
```diff
@@ -0,0 +1,375 @@
...
+            } else {
+                // No overlap, push a new range
+                merged.push(range);
+            }
+        } else {
+            // First element, just add it
+            merged.push(range);
+        }
+
+        merged
```
DavisVaughan, line 159:
```suggestion
        let Some(last) = merged.last_mut() else {
            // First element, just add it
            merged.push(range);
            return merged;
        };

        // if range.start <= last.end {
        if last.contains(&range.start) {
            // Overlap, merge with last range
            last.end = last.end.max(range.end);
        } else {
            // No overlap, push a new range
            merged.push(range);
        }

        merged
```

I felt like it was slightly easier to understand if you get the early exit out of the way

crates/ark/src/analysis/parse_boundaries.rs
```diff
@@ -0,0 +1,375 @@
...
+            if !range.contains(&last_line) {
+                for start in last_line..range.start {
+                    filled.push(range_from(start))
+                }
+            }
+
+            last_line = range.end;
+            filled.push(range);
+        }
+    }
```
DavisVaughan, line 206:
```suggestion
    // Fill gaps between complete expressions
    for range in ranges.into_iter() {
        // We found a gap, fill ranges for lines in that gap
        if !range.contains(&last_line) {
            for start in last_line..range.start {
                filled.push(range_from(start))
            }
        }

        last_line = range.end;
        filled.push(range);
    }
```

Doesn't seem like you need the extra scope

crates/ark/src/analysis/parse_boundaries.rs
```diff
@@ -0,0 +1,375 @@
...
+    let next_boundary = boundaries
+        .incomplete
+        .as_ref()
+        .or(boundaries.error.as_ref())
+        .map(|r| r.start)
+        .unwrap_or(n_lines);
+
+    for start in last_boundary..next_boundary {
+        filled.push(range_from(start))
+    }
```
DavisVaughan, line 219: I guess there can't be trailing whitespace between incomplete/error and eof?

Like, this adds ranges for lines from `[end_last_complete, start_incomplete_or_error]`, but what about `[end_incomplete_or_error, eof]`? I guess `end_incomplete_or_error == eof`?

crates/ark/src/analysis/parse_boundaries.rs
```diff
@@ -0,0 +1,375 @@
...
+    use crate::analysis::parse_boundaries::*;
+    use crate::test::r_test;
+
+    #[test]
+    fn test_parse_boundaries_complete() {
+        r_test(|| {
+            let boundaries = parse_boundaries("foo").unwrap();
+            #[rustfmt::skip]
+            assert_eq!(boundaries.complete, vec![
+                std::ops::Range { start: 0, end: 1 },
```
DavisVaughan, line 238: A custom type would also let us add a `new()` method for something like `LineRange::new(0, 1)`

crates/ark/src/analysis/parse_boundaries.rs
```diff
@@ -0,0 +1,375 @@
...
+use harp::vector::Vector;
+use harp::ParseResult;
+use harp::RObject;
+
+/// Boundaries are ranges over lines of text.
+#[derive(Debug)]
+pub struct ParseBoundaries {
+    pub complete: Vec<std::ops::Range<usize>>,
+    pub incomplete: Option<std::ops::Range<usize>>,
+    pub error: Option<std::ops::Range<usize>>,
```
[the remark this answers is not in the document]
DavisVaughan: I'm just now realizing a better name for this is probably really `LineRange` (`TextRange` is for a byte range in a document)

The fact that it took me until the test section of the PR to figure this out probably does mean a named wrapper like this would be very helpful

```rust
struct LineRange {
  start: usize,
  end: usize,
}
```

crates/ark/src/analysis/parse_boundaries.rs
```diff
@@ -0,0 +1,375 @@
...
+                start: 0,
+                end: 1
+            }]);
+            assert_match!(
+                boundaries.incomplete,
+                Some(std::ops::Range { start: 1, end: 2 })
+            );
+            assert_eq!(boundaries.error, Some(std::ops::Range { start: 2, end: 3 }));
+        });
+    }
```
DavisVaughan, line 374: I didn't see a test quite like this where there are only incomplete and error


```rust
            let boundaries = parse_boundaries("foo +\n  bar +;").unwrap();
            assert_eq!(boundaries.complete.len(), 0);
            assert_eq!(
                boundaries.incomplete,
                Some(std::ops::Range { start: 0, end: 1 })
            );
            assert_eq!(boundaries.error, Some(std::ops::Range { start: 1, end: 2 }));
```

crates/harp/src/object.rs
lionel-: hmm I think it's unsafe because it's taking an untyped `SEXP` and it doesn't return any `Result`.

We might argue that it's fine because it throws R errors if the type isn't right (but note that it's an implementation detail that it does so instead of crashing like other accessors like `CAR()`).

But I think the criterion for a safe function is that it guarantees a crash or an R error can't happen (either by dynamic checks feeding the `Err` path of the `Result` or by taking typed inputs with strong guarantees such that errors or crashes can't happen).

crates/harp/src/parser/parse_data.rs
lionel-: In general I think it's a balance because too much vocabulary can also make code opaque instead of transparent.

But I'm willing to try and make more type aliases to see how this feels.

crates/ark/src/analysis/parse_boundaries.rs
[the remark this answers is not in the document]
lionel-: You're right documenting the contents with a type name would go a long way here.

`ArkRange` is Ark specific rather thant LSP- or TS-specific. It uses the row/col approach instead of byte offsets.

For byte offsets and ranges we may want to use https://github.com/rust-analyzer/text-size/

But that's storing u32 instead of usize, which makes it harder to interface with the rest of Rust, and we might not need

crates/ark/src/analysis/parse_boundaries.rs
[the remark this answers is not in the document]
lionel-: I'll have to get used to that syntax :-)

crates/ark/src/analysis/parse_boundaries.rs
lionel-: Right I think I had a `mut` in there at some point that I wanted a scope for, but it's no longer there.

crates/ark/src/analysis/parse_boundaries.rs
lionel-: I'll make the comment clearer, it's trying to say exactly what you say, the trailing whitespace that needs to be transformed to complete inputs is between the complete expressions and the rest.

crates/harp/src/parser/srcref.rs
lionel-: hmm I'm not sure, it feels strange not to have "official" constructors. But I'll make this ctor private.

crates/harp/src/parse.rs
lionel-: It's no longer used, neither is `ParseData`, but I've left those for later in case they are needed. E.g. maybe it'll be useful to generate rowan trees from the R parser in a more efficient way than traversing R AST.

crates/harp/src/vector/character_vector.rs
lionel-: oops, leftover

crates/ark/src/analysis/parse_boundaries.rs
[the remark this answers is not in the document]
lionel-: yep `current_line` is 9 the first time around in your example

crates/ark/src/analysis/parse_boundaries.rs
lionel-: With the proposed approach we allocate one big string that is then split by R into a vector of lines (see R implementation of `srcfilecopy()`).

We could avoid the unnecessary splicing though. I made it:

```rust
        // Grab all code up to current line. We don't slice the vector in the
        // first iteration as it's not needed.
        let subset = if current_line == n_lines - 1 {
            CharacterVector::try_from(&lines_r.slice()[..=current_line])?
        } else {
            lines_r.clone()
        };
```

To support this, `CharacterVector` is now clonable (it does not deep-copy since none of our wrappers make owning guarantees for the wrapped data).

crates/ark/src/analysis/parse_boundaries.rs
lionel-: Good point though that's arguably straying away from the standard style.

crates/ark/src/analysis/parse_boundaries.rs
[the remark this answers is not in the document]
lionel-: These types make it harder to provide generic algorithms, e.g. for `fn merge_overlapping<T>(ranges: Vec<std::ops::Range<T>>) -> Vec<std::ops::Range<T>>`. Not that we really need the genericity now...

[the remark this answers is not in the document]
lionel-: ah but the ruff types are directly imported from rust-analyzer's text-size crate!

PR #524, opened 2024-09-16
Truncate error calls and backtraces

lionel-:
The logs in https://github.com/posit-dev/positron/issues/4686 can't be read because the error calls and backtraces include 1000s of lines of inlined data.

To prevent this, we now truncate error calls to ~10~ 50 lines and R backtraces to 500 lines.

### DIFF
--- a/crates/harp/src/error.rs
+++ b/crates/harp/src/error.rs
@@ -101,7 +101,7 @@ impl fmt::Display for Error {
                 ..
             } => {
                 if let Some(code) = code {
-                    let code = truncate_lines(code.clone(), 10);
+                    let code = truncate_lines(code.clone(), 50);
                     write!(f, "Error evaluating '{code}': {message}")?;
                 } else {
                     write!(f, "{message}")?;
--- a/crates/harp/src/error.rs
+++ b/crates/harp/src/error.rs
@@ -101,6 +101,7 @@ impl fmt::Display for Error {
                 ..
             } => {
                 if let Some(code) = code {
+                    let code = truncate_lines(code.clone(), 10);
                     write!(f, "Error evaluating '{code}': {message}")?;
                 } else {
                     write!(f, "{message}")?;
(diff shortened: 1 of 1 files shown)


crates/harp/src/error.rs
```diff
@@ -266,3 +268,15 @@ impl From<Utf8Error> for Error {
...
 }
+
+fn truncate_lines(text: String, max_lines: usize) -> String {
+    let n_lines = text.lines().count();
+    if n_lines <= max_lines {
+        return text;
+    }
+
+    let mut text: String = text.lines().take(max_lines).collect();
+    text.push_str(&format!("... *Truncated {} lines*", n_lines - max_lines));
```
DavisVaughan, line 279: Should there be a leading `\n` to put this on its own line?

crates/harp/src/error.rs
```diff
@@ -101,6 +101,7 @@ impl fmt::Display for Error {
                 ..
             } => {
                 if let Some(code) = code {
+                    let code = truncate_lines(code.clone(), 10);
```
DavisVaughan, line 104:
```suggestion
                    let code = truncate_lines(code, 10);
```

crates/harp/src/error.rs
```diff
@@ -111,6 +112,7 @@ impl fmt::Display for Error {
                 // https://users.rust-lang.org/t/why-doesnt-anyhows-debug-formatter-include-the-underlying-debug-formatting/44227
 
                 if !r_trace.is_empty() {
+                    let r_trace = truncate_lines(r_trace.clone(), 500);
```
DavisVaughan, line 115:
```suggestion
                    let r_trace = truncate_lines(r_trace, 500);
```

crates/harp/src/error.rs
DavisVaughan, line 104:
```suggestion
                    let code = truncate_lines(code.clone(), 50);
```

I would be ok with this too

DavisVaughan (approved):

crates/harp/src/error.rs
lionel-: yep!

PR #526, opened 2024-09-16
Protect against errors in `r_size()`

lionel-:
Addresses the crash part of https://github.com/posit-dev/positron/issues/4686 but not the underlying error. It's now turned into a log warning:

Screenshot 2024-09-16 at 12 02 46

### DIFF
--- a/crates/harp/src/size.rs
+++ b/crates/harp/src/size.rs
@@ -30,13 +30,10 @@ use crate::r_typeof;
 pub fn r_size(x: SEXP) -> harp::Result<usize> {
     let mut seen: HashSet<SEXP> = HashSet::new();
 
-    let sizeof_node: f64 = harp::parse_eval0(
-        "as.vector(utils::object.size(quote(expr = )))",
-        R_ENVS.global,
-    )
-    .and_then(|x| x.try_into())?;
+    let sizeof_node: f64 = harp::parse_eval_base("as.vector(utils::object.size(quote(expr = )))")
+        .and_then(|x| x.try_into())?;
 
-    let sizeof_vector: f64 = harp::parse_eval_global("as.vector(utils::object.size(logical()))")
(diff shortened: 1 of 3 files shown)


lionel-:
I'm not happy with `r_task()` panicking like that but as we progress towards our goal of eliminating it, it will become less and less costly to make it return a `Result` that the callers have to handle. cc @DavisVaughan

crates/harp/src/size.rs
```diff
@@ -27,28 +27,28 @@ use crate::r_typeof;
...
 // variables pane which required more performance.
 // See for more info.
-pub fn r_size(x: SEXP) -> usize {
+pub fn r_size(x: SEXP) -> harp::Result<usize> {
     let mut seen: HashSet<SEXP> = HashSet::new();
 
     let sizeof_node: f64 = harp::parse_eval0(
         "as.vector(utils::object.size(quote(expr = )))",
         R_ENVS.global,
     )
```
DavisVaughan, line 36: while you're here, just make this `parse_eval_global()` to match `sizeof_vector`?

dfalbel (approved):

crates/harp/src/size.rs
```diff
@@ -27,28 +27,28 @@ use crate::r_typeof;
...
-        x,
-        R_ENVS.global,
-        sizeof_node as usize,
-        sizeof_vector as usize,
-        &mut seen,
-        0,
-    )
+        .and_then(|x| x.try_into())?;
+
+    harp::try_catch(|| {
```
DavisVaughan, line 42: I think this needs a comment about the try-catch, something about the fact that `obj_size_tree()` has the potential to violate some assumptions about R internals

DavisVaughan (approved):

crates/harp/src/size.rs
lionel-: I made it `parse_eval_base()` which is more adequate here.

PR #530, opened 2024-09-17
Fix `Sys.getenv()` completions, support `Sys.unsetenv()` too

DavisVaughan:
Addresses https://github.com/posit-dev/positron/issues/4677

I think I accidentally broke this with https://github.com/posit-dev/ark/pull/439 because I didn't know we were relying on the argument labels _only_ containing the argument name (they now also contain the default value where applicable).

I _feel_ like the LSP `SignatureHelp` data structure is probably not completely appropriate for what we want here since it is used for UI purposes, and we need something that is still useful for static analysis in other places besides the LSP request handler. But I'm not ready to shave that yak right now.

I've added lots of tests for all the custom completion types we support, so we can't accidentally break this in the future.

### DIFF
--- a/crates/ark/src/lsp/completions/sources/unique/custom.rs
+++ b/crates/ark/src/lsp/completions/sources/unique/custom.rs
@@ -235,7 +235,7 @@ mod tests {
     fn test_completion_custom_library() {
         r_test(|| {
             let n_packages = {
-                let n = harp::parse_eval_global("length(base::.packages(TRUE))").unwrap();
+                let n = harp::parse_eval_base("length(base::.packages(TRUE))").unwrap();
                 let n = i32::try_from(n).unwrap();
                 usize::try_from(n).unwrap()
             };
@@ -267,7 +267,7 @@ mod tests {
         r_test(|| {
             let name = "ARK_TEST_ENVVAR";
 
-            harp::parse_eval_global(format!("Sys.setenv({name} = '1')").as_str()).unwrap();
+            harp::parse_eval_base(format!("Sys.setenv({name} = '1')").as_str()).unwrap();
 
             let assert_has_ark_test_envvar_completion = |text: &str, point: Point| {
                 let document = Document::new(text, None);
@@ -307,7 +307,7 @@ mod tests {
             let completions = completions_from_custom_source(&context).unwrap();
             assert!(completions.is_none());
 
-            harp::parse_eval_global(format!("Sys.unsetenv('{name}')").as_str()).unwrap();
+            harp::parse_eval_base(format!("Sys.unsetenv('{name}')").as_str()).unwrap();
         })
     }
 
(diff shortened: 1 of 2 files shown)


crates/ark/src/lsp/completions/sources/unique/custom.rs
```diff
@@ -106,6 +106,13 @@ pub fn completions_from_custom_source_impl(
         tower_lsp::lsp_types::ParameterLabel::Simple(string) => string,
     };
 
+    // Argument text typically contains the argument name and its default value if there is one.
+    // Extract out just the argument name for matching purposes.
+    let argument = match argument.find("=") {
+        Some(byte) => &argument[..byte].trim(),
```
lionel-, line 112: Minor but this is clearer to me?

```suggestion
        Some(loc) => &argument[..loc].trim(),
```

crates/ark/src/lsp/completions/sources/unique/custom.rs
```diff
@@ -106,6 +106,13 @@ pub fn completions_from_custom_source_impl(
         tower_lsp::lsp_types::ParameterLabel::Simple(string) => string,
     };
 
+    // Argument text typically contains the argument name and its default value if there is one.
+    // Extract out just the argument name for matching purposes.
+    let argument = match argument.find("=") {
```
lionel-, line 111: I'd rename `let argument` to `let formal` or `let param` since "argument" normally refers to the value passed in.

lionel-: oh I see above we use "parameter" which is clearer. Why did we go from param to arg terminology? Since we're analysing a function rather than a function call, there is no argument involved except for the default arg if there is one. I think it helps to be accurate and consistent in this sort of code.

crates/ark/src/lsp/completions/sources/unique/custom.rs
```diff
@@ -248,11 +256,229 @@ mod tests {
...
             assert!(compls.iter().any(|c| c.label == "utils"));
         })
     }
+
+    #[test]
+    fn test_completion_custom_sys_getenv() {
+        r_test(|| {
+            let name = "ARK_TEST_ENVVAR";
+
+            harp::parse_eval_global(format!("Sys.setenv({name} = '1')").as_str()).unwrap();
```
lionel-, line 270: Here and below: Only use global when needed (to create a side effect there or because of a dependency on global var).

```suggestion
            harp::parse_eval_base(format!("Sys.setenv({name} = '1')").as_str()).unwrap();
```

crates/ark/src/lsp/completions/sources/unique/custom.rs
```diff
@@ -248,11 +256,229 @@ mod tests {
...
+                let document = Document::new(text, None);
+                let context = DocumentContext::new(&document, point, None);
+
+                let completions = completions_from_custom_source(&context).unwrap().unwrap();
+                let completion = completions
+                    .into_iter()
+                    .find(|completion| completion.label == name);
+                assert!(completion.is_some());
+
+                // Insert text is NOT quoted! And we get an ` = ` appended.
```
lionel-, line 377: Nice

lionel- (approved):
Great set of tests!

PR #532, opened 2024-09-18
Distinguish whitespace and include syntax error messages

lionel-:
We now distinguish between complete inputs and pure whitespace. This way we don't need to send long streaks of empty lines and comments over the network for evaluation, since there is nothing to evaluate.

For the same reason invalid boundaries now carry an error message.

The data types have been reworked so that parse boundaries are now a simple vector of variant types.

Progress towards https://github.com/posit-dev/positron/issues/1326

### DIFF
--- a/crates/ark/src/analysis/parse_boundaries.rs
+++ b/crates/ark/src/analysis/parse_boundaries.rs
@@ -168,6 +168,7 @@ pub fn parse_boundaries(text: &str) -> anyhow::Result<Vec<ParseBoundary>> {
         boundaries.push(ParseBoundary::incomplete(boundary));
     }
     if let Some(boundary) = invalid {
+        // SAFETY: `invalid_message` has to be `Some()` because `invalid` is `Some()`
         boundaries.push(ParseBoundary::invalid(boundary, invalid_message.unwrap()));
     }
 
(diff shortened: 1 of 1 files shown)


crates/ark/src/analysis/parse_boundaries.rs
```diff
@@ -50,19 +76,20 @@ pub fn parse_boundaries(text: &str) -> anyhow::Result<ParseBoundaries> {
 
     let mut complete: Vec<LineRange> = vec![];
     let mut incomplete: Option<LineRange> = None;
-    let mut error: Option<LineRange> = None;
+    let mut invalid: Option<LineRange> = None;
+    let mut invalid_message: Option<String> = None;
```
DavisVaughan, line 80: Main comment is maybe you could make this an optional tuple like `Option<(LineRange, String)>` so you don't have to `unwrap()` the `invalid_message` later on (which made me stop and think for a second)

crates/ark/src/analysis/parse_boundaries.rs
```diff
@@ -170,190 +200,200 @@ fn merge_overlapping(ranges: Vec<LineRange>) -> Vec<LineRange> {
...
     }
 
     // Fill gaps between complete expressions
-    for range in ranges.into_iter() {
+    for range in complete.into_iter() {
         // We found a gap, fill ranges for lines in that gap
         if !range.contains(last_line) {
             for start in last_line..range.start() {
-                filled.push(range_from(start))
+                filled.push(ParseBoundary::whitespace(range_from(start)))
```
DavisVaughan, line 228: I like how explicit this is now

crates/ark/src/analysis/parse_boundaries.rs
```diff
@@ -170,190 +200,200 @@ fn merge_overlapping(ranges: Vec<LineRange>) -> Vec<LineRange> {
...
+    // (incomplete, invalid, or eof)
+    let last_boundary = filled.last().map(|b| b.range.end()).unwrap_or(0);
+    let next_boundary = incomplete
         .as_ref()
-        .or(boundaries.error.as_ref())
+        .or(invalid.as_ref())
         .map(|r| r.start())
         .unwrap_or(n_lines);
 
     for start in last_boundary..next_boundary {
```
DavisVaughan, line 245:
```suggestion
    for start in last_complete_boundary..next_boundary {
```

Maybe? Having `last_boundary..next_boundary` is mildly confusing because "last" implies its the...last one, but then we are saying there is a "next" one after that

DavisVaughan (approved):

PR #537, opened 2024-09-18
Fix package name extraction from help path

DavisVaughan:
Addresses https://github.com/posit-dev/positron/issues/3775

The TLDR of https://github.com/posit-dev/positron/issues/3775 is that normally libpaths look like this:

```r
> .libPaths()
[1] "/Users/davis/Library/R/arm64/4.4/library"                            
[2] "/Library/Frameworks/R.framework/Versions/4.4-arm64/Resources/library"
```

And help paths look like this

```r
> unclass(help("match"))
[1] "/Library/Frameworks/R.framework/Versions/4.4-arm64/Resources/library/base/help/match"
```

So we would look for `/library/<package>` to extract the package name.

But it turns out that the `.libPaths()` prefix part of the help path can be anything! Here's what renv uses:

```r
> .libPaths()
[1] "/Users/davis/Desktop/myRProject/renv/library/macos/R-4.4/aarch64-apple-darwin20"                       
[2] "/Users/davis/Library/Caches/org.R-project.R/R/renv/sandbox/macos/R-4.4/aarch64-apple-darwin20/f7156815"
```

```r
> unclass(help("glue_data"))
[1] "/Users/davis/Desktop/myRProject/renv/library/macos/R-4.4/aarch64-apple-darwin20/glue/help/glue"
```

See the `/library/macos` there? `macos` isn't a package! So our package name resolution falls over here. We end up passing `package = "macos"` to `tools::Rd2HTML()` and it eventually triggers a warning of `no package 'macos' was found`.

---

I experimented with a few ways to better get at that package name, and ultimately I think the most reliable way is to recognize that `utils:::.getHelpFile()` has to resolve the package name too to figure out which Rd database to pull the Rd file from. It uses the fact that help paths always look like this:

```
<libpath>/<package>/help/<topic>
```

Where:
- `<libpath>/<package>/help` is a true folder on the file system, and the Rd database lives in that folder
- `<libpath>/<package>/help/<topic>` doesn't actually exist, but gives you the topic name. The path structure is just a convention.

So `.getHelpFile()` just does `basename(dirname(dirname(path)))` to get at the package name. I think we can keep it simple and do exactly this, trusting that base R won't change something this critical to the help system.

### DIFF
--- a/crates/ark/src/modules/positron/help.R
+++ b/crates/ark/src/modules/positron/help.R
@@ -159,58 +159,27 @@ getHtmlHelpContentsInstalled <- function(helpFiles, package) {
   paste(contents, collapse = "\n")
 }
 
-getPackageNameFromHelpPath <- function(x) {
-  # Help paths are of the form:
-  # <libpath>/<package>/<etc>/<topic>
+getPackageNameFromHelpPath <- function(path) {
+  # Help paths are always of the form:
+  # <libpath>/<package>/help/<topic>
   #
-  # General idea is to strip the matching `.libPaths()` prefix from the help
-  # path. What remains starts with the package name!
+  # `<libpath>/<package>/help` should be a real path on the file system,
+  # and points to the folder where the Rd database is stored. Note that
+  # the full path of `<libpath>/<package>/help/<topic>` actually does NOT
+  # exist on the file system, it's just a convention!
+  #
+  # `utils:::.getHelpFile()` utilizes the fact that `<package>` is always
+  # relative to `<topic>` in a specific way, so we should be safe to do the
+  # same.
+  path_help <- dirname(path)
+  path_package <- dirname(path_help)
 
-  # Normalize paths everywhere to avoid any irrelevant path differences.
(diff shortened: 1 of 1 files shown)


DavisVaughan:
I had a bit of trouble getting renv to install a package _without_ first installing it into the cache. Ultimately this worked

```r
# Prevent renv from installing glue in the cache
options(renv.config.cache.enabled = FALSE)

renv::install("glue")

library(glue)

# It should show `library/macos` in the path if
# the cache is actually disabled
unclass(help("glue_data"))

glue_data()

```

crates/ark/src/modules/positron/help.R
```diff
@@ -163,6 +159,31 @@ getHtmlHelpContentsInstalled <- function(helpFiles, package) {
   paste(contents, collapse = "\n")
 }
 
+getPackageNameFromHelpPath <- function(path) {
```
DavisVaughan, line 162: This is where having these modules as a little package would be handy, because I wanted to write a few tests for this helper but have no where to put them

kevinushey (approved):
LGTM!

PR #538, opened 2024-09-19
Check immediate bindings before calling `CAR`

dfalbel:
Adresses crashes from:

- https://github.com/posit-dev/positron/issues/4686
- https://github.com/posit-dev/positron/issues/4741


crates/harp/src/size.rs
```diff
@@ -611,4 +619,42 @@ mod tests {
...
     }
+
+    // This is a regression test for:
+    // - https://github.com/posit-dev/positron/issues/4686
+    // - https://github.com/posit-dev/positron/issues/4741
+    // In both cases the R session crashed because we tried to call `CAR`
+    // in a immediate binding. More info in:
+    // https://github.com/posit-dev/positron/issues/4686#issuecomment-2352427239
+    #[test]
+    fn test_shinytest2_size() {
```
lionel-, line 630: hmm... I'd rather not have integration tests of this kind here.
I would prefer a unit test.

What do you think @DavisVaughan?

lionel- (commented):
This approach is fine as a stopgap but in the long term we'll want to avoid depending on R internals.

Romain already added support for SxpInfo stuff actually, could you please use the existing `Sxpinfo::interpret().is_extra()`?

crates/harp/src/size.rs
```diff
@@ -184,14 +184,16 @@ fn obj_size_tree(
...
                     );
-                    size += obj_size_tree(
-                        unsafe { libr::CAR(cons) },
-                        base_env,
-                        sizeof_node,
-                        sizeof_vector,
-                        seen,
-                        depth + 1,
-                    );
+                    if !is_immediate_binding(cons) {
```
lionel-, line 187: I think you can still compute the size by looking at the `TAG` and using `getVar()`.

dfalbel: I somehow understood that since the value is inlined, the size is already computed when we add the node size: https://github.com/posit-dev/ark/blob/eb1260c5f8a7c1c87cfd0c738028a362bbf73a64/crates/harp/src/size.rs#L178-L179

Ie, the value is using the same space as the *SEXPREC would use.

Am I missing something?

crates/harp/src/size.rs
dfalbel: Ok! I'm happy to remove this test. I unfortunatelly can't find a way of constructing these inlined values in a simple way.

crates/harp/src/size.rs
lionel-: oh no you're totally right! Can you please add a comment?

lionel-:
@dfalbel We were planning on removing the dependency on sxpinfo to avoid Ark crashes when internals change. @DavisVaughan would like to go through with that change now: https://github.com/posit-dev/ark/pull/540/files

Could you please rewrite environment traversal so that it uses exposed API? E.g. `findVar()` (gets you values and promises), `R_BindingIsActive()`, `R_ActiveBindingFunction()`, etc.

A tricky part is that it has to work the same way on all R versions.

DavisVaughan:
Since we are actually in a place where we can break the dependency on `Sxpinfo` quite easily, I strongly strongly prefer that we find a different approach here https://github.com/posit-dev/ark/pull/540

lionel-:
@DavisVaughan Agreed, I had no idea we were this close. I guess our new environment iterator implementation went a long way.

I think the right way to approach things here is to create a new environment iterator that provides the binding type along the lines of https://gist.github.com/lionel-/1ebcbd5ec69c0775d514c329522408a3#binding-type. Then you could access this typed binding with specific accessors that we can tweak so they work well on all R versions.

dfalbel:
I updated the PR to use `harp`'s internal environment iterator, which should safely expand immediate bindings when necessary,and correctly avoids the evaluation of promises and active bindings. We'll no longer account the size of internal implementation details of environmentsa, such as the pre-allocated size of the hash table, or the size each chain uses, etc.

crates/harp/src/size.rs
```diff
@@ -235,14 +238,61 @@ fn obj_size_tree(
...
+                        sizeof_node,
+                        sizeof_vector,
+                        seen,
+                        depth + 1,
+                    ),
+                    // `object_size_tree` is aware of promise objects.
+                    // The environment iterator will automatically return `PRVALUE` as
+                    // a `Standard` binding though. So this is only seeing unevaluated promises.
+                    // For evaluated promises, we are not counting the size of `PRCODE`, but hopefully
+                    // their sizes are neglegible, mostly just symbols or small expressions.
```
lionel-, line 273:
```suggestion
                    // their sizes are negligible, mostly just symbols or small expressions.
```

crates/harp/src/size.rs
```diff
@@ -235,14 +238,61 @@ fn obj_size_tree(
...
+                        base_env,
+                        sizeof_node,
+                        sizeof_vector,
+                        seen,
+                        depth + 1,
+                    ),
+                    // `object_size_tree` is aware of promise objects.
+                    // The environment iterator will automatically return `PRVALUE` as
+                    // a `Standard` binding though. So this is only seeing unevaluated promises.
+                    // For evaluated promises, we are not counting the size of `PRCODE`, but hopefully
```
lionel-, line 272: No need to do now unless you want to, but the iterator should provide `PRCODE` too for promises. Luke agreed to provide promise accessors, as long as promises themselves are not exposed and instead accessed via the environment.

lionel- (approved):
Looks good, thanks for fixing!

PR #541, opened 2024-09-21
Fix synchronisation of R accesses in tests

lionel-:
Closes https://github.com/posit-dev/positron/issues/4222: Instability in Ark tests is fixed.

Reverts https://github.com/posit-dev/ark/pull/465 which is no longer needed to mitigate the crashes.

@DavisVaughan determined in https://github.com/posit-dev/positron/issues/4222 that tests can end up in a situation where `r_task()` is called from another thread than the thread where tests are executed. Because of this escape hatch for tests: https://github.com/posit-dev/ark/blob/35215ac5309739840e372bbd79f4662fe43517f3/crates/ark/src/r_task.rs#L149, there is no synchronisation when `r_task()` is called within an `r_test()`.

To fix this:

- We now lock harp's test mutex in `r_task()`.

- We're also initializing R from `r_task()` if not done already.

- The test mutex is now reentrant. This is necessary because we allow for the same leniency in the real `r_task()` where we just run the closure if we're already on the main thread. This should arguably be changed to represent a logic error instead, but it's probably too dangerous to do that before we have drastically reduced our reliance on `r_task()`.

In addition to these changes, I've renamed harp's `r_test()` to `r_task()`, and I've removed Ark's `r_test()` in favour to `r_task()`. This simplifies the API as we now only have one operation to run R code from other threads (two really, one in Harp and one in Ark, the latter being necessary in tests for Ark-specific initialization).

These changes allow for a new workflow for testing comms:

- Don't wrap your test in `r_task()` (previously `r_test()`)
- Instead use `r_task()` selectively when you need to access the R API.

This lets other threads, such as a comm thread, use `r_task()` to get access to the R API in a synchronised way and without causing deadlocks.

One big behaviour change is that tests that are no longer wrapped in `r_task()` no longer run sequentially. This is problematic for the data explorer tests because they depend on state in the global env. Ideally they should be rewritten to remove these dependencies @dfalbel.

In the mean time I've added `r_test_lock()` which uses a second lock to force these tests to run sequentially.


crates/ark/tests/data_explorer.rs
```diff
@@ -70,24 +71,14 @@ use itertools::Itertools;
 use libr::R_GlobalEnv;
 use libr::Rf_eval;
 
-struct DataExplorerSocket {
-    socket: socket::comm::CommSocket,
-}
-
-impl Drop for DataExplorerSocket {
```
dfalbel, line 77: When the Data Explorer is opened from a binding name, we'll need to make sure that at the end of the test they get deleted and we send a `EVENTS.console_prompt.emit(());`, otherwise there will be too many data explorer loops running at some point, which can slow down test execution.

lionel-: It would also be nice if these tests did not rely on global environment bindings and instead were self-contained.

Do you think we should preserve `DataExplorerSocket` in the mean time? I'd be happy to revert the revert commit.

dfalbel: I can work on that in a follow up PR.

crates/harp/src/fixtures/mod.rs
```diff
@@ -61,19 +52,24 @@ pub static IS_TESTING: bool = cfg!(feature = "testing");
...
 
-pub fn r_test<F: FnOnce()>(f: F) {
-    let guard = unsafe { R_RUNTIME_LOCK.lock() };
+/// Run code accessing the R API in a safe context.
+///
+/// Takes a lock on `R_TEST_LOCK` and ensures R is initialized.
+///
+/// Note: `harp::r_task()` should only be used in Harp tests. Use
+/// `ark::r_task()` in Ark tests so that Ark initialisation also takes place.
+pub fn r_task<F: FnOnce()>(f: F) {
```
DavisVaughan, line 61:
```suggestion
#[cfg(test)]
pub(crate) fn r_task<F: FnOnce()>(f: F) {
```

If you do this, then `harp::fixtures::r_task()` is only available in harp's unit tests - preventing us from accidentally using it from ark, as we've done in signature-help right now.

Harp doesn't have any integration tests, so I think this is good for now.

(Without `#[cfg(test)]` i get yelled at by RA because `r_task()` isn't used anywhere outside the tests)

crates/ark/src/r_task.rs
```diff
@@ -252,8 +254,7 @@ where
     F: FnOnce() -> Fut + 'static + Send,
     Fut: Future<Output = ()> + 'static,
 {
-    // Idle tasks are always run from the read-console loop
-    if unsafe { R_TASK_BYPASS } {
+    if harp::IS_TESTING {
         // Escape hatch for unit tests
         futures::executor::block_on(fun());
```
DavisVaughan, line 259: This almost certainly needs the `R_TEST_LOCK` and `r_test_init()` too right? Even though we aren't using it yet.

crates/harp/src/fixtures/mod.rs
lionel-: Sounds good. The other way to fix if we add integration tests for harp is to export with `pub`, which would be needed anyway.

lionel-: Good idea. If we do end up exporting, we could consistently use the namespaced form `crate::r_task()` to ensure we use the correct variant.

lionel-: I replaced everything with `crate::r_task()` in tests so that it's obvious we're using the right symbol.

crates/harp/src/fixtures/mod.rs
```diff
@@ -27,13 +21,10 @@ use stdext::cargs;
...
-// Escape hatch for unit tests. We need this because the default
-// implementation of `r_task()` needs a fully formed `RMain` to send the
-// task to, which we don't have in unit tests. Consequently tasks run
-// immediately in the current thread in unit tests. Since each test has its
-// own thread, they are synchronised via the `R_RUNTIME_LOCK` mutex.
-pub static mut R_TASK_BYPASS: bool = false;
-static mut R_RUNTIME_LOCK: Mutex<()> = Mutex::new(());
+// This needs to be a reentrant mutex because many of our tests are wrapped in
+// `r_test()` which takes the R lock. Without a reentrant mutex, we'd get
+// deadlocked when we cause some other background thread to use an `r_task()`.
```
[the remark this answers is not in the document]
lionel-: Since we're no longer wrapping in `r_test()`, I removed the reentrant lock. We only need reentrancy in production builds to avoid deadlocking if there is an unexpected nested `r_task()`. But that should not happen: either we are on the main thread, or we are not. A nested `r_task()` indicates code that is confused about where it should run.

I think we may have initially envisioned having tools that could be called both from the R thread and background threads, but that was misguided. The `r_task()` should be explicit and only used in background thread code.

crates/ark/src/r_task.rs
lionel-: I left that part as is because it confused me and I did not get to the bottom of it, but it's currently only running on the main thread.

You're right we should be more careful though, because that might not be the case in the future, so we now take the lock here too.

crates/harp/src/fixtures/mod.rs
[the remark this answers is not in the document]
lionel-: Well it's now a reentrant lock again for https://github.com/posit-dev/ark/pull/541#discussion_r1782580736

PR #543, opened 2024-09-23
Add dotty assignment support

DavisVaughan:
Closes https://github.com/posit-dev/ark/pull/511
Addresses https://github.com/posit-dev/positron/issues/4544

A rework of https://github.com/posit-dev/ark/pull/511 on top of current `main`. The tree-sitter sync up made it rather complicated to merge/rebase, and I had a number of comments so I just did a rewrite here, using Kevin's PR as a very solid base to work from.

### DIFF
--- a/crates/ark/src/lsp/completions/sources/composite/pipe.rs
+++ b/crates/ark/src/lsp/completions/sources/composite/pipe.rs
@@ -5,20 +5,15 @@
 //
 //
 
-use anyhow::Result;
 use harp::error::Error;
 use harp::eval::RParseEvalOptions;
 use harp::object::RObject;
-use stdext::local;
-use stdext::IntoOption;
 use tower_lsp::lsp_types::CompletionItem;
 use tree_sitter::Node;
 
 use crate::lsp::completions::sources::utils::completions_from_object_names;
 use crate::lsp::document_context::DocumentContext;
 use crate::lsp::traits::rope::RopeExt;
-use crate::treesitter::BinaryOperatorType;
-use crate::treesitter::NodeType;
 use crate::treesitter::NodeTypeExt;
 
 #[derive(Clone)]
@@ -30,7 +25,9 @@ pub(super) struct PipeRoot {
     pub(super) object: Option<RObject>,
 }
 
-pub(super) fn completions_from_pipe(root: Option<PipeRoot>) -> Result<Option<Vec<CompletionItem>>> {
+pub(super) fn completions_from_pipe(
+    root: Option<PipeRoot>,
+) -> anyhow::Result<Option<Vec<CompletionItem>>> {
     let Some(root) = root else {
         // No pipe
         return Ok(None);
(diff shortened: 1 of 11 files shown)


crates/ark/src/lsp/completions/sources/composite/pipe.rs
```diff
@@ -124,69 +121,51 @@ fn eval_pipe_root(name: &str) -> Option<RObject> {
     Some(value)
 }
 
-fn find_pipe_root_name(context: &DocumentContext, node: &Node) -> Option<String> {
-    // Try to figure out the code associated with the 'root' of the pipe expression.
-    let root = local! {
-
-        let root = find_pipe_root_node(context, *node)?;
-        is_pipe_operator(context, &root).into_option()?;
+fn find_pipe_root_name(context: &DocumentContext, node: &Node) -> anyhow::Result<Option<String>> {
```
DavisVaughan, line 124: I was prompted to clean this up by the fact that we now propagate up the `Result`, and I think it cleaned up nicely.

crates/ark/src/lsp/diagnostics.rs
```diff
@@ -1094,4 +1218,144 @@ foo
...
+
+            let document = Document::new(code, None);
+
+            let diagnostics = generate_diagnostics(document.clone(), DEFAULT_STATE.clone());
+            assert_eq!(diagnostics.len(), 1);
+
+            let diagnostic = diagnostics.get(0).unwrap();
+            assert_eq!(diagnostic.range.start.line, 2);
+            insta::assert_snapshot!(diagnostic.message);
+        })
```
DavisVaughan, line 1359: We have kind of a general issue here, so I've recorded this "supposed to be failing" test. Definitely a tough one though, i.e. NSE really throws this off.

crates/ark/src/treesitter.rs
```diff
@@ -386,6 +389,36 @@ impl NodeTypeExt for Node<'_> {
     fn is_binary_operator(&self) -> bool {
         matches!(self.node_type(), NodeType::BinaryOperator(_))
     }
+
+    fn is_native_pipe_operator(&self) -> bool {
+        self.node_type() == NodeType::BinaryOperator(BinaryOperatorType::Pipe)
+    }
+
+    fn is_magrittr_pipe_operator(&self, contents: &ropey::Rope) -> anyhow::Result<bool> {
```
DavisVaughan, line 397: One thing I did differently was to propagate this `Result` upward, which affected some design decisions elsewhere. It seems like the caller should get to decide what to do here.

crates/ark/src/lsp/diagnostics.rs
```diff
@@ -1094,4 +1218,144 @@ foo
...
+    }
+
+    #[test]
+    fn test_dotty_assignment_named() {
+        r_test(|| {
+            // `x` should not be defined
+            let code = "
+                .[apple = x, banana] <- list(w = 1, x = 2, y = 3, z = 4)
```
DavisVaughan, line 1267: Added a named test

crates/ark/src/lsp/diagnostics.rs
```diff
@@ -471,6 +470,131 @@ fn recurse_assignment(
...
+            context.add_defined_variable(name.as_str(), range);
+            continue;
+        }
+
+        // i.e. `.[x, .[y]]` where unpacking is recursive
+        if value.is_subset() {
+            handle_assignment_dotty(&value, context)?;
+            continue;
+        }
+    }
```
DavisVaughan, line 559: Most of my changes were adding example based docs from the perspective of someone that isn't that familiar with dotty, and also tweaking the loop structure to make it a little easier to read (i think)

DavisVaughan (commented):
@kevinushey can you take a quick look at how I reworked this on top of current main? Mostly the same, changes are called out inline

kevinushey (approved):
LGTM!

PR #544, opened 2024-09-24
Use correct setter to set vector value

lionel-:
Fixes intermittent failure:

```
Error: bad LGLSXP vector
```

### DIFF
--- a/crates/ark/src/lsp/signature_help.rs
+++ b/crates/ark/src/lsp/signature_help.rs
@@ -702,7 +702,7 @@ fn <- function(
         r_test(|| {
             let x = RObject::from(r_alloc_integer(12));
             for i in 0..12 {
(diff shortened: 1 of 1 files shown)


crates/ark/src/lsp/signature_help.rs
```diff
@@ -702,7 +702,7 @@ fn <- function(
         r_test(|| {
             let x = RObject::from(r_alloc_integer(12));
             for i in 0..12 {
-                r_lgl_poke(x.sexp, R_xlen_t::try_from(i).unwrap(), i);
+                r_int_poke(x.sexp, R_xlen_t::try_from(i).unwrap(), i);
```
lionel-, line 705: @DavisVaughan I think we see this with local builds of R used in tests where strict checking is enabled: https://github.com/r-devel/r-svn/blob/7b9e7d2f5c0fe3765feb85fbb12ee0f5d33ec150/src/include/Rinlinedfuns.h#L158

PR #545, opened 2024-09-25
Add testing on Windows, macOS, and R 4.2

DavisVaughan:
This PR has the goal of getting us the following CI structure:
- Windows (stable rust, release R)
- macOS (stable rust, release R)
- Linux (stable rust, release R)
- Linux (stable rust, 4.2 R)
- Linux (nightly rust, release R)

Adding support for macOS was incredibly simple.

Adding support for R 4.2 revealed one bug that I've fixed and called out below.

Adding support for Windows has been a nightmare. Further details below.

You should be able to trigger each OS's workflow manually through workflow dispatch, and you should also be able to turn on SSH-ing into the Linux workflow (it will pause right before running `cargo test`).

## Application manifest

I noticed that during tests on Windows, `sessionInfo()` reports that we are _not_ in a UTF-8 locale and we are on Windows Server 2012, even though the github machine is Windows Server 2022. This is a classic sign that our application manifest file wasn't being used https://github.com/posit-dev/ark/pull/178.

The manifest is being used for our main binary, `ark.exe`, but it wasn't being embedded into our test binary, like `ark-<stuff>.exe`. Making that happen was a little complicated. The crux of it is:
- `embed_resource::compile()` sets `cargo:rustc-link-arg-bins`, which doesn't target test binaries
- `embed_resource::compile_for_tests()` sets `cargo:rustc-link-arg-tests`, but that doesn't target _unit_ test binaries due to a Rust bug
- `embed_resource::compile_for_everything()` sets `cargo:rustc-link-ark`, which is the "just do it for everything" approach, and that does work. This didn't exist though, I had to ask the embed-resource maintainer for it!

So that allowed ark's unit tests to start up R in a way that R was in a UTF-8 locale and ran under Windows Server 2022.

But harp also starts R for its unit tests! And it also tests UTF-8 behavior! So I had to add a `build.rs` script to harp to also embed the manifest file there as well.

## Open comms timing issues

See https://github.com/posit-dev/ark/pull/548, I seemed to hit this error every now and then on the Windows CI and my local Windows VM

## Outstanding issues

Occasionally the test suite will hang on windows after running `test_kernel`, but I cannot figure out why. I get no information except eventually it says something like:

```
test test_kernel has been running for over 60 seconds
```

My guess is that a zmq `send()` or `recv()` is stuck, so I've set `set_sndtimeo()` and `set_rectimeo()` to try and turn these cases into a panic at test time, but I actually haven't been able to reproduce the issue since I added them, and can't reproduce this one locally.


DavisVaughan:
There seems to be some kind of issue occasionally. I sometimes see this locally

```
[2024-09-26T21:24:52Z ERROR amalthea::kernel] While forwarding outbound message: ZeroMQ protocol error on Stdin socket: Host unreachable
test test_kernel has been running for over 60 seconds
``` 

and one of our builds above shows it too

lionel-:
hmm that's surprising these dummy frontend sockets are on `tcp://127.0.0.1`:

https://github.com/posit-dev/ark/blob/77d94003d2d400b643a53f5f2d39fc7a9d23e714/crates/amalthea/src/fixtures/dummy_frontend.rs#L88-L95

Since you can see it locally, maybe you could set a breakpoint and explore with e.g. `telnet` whether a specific port might be at fault?

crates/ark/src/modules/positron/r_data_explorer.R
```diff
@@ -400,7 +425,9 @@ profile_histogram <- function(x, method = c("fixed", "sturges", "fd", "scott"),
 
   # For dates, we convert back the breaks to the date representation.
   if (inherits(x, "POSIXct")) {
-    bin_edges <- as.POSIXct(h$breaks, tz = attr(x, "tzone"))
+    # Must supply an `origin` on R <= 4.2
+    origin <- as.POSIXct("1970-01-01", tz = "UTC")
+    bin_edges <- as.POSIXct(h$breaks, tz = attr(x, "tzone"), origin = origin)
```
DavisVaughan, line 430: Bug found when adding 4.2 CI

crates/ark/src/modules/positron/r_data_explorer.R
```diff
@@ -327,13 +327,38 @@ export_selection <- function(x, format = c("csv", "tsv", "html"), include_header
...
+write_delim_impl <- function(x, path, delim, include_header) {
+    # Scope the `con` lifetime to just this helper.
+    # We need to `close()` the connection before we try and get the
+    # `file.info()$size`.
+
+    # Must open in binary write mode, otherwise even though we set
+    # `eol = "\n"` on Windows it will still write `\r\n` due to the C
+    # level `vfprintf()` call.
+    con <- file(path, open = "wb")
+    defer(close(con))
```
DavisVaughan, line 350: Bug found when adding Windows CI. Tests expect `\n` as the line ending

And the way we do `readChar(path, file.info(path)$size - 1L, useBytes = TRUE)` above expects 1 char for the trailing line (i.e. `\n` not `\r\n`)

crates/amalthea/src/socket/socket.rs
```diff
@@ -133,6 +133,15 @@ impl Socket {
...
         }
 
+        // 10 second timeout on `recv()` operations (for testing for now)
+        // TODO: Expose `IS_TESTING` in its own lightweight crate?
+        if let Err(error) = socket.set_rcvtimeo(10000) {
+            return Err(Error::CreateSocketFailed(name, error));
+        }
+        if let Err(error) = socket.set_sndtimeo(10000) {
+            return Err(Error::CreateSocketFailed(name, error));
+        }
```
DavisVaughan, line 143: TODO: Remove or hide behind test flag

The goal is to avoid an infinite hang on send/recv issues during kernel tests

crates/ark/src/sys/windows/locale.rs
```diff
@@ -0,0 +1,33 @@
...
+            let utf8 = bool::try_from(utf8).unwrap();
+            assert!(utf8);
+
+            let codepage = harp::parse_eval_base("l10n_info()$codepage").unwrap();
+            let codepage = i32::try_from(codepage).unwrap();
+            assert_eq!(codepage, 65001);
+
+            let system_codepage = harp::parse_eval_base("l10n_info()$system.codepage").unwrap();
+            let system_codepage = i32::try_from(system_codepage).unwrap();
+            assert_eq!(system_codepage, 65001);
```
DavisVaughan, line 30: Tests on both the ark and harp side that the application manifest was correctly embedded into the test binary

scripts/manifest/generate.R
```diff
@@ -0,0 +1,42 @@
...
+#'
+#' We use two files to embed an application manifest on Windows:
+#' - `{crate}-manifest.rc`
+#' - `{crate}.exe.manifest`
+#'
+#' And we generate these two files for both `harp` and `ark`.
+#'
+#' They are effectively the same, we just need to swap out the crate names,
+#' so we use this script to write the files in a consistent manner.
+write_manifest <- function(root, crate) {
```
DavisVaughan, line 11: Because the manifest files are almost identical between ark and harp, we generate them from templates

.github/workflows/test-linux.yml
```diff
@@ -0,0 +1,52 @@
...
+          use-public-rspm: true
+
+      - name: Setup Build Environment
+        run: |
+          sudo apt-get update
+
+      - name: Setup SSH access
+        uses: mxschmitt/action-tmate@v3
+        if: ${{ inputs.ssh }}
+        timeout-minutes: 30
```
lionel-, line 47: Cool!

.github/workflows/test-linux.yml
```diff
@@ -0,0 +1,52 @@
...
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: dtolnay/rust-toolchain@nightly
+        if: matrix.config.rust == 'nightly'
+
+      - uses: r-lib/actions/setup-r@v2
+        with:
+          r-version: ${{ matrix.config.r }}
+          use-public-rspm: true
```
lionel-, line 38: Does this mean we can install binary packages from p3m?

By the way there's a bunch of tests skipped because data.table and tibble are not installed on CI. We could now easily fix that right?

E.g. using https://github.com/r-lib/actions/tree/v2/setup-r-dependencies with `extra-packages` or something (hopefully it still works without a `DESCPRIPTION` file but we could add one if needed).

crates/ark/build.rs
```diff
@@ -37,8 +37,23 @@ fn main() {
 
     // Embed an Application Manifest file on Windows.
```
lionel-, line 38: Should explain what the manifest does to our main binary, i.e. enable UTF-8.

crates/ark/src/modules/positron/r_data_explorer.R
```diff
@@ -327,13 +327,38 @@ export_selection <- function(x, format = c("csv", "tsv", "html"), include_header
...
+}
+
+write_delim_impl <- function(x, path, delim, include_header) {
+    # Scope the `con` lifetime to just this helper.
+    # We need to `close()` the connection before we try and get the
+    # `file.info()$size`.
+
+    # Must open in binary write mode, otherwise even though we set
+    # `eol = "\n"` on Windows it will still write `\r\n` due to the C
+    # level `vfprintf()` call.
```
lionel-, line 348: 😬

crates/ark/build.rs
lionel-: And link to `crates/harp/resources/manifest/harp.exe.manifest` for reference.

lionel- (approved):
Thanks for the hard work!

.github/workflows/test-linux.yml
DavisVaughan: https://github.com/posit-dev/ark/pull/545/commits/b016ea97557734893018080ae6a3ac1a175cc10f

PR #553, opened 2024-09-30
Make optional parts of `LanguageInfo` actually optional!

DavisVaughan:
I noticed in https://github.com/posit-dev/positron/issues/2098#issuecomment-2384064069 that in a jupyter console we get this pygment error:

Screenshot 2024-09-30 at 4 02 18 PM

See that `for language ''`? That implies we are actually sending _the empty string_ over for the `pygments_lexer` part of the `language_info` part of a `kernel_info_reply`!
https://jupyter-client.readthedocs.io/en/stable/messaging.html#kernel-info

Notice that it says

```
        # Pygments lexer, for highlighting
        # Only needed if it differs from the 'name' field.
        'pygments_lexer': str,
```

I'm interpreting this as meaning that its actually an optional field. We were treating it as a required `String`, but we sent over `String::new()`.

And indeed jupyter_console treats it as an optional field!
https://github.com/jupyter/jupyter_console/blob/fddbc42d2e0be85feace1fe783a05e2b569fceae/jupyter_console/ptshell.py#L541-L542

If `pygments_lexer` is not there, it falls back to `name` of `"R"` which is exactly what we wanted all along.

---

Now for the cool part. In jupyter console, `pygments_lexer` is used to look up syntax highlighting using the pygments library. So fixing this actually enables R syntax highlighting in jupyter console for ark!

It uses this, with the key being `short name: r`
https://pygments.org/docs/lexers/#pygments.lexers.r.SLexer

Screenshot 2024-09-30 at 4 32 06 PM

### DIFF
--- a/crates/amalthea/src/wire/language_info.rs
+++ b/crates/amalthea/src/wire/language_info.rs
@@ -24,14 +24,14 @@ pub struct LanguageInfo {
     /// The file extension for script files in the language
     pub file_extension: String,
 
-    /// Pygments lexer (for highlighting), if different than name
-    pub pygments_lexer: String,
+    /// Pygments lexer (for highlighting), if different than `name`
+    pub pygments_lexer: Option<String>,
 
-    /// Codemirror mode (for editing), if different than name
-    pub codemirror_mode: String,
+    /// Codemirror mode (for editing), if different than `name`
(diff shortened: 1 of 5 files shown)


lionel- (approved):
Cool!

PR #536, opened 2024-09-18
Split inputs by newlines

DavisVaughan:
With @lionel- 

Addresses https://github.com/posit-dev/positron/issues/4745
Addresses https://github.com/posit-dev/positron/issues/4742
Closes https://github.com/posit-dev/ark/issues/557

PR #558, opened 2024-10-02
Fix hanging issues in integration tests

### LINKED ISSUE #551: Ark: Panics in dummy kernel threads should be propagated to test harness
Currently, if a panic occurs in integration tests on a background thread involved in sending server-side zmq messages, the thread gets poisoned silently and the test harness hangs because the client is waiting for zmq messages that can no longer be sent. This makes it hard to debug issues both locally and on CI.

To fix that, we could use the same approach as for the Ark binary where we propagate panics to the main thread and then panic from there:

https://github.com/posit-dev/ark/blob/c6df0b2419dd3b5ee5a6a51bcef8d452b0245f9e/crates/ark/src/main.rs#L245-L249

lionel-:
Closes #551.
Branched from #547.

I've been facing issues locally with instability which in turned caused integration tests to hang indefinitely. And on CI we now also test on Windows, which is causing similar hang issues in the Amalthea integration tests (but due to other instability causes). This PR addresses my local instability issue and attempts to better propagate panics and problems in server-side threads so that the test threads do not hang waiting for messages.

It turns out that the instability was being caused by this top-level callback installed by the cli package: https://github.com/r-lib/cli/blob/1220ed092c03e167ff0062e9839c81d7258a4600/R/onload.R#L33-L40. This caused extra output containing an ANSI escape sequence to be streamed on IOpub at random times. To avoid this, we now pass `--vanilla` to R on startup. (Note that Positron will not see these "show cursor" escapes because we redirect stdout in production builds, which causes `istty(stdout())` to be `FALSE` and cli does not register that task callback.)

We now also use `assert_matches()` to get better information about actual incoming messages when they don't match what is expected. Unlike `assert_match()`, this shows the contents of the failing input which helps a lot.

Regarding tests hanging, the key point is that when a background thread panics, a panic hook is run to record information about the panic, the thread is unwound, but then nothing else happens if the thread is not joined. So when a test thread is waiting for an expected message and the server thread has panicked, the panic is not propagated to the client-side thread, which waits forever.

What did not work is setting a panic hook to propagate background panics to the main thread:

  - Notifying other threads from the panic hook is feasible but we need to specifically wait for notifications at the other end.
  - Another approach is to flip a global flag that other threads can check periodically, but that's manual and error-prone.
  - Calling the original panic hook and then exiting the process with `exit()` or `abort()` does not work out because the hard exit prevents the test runner from printing a summary of the test results.

Things that this PR does instead to (partially) address this:

- We really need the client test thread to panic when something goes wrong in the server threads. Propagating panics from the server threads is hard, so instead we now poll for expected incoming messages with a timeout. After one second the client thread panics.

- In the zmq/crossbeam merging threads, we now panic instead of logging if we're running tests and something unexpected happens, e.g. when an expected message doesn't come in or when a socket connection fails. This allows the test runner's panic hook to take note of the panic.

  Whether to panic is implemented by checking for `cfg(debug_assertions)`. This way we remain lenient (but verbose) in production builds.

- Sometimes the panic occurs even before the kernel starts up, e.g. because a zmq socket fails to connect. This is not a problem for the main Ark binary because this all happens on the main thread, where panics cause an exit. However in integration tests we run R in a background thread so it can be reused across multiple test threads. An early panic would then be silently ignored and the tests would be waiting indefinitely for R initialization.

  To fix this, I split `RMain::setup()` from `RMain::start()`. This allows the test init to run the former (via `start_kernel()`) independently on its own thread, before starting the R REPL on a background thread. Most init problems can then be propagated correctly to the test runner.

### DIFF
--- a/crates/ark/src/fixtures/dummy_frontend.rs
+++ b/crates/ark/src/fixtures/dummy_frontend.rs
@@ -42,23 +42,24 @@ impl DummyArkFrontend {
         let frontend = DummyFrontend::new();
         let connection_file = frontend.get_connection_file();
 
-        stdext::spawn!("dummy_kernel", || {
-            crate::start::start_kernel(
-                connection_file,
-                vec![
-                    String::from("--interactive"),
-                    String::from("--vanilla"),
-                    String::from("--no-save"),
-                    String::from("--no-restore"),
-                ],
-                None,
-                SessionMode::Console,
-                false,
-            );
-        });
+        // Start the kernel in this thread so that panics are propagated
+        crate::start::start_kernel(
+            connection_file,
+            vec![
+                String::from("--interactive"),
+                String::from("--vanilla"),
+                String::from("--no-save"),
+                String::from("--no-restore"),
+            ],
+            None,
+            SessionMode::Console,
+            false,
+        );
 
-        // Wait for startup to complete
-        RMain::wait_r_initialized();
+        // Start the REPL in a background thread, does not return and is never joined
+        stdext::spawn!("dummy_kernel", || {
+            RMain::start();
+        });
 
         frontend.complete_initialization();
         frontend
--- a/crates/ark/src/fixtures/dummy_frontend.rs
+++ b/crates/ark/src/fixtures/dummy_frontend.rs
@@ -45,7 +45,12 @@ impl DummyArkFrontend {
         stdext::spawn!("dummy_kernel", || {
             crate::start::start_kernel(
                 connection_file,
-                vec![String::from("--no-save"), String::from("--no-restore")],
+                vec![
+                    String::from("--interactive"),
+                    String::from("--vanilla"),
+                    String::from("--no-save"),
+                    String::from("--no-restore"),
+                ],
                 None,
                 SessionMode::Console,
                 false,
--- a/crates/ark/src/interface.rs
+++ b/crates/ark/src/interface.rs
@@ -287,7 +287,7 @@ impl RMain {
     /// starting the R REPL. Since the REPL does not return, it might be
     /// launched in a background thread (which we do in integration tests). The
     /// setup can still be done in your main thread so that panics may propagate
-    /// as expected. Call `RMain::start_repl()` after this to actually start the
+    /// as expected. Call `RMain::start()` after this to actually start the
     /// R REPL.
     pub fn setup(
         r_args: Vec<String>,
--- a/crates/ark/src/interface.rs
+++ b/crates/ark/src/interface.rs
@@ -282,9 +282,14 @@ pub enum ConsoleResult {
 }
 
 impl RMain {
-    /// Starts the main R thread and initializes the `R_MAIN` singleton.
-    /// Doesn't return. Must be called only once.
-    pub fn start(
+    /// Sets up the main R thread and initializes the `R_MAIN` singleton. Must
+    /// be called only once. This is doing as much setup as possible before
+    /// starting the R REPL. Since the REPL does not return, it might be
+    /// launched in a background thread (which we do in integration tests). The
+    /// setup can still be done in your main thread so that panics may propagate
+    /// as expected. Call `RMain::start_repl()` after this to actually start the
+    /// R REPL.
+    pub fn setup(
         r_args: Vec<String>,
         startup_file: Option<String>,
         kernel_mutex: Arc<Mutex<Kernel>>,
@@ -423,8 +428,13 @@ impl RMain {
         if !ignore_user_r_profile {
             startup::source_user_r_profile();
         }
+    }
 
-        // Does not return!
+    /// Start the REPL. Does not return!
+    pub fn start() {
+        // Update the main thread ID in case the REPL is started in a background
+        // thread (e.g. in integration tests)
+        unsafe { R_MAIN_THREAD_ID = Some(std::thread::current().id()) };
         crate::sys::interface::run_r();
     }
 
--- a/crates/ark/src/interface.rs
+++ b/crates/ark/src/interface.rs
@@ -140,11 +140,8 @@ pub enum SessionMode {
 // We use the `once_cell` crate for init synchronisation because the stdlib
 // equivalent `std::sync::Once` does not have a `wait()` method.
 
-/// Ensures that the kernel is only ever initialized once. Used to wait for the
-/// `RMain` singleton initialization in `RMain::wait_initialized()`.
-static R_MAIN_INIT: once_cell::sync::OnceCell<()> = once_cell::sync::OnceCell::new();
-
-/// Used to wait for complete R startup in `RMain::wait_r_initialized()`.
+/// Used to wait for complete R startup in `RMain::wait_initialized()` or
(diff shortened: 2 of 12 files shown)


crates/ark/src/interface.rs
```diff
@@ -282,9 +282,14 @@ pub enum ConsoleResult {
...
 
 impl RMain {
-    /// Starts the main R thread and initializes the `R_MAIN` singleton.
-    /// Doesn't return. Must be called only once.
-    pub fn start(
+    /// Sets up the main R thread and initializes the `R_MAIN` singleton. Must
+    /// be called only once. This is doing as much setup as possible before
+    /// starting the R REPL. Since the REPL does not return, it might be
+    /// launched in a background thread (which we do in integration tests). The
+    /// setup can still be done in your main thread so that panics may propagate
```
DavisVaughan, line 289: "so that panics" - do you mean panics during setup? or panics in general?

crates/ark/src/interface.rs
```diff
@@ -423,8 +428,13 @@ impl RMain {
...
             startup::source_user_r_profile();
         }
+    }
 
-        // Does not return!
+    /// Start the REPL. Does not return!
+    pub fn start() {
+        // Update the main thread ID in case the REPL is started in a background
+        // thread (e.g. in integration tests)
+        unsafe { R_MAIN_THREAD_ID = Some(std::thread::current().id()) };
```
DavisVaughan, line 437: Should we even set `R_MAIN_THREAD_ID` from `setup()` then? It feels weird to do so twice to me.

crates/ark/src/fixtures/dummy_frontend.rs
```diff
@@ -42,19 +42,25 @@ impl DummyArkFrontend {
...
-                vec![String::from("--no-save"), String::from("--no-restore")],
-                None,
-                SessionMode::Console,
-                false,
-            );
+            RMain::start();
         });
 
-        // Wait for startup to complete
-        RMain::wait_r_initialized();
```
DavisVaughan, line 56: Remove `wait_r_initialized()` now?

crates/ark/src/fixtures/dummy_frontend.rs
```diff
@@ -42,19 +42,25 @@ impl DummyArkFrontend {
...
+        // Start the REPL in a background thread, does not return and is never joined
         stdext::spawn!("dummy_kernel", || {
-            crate::start::start_kernel(
-                connection_file,
-                vec![String::from("--no-save"), String::from("--no-restore")],
-                None,
-                SessionMode::Console,
-                false,
-            );
+            RMain::start();
```
DavisVaughan, line 61: I am mildly concerned about the fact that `R_INIT` sets "set" in `setup()` at the very end.

But we typically think of `R_INIT` as "R is ready to receive input", but that really isn't true because we haven't started the REPL yet (not until we call `start()` here).

I feel like its probably ok? Maybe `R_INIT` should be `R_STARTING` or something, I'm not sure. And the remaining helper would be `is_r_starting()`. If you look at the 1 place `is_r_initialized()` is currently being used, it is to capture output that is part of the startup banner. Switching the name to `is_r_starting()` seems like it would still make sense there.

crates/ark/src/fixtures/dummy_frontend.rs
```diff
@@ -42,19 +42,25 @@ impl DummyArkFrontend {
...
         });
```
DavisVaughan, line 62: I also wonder if we really need `R_MAIN_INIT`.

Can it _just_ be `R_INIT`/`R_STARTING`?

`RMain::is_initialized()` is only used in 2 places, both of which feel like it would be ok to wait _just a little longer_ until `R_INIT` is also set (UI comm for the "is busy" response, and connection pane)

DavisVaughan (approved):

crates/ark/src/interface.rs
lionel-: I feel the same but we have to because `write_console()`, called during startup when the banner is emitted, calls `RMain::get_mut()`.

crates/ark/src/fixtures/dummy_frontend.rs
lionel-: I was wondering about that too, but I couldn't move it to the `start()` method.

But I think that's ok because R _is_ ready to receive input or be accessed with well protected (top-level exec) API calls.

crates/ark/src/fixtures/dummy_frontend.rs
lionel-: I'd be happy to do that simplification, I'll try it out.

crates/ark/src/interface.rs
lionel-: during setup, I'll specify

crates/ark/src/fixtures/dummy_frontend.rs
```diff
@@ -42,19 +42,25 @@ impl DummyArkFrontend {
         let frontend = DummyFrontend::new();
         let connection_file = frontend.get_connection_file();
 
+        // Start the kernel in this thread so that panics are propagated
+        crate::start::start_kernel(
+            connection_file,
+            vec![
+                String::from("--interactive"),
+                String::from("--vanilla"),
```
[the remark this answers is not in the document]
lionel-:
```
 --vanilla		Combine --no-save, --no-restore, --no-site-file,
			--no-init-file and --no-environ
```

So without an environ file, site file, or init file, `.libPaths()` should only contain packages from the default install (unless some have been manually installed there).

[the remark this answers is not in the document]
lionel-: We've determined that what I said is only true with the old workflow of setting a libpath in `Renviron`. With the default rig workflow, a user-specific library is used even with `--vanilla`. So we've worked around the cli issue by setting `R_CLI_HIDE_CURSOR`: https://github.com/posit-dev/ark/pull/536/commits/c7e92cc45cf4bb753280062f6f69c11682b5c79e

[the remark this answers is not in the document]
lionel-: `--vanilla` is still good for preventing the local `Rprofile` from running.

crates/ark/src/fixtures/dummy_frontend.rs
lionel-: Now done. This might prevent us from establishing some state on `RMain` from other threads while R is initializing, which could take a while with some init files, but we can work this out later if needed

crates/ark/src/interface.rs
lionel-: I've now:

- Split `write_console()` into pre-init and post-init handlers, as you suggested yesterday. The pre-init handler pushes output to a new global variable `R_BANNER` so it doesn't need to call `RMain::get_mut()`.

- Added the `RMain::positron_ns` field. The modules watchers now call `RMain::get_mut()` to access the namespace. This solves issues having to do with the creation of an `RThreadSafe` object which requires `RMain::get()` which would fail if the R thread ID is not initialized.

This makes it possible to set up the thread ID only once, in `RMain::start()`.

PR #559, opened 2024-10-02
Test multiple expressions in console and notebook modes

lionel-:
Branched from #558.

- Add a Notebook variant of our dummy kernel. It's tested in its own `tests/` file.

- Test multiple expressions in console and notebook modes. In notebook mode, intermediate results are not printed. In console mode, they are. This is relevant for the case of selection sent to the console in Positron, until posit-dev/positron#1326 is fixed.


crates/ark/src/fixtures/dummy_frontend.rs
```diff
@@ -31,10 +41,13 @@ impl DummyArkFrontend {
     }
 
     fn get_frontend() -> &'static Arc<Mutex<DummyFrontend>> {
-        FRONTEND.get_or_init(|| Arc::new(Mutex::new(DummyArkFrontend::init())))
+        // These are the hard-coded defaults. Call `init()` explicitely to
```
DavisVaughan, line 44:
```suggestion
        // These are the hard-coded defaults. Call `init()` explicitly to
```

DavisVaughan (approved):

PR #561, opened 2024-10-02
Extending the variables pane - Inspect

dfalbel:
This a follow up for #560. See there for more details.

This PR adds support for custom inspect behavior using the `ark_variable_get_child()` and `ark_variable_get_children()` methods. It's made in a separate PR as the large diff is confusing to review. Most of it though, is a refactor to make the recursive behavior of `resolve_object_from_path()` more explicit.

### DIFF
--- a/crates/ark/src/variables/variable.rs
+++ b/crates/ark/src/variables/variable.rs
@@ -820,8 +820,8 @@ impl PositronVariable {
             EnvironmentVariableNode::R6Node { object, name } => match name.as_str() {
                 "<private>" => {
                     let env = Environment::new(object);
-                    let enclos = Environment::new(RObject::view(env.find(".__enclos_env__")?));
-                    let private = RObject::view(enclos.find("private")?);
+                    let enclos = Environment::new(RObject::new(env.find(".__enclos_env__")?));
+                    let private = RObject::new(enclos.find("private")?);
 
                     Self::inspect_environment(private)
                 },
--- a/crates/ark/src/variables/variable.rs
+++ b/crates/ark/src/variables/variable.rs
@@ -165,7 +165,7 @@ impl WorkspaceVariableDisplayValue {
     fn from_closure(value: SEXP) -> Self {
         unsafe {
             let args = RFunction::from("args").add(value).call().unwrap();
-            let formatted = RFunction::from("format").add(*args).call().unwrap();
+            let formatted = RFunction::from("format").add(args.sexp).call().unwrap();
             let formatted = CharacterVector::new_unchecked(formatted);
             let out = formatted
                 .iter()
@@ -437,7 +437,7 @@ impl WorkspaceVariableDisplayType {
                                 .add(value)
                                 .call()
                                 .unwrap();
-                            let shape = FormattedVector::new(*dim).unwrap().iter().join(", ");
+                            let shape = FormattedVector::new(dim.sexp).unwrap().iter().join(", ");
                             let display_type = format!("{} [{}]", dfclass, shape);
                             Self::simple(display_type)
                         },
@@ -844,23 +844,23 @@ impl PositronVariable {
                 }
 
                 if object.is_s4() {
-                    Self::inspect_s4(*object)
+                    Self::inspect_s4(object.sexp)
                 } else {
-                    match r_typeof(*object) {
-                        VECSXP | EXPRSXP => Self::inspect_list(*object),
-                        LISTSXP => Self::inspect_pairlist(*object),
+                    match r_typeof(object.sexp) {
+                        VECSXP | EXPRSXP => Self::inspect_list(object.sexp),
+                        LISTSXP => Self::inspect_pairlist(object.sexp),
                         ENVSXP => {
-                            if r_inherits(*object, "R6") {
+                            if r_inherits(object.sexp, "R6") {
                                 Self::inspect_r6(object)
                             } else {
                                 Self::inspect_environment(object)
                             }
                         },
                         LGLSXP | RAWSXP | STRSXP | INTSXP | REALSXP | CPLXSXP => {
-                            if r_is_matrix(*object) {
-                                Self::inspect_matrix(*object)
+                            if r_is_matrix(object.sexp) {
+                                Self::inspect_matrix(object.sexp)
                             } else {
-                                Self::inspect_vector(*object)
+                                Self::inspect_vector(object.sexp)
                             }
                         },
                         _ => Ok(vec![]),
@@ -869,7 +869,7 @@ impl PositronVariable {
             },
 
             EnvironmentVariableNode::Matrixcolumn { object, index } => {
-                Self::inspect_matrix_column(*object, index)
+                Self::inspect_matrix_column(object.sexp, index)
             },
             EnvironmentVariableNode::AtomicVectorElement { .. } => Ok(vec![]),
         }
@@ -884,31 +884,33 @@ impl PositronVariable {
 
         match node {
             EnvironmentVariableNode::Concrete { object } => {
-                if r_is_data_frame(*object) {
+                if r_is_data_frame(object.sexp) {
                     let formatted = RFunction::from(".ps.environment.clipboardFormatDataFrame")
                         .add(object)
                         .call()?;
 
-                    Ok(FormattedVector::new(*formatted)?.iter().join("\n"))
-                } else if r_typeof(*object) == CLOSXP {
-                    let deparsed: Vec<String> =
-                        RFunction::from("deparse").add(*object).call()?.try_into()?;
+                    Ok(FormattedVector::new(formatted.sexp)?.iter().join("\n"))
+                } else if r_typeof(object.sexp) == CLOSXP {
+                    let deparsed: Vec<String> = RFunction::from("deparse")
+                        .add(object.sexp)
+                        .call()?
+                        .try_into()?;
 
                     Ok(deparsed.join("\n"))
                 } else {
-                    Ok(FormattedVector::new(*object)?.iter().join(" "))
+                    Ok(FormattedVector::new(object.sexp)?.iter().join(" "))
                 }
             },
             EnvironmentVariableNode::R6Node { .. } => Ok(String::from("")),
             EnvironmentVariableNode::AtomicVectorElement { object, index } => {
-                let formatted = FormattedVector::new(*object)?;
+                let formatted = FormattedVector::new(object.sexp)?;
                 Ok(formatted.get_unchecked(index))
             },
             EnvironmentVariableNode::Matrixcolumn { object, index } => unsafe {
-                let dim = IntegerVector::new(Rf_getAttrib(*object, R_DimSymbol))?;
+                let dim = IntegerVector::new(Rf_getAttrib(object.sexp, R_DimSymbol))?;
                 let n_row = dim.get_unchecked(0).unwrap() as usize;
 
-                let clipped = FormattedVector::new(*object)?
+                let clipped = FormattedVector::new(object.sexp)?
                     .iter()
                     .skip(index as usize * n_row)
                     .take(n_row)
@@ -936,7 +938,7 @@ impl PositronVariable {
         access_key: &String,
     ) -> harp::Result<EnvironmentVariableNode> {
         let symbol = unsafe { r_symbol!(access_key) };
-        let mut x = unsafe { Rf_findVarInFrame(*object, symbol) };
+        let mut x = unsafe { Rf_findVarInFrame(object.sexp, symbol) };
 
         if r_typeof(x) == PROMSXP {
             // if we are here, it means the promise is either evaluated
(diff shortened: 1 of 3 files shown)


crates/ark/src/variables/variable.rs
```diff
@@ -1354,6 +1529,12 @@ pub fn plain_binding_force_with_rollback(binding: &Binding) -> anyhow::Result<RO
     }
 }
 
+fn parse_index(x: &String) -> Result<isize, harp::error::Error> {
```
DavisVaughan, line 1532:
```suggestion
fn parse_index(x: &String) -> harp::Result<isize> {
```

DavisVaughan (approved):
I definitely don't feel like an expert on this part of the codebase, but where it seemed like new code it seems reasonable to me.

If you rebase on main hopefully the Windows failures will go away - I think we fixed the hang there.

crates/ark/src/variables/variable.rs
```diff
@@ -1326,6 +1420,87 @@ impl PositronVariable {
...
+            .try_dispatch(value, vec![])
+            .map_err(|err| harp::Error::Anyhow(err))?;
+
+        match result {
+            None => Ok(None),
+            Some(value) => {
+                // Make sure value is a list before using inspect_list
+                if !r_typeof(value.sexp) == LISTSXP {
+                    return Err(harp::Error::Anyhow(anyhow!(
+                        "Expected `ark_variable_get_children` to return a list."
```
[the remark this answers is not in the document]
dfalbel: Done!

crates/ark/src/variables/variable.rs
dfalbel: Done!

crates/ark/src/variables/variable.rs
```diff
@@ -908,132 +918,216 @@ impl PositronVariable {
...
-                                    EnvironmentVariableNode::Concrete {
-                                        object: RObject::view(x),
-                                    }
-                                }
-                            },
+        if r_typeof(x) == PROMSXP {
+            // if we are here, it means the promise is either evaluated
+            // already, i.e. PRVALUE() is bound or it is a promise to
+            // something that is not a call or a symbol because it would
+            // have been handled in Binding::new()
```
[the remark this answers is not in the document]
dfalbel: I didn't write this code unfortunatelly, this appears as lines changed due to moving for refactoring.

crates/ark/src/variables/variable.rs
```diff
@@ -908,132 +918,216 @@ impl PositronVariable {
...
 
-                    // TODO: use ? here, but this does not return a crate::error::Error, so
-                    //       maybe use anyhow here instead ?
-                    let row_index = path_element.parse::<isize>().unwrap();
+        // For S4 objects, we acess child nodes using R_do_slot.
+        if object.is_s4() {
+            let name = unsafe { r_symbol!(access_key) };
+            let child: RObject =
+                harp::try_catch(|| unsafe { R_do_slot(object.sexp, name) }.into())?;
+            return Ok(EnvironmentVariableNode::Concrete { object: child });
```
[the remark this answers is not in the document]
dfalbel: I also didn't change that. I'm think the variables.rs file definitely deserves this kind of look, but we might want to do it in a separate PR, so we can track possible problems easier.

dfalbel:
I lacked git skills here , so I had to squash the changes in this PR. The history became completely meaningless due to conflicts when rebasing.

crates/ark/src/variables/variable.rs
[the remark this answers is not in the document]
lionel-: I guess parser literals could be stored in prexpr, e.g. in `identity(1)`, in non-compiled code? (The compiler unwraps promises to constants)

crates/ark/src/variables/variable.rs
```diff
@@ -918,132 +931,216 @@ impl PositronVariable {
...
+        access_key: &String,
+    ) -> harp::Result<EnvironmentVariableNode> {
+        // Concrete nodes are objects that are treated as is. Accessing an element from them
+        // might result in special node types.
+
+        // First try to get child using a generic method
+        // When building the children list of nodes that use a custom `get_children` method, the access_key is
+        // formatted as "custom-{index}-{length(name)}-{name}". If the access_key has this format, we call the custom `get_child_at`,
+        // method, if there's one available:
+        let result = local!({
```
lionel-, line 974: We avoid `local!` since blocks in macros cause difficulties with formatting / assistance.

lionel-: Also I don't see a `?` so `local` is not needed?

crates/ark/src/variables/variable.rs
```diff
@@ -918,132 +931,216 @@ impl PositronVariable {
...
+            let parsed_access_key: Vec<&str> = access_key.splitn(4, '-').collect();
+
+            if parsed_access_key.len() != 4 {
+                return Ok(None);
```
lionel-, line 978: Should that be logged? Same question for other failure modes.

lionel-: Or I guess: Should that be an `Err()`?

crates/ark/src/variables/variable.rs
```diff
@@ -918,132 +931,216 @@ impl PositronVariable {
...
+            let name = unsafe { r_symbol!(access_key) };
+            let child: RObject =
+                harp::try_catch(|| unsafe { R_do_slot(object.sexp, name) }.into())?;
+            return Ok(EnvironmentVariableNode::Concrete { object: child });
+        }
 
-                    // TODO: use ? here, but this does not return a crate::error::Error, so
-                    //       maybe use anyhow here instead ?
-                    let row_index = path_element.parse::<isize>().unwrap();
+        // R6 objects may be accessed with special elements called <methods> and <private>.
```
lionel-, line 1035: Do you think we could move all that code to the new Ark methods?

It'd be nice to move all that complexity to the R side.

crates/ark/src/variables/variable.rs
```diff
@@ -918,132 +931,216 @@ impl PositronVariable {
...
+        match node {
+            EnvironmentVariableNode::Concrete { object } => {
+                Self::get_concrete_child_node(object, path_elt)
+            },
+
+            EnvironmentVariableNode::R6Node { object, name } => {
+                match name.as_str() {
+                    "<private>" => {
+                        let env = Environment::new(object);
+                        let enclos = Environment::new(RObject::view(env.find(".__enclos_env__")?));
```
lionel-, line 1094: Could you change all these `view()` to `new()` please? It's too much work to assess safety. In general `view()` should only be used for dispatch.

crates/ark/src/variables/variable.rs
```diff
@@ -918,132 +931,216 @@ impl PositronVariable {
...
+                        let private = Environment::new(RObject::view(enclos.find("private")?));
+
+                        // TODO: it seems unlikely that private would host active bindings
```
lionel-, line 1097: I don't remember, does `find()` return the active binding function? I vaguely remember it depends on the R version, with older versions returning the function and newer returning the value.

It seems like this could all be cleaned up by using safer / more explicit accessors, because this code currently raises lots of questions.

crates/ark/src/variables/variable.rs
```diff
@@ -918,132 +931,216 @@ impl PositronVariable {
...
+        // For them, we'll have to build the next node artifically.
+        if r_inherits(object.sexp, "R6") && access_key.starts_with("<") {
+            return Ok(EnvironmentVariableNode::R6Node {
+                object,
+                name: access_key.clone(),
+            });
+        }
 
-                    EnvironmentVariableNode::VectorElement {
+        match r_typeof(*object) {
```
lionel-, line 1044: While you're in there, could you please change all `*obj` to the clearer `obj.sexp`?

crates/ark/src/variables/variable.rs
```diff
@@ -918,132 +931,216 @@ impl PositronVariable {
...
+                return Err(harp::Error::Anyhow(anyhow!(
+                    "Can't subset an atomic vector even further, got {path_elt}"
+                )));
+            },
+
+            EnvironmentVariableNode::Matrixcolumn { object, index } => unsafe {
+                let dim = IntegerVector::new(Rf_getAttrib(*object, R_DimSymbol))?;
+                let n_row = dim.get_unchecked(0).unwrap() as isize;
+
+                // TODO: use ? here, but this does not return a crate::error::Error, so
```
lionel-, line 1124: yep it's fine to return an `Error::Anyhow` variant.

crates/ark/src/variables/variable.rs
```diff
@@ -1336,6 +1433,88 @@ impl PositronVariable {
...
+        let result: Option<RObject> = ArkGenerics::VariableGetChildren
+            .try_dispatch(value, vec![])
+            .map_err(|err| harp::Error::Anyhow(err))?;
+
+        match result {
+            None => Ok(None),
+            Some(value) => {
+                // Make sure value is a list before using inspect_list
+                if !r_typeof(value.sexp) == LISTSXP {
+                    return Err(harp::Error::Anyhow(anyhow!(
```
lionel-, line 1447: Should this method return an `anyhow::Result` instead of `harp::Result`?

crates/ark/src/variables/variable.rs
```diff
@@ -1336,6 +1433,88 @@ impl PositronVariable {
...
+                let list = List::new(value.sexp)?;
+                let n = unsafe { list.len() };
+
+                let names = local!({
+                    let r_names = unsafe { RObject::new(Rf_getAttrib(value.sexp, R_NamesSymbol)) };
+                    if r_is_null(r_names.sexp) {
+                        return vec![None; n];
+                    }
+
+                    let names = unsafe { CharacterVector::new_unchecked(r_names) };
```
lionel-, line 1466: Simplify with `RObject::attr()` or `RObject::names()`?

crates/ark/src/variables/variable.rs
```diff
@@ -1336,6 +1433,88 @@ impl PositronVariable {
...
+                }
+
+                // This is essentially the same as Self::inspect_list but with modified `access_key`
+                // that adds more information about the object:
+                // 1. Provide the name and the index for the `get_child_at` method.
+                // 2. (Not necessary) Given an access key, we can detect if we want to apply a custom get_child_method.
+                let list = List::new(value.sexp)?;
+                let n = unsafe { list.len() };
+
+                let names = local!({
```
lionel-, line 1460: inline `local`

lionel- (approved):
Looking good. The refactored code still feels nested, it'd be nice to clean it up a bit further when we get a chance.

crates/ark/src/variables/variable.rs
dfalbel: In this first specific case no, because that just means that the access key does not come from a 'custom' inspect method.

The others could be logged, but in theory they could still be false positives (even though unlikely), an object could have an access key that's `custom-key-a-b` and get parsed until the below, so my hunch is that we don't need to do anything here.

crates/ark/src/variables/variable.rs
dfalbel: Yes, I have an initial implementation here: https://github.com/dfalbel/testVariablesPaneExtension/blob/main/R/R6.R
It's still pretty much a draft and depends a lot on rlang, but we could move it to ark and inron it out. I'd rather do it in a follow up PR though.

crates/ark/src/variables/variable.rs
dfalbel: Done in https://github.com/posit-dev/ark/pull/561/commits/617ced8acb1015cbe2085d27e3b9631b5092f600 but further simplified in https://github.com/posit-dev/ark/pull/561/commits/c7ae20e423fd3921d88fef90c12926dfc156cf4a

crates/ark/src/variables/variable.rs
dfalbel: Done in https://github.com/posit-dev/ark/pull/561/commits/c7ae20e423fd3921d88fef90c12926dfc156cf4a

crates/ark/src/variables/variable.rs
dfalbel: Yes! https://github.com/posit-dev/ark/pull/561/commits/98f9996d47ccc65882b97f89ea6afdcecf303724

crates/ark/src/variables/variable.rs
dfalbel: Fixed in https://github.com/posit-dev/ark/pull/561/commits/c57868a9258e8c9bb697cc718d3a8bb342930585

crates/ark/src/variables/variable.rs
dfalbel: Done in https://github.com/posit-dev/ark/pull/561/commits/4c9b962340639f3d07ac7628ddf717ba7603c2d1

crates/ark/src/variables/variable.rs
dfalbel: Fixed in https://github.com/posit-dev/ark/pull/561/commits/82fe4409c2be208c3b50f46a7a6d91e64d013062

crates/ark/src/variables/variable.rs
dfalbel: I'm not sure too, this piece of code is not new thugh. Should we change it a follow up PR, maybe just rewriting in R: https://github.com/posit-dev/ark/pull/561#discussion_r1810922083

Issue #562, opened 2024-10-03
Use actual Ark binary in integration tests?


lionel-:
We currently launch Ark in a background thread but we're seeing some weirdness on Windows CI, such as stack overflows: https://github.com/posit-dev/ark/actions/runs/11143776136/job/30969594068

```
thread 'dummy_kernel' panicked at crates\ark\src\interface.rs:1653:9:
Unexpected longjump while reading console: Unexpected longjump
Likely caused by:
Error: Error: C stack usage  2069919488 is too close to the limit

...

Caused by:
  process didn't exit successfully: `D:\a\ark\ark\target\debug\deps\kernel-55e8a51febeb10a6.exe` (exit code: 0xc0000409, STATUS_STACK_BUFFER_OVERRUN)
```

It's _possible_ running R in the main thread (by running a normal Ark process) would solve this. Also it would make the integration tests a bit closer to real usage.

We're almost there since the integration tests already start a pretty normal Jupyter kernel that we connect to via zmq.

DavisVaughan:
We've had issues where we've broken something in release ark and can't catch it during testing because testing behavior works slightly differently https://github.com/posit-dev/ark/pull/566#issuecomment-2394939383 - in theory this would help avoid that

lionel-:
We're now happy with our setup where:

- We spawn a full R-bearing kernel and communicate with it through our `DummyFrontend` client.
- Use nextest as our test harness to isolate each test in its own process

Spawning an actual Ark process for tests would make them marginally "truer", but for little additional benefit compared to our current setup.

PR #566, opened 2024-10-04
Fix `r_task::initialize()` timing issue

DavisVaughan:
Both `modules::initialize()` and `resource_loaded_namespaces()` call `r_task::spawn_idle()`. Even though that is async, it blocks until the task channels are initialized, so `r_task::initialize()` has to run first!

---

We didn't catch this during testing because `r_task::spawn_idle()` has that escape hatch during testing and just runs immediately. But CI in Positron caught an issue, and it was easy to reproduce by just starting up R with the latest version of ark in Positron. Things crashed immediately.

### DIFF
--- a/crates/ark/src/interface.rs
+++ b/crates/ark/src/interface.rs
@@ -387,7 +387,13 @@ impl RMain {
                     .or_log_error(&format!("Failed to source startup file '{file}' due to"));
             }
 
-            // Initialize support functions (after routine registration)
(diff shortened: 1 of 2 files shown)


lionel-:
> We didn't catch this during testing because r_task::spawn_idle() has that escape hatch during testing and just runs immediately.

That's a good reason to switch to an independent binary in integration tests, so we don't hit any testing path at all: #562.

PR #568, opened 2024-10-04
Factor out `handle_active_request()` to correct multiline input request issue

DavisVaughan:
Addresses https://github.com/posit-dev/positron/issues/4901

The problem was that in something like:

```r
val <- readline("prompt>")
paste0(val, "-suffix")
```

if you send that whole selection, then we'd process the `val <- readline("prompt>")` line, but then R calls `read_console()` back and we'd just immediately shove `paste0(val, "-suffix")` through as the reply to the readline request! We need to use the `prompt_info()` to recognize that an intermediate expression has put us into an `input_request` state, and handle that before we handle `pending_lines`.

The ordering of our state machine is now:
- Handle input requests
    - Falls through to event loop to wait for input reply
- Then pending lines
- Then close out active requests
    - Falls through to event loop to wait for next user input

I've accomplished this by factoring out `handle_active_request()`. This takes that big if/else branch related to the active request and gives it its own function. The logic in that if/else was actually pretty tricky, and I think it is much cleaner now. It also allows us to sneak in `handle_pending_lines()` between the input request check and closing out the active request.

### DIFF
--- a/crates/ark/src/interface.rs
+++ b/crates/ark/src/interface.rs
@@ -649,18 +649,6 @@ impl RMain {
         // debug call text to maintain the debug state.
         self.dap.finalize_call_text();
 
-        // TODO: Can we remove this below code?
-        // If the prompt begins with "Save workspace", respond with (n)
-        //
-        // NOTE: Should be able to overwrite the `Cleanup` frontend method.
-        // This would also help with detecting normal exits versus crashes.
-        if info.input_prompt.starts_with("Save workspace") {
-            match Self::on_console_input(buf, buflen, String::from("n")) {
-                Ok(()) => return ConsoleResult::NewInput,
-                Err(err) => return ConsoleResult::Error(err),
-            }
-        }
-
         // We get called here everytime R needs more input. This handler
         // represents the driving event of a small state machine that manages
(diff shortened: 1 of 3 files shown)


crates/ark/src/interface.rs
```diff
@@ -640,20 +640,34 @@ impl RMain {
...
         _hist: c_int,
     ) -> ConsoleResult {
+        let info = Self::prompt_info(prompt);
+        log::trace!("R prompt: {}", info.input_prompt);
+
+        // Upon entering read-console, finalize any debug call text that we were capturing.
+        // At this point, the user can either advance the debugger, causing us to capture
+        // a new expression, or execute arbitrary code, where we will reuse a finalized
+        // debug call text to maintain the debug state.
+        self.dap.finalize_call_text();
```
DavisVaughan, line 650: I'm not entirely sure if it matters that this now runs _before_ pending lines are handled 🤔

lionel- (approved):
LGTM! It now makes sense to me to do that since input requests are nested in execute-requests (https://github.com/posit-dev/positron/issues/4901#issuecomment-2396438954).

PR #571, opened 2024-10-06
Primitive functionality to display comment sections in outlines

kv9898:
Attempt to (partly) fix https://github.com/posit-dev/positron/issues/3822

image

I've managed to make comments which are in the shape of "## Title ####/----/====" appear in the outline with the changes in this pull request, but I struggled to make them appear in a hierarchical form/foldable.

### DIFF
--- a/crates/ark/src/lsp/symbols.rs
+++ b/crates/ark/src/lsp/symbols.rs
@@ -146,7 +146,7 @@ fn index_node(
         let comment_text = contents.node_slice(&node)?.to_string();
 
         // Check if the comment starts with one or more '#' followed by any text and ends with 4+ punctuations
-        if let Some((level, title)) = parse_comment_as_section(&comment_text) {
+        if let Some((_level, title)) = parse_comment_as_section(&comment_text) {
             // Create a symbol based on the parsed comment
             let start = convert_point_to_position(contents, node.start_position());
             let end = convert_point_to_position(contents, node.end_position());
--- a/crates/ark/src/lsp/symbols.rs
+++ b/crates/ark/src/lsp/symbols.rs
@@ -60,7 +60,7 @@ pub fn symbols(params: &WorkspaceSymbolParams) -> anyhow::Result<Vec<SymbolInfor
             IndexEntryData::Section { level: _, title } => {
                 info.push(SymbolInformation {
                     name: title.to_string(),
-                    kind: SymbolKind::MODULE,
+                    kind: SymbolKind::STRING,
                     location: Location {
                         uri: Url::from_file_path(path).unwrap(),
                         range: entry.range,
--- a/crates/ark/src/lsp/symbols.rs
+++ b/crates/ark/src/lsp/symbols.rs
@@ -170,7 +170,6 @@ fn index_node(
         }
     }
 
-    // Existing code for indexing assignments or other nodes...
     if matches!(
         node.node_type(),
         NodeType::BinaryOperator(BinaryOperatorType::LeftAssignment) |
--- a/crates/ark/src/lsp/symbols.rs
+++ b/crates/ark/src/lsp/symbols.rs
@@ -124,9 +124,7 @@ fn is_indexable(node: &Node) -> bool {
 fn parse_comment_as_section(comment: &str) -> Option<(usize, String)> {
     // Match lines starting with one or more '#' followed by some non-empty content and must end with 4 or more '-', '#', or `=`
     // Ensure that there's actual content between the start and the trailing symbols.
-    let comment_re = regex::Regex::new(r"^(#+)\s*([^\s#-].+?)\s*(#{4,}|-{4,}|={4,})$").unwrap();
-
-    if let Some(caps) = comment_re.captures(comment) {
+    if let Some(caps) = indexer::RE_COMMENT_SECTION.captures(comment) {
         let hashes = caps.get(1)?.as_str().len(); // Count the number of '#'
         let title = caps.get(2)?.as_str().trim().to_string(); // Extract the title text without trailing punctuations
         return Some((hashes, title)); // Return the level based on the number of '#' and the title
--- a/crates/ark/src/lsp/symbols.rs
+++ b/crates/ark/src/lsp/symbols.rs
@@ -127,9 +127,9 @@ fn parse_comment_as_section(comment: &str) -> Option<(usize, String)> {
     let comment_re = regex::Regex::new(r"^(#+)\s*([^\s#-].+?)\s*(#{4,}|-{4,}|={4,})$").unwrap();
 
     if let Some(caps) = comment_re.captures(comment) {
-        let hashes = caps.get(1)?.as_str().len();  // Count the number of '#'
-        let title = caps.get(2)?.as_str().trim().to_string();  // Extract the title text without trailing punctuations
-        return Some((hashes, title));  // Return the level based on the number of '#' and the title
+        let hashes = caps.get(1)?.as_str().len(); // Count the number of '#'
+        let title = caps.get(2)?.as_str().trim().to_string(); // Extract the title text without trailing punctuations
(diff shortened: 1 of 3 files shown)


crates/ark/src/lsp/symbols.rs
```diff
@@ -120,29 +120,73 @@ fn is_indexable(node: &Node) -> bool {
...
+    if let Some(caps) = comment_re.captures(comment) {
+        let hashes = caps.get(1)?.as_str().len();  // Count the number of '#'
+        let title = caps.get(2)?.as_str().trim().to_string();  // Extract the title text without trailing punctuations
+        return Some((hashes, title));  // Return the level based on the number of '#' and the title
+    }
+
+    None
+}
+
 fn index_node(
```
lionel-, line 138: Some tests are needed for this function too. You can do this to get a `Rope` and node:

```rs
let code = "my_code";
let document = Document::new(code, None);
let node = ast.root_node();
```

lionel- (commented):
Thanks for taking a look at this!

Regarding folding, see @juliasilge's comment in https://github.com/posit-dev/positron/pull/4174#issuecomment-2258999703. Have you looked at https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_foldingRange?

But I wonder if we can provide ranges both from the rules in our extension's json config, and from the LSP?

In any case, even without ranges it seems like a good progress on this issue.

lionel-:
@kv9898 No worries, I've added a couple tests, feel free to add more from these examples.

kv9898:
@lionel-  Thank you so much for helping me out!

crates/ark/src/lsp/symbols.rs
```diff
@@ -120,13 +120,56 @@ fn is_indexable(node: &Node) -> bool {
     true
 }
 
+// Function to parse a comment and return the section level and title
+fn parse_comment_as_section(comment: &str) -> Option<(usize, String)> {
+    // Match lines starting with one or more '#' followed by some non-empty content and must end with 4 or more '-' or '#'
+    // Ensure that there's actual content between the start and the trailing symbols.
+    let comment_re = regex::Regex::new(r"^(#+)\s*([^\s#-].+?)\s*(#{4,}|-{4,}|={4,})$").unwrap();
```
DavisVaughan, line 127:
```suggestion
    let comment_re = regex::Regex::new(r"^(#+)\s*([^\s#-=].+?)\s*(#{4,}|-{4,}|={4,})$").unwrap();
```

Right? Since 4+ `=` are also allowed as the section signal?

crates/ark/src/lsp/symbols.rs
```diff
@@ -120,13 +120,56 @@ fn is_indexable(node: &Node) -> bool {
     true
 }
 
+// Function to parse a comment and return the section level and title
+fn parse_comment_as_section(comment: &str) -> Option<(usize, String)> {
+    // Match lines starting with one or more '#' followed by some non-empty content and must end with 4 or more '-' or '#'
```
DavisVaughan, line 125:
```suggestion
    // Match lines starting with one or more '#' followed by some non-empty content and must end with 4 or more '-', '#', or `=`
```

crates/ark/src/lsp/symbols.rs
DavisVaughan, line 127: It turns out that regex are fairly slow to compile on the fly. We typically compile them 1 time on first access and reuse them. The best example is `RE_COMMENT_SECTION`, which uses new std library features for this:
https://github.com/posit-dev/ark/blob/d6ab389bb828db56850948f10b27e3161285be47/crates/ark/src/lsp/indexer.rs#L58-L59

It turns out that `RE_COMMENT_SECTION` is trying to basically do the same thing as what you do here, so I'd just reuse that name here.

crates/ark/src/lsp/symbols.rs
kv9898: That's right! Thank you for pointing this out. `=` is supported, in line with RStudio's behavior.

crates/ark/src/lsp/symbols.rs
```diff
@@ -120,13 +120,56 @@ fn is_indexable(node: &Node) -> bool {
...
+        let title = caps.get(2)?.as_str().trim().to_string(); // Extract the title text without trailing punctuations
+        return Some((hashes, title)); // Return the level based on the number of '#' and the title
+    }
+
+    None
+}
+
 fn index_node(
     node: &Node,
     contents: &Rope,
```
DavisVaughan, line 140: Meta comment

Looks like our `indexer`, which watches the whole "workspace" (i.e. multiple folders of files), also does something similar. i.e. see `index_comment()` there.

That indexer powers `symbols()` in this file, which powers _workspace symbols_. i.e. when you hit `CMD + T` then it pops up all the symbols in the workspace - even in closed files.

Screenshot 2024-10-11 at 3 17 20 PM

What we are changing here powers _document symbols_, i.e. symbols within a single document. This powers the Outline editor view we have been talking about, and it powers this Outline quick pick

Screenshot 2024-10-11 at 3 19 04 PM

---

It will eventually probably be good to merge all the infrastructure here. Ideally we'd have one set of code that we can run on a file that returns all its symbols, and then we map them into either a workspace symbol or document symbol (its essentially `indexer::index_file()` i think, but it doesn't know how to create _nested_ outlines, which is important).

I don't think this should block anything, but we should keep this in mind and open an issue to merge all of this code

crates/ark/src/lsp/symbols.rs
```diff
@@ -120,13 +120,56 @@ fn is_indexable(node: &Node) -> bool {
...
+
+        // Check if the comment starts with one or more '#' followed by any text and ends with 4+ punctuations
+        if let Some((_level, title)) = parse_comment_as_section(&comment_text) {
+            // Create a symbol based on the parsed comment
+            let start = convert_point_to_position(contents, node.start_position());
+            let end = convert_point_to_position(contents, node.end_position());
+
+            let symbol = DocumentSymbol {
+                name: title,                // Use the title without the trailing '####' or '----'
+                kind: SymbolKind::STRING,   // Treat it as a string section
```
DavisVaughan, line 156: I noticed that at the top of this file in `symbols()` we represent Sections with `SymbolKind::MODULE`.

I do think `SymbolKind::STRING` looks a little better, so could you please change `MODULE` to `STRING` to keep them in sync?

crates/ark/src/lsp/symbols.rs
kv9898: That sounds very desirable! I would really appreciate the nested outline and foldable comment functionalities. Sadly, as a student primarily working on R-based data analysis projects, I'm new to Rust and unfamiliar with the ark and LSP architecture, I'm struggling to carry this out.

DavisVaughan: We can take care of it!

crates/ark/src/lsp/symbols.rs
```diff
@@ -120,13 +120,54 @@ fn is_indexable(node: &Node) -> bool {
...
+    // Match lines starting with one or more '#' followed by some non-empty content and must end with 4 or more '-', '#', or `=`
+    // Ensure that there's actual content between the start and the trailing symbols.
+    if let Some(caps) = indexer::RE_COMMENT_SECTION.captures(comment) {
+        let hashes = caps.get(1)?.as_str().len(); // Count the number of '#'
+        let title = caps.get(2)?.as_str().trim().to_string(); // Extract the title text without trailing punctuations
+        return Some((hashes, title)); // Return the level based on the number of '#' and the title
+    }
+
+    None
+}
```
kv9898, line 134: Here the function does return the number of hashes (level of comment). I wanted this to help with the foldable comments/nested outlines. But I struggled to get this to work.

lionel-:
Thanks a lot @kv9898!

PR #575, opened 2024-10-08
Streamline handling of Shell replies

lionel-:
Some clean ups following discussion in #547.

- Shell handlers now return an `amalthea::Result`.

- New `amalthea::Error::ShellErrorReply` variant for error replies. When a handler returns one, it's forwarded to the client as a Shell reply.

- The `amalthea::Error::ShellErrorExecuteReply` variant is specifically for execution replies as they are the only ones that need an execution count.
 
  (This count is currently created by the R thread. We might want to move some of this handling to the Shell socket: it could create it, bump it when appropriate, etc. The Shell socket might also become in charge of sending Inputs and Results to IOPub to remove as many protocol details from Ark as possible).

- Removed the `ExecuteResponse` enum as we now pass that information through an `amalthea::Result`.

- New `Shell::send_execute_error()` method. Sends errors for execute requests with the execution count.

- `Shell::handle_request()` is now in charge of sending replies from handlers. All errors, including internal (i.e. not the two variants discussed above) are forwarded to the frontend as replies.


crates/ark/src/interface.rs
```diff
@@ -1328,7 +1326,7 @@ impl RMain {
     fn make_execute_response_result(
```
DavisVaughan, line 1326: Can we get rid of all usage of "response" in this file? I find it confusing

crates/amalthea/src/language/shell_handler.rs
```diff
@@ -31,7 +29,7 @@ pub trait ShellHandler: Send {
     async fn handle_info_request(
         &mut self,
         req: &KernelInfoRequest,
-    ) -> Result<KernelInfoReply, Exception>;
+    ) -> crate::Result<KernelInfoReply>;
```
DavisVaughan, line 32: Hmmmmmmmmmm

I actually disagree with this strongly enough to request that we talk about this one first.

This change implies that rather than encoding that we return a well-defined `Exception`, we now return an arbitrary `amalthea::Error`. That is not what the jupyter spec declares as the possible return value here.

---

Oh hmm. I see now that what you are doing in `handle_request()` is taking that `amalthea::Error` and mapping it into the appropriate `Exception` and then you send that out.

I thought for sure we instead wanted to return something like this `KernelInfoReply` here:

```rust
pub enum KernelInfoReply {
    Ok(KernelInfoReplyData),
    Error(ReplyException),
}

pub struct KernelInfoReplyData {
    /// The execution status ("ok" or "error")
    pub status: Status,

    /// Version of messaging protocol
    pub protocol_version: String,

    /// Information about the language the kernel supports
    pub language_info: LanguageInfo,

    /// A startup banner
    pub banner: String,

    /// Whether debugging is supported
    pub debugger: bool,

    /// A list of help links
    pub help_links: Vec<HelpLink>,
}

pub struct ReplyException {
    /// The name of the exception
    pub ename: String,

    /// The value/description of the exception
    pub evalue: String,

    /// List of traceback frames, as strings
    pub traceback: Vec<String>,
}
```

Where `ReplyException` is reused across reply types, and `ExecuteReplyException` is a special variant with the execution count.

crates/amalthea/src/socket/shell.rs
```diff
@@ -145,7 +147,7 @@ impl Shell {
                 self.handle_request(req, |h, r| self.handle_comm_info_request(h, r))
             },
             Message::CommOpen(req) => self.handle_comm_open(req),
-            Message::CommMsg(req) => self.handle_request(req, |h, r| self.handle_comm_msg(h, r)),
+            Message::CommMsg(req) => self.handle_comm_msg(req),
```
DavisVaughan, line 150: There are 2 very important things I want to call out here:
- Prior to this change, we were sending idle/busy messages _twice_ for each comm msg! Once in `handle_request()` and once in `handle_comm_msg()`! (any way to add an integration test?)
- Are we 100% sure that `handle_comm_msg()` does not need the lock on the `shell_handler`? It previously got the lock but was not actively using it, but could it have been passively using it?

crates/amalthea/src/socket/shell.rs
```diff
@@ -145,7 +147,7 @@ impl Shell {
                 self.handle_request(req, |h, r| self.handle_comm_info_request(h, r))
```
DavisVaughan, line 147: Your change to remove `handle_request()` from `handle_comm_msg()` below implies we could also do it here with `handle_comm_info_request()`, because it doesn't use the `shell_handler` either?

But unlike `handle_comm_msg()`, `handle_comm_info_request()` _does_ have a return value it needs to send back. So that makes things complicated.

crates/amalthea/src/socket/shell.rs
```diff
@@ -452,25 +446,7 @@ impl Shell {
...
-                        req.send_error::<CommWireMsg>(err, &self.socket)?;
-
-                        // Return an error to the caller indicating that the
-                        // comm could not be opened due to the invalid open
-                        // call.
-                        return Err(Error::InvalidCommMessage(
-                            req.content.target_name.clone(),
-                            data_str,
-                            errname,
-                        ));
```
DavisVaughan, line 471: Did we do lose this subtle behavior?

i.e. did we lose the `CommWireMsg` being sent back? And did we lose the special `InvalidCommMessage`?

crates/amalthea/src/socket/shell.rs
lionel-: yep I noticed the double status too. I'll look into a test.

I don't think we need a lock on the shell handler if we're not calling shell methods. The shell handler will not be called by anyone else until the request handler returns.

Ideally we'd set up things in such a way we don't need a shell handler lock at all. Well after taking a quick look it does seem the shell thread can fully own the handler.

crates/amalthea/src/socket/shell.rs
lionel-: yes but these are not conform to the protocol. There is no reply to a `comm_open` so the error is logged on our end instead. We should reply a `comm_close` though. Here is an example impl: https://github.com/ipython/ipykernel/blob/3089438d84447bf859556b89f528ef59332803c5/ipykernel/comm/manager.py#L30

crates/amalthea/src/socket/shell.rs
lionel-: Looks like our existing test purposefully ignored status messages: https://github.com/posit-dev/ark/blob/ba2bf26debd055e84c1bb61171dd956be06959f0/crates/amalthea/tests/client.rs#L403-L429

That's likely because there the ordering of the Idle and comm reply messages is undetermined since the latter is emitted on a different thread and there is no waiting mechanism for comm message handling.

I've refactored the loop so that we do test for status messages without introducing undeterminism.

crates/amalthea/src/socket/shell.rs
lionel-: We now reply a `comm_close` and this is now covered by a test.

crates/amalthea/src/socket/shell.rs
DavisVaughan: > Since Shell is supposed to be locked while the kernel is busy anyway, I think the kernel should own the handling. So the message handling would run on the kernel/RMain thread (which would own the shell socket) rather than on the shell thread (which would be removed).

But not all Shell messages require the lock right? Like `handle_comm_close`. So we'd be slowing those down potentially?

lionel-: There would be no lock, just blocking handling, and `handle_comm_close` would be a very quick handler.

crates/amalthea/src/kernel.rs
```diff
@@ -104,7 +104,6 @@ pub fn connect(
     )?;
     let shell_port = port_finalize(&shell_socket, connection_file.shell_port)?;
 
-    let shell_clone = shell_handler.clone();
     let iopub_tx_clone = iopub_tx.clone();
     let lsp_handler_clone = lsp_handler.clone();
     let dap_handler_clone = dap_handler.clone();
```
DavisVaughan, line 109: While we are here - these don't need to be clones either!

crates/amalthea/src/wire/exception.rs
```diff
@@ -20,3 +20,13 @@ pub struct Exception {
     /// List of traceback frames, as strings
     pub traceback: Vec<String>,
 }
+
+impl Exception {
+    pub fn internal_error(evalue: String) -> Self {
+        Exception {
```
DavisVaughan, line 26:
```suggestion
        Self {
```

crates/amalthea/src/wire/input_request.rs
```diff
@@ -49,14 +49,14 @@ pub struct UiCommFrontendRequest {
     pub originator: Originator,
 
     /// The response channel for the request
```
DavisVaughan, line 51:
```suggestion
    /// The reply channel for the request
```

crates/ark/src/interface.rs
```diff
@@ -230,8 +228,8 @@ pub struct RMain {
 struct ActiveReadConsoleRequest {
     exec_count: u32,
     request: ExecuteRequest,
-    originator: Originator,
-    response_tx: Sender<ExecuteResponse>,
+    orig: Originator,
```
DavisVaughan, line 231: I actually purposefully made this `originator` everywhere to avoid shorthand notation

crates/amalthea/src/socket/shell.rs
```diff
@@ -88,14 +81,14 @@ impl Shell {
...
-        shell_handler: Arc<Mutex<dyn ShellHandler>>,
+        shell_handler: Box<dyn ShellHandler>,
         lsp_handler: Option<Arc<Mutex<dyn ServerHandler>>>,
         dap_handler: Option<Arc<Mutex<dyn ServerHandler>>>,
     ) -> Self {
         Self {
             socket,
             iopub_tx,
-            shell_handler,
+            shell_handler: RefCell::new(shell_handler),
```
DavisVaughan, line 91: Maybe create it on its own line, with a comment about the need for RefCell? (because some handler methods take a `mut` self?)

crates/amalthea/src/socket/shell.rs
```diff
@@ -317,25 +265,30 @@ impl Shell {
...
+        let result = self.open_comm(shell_handler, req.clone());
+
+        // There is no error reply for a comm open request. Instead we must send
+        // a `comm_close` message as soon as possible. The error is logged on our side.
+        if let Err(err) = result {
+            let reply = IOPubMessage::CommClose(CommClose {
+                comm_id: req.content.comm_id.clone(),
+            });
+            self.iopub_tx.send(reply).unwrap();
+            log::warn!("Failed to open comm: {err:?}");
```
DavisVaughan, line 277: warn or error?

DavisVaughan (approved):

crates/amalthea/src/kernel.rs
lionel-: I wish the compiler would let us know.

crates/amalthea/src/socket/shell.rs
lionel-: Warn otherwise this shows up in the test output.

PR #576, opened 2024-10-08
Add support for alternative `RegistrationFile`

### LINKED ISSUE #563: Fix Jupyter connection race condition
We're seeing issues in Windows CI that look like:

```
thread 'dummy_kernel' panicked at crates/ark/src/start.rs:112:9:
Couldn't connect to frontend: SocketBindError("Control", "tcp://127.0.0.1:23393", Address already in use)
```

I can also reproduce it locally by running integration tests in a loop. It's a mystery why it happens so often on the Windows CI though.

This might be due to the "classic jupyter race condition": https://github.com/jupyter/jupyter_client/issues/487. In Jupyter's connection scheme, the client searches for available ports, communicates those to the server which then tries to bind to them. This fails if any of the ports end up getting used up in the meantime.

There is no other solution than to let the kernel pick the ports. In the linked issue, they suggest implementing this scheme:

> The client opens a socket A, passes the port of this socket to the kernel that it launches and waits the kernel starts, finds free ports to bind shell, control, stdin, heartbeat and iopub sockets. Then it connects to the socket A of the client, sends a message containing these ports, and close the connection to socket A. Upon reception of this message, the client connects to the kernel and closes the socket A.

Essentially the client would pick a port for a handshake socket, bind to it, and send this connection info:

```json
{
  "transport": "tcp",
  "signature_scheme": "hmac-sha256",
  "ip": "127.0.0.1",
  "key": "a0436f6c-1916-498b-8eb9-e81ab9368e84",
  "handshake_port": 40885
}
```

And the server would connect to the handshake socket and send back:

```json
{
  "control_port": 50160,
  "shell_port": 57503,
  "stdin_port": 52597,
  "hb_port": 42540,
  "iopub_port": 40885,
}
```

On the server side, it looks like we can use `:*` or `:0` to let the OS pick a port: https://stackoverflow.com/questions/16699890/connect-to-first-free-port-with-tcp-using-0mq

Positron could also use this to make the initial connection to Ark more robust, cc @jmcphers.

DavisVaughan:
Closes #563 
Joint work with @lionel- 

This PR implements the alternative `RegistrationFile` approach outlined in JEP 66 that allows for a "handshake" to occur between the client and server on startup. In particular, it allows the server to be in charge of picking the ports, and immediately binds to them as it picks them, avoiding any race conditions here.

If the `--connection_file` argument provided to ark can parse into this structure:

```rust
pub struct RegistrationFile {
    /// The transport type to use for ZeroMQ; generally "tcp"
    pub transport: String,

    /// The signature scheme to use for messages; generally "hmac-sha256"
    pub signature_scheme: String,

    /// The IP address to bind to
    pub ip: String,

    /// The HMAC-256 signing key, or an empty string for an unauthenticated
    /// connection
    pub key: String,

    /// ZeroMQ port: Registration messages (handshake)
    pub registration_port: u16,
}
```

Then we assume we are going to be using the handshake method of connecting. Otherwise we parse into the typical `ConnectionFile` structure and assume the Client picked the ports.

We expect that the Client _binds_ to a `zmq::REP` socket on `registration_port`. Ark, as the Server, will then _connect_ to this `registration_port` as a `zmq::REQ` socket.

Ark will pick ports, bind to them, and send this message over the registration socket:

```rust
pub struct HandshakeRequest {
    /// ZeroMQ port: Control channel (kernel interrupts)
    pub control_port: u16,

    /// ZeroMQ port: Shell channel (execution, completion)
    pub shell_port: u16,

    /// ZeroMQ port: Standard input channel (prompts)
    pub stdin_port: u16,

    /// ZeroMQ port: IOPub channel (broadcasts input/output)
    pub iopub_port: u16,

    /// ZeroMQ port: Heartbeat messages (echo)
    pub hb_port: u16,
}
```

Ark will then _immediately_ block, waiting for this `HandshakeReply`:

```rust
pub struct HandshakeReply {
    /// The execution status ("ok" or "error")
    pub status: Status,
}
```

This is just a receipt from the Client that confirms that it received the socket information.

If ark does not receive this reply after a few seconds, it will shut itself down.

Ark disconnects from the registration socket after receiving the `HandshakeReply`, and the kernel proceeds to start up.

Co-authored-by: Lionel Henry <lionel@posit.co>
Co-authored-by: Davis Vaughan <davis@posit.co>

### DIFF
--- a/crates/amalthea/src/fixtures/dummy_frontend.rs
+++ b/crates/amalthea/src/fixtures/dummy_frontend.rs
@@ -185,6 +185,15 @@ impl DummyFrontend {
         )
         .unwrap();
 
+        // TODO!: Without this sleep, `IOPub` `Busy` messages sporadically
+        // don't arrive when running integration tests. I believe this is a result
+        // of PUB sockets dropping messages while in a "mute" state (i.e. no subscriber
+        // connected yet). Even though we run `iopub_socket.subscribe()` to subscribe,
+        // it seems like we can return from this function even before our socket
+        // has fully subscribed, causing messages to get dropped.
+        // https://libzmq.readthedocs.io/en/latest/zmq_socket.html
+        std::thread::sleep(std::time::Duration::from_millis(500));
+
         Self {
             _control_socket,
             shell_socket,
--- a/crates/amalthea/src/fixtures/dummy_frontend.rs
+++ b/crates/amalthea/src/fixtures/dummy_frontend.rs
@@ -54,7 +54,7 @@ impl DummyFrontend {
         let shell_id = rand::thread_rng().gen::<[u8; 16]>();
 
         // Create a new kernel session from the key
-        let session = Session::create(key.clone()).unwrap();
+        let session = Session::create(&key).unwrap();
 
         let ctx = zmq::Context::new();
 
--- a/crates/amalthea/src/fixtures/dummy_frontend.rs
+++ b/crates/amalthea/src/fixtures/dummy_frontend.rs
@@ -30,6 +30,9 @@ pub struct DummyConnection {
     pub ctx: zmq::Context,
     pub session: Session,
     pub key: String,
+    pub ip: String,
+    pub transport: String,
+    pub signature_scheme: String,
 }
 
 pub struct DummyFrontend {
@@ -57,8 +60,12 @@ impl DummyConnection {
         // Create a zmq context for all sockets we create in this session
         let ctx = zmq::Context::new();
 
+        let ip = String::from("127.0.0.1");
+        let transport = String::from("tcp");
+        let signature_scheme = String::from("hmac-sha256");
+
         // Bind to a random port using `:0`
-        let endpoint = String::from("tcp://127.0.0.1:0");
+        let endpoint = String::from("{transport}://{ip}:0");
 
         let registration_socket = Socket::new(
             session.clone(),
@@ -75,6 +82,9 @@ impl DummyConnection {
             ctx,
             session,
             key,
+            ip,
+            transport,
+            signature_scheme,
         }
     }
 
@@ -82,18 +92,12 @@ impl DummyConnection {
     /// this synthetic frontend. Uses a handshake through a registration
     /// file to avoid race conditions related to port binding.
     pub fn get_connection_files(&self) -> (ConnectionFile, RegistrationFile) {
-        let ip = String::from("127.0.0.1");
-        let transport = String::from("tcp");
-        let signature_scheme = String::from("hmac-sha256");
-        let key = self.key.clone();
-        let registration_port = crate::kernel::port_from_socket(&self.registration_socket).unwrap();
-
         let registration_file = RegistrationFile {
-            ip,
-            transport,
-            signature_scheme,
-            key,
-            registration_port,
+            ip: self.ip.clone(),
+            transport: self.transport.clone(),
+            signature_scheme: self.signature_scheme.clone(),
+            key: self.key.clone(),
+            registration_port: crate::kernel::port_from_socket(&self.registration_socket).unwrap(),
         };
 
         let connection_file = registration_file.as_connection_file();
@@ -151,7 +155,8 @@ impl DummyFrontend {
         )
         .unwrap();
 
-        // Subscribe to IOPub!
+        // Subscribe to IOPub! Server is the one that sent us this port,
+        // so its already connected on its end.
         iopub_socket.subscribe().unwrap();
 
         let stdin_socket = Socket::new(
--- a/crates/amalthea/src/fixtures/dummy_frontend.rs
+++ b/crates/amalthea/src/fixtures/dummy_frontend.rs
@@ -6,13 +6,16 @@
  */
 
 use assert_matches::assert_matches;
+use rand::Rng;
 use serde_json::Value;
 
 use crate::connection_file::ConnectionFile;
+use crate::registration_file::RegistrationFile;
 use crate::session::Session;
 use crate::socket::socket::Socket;
 use crate::wire::execute_input::ExecuteInput;
 use crate::wire::execute_request::ExecuteRequest;
+use crate::wire::handshake_reply::HandshakeReply;
 use crate::wire::input_reply::InputReply;
 use crate::wire::jupyter_message::JupyterMessage;
 use crate::wire::jupyter_message::Message;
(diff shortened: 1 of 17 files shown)


lionel-:
TODO for JEP65:

### Subscribe to all subscription messages

```c++
// Set xpub_verbose option to 1 to pass all subscription messages (not only unique ones).
m_publisher.set(zmq::sockopt::xpub_verbose, 1);
```

From the guide (): https://zguide.zeromq.org/docs/chapter5/

> by default, the XPUB socket does not report duplicate subscriptions, which is what you want when you’re naively connecting an XPUB to an XSUB. Our example sneakily gets around this by using random topics so the chance of it not working is one in a million. In a real LVC proxy, you’ll want to use the ZMQ_XPUB_VERBOSE option that we implement in Chapter 6 - The ZeroMQ Community as an exercise.


### IOPub needs to listen on its socket

How do we listen to both XPUB and channel messages?

Have the forwarding thread own the socket? Need to send messages via that thread too...

crates/amalthea/src/fixtures/dummy_frontend.rs
```diff
@@ -22,126 +25,189 @@ use crate::wire::status::ExecutionState;
...
         .unwrap();
 
+        // TODO!: Without this sleep, `IOPub` `Busy` messages sporadically
+        // don't arrive when running integration tests. I believe this is a result
+        // of PUB sockets dropping messages while in a "mute" state (i.e. no subscriber
+        // connected yet). Even though we run `iopub_socket.subscribe()` to subscribe,
+        // it seems like we can return from this function even before our socket
+        // has fully subscribed, causing messages to get dropped.
+        // https://libzmq.readthedocs.io/en/latest/zmq_socket.html
+        std::thread::sleep(std::time::Duration::from_millis(500));
```
DavisVaughan, line 195: Addressed in https://github.com/posit-dev/ark/pull/577

crates/amalthea/src/wire/handshake_request.rs
```diff
@@ -0,0 +1,36 @@
...
+ * Copyright (C) 2024 Posit Software, PBC. All rights reserved.
+ *
+ */
+
+use serde::Deserialize;
+use serde::Serialize;
+
+use crate::wire::jupyter_message::MessageType;
+
+/// Represents a reply to a handshake_request
```
DavisVaughan, line 13:
```suggestion
/// Represents a handshake_request
```

crates/amalthea/tests/client.rs
```diff
@@ -69,25 +71,36 @@ fn test_kernel() {
...
-        StreamBehavior::None,
-        stdin_request_rx,
-        stdin_reply_tx,
-    ) {
-        panic!("Error connecting kernel: {err:?}");
-    };
-
-    // Give the kernel a little time to start up
-    info!("Waiting 500ms for kernel startup to complete");
-    std::thread::sleep(std::time::Duration::from_millis(500));
```
DavisVaughan, line 86: Notably I don't think we need this sleep anymore

crates/ark/src/fixtures/dummy_frontend.rs
```diff
@@ -52,36 +53,39 @@ impl DummyArkFrontend {
...
+                    String::from("--no-save"),
+                    String::from("--no-restore"),
+                ],
+                None,
+                session_mode,
+                false,
+            );
+
             RMain::start();
         });
```
DavisVaughan, line 85: @lionel- since we are moving the kernel `setup()` back onto the same thread as the `RMain::start()`, I think we can move `R_MAIN_THREAD_ID` back into the `setup()` method to solve https://github.com/posit-dev/positron/issues/4973

In fact, after we merge this we can probably combine the `RMain::setup()` and `RMain::start()` again. I see no reason they should be separate now.

crates/amalthea/src/kernel.rs
```diff
@@ -78,418 +55,563 @@ pub enum StreamBehavior {
...
+        handshake(
+            registration_file,
+            &ctx,
+            &session,
+            control_port,
+            shell_port,
+            stdin_port,
+            iopub_port,
+            hb_port,
         )?;
```
DavisVaughan, line 260: After we bind to all our sockets, we do this optional handshake with the client

crates/amalthea/src/kernel.rs
```diff
@@ -78,418 +55,563 @@ pub enum StreamBehavior {
...
+    };
+    let message = JupyterMessage::create(message, None, &session);
+
+    message.send(&registration_socket)?;
+
+    // Wait for the handshake reply with a 5 second timeout.
+    // If we don't get a handshake reply, we are going to eventually panic and shut down.
+    if !registration_socket
+        .poll_incoming(5000)
+        .map_err(|err| Error::ZmqError(registration_socket.name.clone(), err))?
```
DavisVaughan, line 541: Ark waits 5 seconds before self imploding

crates/amalthea/src/kernel.rs
```diff
@@ -78,418 +55,563 @@ pub enum StreamBehavior {
...
+        Ok(port) => port,
+        Err(err) => {
+            return Err(crate::anyhow!(
+                "Can't parse port '{port}' into a `u16` due to {err:?}"
+            ));
+        },
+    };
+
+    Ok(port)
+}
```
DavisVaughan, line 617: This was kind of ugly but should work for TCP sockets. The only other option is to do what the Python implementation does when they pick a random port, which is to just choose some random port number and try to `bind()` to it. If it works, great, and they know the port number. If it doesn't they pick another one.

We just use `0` and let the OS pick a port, but its kind of hard to get at the port number.

crates/amalthea/src/kernel.rs
```diff
@@ -78,418 +55,563 @@ pub enum StreamBehavior {
...
+    // for client messages.
+    let shell_socket = Socket::new(
+        session.clone(),
+        ctx.clone(),
+        String::from("Shell"),
+        zmq::ROUTER,
+        None,
+        connection_file.endpoint(connection_file.shell_port),
+    )?;
+    let shell_port = port_finalize(&shell_socket, connection_file.shell_port)?;
```
lionel-, line 104: Since the port might be chosen randomly by the OS, depending on the details of the connection file, we now potentially have to retrieve the port from the socket with `port_finalize()`.

crates/amalthea/src/kernel.rs
```diff
@@ -33,42 +35,17 @@ use crate::socket::socket::Socket;
...
-    /// Receives message sent to the IOPub socket
-    iopub_rx: Option<Receiver<IOPubMessage>>,
-
-    /// Sends notifications about comm changes and events to the comm manager.
-    /// Use `create_comm_manager_tx` to access it.
-    comm_manager_tx: Sender<CommManagerEvent>,
-
-    /// Receives notifications about comm changes and events
-    comm_manager_rx: Receiver<CommManagerEvent>,
-}
```
lionel-, line 70: We were able to remove all of this state from Amalthea's `Kernel`. For instance the IOPub and CommManager channels are now created by the caller of`connect()`.

That's why the diff in this file is pretty big. We removed the `Kernel` struct altogether and all impls are now free functions in the `kernel` module. Most of the diff comes from the ensuing unnesting.

Issue #569, opened 2024-10-05
Race condition between the R and Amalthea Shell threads


lionel-:
One run of Windows CI failed because a `status` message came in on IOPub before the expected `execute_result`:

```
thread 'test_notebook_execute_request_incomplete_multiple_lines' panicked at crates\amalthea\src\fixtures\dummy_frontend.rs:219:9:
assertion failed: `ExecuteError(JupyterMessage {
  zmq_identities: [],
  header: JupyterHeader {
    msg_id: "45d7303b-4f2b-49de-ad9a-561425034687",
    session: "bc8fb405-6169-4601-9caf-507488d75ee6",
    username: "kernel",
    date: "2024-10-04T14:48:27.391849800+00:00",
    msg_type: "error",
    version: "5.3"
  },
  parent_header: Some(JupyterHeader {
    msg_id: "38b5f3ed-3087-4c14-a0c4-38af2ddb1f87",
    session: "59960967-e8b2-4ba6-86f9-c7cbda70e3f6",
    username: "kernel",
    date: "2024-10-04T14:48:27.145261900+00:00",
    msg_type: "execute_request", version: "5.3"
    }),
  content: ExecuteError {
    exception: Exception {
      ename: "",
      evalue: "Error:\n! \nCan't execute incomplete input:\n1 +\n2 +",
      traceback: ["Error:\n! \nCan't execute incomplete input:\n1 +\n2 +"] } } })`
      does not match `Message::Status(data)`
```

<details>

I think it's caused by us handling requests on two different threads (Amalthea Shell and R) interacting with a third thread (Amalthea IOPub).

When a Shell request comes in, the Amalthea's Shell socket thread (dispatch in https://github.com/posit-dev/ark/blob/cac78eb3d92d54687845ddba9402054ba907b510/crates/amalthea/src/socket/shell.rs#L160, `execute_request` handler in https://github.com/posit-dev/ark/blob/cac78eb3d92d54687845ddba9402054ba907b510/crates/amalthea/src/socket/shell.rs#L221) performs this sequence:

- Send a `Busy` message on IOPub
- Invoke Ark's handler for Shell messages
- Block until Ark's handler responds
- Send the response on its Shell socket ()
- Send `Idle` to IOPub.

Note that IOPub's messages are not sent immediately on the IOPub socket, they are sent to the IOPub thread which relays the messages to the IOPub socket which it owns.

[...]

PR #577, opened 2024-10-09
Implement JEP 65

lionel-:
Closes https://github.com/posit-dev/ark/issues/569

This PR fixes a race condition regarding subscriptions to IOPub that causes clients to miss IOPub messages:

- On startup a client connects to the server sockets of a kernel.
- The client sends a request on Shell.
- The kernel starts processing the request and emits busy on IOPub.

If the client hasn't been able to fully subscribe to IOPub, messages can be lost, in particular the Busy message that encloses the request output.

On the Positron side we fixed it by sending kernel-info requests in a loop until we get a Ready message on IOPub. This signals Positron that the kernel is fully connected and in the Ready state: https://github.com/posit-dev/positron/pull/2207. We haven't implemented a similar fix in our dummy clients for integration tests and we believe this is what is causing the race condition described in #569.

As noted in https://github.com/posit-dev/positron/pull/2207, there is an accepted JEP proposal (JEP 65) that aims at solving this problem by switching to XPUB.

https://jupyter.org/enhancement-proposals/65-jupyter-xpub/jupyter-xpub.html
https://github.com/jupyter/enhancement-proposals/pull/65

The XPUB socket allows the server to get notified of all new subscriptions. A message of type `iopub_welcome` is sent to all connected clients. They should generally ignore it but clients that have just started up can use it as a cue that IOPub is correctly connected and that they won't miss any output from that point on.

Approach:

The subscription notification comes in as a message on the IOPub socket. This is problematic because the IOPub thread now needs to listens to its crossbeam channel and to the 0MQ socket at the same time, which isn't possible without resorting to timeout polling. So we use the same approach and infrastructure that we implemented in https://github.com/posit-dev/ark/pull/58 for listeing to both input replies on the StdIn socket and interrupt notifications on a crossbeam channel. The forwarding thread now owns the IOPub socket and listens for subscription notifications and fowrards IOPub messages coming from the kernel components.

Co-authored-by: Davis Vaughan <davis@posit.co>
Co-authored-by: Lionel Henry <lionel@posit.co>

### DIFF
--- a/crates/amalthea/src/kernel.rs
+++ b/crates/amalthea/src/kernel.rs
@@ -381,6 +381,26 @@ fn stdin_thread(
 
 /// Starts the thread that forwards 0MQ messages to Amalthea channels
 /// and vice versa.
+///
+/// This is a solution to the problem of polling/selecting from 0MQ sockets and
+/// crossbeam channels at the same time. Message events on crossbeam channels
+/// are emitted by the notifier thread (see below) on a 0MQ socket. The
+/// forwarding thread is then able to listen on 0MQ sockets (e.g. StdIn replies
+/// and IOPub subscriptions) and the notification socket at the same time.
+///
+/// Part of the problem this setup solves is that 0MQ sockets can only be owned
+/// by one thread at a time. Take IOPUb as an example: we need to listen on that
+/// socket for subscription events. We also need to listen for new IOPub
+/// messages to send to the client, sent via Crossbeam channels. So we need at
+/// least two threads listening for these two different kinds of events. But the
+/// forwarding thread has to fully own the socket to be able to listen to it. So
+/// it's also in charge of sending IOPub messages on that socket. When an IOPub
+/// message comes in, the notifier thread wakes up the forwarding thread which
+/// then pulls messages from the channel and forwards them to the IOPub socket.
+///
+/// Terminology:
+/// - Outbound means that a crossbeam message needs to be forwarded to a 0MQ socket.
+/// - Inbound means that a 0MQ message needs to be forwarded to a crossbeam channel.
 fn zmq_forwarding_thread(
     outbound_notif_socket: Socket,
     stdin_socket: Socket,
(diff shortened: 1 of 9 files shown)


crates/amalthea/src/kernel.rs
```diff
@@ -419,17 +436,28 @@ fn zmq_forwarding_thread(
...
+            Ok(())
+        };
+
+    // Forwards special 0MQ XPUB subscription message from the frontend to the IOPub thread.
+    let forward_inbound_subscription = |socket: &Socket,
+                                        inbound_tx: &Sender<crate::Result<SubscriptionMessage>>|
+     -> anyhow::Result<()> {
+        let msg = SubscriptionMessage::read_from_socket(socket);
+        inbound_tx.send(msg)?;
         Ok(())
```
DavisVaughan, line 452: We need a special inbound helper here, since this is a special `SubscriptionMessage` not a typical `Message`. It has its own format that it reads off the wire

crates/amalthea/src/wire/subscription_message.rs
```diff
@@ -0,0 +1,81 @@
...
+        let bufs = socket.recv_multipart()?;
+        Self::from_buffers(bufs)
+    }
+
+    /// Parse a SubscriptionMessage from an array of buffers (from a ZeroMQ message)
+    ///
+    /// Always a single frame (i.e. `bufs` should be length 1).
+    /// Either `1{subscription}` for subscription.
+    /// Or `0{subscription}` for unsubscription.
+    fn from_buffers(bufs: Vec<Vec<u8>>) -> Result<SubscriptionMessage, Error> {
```
DavisVaughan, line 40: This was fun

crates/amalthea/src/wire/welcome.rs
```diff
@@ -0,0 +1,35 @@
...
+    /// message.
+    pub subscription: String,
+}
+
+// Message type comes from copying what xeus and jupyter_kernel_test use:
+// https://github.com/jupyter-xeus/xeus-zmq/pull/31
+// https://github.com/jupyter/jupyter_kernel_test/blob/5f2c65271b48dc95fc75a9585cb1d6db0bb55557/jupyter_kernel_test/__init__.py#L449-L450
+impl MessageType for Welcome {
+    fn message_type() -> String {
+        String::from("iopub_welcome")
```
DavisVaughan, line 33: `"iopub_welcome"` is the name recommended by the JEP

DavisVaughan (approved):

PR #580, opened 2024-10-11
Set `R_MAIN_THREAD_ID` at the very beginning of setup again

DavisVaughan:
Addresses https://github.com/posit-dev/positron/issues/4973

Also creates a new frontend variant called `DummyArkFrontendRprofile` - a `Console` ark that supports loading `.Rprofile`. This allows us to test that we can load `.Rprofile`s correctly on startup. Notably you can only call `DummyArkFrontendRprofile::lock()` once per process, because you can only load an `.Rprofile` once! This is enforced with a panic if you call `lock()` twice in a single integration test.

### DIFF
--- a/crates/ark/src/fixtures/dummy_frontend.rs
+++ b/crates/ark/src/fixtures/dummy_frontend.rs
@@ -187,10 +187,11 @@ impl DummyArkFrontendRprofile {
     /// Lock a frontend that supports `.Rprofile`s.
     ///
     /// NOTE: This variant can only be called exactly once per process,
-    /// because you can only load an `.Rprofile` one time.
-    ///
-    /// NOTE: Only one `DummyArkFrontend` variant should call `lock()` within
-    /// a given process.
+    /// because you can only load an `.Rprofile` one time. Additionally,
+    /// only one `DummyArkFrontend` variant should call `lock()` within
+    /// a given process. Practically, this ends up meaning you can only
+    /// have 1 test block per integration test that uses a
+    /// `DummyArkFrontendRprofile`.
     pub fn lock() -> Self {
         Self::init();
 
--- a/crates/ark/src/fixtures/dummy_frontend.rs
+++ b/crates/ark/src/fixtures/dummy_frontend.rs
@@ -23,6 +23,14 @@ pub struct DummyArkFrontend {
     guard: MutexGuard<'static, DummyFrontend>,
 }
 
+struct DummyArkFrontendOptions {
+    interactive: bool,
+    site_r_profile: bool,
+    user_r_profile: bool,
+    r_environ: bool,
+    session_mode: SessionMode,
+}
+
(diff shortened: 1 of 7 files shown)


DavisVaughan:
I'm mildly surprised that the test passes on Windows, I thought the Rprofile would actually run twice due to https://github.com/posit-dev/positron/issues/4253, causing multiple IOPub messages

jennybc:
FWIW this gets me back to a happy place in a positron dev build 🎉

crates/ark/src/fixtures/dummy_frontend.rs
```diff
@@ -144,3 +182,54 @@ impl DerefMut for DummyArkFrontendNotebook {
...
 }
+
+impl DummyArkFrontendRprofile {
+    /// Lock a frontend that supports `.Rprofile`s.
+    ///
+    /// NOTE: This variant can only be called exactly once per process,
+    /// because you can only load an `.Rprofile` one time.
+    ///
+    /// NOTE: Only one `DummyArkFrontend` variant should call `lock()` within
+    /// a given process.
```
lionel-, line 193: Do these two notes need to be merged?

crates/ark/tests/r-profile-cat.rs
```diff
@@ -0,0 +1,34 @@
...
+    let path = file.path();
+    let path = path.to_str().unwrap();
+
+    unsafe { std::env::set_var("R_PROFILE_USER", path) };
+
+    // Ok, load up R now. It should `cat()` the `message` over iopub.
+    let frontend = DummyArkFrontendRprofile::lock();
+
+    frontend.recv_iopub_stream_stdout(message)
+}
```
lionel-, line 34: I guess we should also add a test for loading a package to exercise our load hooks.

hmm could we disable `IS_TESTING` in integration tests?

crates/ark/src/fixtures/dummy_frontend.rs
lionel-: IIUC, we could potentially have one global Rprofile per test process, but we still could not have multiple test blocks because we need to exhaust all emitted messages within the first block?

crates/ark/src/fixtures/dummy_frontend.rs
```diff
@@ -24,6 +23,14 @@ pub struct DummyArkFrontend {
     guard: MutexGuard<'static, DummyFrontend>,
 }
 
+struct DummyArkFrontendOptions {
```
lionel-, line 26: It would be helpful to implement `Default` here for self-documentation purpose (and use it in the default paths for clarity / resilience to typos)

lionel- (approved):

crates/ark/tests/r-profile-cat.rs
DavisVaughan: We definitely dont want to disable `IS_TESTING` everywhere in integration tests

PR #583, opened 2024-10-12
Move UI Comm sender ownership to `RMain`

DavisVaughan:


crates/ark/src/interface.rs
```diff
@@ -890,18 +894,25 @@ impl RMain {
...
+            // Check for pending graphics updates
+            // (Important that this occurs while in the "busy" state of this ExecuteRequest
+            // so that the `parent` message is set correctly in any Jupyter messages)
+            unsafe {
+                graphics_device::on_did_execute_request(
+                    self.comm_manager_tx.clone(),
+                    self.iopub_tx.clone(),
+                    self.is_ui_comm_connected() && self.session_mode == SessionMode::Console,
+                )
+            };
```
DavisVaughan, line 915: Previously we did a "prompt" refresh here right before unblocking shell, but over on shell when we got unblocked we then did a working directory refresh and a graphics device refresh.

We now do _all_ 3 of those here, which seems way more appropriate

crates/ark/src/interface.rs
```diff
@@ -1156,6 +1167,77 @@ impl RMain {
...
+    {
+        match self.get_ui_comm_tx() {
+            Some(ui_comm_tx) => f(ui_comm_tx),
+            None => {
+                // Trace level logging, its typically not a bug if the frontend
+                // isn't connected. Happens in all Jupyter use cases.
+                log::trace!("UI comm isn't connected.");
+            },
+        }
+    }
```
DavisVaughan, line 1221: The idea is that in `interface.rs` we use `with_ui_comm_tx()` quite a bit because it doesn't log an _error_ if the UI comm isn't connected - which is correct, because this code runs when in Jupyter mode too.

But in specific UI methods in other files (which only run in Positron mode), we use `get_ui_comm_tx()` which allows us to promote a `None` sender into an error, because that would be very unexpected

crates/ark/src/interface.rs
```diff
@@ -1685,26 +1758,26 @@ impl RMain {
     }
 
     pub fn call_frontend_method(&self, request: UiFrontendRequest) -> anyhow::Result<RObject> {
-        log::trace!("Calling frontend method '{request:?}'");
+        log::trace!("Calling frontend method {request:?}");
+
+        let ui_comm_tx = self.get_ui_comm_tx().ok_or_else(|| {
+            anyhow::anyhow!("UI comm is not connected. Can't execute request {request:?}")
+        })?;
```
DavisVaughan, line 1765: Should only be called when we are in Positron mode, so this feels like an error

crates/ark/src/interface.rs
```diff
@@ -1719,31 +1792,29 @@ impl RMain {
                     if let Err(err) =
                         ui_frontend_reply_from_value(response.result.clone(), &request)
                     {
-                        anyhow::bail!("Can't deserialize RPC response for {request:?}:\n{err:?}");
+                        return Err(anyhow::anyhow!(
+                            "Can't deserialize RPC response for {request:?}:\n{err:?}"
+                        ));
```
DavisVaughan, line 1797: Fixing up a few bail!s

crates/ark/src/shell.rs
```diff
@@ -229,24 +203,6 @@ impl ShellHandler for Shell {
             ExecuteResponse::ReplyException(err) => Err(err),
         };
 
-        let mut kernel = self.kernel.lock().unwrap();
```
DavisVaughan, line 232: I really really like that we now just:
- send request
- get response
- return response

With no extra funny business

crates/ark/src/ui/events.rs
```diff
@@ -25,9 +25,12 @@ pub unsafe extern "C" fn ps_ui_show_message(message: SEXP) -> anyhow::Result<SEX
         message: RObject::view(message).try_into()?,
     };
 
-    let main = RMain::get();
     let event = UiFrontendEvent::ShowMessage(params);
-    main.send_frontend_event(event);
+
+    let main = RMain::get();
+    let ui_comm_tx = main.get_ui_comm_tx().ok_or_else(ui_comm_not_connected)?;
+    ui_comm_tx.send_event(event);
```
DavisVaughan, line 32: Previously would eventually just do

```
log::info!("Discarding message {msg:?}; no frontend UI comm connected");
```

but i think we should actually error here

crates/ark/src/ui/sender.rs
```diff
@@ -0,0 +1,103 @@
...
+        }
+    }
+
+    pub fn send_refresh(&mut self, input_prompt: String, continuation_prompt: String) {
+        self.refresh_prompt_info(input_prompt, continuation_prompt);
+
+        if let Err(err) = self.refresh_working_directory() {
+            log::error!("Can't refresh working directory: {err:?}");
+        }
+    }
```
DavisVaughan, line 67: All UI related refreshes should be here now - like an LSP refresh

crates/ark/src/interface.rs
```diff
@@ -1556,29 +1638,24 @@ impl RMain {
...
+        // Send updated state to the frontend over the UI comm
+        self.with_ui_comm_tx(|ui_comm_tx| {
+            ui_comm_tx.send_event(UiFrontendEvent::Busy(BusyParams { busy }));
+        });
     }
 
     /// Invoked by R to show a message to the user.
     fn show_message(&self, buf: *const c_char) {
         let message = unsafe { CStr::from_ptr(buf) };
+        let message = message.to_str().unwrap().to_string();
```
lionel-, line 1653: Same here, we'd run component-supplied handlers.

crates/ark/src/shell.rs
```diff
@@ -5,14 +5,10 @@
 //
 //
```
lionel-, line 6: Nice to see all these deletions!

lionel- (commented):
Nice progress! With a bit more effort I think we could get in an even better place with all UI-related stuff living in the UI comm module.

lionel- (approved):

crates/ark/src/interface.rs
```diff
@@ -888,18 +892,25 @@ impl RMain {
...
             // FIXME: Race condition between the comm and shell socket threads.
             //
-            // Send info for the next prompt to frontend. This handles
-            // custom prompts set by users, e.g. `options(prompt = ,
-            // continue = )`, as well as debugging prompts, e.g. after a
-            // call to `browser()`.
-            let event = UiFrontendEvent::PromptState(PromptStateParams {
-                input_prompt: info.input_prompt.clone(),
-                continuation_prompt: info.continuation_prompt.clone(),
+            // Perform a refresh of the frontend state
```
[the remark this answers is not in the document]
DavisVaughan: https://github.com/posit-dev/ark/issues/587

crates/ark/src/interface.rs
```diff
@@ -1154,6 +1165,77 @@ impl RMain {
...
+        // Go ahead and do an initial refresh
+        self.with_mut_ui_comm_tx(|ui_comm_tx| {
+            let input_prompt = info.input_prompt.clone();
+            let continuation_prompt = info.continuation_prompt.clone();
+
+            ui_comm_tx.send_refresh(input_prompt, continuation_prompt);
+        });
+    }
+
+    pub fn get_ui_comm_tx(&self) -> Option<&UiCommSender> {
```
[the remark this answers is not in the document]
DavisVaughan: This is a convention we do use elsewhere in this file (`get_comm_manager_tx()`) so im inclined to keep it until we nuke the `ui_comm_tx` entirely

PR #590, opened 2024-10-16
Fix broken output pane caused by lines with hashtags only

kv9898:
Addresses https://github.com/posit-dev/positron/issues/5036


kv9898:
> Thanks! Could you add the relevant unit tests please?

Yes! Thanks to the example tests you created, I added 2 lines to your functions.

kv9898:
@lionel- I made another PR at #593, which supports the nested outline feature and also fixes this very issue, but *much* more changes are involved. On the other hand, this PR is a quick fix and can be merged/reviewed easily.

lionel-:
wow that's great, thanks so much!

PR #597, opened 2024-10-21
Refactor symbol search

lionel-:
Progress towards https://github.com/posit-dev/positron/issues/3822.

- Errors are now consistently propagated as `Err` and only logged at the very top.

- We no longer recurse blindly into all node types except argument lists (this is the current restriction but note that we should recurse in arg lists in the future for e.g. `lapply` and `test_that`). Instead, we only recurse in handlers of specific node types. For instance, comment sections are only expected in expression lists, and thus only handled in the expression list handler.

- We no longer pass around a mut ref of the parent. Instead we pass a store of children by value and return it once done. If a handler inserts a note in the outline, it creates a new store, recurse into its children with this new store, and adds itself to the store of its caller.

  This simpler approach is easier to understand and should solve the lifetime issues encountered in https://github.com/posit-dev/ark/pull/593.

- Bunch of tests.

### DIFF
--- a/crates/ark/src/lsp/symbols.rs
+++ b/crates/ark/src/lsp/symbols.rs
@@ -131,39 +131,45 @@ fn index_node(
     mut store: Vec<DocumentSymbol>,
     contents: &Rope,
 ) -> anyhow::Result<Vec<DocumentSymbol>> {
-    // Check if the node is a comment and matches the markdown-style comment patterns
-    if node.node_type() == NodeType::Comment {
-        let comment_text = contents.node_slice(&node)?.to_string();
-
-        // Check if the comment starts with one or more '#' followed by any text and ends with 4+ punctuations
-        if let Some((_level, title)) = parse_comment_as_section(&comment_text) {
-            // Create a symbol based on the parsed comment
-            let start = convert_point_to_position(contents, node.start_position());
-            let end = convert_point_to_position(contents, node.end_position());
-
-            let symbol = new_symbol(title, SymbolKind::STRING, None, Range { start, end });
-            store.push(symbol);
-
-            // Return early to avoid further processing
-            return Ok(store);
-        }
-    }
-
-    if matches!(
-        node.node_type(),
-        NodeType::BinaryOperator(BinaryOperatorType::LeftAssignment) |
-            NodeType::BinaryOperator(BinaryOperatorType::EqualsAssignment)
-    ) {
-        return index_assignment(node, store, contents);
-    }
-
-    // Recurse into children. We're in the same outline section so use the same store.
     let mut cursor = node.walk();
+
     for child in node.children(&mut cursor) {
-        if is_indexable(&child) {
-            store = index_node(&child, store, contents)?;
-        }
+        store = match child.node_type() {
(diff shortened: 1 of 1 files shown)


kv9898:
Thanks for the work done, @lionel- . One thing I have been struggling with in adding nested outline to this new code structure is this:

let me illustrate with this simple commented code structure:
```r
# first ####
## second ####
a <- 1
```
The tricky problem is that `first` `second` `a` are nodes belonging to the *same parent*  in the `node.walk()`, but we need to add `first` as a child of `first`, and `a` as a child of `second`.

This raises two problems. First, we need pass on additional information to "tell" `a` that its symbol should be added to the children of its node brother, `second`. It seems to me that this involves an additional parameter passing between `index_node`, `index_expression_list`, `index_comment` and `index_assignment`, etc. Otherwise, the `index_` argument does not know which children of children of `store` it should add the symbol to. It seems that my current PR implementation of only passing `section_stack` between `index_node` and `index_comment` is simpler.

Second, adding element to the children of children of the last element in `store` did not seem fun. I couldn't avoid repeatedly borrowing `&mut` references, which is not allowed. I think this really is the lifetime issue.

lionel-:
I'm thinking that only `index_expression_list()` needs to keep track of the section stack and pass the currently active store to its children.

Now that all methods _own_ their store, we no longer need any ref or mut ref. So `index_expression_list()` will be able to push and pop section stores out of its stack and we only work with owned values that we move around.

Does that makes sense?

crates/ark/src/lsp/symbols.rs
```diff
@@ -33,6 +31,30 @@ use crate::treesitter::BinaryOperatorType;
...
+fn new_symbol_node(
+    name: String,
+    kind: SymbolKind,
+    range: Range,
+    children: Vec<DocumentSymbol>,
+) -> DocumentSymbol {
+    let mut symbol = new_symbol(name, kind, range);
+    symbol.children = Some(children);
+    symbol
+}
```
DavisVaughan, line 56: One thing that I really like from the Biome group is the `with_<optional_feature>()` pattern

```rust
fn with_children(mut self, children) -> Self {
  self.children = children;
  self
}
```

So it becomes

```rust
new_symbol(symbol, kind, range).with_children(children)
```

It would require a `DocumentSymbolExt` trait so this is fine if you want to keep it as is, just thought id point it out as a pattern to keep in mind

DavisVaughan (approved):
Looks like a nice simplification!

kv9898:
> I'm thinking that only `index_expression_list()` (added in this PR) needs to keep track of the section stack and pass the currently active store to its children.
> 
> Now that all methods _own_ their store, we no longer need any ref or mut ref. So `index_expression_list()` will be able to push and pop section stores out of its stack and we only work with owned values that we move around.
> 
> Does that makes sense?

Oh yes it does. I was being kinda silly. Now I think I know what to do.

crates/ark/src/lsp/symbols.rs
lionel-: That does seem like a nice pattern!

PR #599, opened 2024-10-21
Add `r_command()` for executing commands over `R` and `R.bat`

DavisVaughan:
Addresses https://github.com/posit-dev/positron/issues/4831

We now iterate over both `Command::new("R")` and `Command::new("R.bat")` and either:
- Exit on the first success
- Or return the error of the last failure if all fail

The main win here is being able to run `cargo test` out of the box on Windows when you have a Rig version of R installed, which puts R on the PATH as `R.bat` rather than `R.exe`

### DIFF
--- a/crates/harp/src/command.rs
+++ b/crates/harp/src/command.rs
@@ -1,0 +1,47 @@
+//
+// command.rs
+//
+// Copyright (C) 2024 Posit Software, PBC. All rights reserved.
+//
+//
+
+use std::io;
+use std::process::Command;
+use std::process::Output;
+
+use crate::sys::command::COMMAND_R_LOCATIONS;
+
+/// Execute a `Command` for R, trying multiple locations where R might exist
+///
+/// - For unix, this look at `R`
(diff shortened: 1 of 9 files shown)


crates/harp/src/command.rs
```diff
@@ -0,0 +1,47 @@
...
+            // Found R, executed command successfully
+            return out;
+        }
+        if i == n - 1 {
+            // On last location, but still couldn't find R, return error
+            return out;
+        }
+    }
+
+    unreachable!("`assert!` ensures at least 1 program location is provided.");
```
lionel-, line 46: It might be cleaner/simpler to set an `out` variable defined outside the loop and return it at the end? This way it's clear what's going to be returned and we get rid of the unreachable.

lionel- (approved):

crates/harp/src/command.rs
DavisVaughan: At some point in the past I convinced myself this was better because it avoided an `unwrap()` but I think you are probably right

PR #600, opened 2024-10-21
Modernize all GHA dependencies

DavisVaughan:
Addresses https://github.com/posit-dev/positron/issues/4821

Here is the current 0.1.145

Screenshot 2024-10-21 at 4 51 31 PM

Here is a manual trigger of 0.1.146 with this PR at https://github.com/posit-dev/ark/actions/runs/11448395058. Note, no warnings at the bottom of this runs page!

Screenshot 2024-10-21 at 5 00 36 PM

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -86,16 +86,13 @@ jobs:
         needs: [do_release, get_version, build_macos, build_windows, build_linux]
         env:
             GITHUB_TOKEN: ${{ github.token }}
-        outputs:
-            upload_url: ${{ steps.create_release.outputs.upload_url }}
         steps:
             - name: Create release
-              uses: actions/create-release@v1
+              uses: softprops/action-gh-release@v2
               id: create_release
               with:
                   draft: false
                   prerelease: true
-                  release_name: ${{ needs.get_version.outputs.ARK_VERSION }}
                   tag_name: ${{ needs.get_version.outputs.ARK_VERSION }}
 
     # Uploads binaries, if we created a release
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -135,44 +132,36 @@ jobs:
...
+              uses: softprops/action-gh-release@v2
               env:
                   GITHUB_TOKEN: ${{ github.token }}
               with:
-                  upload_url: ${{ needs.create_release.outputs.upload_url }}
-                  asset_path: ark-${{ needs.get_version.outputs.ARK_VERSION }}-${{ matrix.flavor }}-windows-x64.zip
-                  asset_name: ark-${{ needs.get_version.outputs.ARK_VERSION }}${{ env.DEBUG_FLAG }}-windows-x64.zip
-                  asset_content_type: application/octet-stream
+                  tag_name: ${{ needs.get_version.outputs.ARK_VERSION }}
+                  files: ark-${{ needs.get_version.outputs.ARK_VERSION }}${{ env.DEBUG_FLAG }}-windows-x64.zip
```
DavisVaughan, line 148: The main difficulty is that this new softprops action does not distinguish between `asset_path` (where it came from) and `asset_name` (what you want to upload as).

Previously for Windows and Linux we _slightly_ tweaked the `asset_path` name to remove `-release` for release builds. That's the `env.DEBUG_FLAG` usage above. Note that on macOS, we were already uploading the asset at `asset_path` with `-release` removed, so no tweaking was being done there.

An easy solution is to simply upload the intermediate zips in `release-windows.yml` and `release-linux.yml` with the finalized names that we want to upload to the actual release - i.e. we follow what `release-macos.yml` was already doing and go ahead and remove the `-release` from the name when it isn't a debug build.

DavisVaughan:
It was deprecated in 2021 without replacement https://github.com/actions/upload-release-asset

That official deprecated repo recommends this one
https://github.com/actions/upload-release-asset?tab=readme-ov-file#github-action---releases-api

It is also what Positron has been using for some time now, so I think it is okay

lionel- (approved):

PR #602, opened 2024-10-22
Add issue template for github

lionel-:


juliasilge (approved):
Looks good!

Do you want to add anything to elicit versions from folks? Or not so useful here in ark?

lionel-:
I was thinking it's currently mostly relevant for the positron side.

DavisVaughan (approved):

PR #603, opened 2024-10-22
Shutdown reticulate finalizer

dfalbel:
Make sure that exiting the R session will first finalize the reticulate Python session.
See https://github.com/posit-dev/positron/pull/5100 for more details.


lionel-:
I would avoid using finalizers for this, and also blocking in the finalizer.

Could we shut it down from the positron side via comm msg?
If not, do it from Ark's shutdown message handler, perhaps with some new handler mechanism that comms could register.

PR #605, opened 2024-10-23
Export `.rs.api.getVersion()` and `.rs.api.getMode()`

DavisVaughan:
Addresses https://github.com/posit-dev/positron/issues/5094

`rstudioapi::getMode()` is new as of rstudioapi 0.17.0

For `rstudioapi::getVersion()`, this function did exist but it used to be this:

```
> rstudioapi::getVersion
function () 
{
    verifyAvailable()
    callFun("versionInfo")$version
}
```

Note the `getVersion()` function calls `callFun("versionInfo")` to indirectly get at `$version`.

We do provide `.rs.api.versionInfo()`, so this was working before.

Now it more directly calls:

```
> rstudioapi::getVersion
function () 
{
    if (hasFun("getVersion")) 
        return(callFun("getVersion"))
    verifyAvailable()
    base <- .BaseNamespaceEnv
    version <- base$.Call("rs_rstudioVersion", PACKAGE = "(embedding)")
    package_version(version)
}
```

So we need to provide the more direct accessor of `callFun("getVersion")` / `.rs.api.getVersion()`, which this PR does.

---

I have tested that `getMode()` and `getVersion()` work with CRAN rstudioapi, and that `getVersion()` continues to work with older versions of rstudioapi too.

### DIFF
--- a/crates/ark/tests/rstudioapi.rs
+++ b/crates/ark/tests/rstudioapi.rs
@@ -1,0 +1,87 @@
+use amalthea::fixtures::dummy_frontend::ExecuteRequestOptions;
+use ark::fixtures::DummyArkFrontend;
+
+#[test]
+fn test_get_version() {
+    let frontend = DummyArkFrontend::lock();
+
+    if !has_rstudioapi(&frontend) {
+        report_skipped("test_get_version");
+        return;
+    }
+
+    let value = "1.0.0";
+    std::env::set_var("POSITRON_VERSION", value);
+
+    let code = "as.character(rstudioapi::getVersion())";
+    frontend.send_execute_request(code, ExecuteRequestOptions::default());
+    frontend.recv_iopub_busy();
+
+    let input = frontend.recv_iopub_execute_input();
+    assert_eq!(input.code, code);
+    assert_eq!(
+        frontend.recv_iopub_execute_result(),
+        format!("[1] \"{value}\"")
+    );
+
+    frontend.recv_iopub_idle();
+
+    assert_eq!(frontend.recv_shell_execute_reply(), input.execution_count)
+}
+
+#[test]
+fn test_get_mode() {
+    let frontend = DummyArkFrontend::lock();
+
+    if !has_rstudioapi(&frontend) {
+        report_skipped("test_get_mode");
+        return;
+    }
+
+    let value = "desktop";
+    std::env::set_var("POSITRON_MODE", value);
+
+    let code = "rstudioapi::getMode()";
+    frontend.send_execute_request(code, ExecuteRequestOptions::default());
+    frontend.recv_iopub_busy();
+
+    let input = frontend.recv_iopub_execute_input();
+    assert_eq!(input.code, code);
+    assert_eq!(
+        frontend.recv_iopub_execute_result(),
+        format!("[1] \"{value}\"")
+    );
+
+    frontend.recv_iopub_idle();
+
+    assert_eq!(frontend.recv_shell_execute_reply(), input.execution_count)
+}
+
+fn has_rstudioapi(frontend: &DummyArkFrontend) -> bool {
+    let code = ".ps.is_installed('rstudioapi')";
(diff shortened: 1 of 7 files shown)


DavisVaughan:
Oooh I wonder if I should try and write an integration test for this! Would probably need to get the rstudioapi package installed on CI but that's easy these days

DavisVaughan:
Includes some nice little stdout when the test is skipped

Screenshot 2024-10-23 at 10 43 43 AM

crates/ark/tests/rstudioapi.rs
```diff
@@ -0,0 +1,87 @@
...
+    };
+
+    frontend.recv_iopub_idle();
+
+    assert_eq!(frontend.recv_shell_execute_reply(), input.execution_count);
+
+    out
+}
+
+fn report_skipped(f: &str) {
```
lionel-, line 85: Simple but effective 😄

crates/harp/src/envvar.rs
```diff
@@ -0,0 +1,109 @@
...
+/// This is a safe alternative to [std::env::set_var()]. On Windows in particular,
+/// using [std::env::set_var()] after R has started up (i.e. after Ark opens the
+/// R DLL and calls `setup_Rmainloop()`) has no effect. We believe this is because
+/// [std::env::set_var()] ends up calling `SetEnvironmentVariableW()` to set the
+/// environment variable, which only affects the Windows API environment space and
+/// not the C environment space that R seems to have access to after it has started
+/// up. This matters because R's `Sys.getenv()` uses the C level `getenv()` to
+/// access environment variables, which only looks at C environment space.
+///
+/// To affect the C environment space, [std::env::set_var()] would have had to call
```
lionel-, line 25: To be clear `Sys.setenv()` uses `putenv` as well and so has the same issue. I think what we are doing is using the R thread to synchronise all writes to environment variables. Probably worth mentioning this here.

crates/harp/src/envvar.rs
```diff
@@ -0,0 +1,109 @@
...
+/// [libc::putenv()], but that has issues with thread safety, and we have seen this
+/// crash R before, so we'd like to stay away from using that ourselves.
+///
+/// The easiest solution to this problem is to just go through R's `Sys.setenv()`
+/// to ensure that R picks up the environment variable update.
+///
+/// If R has not started up yet, you should be safe to call [std::env::set_var()].
+/// For example, we do this for `R_HOME` during the startup process and for
+/// `R_PROFILE_USER` in some tests.
```
lionel-, line 34: This means the C runtime environment variables are synchronised with the Windows API ones during the R startup (though we don't know exactly how).

crates/harp/src/envvar.rs
```diff
@@ -0,0 +1,109 @@
...
+        let key = r_symbol!(key);
+        protect.add(key);
+
+        let value = r_string!(value, &mut protect);
+
+        let call = r_lang!(r_symbol!("Sys.setenv"), !!key = value);
+        protect.add(call);
+
+        libr::Rf_eval(call, libr::R_BaseEnv);
+    }
```
lionel-, line 48: Use the safer `RFunction`?

crates/harp/src/envvar.rs
```diff
@@ -0,0 +1,109 @@
...
+
+        let key = r_string!(key, &mut protect);
+        protect.add(key);
+
+        let call = r_lang!(r_symbol!("Sys.getenv"), key);
+        protect.add(call);
+
+        let out = RObject::new(libr::Rf_eval(call, libr::R_BaseEnv));
+
+        // SAFETY: Input is length 1 string, so output must be a length 1 string.
```
lionel-, line 64: By the way I was thinking about keeping all-caps `SAFETY:` comments for `unsafe {}` blocks, and use `Safety:` for things that panic. This way we don't get desensitised to the former style.

lionel- (approved):
Looks good but I'd prefer a higher-level implementation of the R calls

PR #606, opened 2024-10-23
Add nested outline to refactored code

kv9898:
With code:
```r
a <- 1 # important to make sure initial non-comment symbols are handled properly
b <- 2
#first level####
###########################
#==========================
#=#=#
## second level ####
#### jump to fourth level####
### Back to third level ####
## another second level ----
some_fucntion <- function(x){
  c <- 3
  # inside first level ####
  ## inside second level####
  #### inside fourth level####
  d <- 4
  ### inside third level ####
  e <- 5
}
```
The outline with the refactored code now looks like:
image

2 caveats:
1. We probably need more testing for the nested outline. Maybe we can use some sample code like the one above?
2. `OBJECT` rathern than 'VARIABLE' symbols are still used. I think it may be better left for you to merge the updates from `main`

### DIFF
--- a/crates/ark/src/lsp/symbols.rs
+++ b/crates/ark/src/lsp/symbols.rs
@@ -9,8 +9,6 @@
 
 use std::result::Result::Ok;
 
-use anyhow::*;
-use log::*;
 use ropey::Rope;
 use stdext::unwrap::IntoResult;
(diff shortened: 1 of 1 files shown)


lionel-:
I think since your pull request is from another repository, the base branch was not automatically updated to main when I merged and your PR got closed. I restored things manually.

Issue #609, opened 2024-10-24
`rstudioapi.rs` test is accessing R off the main thread


DavisVaughan:
The problem is the unguarded `harp::envvar::set_var("POSITRON_VERSION", value)` which talks to R.

This is awkward because there _is_ an `R_MAIN` but this is an integration test (unlike `data_explorer.rs` which is an integration test but without an `R_MAIN`).

So in theory we do want to use an `r_task()`, but right now `IS_TESTING` is `true` so it would just say "oh i guess i can run this right here", but really it can't. It needs to send it to the main R thread, which is probably not the thread the test is running on.

Maybe we can also look at `RMain::is_initialized()`, but I feel like this still isn't perfect:
- `R_MAIN` could be setting up, but not fully initialized yet
- We send an r-task right away
- It sees `IS_TESTING && (RMain::is_initialized() = false)` because it hasn't fully set up yet, so in that case it would still run the r-task on the current thread. Frustrating!

From https://github.com/posit-dev/ark/actions/runs/11502602809/job/32017966783:

```
running 2 tests
thread 'test_get_version' panicked at library/core/src/panicking.rs:221:5:
thread 'test_get_version' panicked at crates/ark/src/interface.rs:561:13:
panic in a function that cannot unwind
stack backtrace:
Must access `R_MAIN` on the main R thread, not thread 'test_get_version'.
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
   0:     0x55b32c02aef5 - std::backtrace_rs::backtrace::libunwind::trace::h649ab3318d3445c5
                               at /rustc/eeb90cda1969383f56a2637cbd3037bdf598841c/library/std/src/../../backtrace/src/backtrace/libunwind.rs:116:5
   1:     0x55b32c02aef5 - std::backtrace_rs::backtrace::trace_unsynchronized::hf4bb60c3387150c3
                               at /rustc/eeb90cda1969383f56a2637cbd3037bdf598841c/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:     0x55b32c02aef5 - std::sys::backtrace::_print_fmt::hd9186c800e44bd00
                               at /rustc/eeb90cda1969383f56a2637cbd3037bdf598841c/library/std/src/sys/backtrace.rs:65:5
   3:     0x55b32c02aef5 - <std::sys::backtrace::BacktraceLock::print::DisplayBacktrace as core::fmt::Display>::fmt::h1b9dad2a88e955ff
[...]

PR #611, opened 2024-10-28
Add nested outline to refactored code

kv9898:
Should solve the majority of https://github.com/posit-dev/positron/issues/3822 except for the foldable comment sections.

Also replace `OBJECT`s in tests with `VARIABLE`s.

With code:
```r
a <- 1 # important to make sure initial non-comment symbols are handled properly
b <- 2
#first level####
###########################
#==========================
#=#=#
## second level ####
#### jump to fourth level####
### Back to third level ####
## another second level ----
some_fucntion <- function(x){
  c <- 3
  # inside first level ####
  ## inside second level####
  #### inside fourth level####
  d <- 4
  ### inside third level ####
  e <- 5
}
```

The outline with the refactored code now looks like:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 - The document symbol kind for assigned variables is now `VARIABLE` (@kv9898, posit-dev/positron#5071). This produces a clearer icon in the outline.
 
-- Added partial support for outline headers in comments (@kv9898, posit-dev/positron#3822).
+- Added support for outline headers in comments (@kv9898, posit-dev/positron#3822).
 
 - Sending long inputs of more than 4096 bytes no longer fails (posit-dev/positron#4745).
 
--- a/crates/ark/src/lsp/snapshots/ark__lsp__symbols__tests__symbol_assignment_function_nested_section.snap
+++ b/crates/ark/src/lsp/snapshots/ark__lsp__symbols__tests__symbol_assignment_function_nested_section.snap
@@ -1,0 +1,221 @@
+---
+source: crates/ark/src/lsp/symbols.rs
+expression: "test_symbol(\"\n## title0 ----\nfoo <- function() {\n  # title1 ----\n  ### title2 ----\n  ## title3 ----\n  # title4 ----\n}\n# title5 ----\")"
+---
+[
+    DocumentSymbol {
+        name: "title0",
+        detail: None,
+        kind: String,
+        tags: None,
+        deprecated: None,
+        range: Range {
+            start: Position {
+                line: 1,
+                character: 0,
+            },
+            end: Position {
+                line: 1,
+                character: 14,
+            },
+        },
+        selection_range: Range {
+            start: Position {
+                line: 1,
+                character: 0,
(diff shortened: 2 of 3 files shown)


kv9898:
Moved to here as https://github.com/posit-dev/ark/pull/606 was accidentally closed.

lionel-:
Hi @kv9898. Could you take a look at the changes I've made? I kept your general outline of the algorithm but I've simplified the way we manage the section stack. I've also added a test using `insta` snapshots. It was getting tedious to define the expected outputs 😄

kv9898:
> Hi @kv9898. Could you take a look at the changes I've made? I kept your general outline of the algorithm but I've simplified the way we manage the section stack. I've also added a test using `insta` snapshots. It was getting tedious to define the expected outputs 😄

The changes look great! I was surprised by the snapshot test - must have been a big effort!

I noted that two of the tests failed, but it seemed that they were unrelated to our changes, right? I also noted that when using the nightly Rust (I tried with 1.84 on Windows), the ark kernel keeps crashing after 1 second or so. I (not the programme) panicked about some potential fatal errors I wrote. Things are good with 1.80.

lionel-:
> I was surprised by the snapshot test - must have been a big effort!

Luckily very easy, I switched to using https://docs.rs/insta/latest/insta/ to generate the snapshot from the `Debug` format impls for our data structures. From now on we probably should default to using this sort of tests.

> I noted that two of the tests failed, but it seemed that they were unrelated to our changes, right?

These failures are from https://github.com/posit-dev/ark/issues/609

>  It seems that once we initiate the Folding Range Capability, other folding rules are overridden (i.e. we need to cover all existing folding rules to prevent a regression). I'm thinking of using the same node-walking strategy to create foldable comment sections, but I'm not sure whether I can reuse any symbols infrastructure as folding range is a separate API. Also, is the node-walking strategy good for other foldings (e.g. regions, brackets)?

I think node walking would be a good strategy for folding ranges too. The ranges will be nested and this nesting is consistent with the nesting of the syntax tree.

I guess there are two options:

- Either start from scratch with a new traversal of the tree. That's a bit wasteful in terms of performance but would lead to simpler code, and perhaps a bit of duplication.

- Or collect folding range from the same place we colect symbols.

Note that we'll change the representation of the syntax tree in the coming weeks to https://github.com/rust-analyzer/rowan. Don't worry about implementing new code with Tree-sitter, just be aware that at some point this code will have to be rewritten. The TS and rowan representations will coexist for some time so there won't be any rush to switch.

But I think this tells me that we can keep things simple for the folding range algorithm, i.e. have a separate traversal. Make a good set of tests for it. And then when we rewrite these to use the new representation we can think about merging them. But if you see a simple way to colect folding ranges in the symbol traversal, feel free to go for it!

In either case it's probably best to start small. Since implementing folding ranges disables the existing rules we have, maybe this could be implemented under a feature flag that's disabled by default. This would allow you to implement incrementally with simple PRs.

lionel-:
Thanks for your work on this!

PR #615, opened 2024-10-30
Folding Range LSP support

kv9898:
It seems that the call for folding ranges has been quite a while: https://github.com/posit-dev/positron/issues/18, https://github.com/posit-dev/positron/issues/2924, https://github.com/posit-dev/positron/issues/3822.

I was initially only thinking about adding foldable comments, but it seems that doing this will disable the existing folding support for things like regions and brackets. Therefore, I rewrote these functionalities also.

The PR already supports the folding-range-handling of brackets, regions, code cells and *nested* comment sections:

image

For the handling of regions and code cells, I'm currently implementing a simple lookup of regex, which respects the existing behaviour and disregards any nested structures. I note that this kinds of get messy when you are also using the nested comments. But maybe that's a separate issue?

### DIFF
--- a/crates/ark/src/lsp/folding_range.rs
+++ b/crates/ark/src/lsp/folding_range.rs
@@ -1,12 +1,10 @@
-use std::cell;
-
 use regex::Regex;
 use tower_lsp::lsp_types::FoldingRange;
 use tower_lsp::lsp_types::FoldingRangeKind;
 
 use crate::lsp::documents::Document;
 use crate::lsp::log_error;
-use crate::lsp::log_info;
+// use crate::lsp::log_info; // Uncomment to enable logging
 use crate::lsp::symbols::parse_comment_as_section;
 
 /// Detects and returns folding ranges for comment sections and curly-bracketed blocks
@@ -55,11 +53,11 @@ pub fn folding_range(document: &Document) -> anyhow::Result<Vec<FoldingRange>> {
     }
 
     // Log the final folding ranges and comment stacks
-    log_info!("folding_ranges: {:#?}", folding_ranges);
-    log_info!("comment_stack: {:#?}", comment_stack); // Should be empty
-    log_info!("bracket_stack: {:#?}", bracket_stack); // Should be empty
-    log_info!("region_marker: {:#?}", region_marker); // Should be None
-    log_info!("cell_marker: {:#?}", cell_marker); // Should be None
+    // log_info!("folding_ranges: {:#?}", folding_ranges); // Contains all folding ranges
+    // log_info!("comment_stack: {:#?}", comment_stack); // Should be empty
+    // log_info!("bracket_stack: {:#?}", bracket_stack); // Should be empty
+    // log_info!("region_marker: {:#?}", region_marker); // Should be None
+    // log_info!("cell_marker: {:#?}", cell_marker); // Should be None
 
     Ok(folding_ranges)
 }
--- a/crates/ark/src/lsp/folding_range.rs
+++ b/crates/ark/src/lsp/folding_range.rs
@@ -1,3 +1,5 @@
+use std::cell;
+
 use regex::Regex;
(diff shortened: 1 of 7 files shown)


kv9898:
I also note that these commands are not working except for `insert section`:
image

kv9898:
@lionel- anything else I need to do?

kv9898:
@lionel- is there anything else I can help to do?

lionel-:
@kv9898 We just need to get some time to look at this. Probably next week.

kv9898:
> @kv9898 We just need to get some time to look at this. Probably next week.

Thank you! I was worried that you were working on some breaking changes that would cause a merge conflict. It's a little bit difficult to rebase everything at this stage.😄

crates/ark/src/lsp/folding_range.rs
```diff
@@ -0,0 +1,299 @@
+use regex::Regex;
```
lionel-, line 1: Needs a Posit copyright header as in the other files.

crates/ark/src/lsp/folding_range.rs
```diff
@@ -0,0 +1,299 @@
...
+
+use crate::lsp::documents::Document;
+use crate::lsp::log_error;
+// use crate::lsp::log_info; // Uncomment to enable logging
+use crate::lsp::symbols::parse_comment_as_section;
+
+/// Detects and returns folding ranges for comment sections and curly-bracketed blocks
+pub fn folding_range(document: &Document) -> anyhow::Result<Vec<FoldingRange>> {
+    let mut folding_ranges: Vec<FoldingRange> = Vec::new();
+    let text = &document.contents; // Assuming `contents()` gives the text of the document
```
lionel-, line 13:
```suggestion
    let text = &document.contents;
```

crates/ark/src/lsp/folding_range.rs
```diff
@@ -0,0 +1,299 @@
...
+
+    // Iterate over each character in line_text to find the positions of `{` and `}`
+    for (char_idx, c) in line_text.char_indices() {
+        match c {
+            '{' | '(' | '[' => {
+                bracket_stack.push((line_idx, char_idx));
+                comment_stack.push(Vec::new());
+            },
+            '}' | ')' | ']' => {
+                (folding_ranges, comment_stack) =
```
lionel-, line 88: Matching on characters like this will work incorrectly with delimiter characters in strings for instance.

lionel- (commented):
Hi @kv9898. Thanks a lot for looking into this and I apologise for only taking a look at this now.

I think parsing for folding ranges need to be done by walking down the syntax tree. We need that structure to properly assess whether a character is indeed a delimiter or if it's something else. For instance in `"foo}"` the `}` character is not a delimiter.

As you consider reworking this PR, please know that we plan for LSP functionality to move to https://github.com/posit-dev/air/. In that project, we use Rowan (see https://github.com/rust-analyzer/rowan but we use Biome's fork) to represent the syntax tree instead of Tree-Sitter (we still parse with tree-sitter for now but convert the tree to Rowan). This means that if you do write a folding range algorithm on top of the TS tree in Ark, it will eventually have to be rewritten on top of Rowan (and/or the typed AST we generate, see https://github.com/posit-dev/air/blob/0.1.0/crates/air_r_syntax/src/generated/nodes.rs).

The advantage of implementing this in Air instead of Ark is that the functionality will be available from other editors like VS Code, Vim, Zed. And in Positron it will also be available when the R session is shut down.

If you'd like we could go over Air internals together in a pairing session. I think you're in the UK timezone? I'm in CET. We use the Zed editor to collaborate btw. My email is lionel à posit.co if you'd like to arrange that.

crates/ark/src/lsp/folding_range.rs
kv9898: Ohh I see! In this case it is indeed better to do it with a syntax tree.

PR #616, opened 2024-10-31
Add package information in completions

lionel-:
Addresses https://github.com/posit-dev/positron/issues/5225

### DIFF
--- a/crates/ark/src/lsp/completions/completion_item.rs
+++ b/crates/ark/src/lsp/completions/completion_item.rs
@@ -31,6 +31,7 @@ use stdext::*;
 use tower_lsp::lsp_types::Command;
 use tower_lsp::lsp_types::CompletionItem;
 use tower_lsp::lsp_types::CompletionItemKind;
+use tower_lsp::lsp_types::CompletionItemLabelDetails;
 use tower_lsp::lsp_types::CompletionTextEdit;
 use tower_lsp::lsp_types::Documentation;
 use tower_lsp::lsp_types::InsertTextFormat;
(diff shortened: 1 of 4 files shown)


DavisVaughan:
I think we also need to let the client know we support this
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_completion

Look for `labelDetailsSupport` there.

It's possible we could be working with a client that _doesn't_ support this, but it is >3 years old now so its probably ok to ignore the client supplied option for now

crates/ark/src/lsp/completions/completion_item.rs
```diff
@@ -175,8 +176,9 @@ pub(super) fn completion_item_from_function<T: AsRef<str>>(
 
     item.kind = Some(CompletionItemKind::FUNCTION);
 
-    let detail = format!("{}({})", name, parameters.joined(", "));
-    item.detail = Some(detail);
+    let detail = format!("({})", parameters.joined(", "));
+    let label_details = item_details(Some(detail), package);
+    item.label_details = Some(label_details);
```
DavisVaughan, line 181: I agree that we have many ways to see the function parameters, and by far the most useless one is seeing the function parameters in this completion item menu - the parameters are very scrunched together and you can't see their defaults and they are often cut off

I vote for just showing the package name, and doing a little work to strip out `parameters` here all together, since its probably a non trivial amount of work to compute it and we won't need it anymore.

DavisVaughan (approved):

crates/ark/src/lsp/completions/completion_item.rs
ntluong95: Not sure if it is relevant but does this also apply for Python package?

juliasilge: No, completions for Python are not generated from ark; ark is a Jupyter kernel for R.

lionel-:
Another change I've made is to add `::` as label details for namespace completions.

Before:

Screenshot 2024-11-04 at 13 28 00

After:

Screenshot 2024-11-04 at 13 29 40

The goal is to make it clearer that a namespace is being completed and `::` will be inserted.

DavisVaughan (approved):
I think this feels great, love the `::`

PR #617, opened 2024-11-01
Avoid panics due to LSP connectivity issues

jmcphers:
This change addresses an issue in ark when reading LSP events and sending notifications. If these operations fail, the kernel crashes and all user data is lost. 

The change minimally avoids these crashes as follows:

- If a notification can't be sent, we log instead of panicking.
- If the event loops don't receive events, we end the loops instead of panicking. 

Addresses https://github.com/posit-dev/positron/issues/4959.

### DIFF
--- a/crates/ark/src/interface.rs
+++ b/crates/ark/src/interface.rs
@@ -1735,7 +1735,9 @@ impl RMain {
 
     fn send_lsp_notification(&self, event: KernelNotification) {
         if let Some(ref tx) = self.lsp_events_tx {
-            tx.send(Event::Kernel(event)).unwrap();
+            if let Err(err) = tx.send(Event::Kernel(event)) {
+                log::error!("Failed to send LSP notification: {err:?}");
+            }
         }
     }
 
--- a/crates/ark/src/lsp/main_loop.rs
+++ b/crates/ark/src/lsp/main_loop.rs
@@ -71,6 +71,7 @@ pub(crate) enum AuxiliaryEvent {
     Log(lsp_types::MessageType, String),
     PublishDiagnostics(Url, Vec<Diagnostic>, Option<i32>),
     SpawnedTask(JoinHandle<anyhow::Result<Option<AuxiliaryEvent>>>),
+    Close,
 }
 
 /// Global state for the main loop
@@ -184,15 +185,24 @@ impl GlobalState {
     /// requests, notifications, and other internal events.
     async fn main_loop(mut self) {
         loop {
-            let event = self.next_event().await;
+            let event = match self.next_event().await {
+                Some(event) => event,
+                None => {
+                    lsp::log_info!("Main loop event channel closed; closing auxiliary loop");
+                    if let Err(err) = auxiliary_tx().send(AuxiliaryEvent::Close) {
+                        lsp::log_error!("Failed to close auxiliary loop:\n{err:?}");
+                    }
+                    return;
+                },
+            };
             if let Err(err) = self.handle_event(event).await {
                 lsp::log_error!("Failure while handling event:\n{err:?}")
             }
         }
     }
 
-    async fn next_event(&mut self) -> Event {
-        self.events_rx.recv().await.unwrap()
+    async fn next_event(&mut self) -> Option<Event> {
+        self.events_rx.recv().await
     }
 
     #[rustfmt::skip]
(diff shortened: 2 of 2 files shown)


crates/ark/src/interface.rs
```diff
@@ -1735,7 +1735,9 @@ impl RMain {
 
     fn send_lsp_notification(&self, event: KernelNotification) {
         if let Some(ref tx) = self.lsp_events_tx {
-            tx.send(Event::Kernel(event)).unwrap();
+            if let Err(err) = tx.send(Event::Kernel(event)) {
```
lionel-, line 1738: This is the main change that fixes the crash right?

To avoid multiple logged errors, we should also set `self.lsp_events_tx` to `None` when it's dropped. A bit tricky because this requires access to `RMain`.

Ideally we'd change this to a a broadcast channel so that multiple clients could be notified. And a broadcast would always succeed no matter the number of connected clients (0, 1, or more).

Perhaps we should do that now even if we don't support multiple clients yet?

lionel- (commented):
I think the changes in `main_loop.rs` are not necessary. In principle I'd prefer to document invariants than creating safety handling that never gets called. Or at least document that these handling paths should be unreachable, perhaps with a macro that logs in release builds and panics in dev ones.

jmcphers:
> Unless I'm mistaken the changes in main_loop.rs should not be necessary. In principle I'd prefer to document invariants than creating safety handling that never gets called.

They shouldn't be necessary, but the changes are not speculative -- I observed a couple of runtime crashes on these lines (in the `.unwrap()` that I'm adding error handling too) before I made the change.

crates/ark/src/interface.rs
jmcphers: > Ideally we'd change this to a a broadcast channel so that multiple clients could be notified.

That's an interesting idea! I started down this road, but it began metastasizing into a much bigger change than I wanted. In particular the LSP events use a MPSC model where there are many copies of the `Sender` passed around to the different spots in the code where we need to send notifications. The `Bus` broadcast model is single producer, multi consumer -- you can't `clone()` a `Bus` or generate additional senders. We _could_ pass around a Arc/Mutex containing a Bus instance, or use a MPSC => SPMC design to fan in/fan out the messages, but that's all adding more complexity than I think this change warrants.

crates/ark/src/lsp/main_loop.rs
```diff
@@ -432,26 +442,38 @@ impl AuxiliaryState {
...
+            };
+            match event {
                 AuxiliaryEvent::Log(level, message) => self.log(level, message).await,
                 AuxiliaryEvent::SpawnedTask(handle) => self.tasks.push(Box::pin(handle)),
                 AuxiliaryEvent::PublishDiagnostics(uri, diagnostics, version) => {
                     self.client
                         .publish_diagnostics(uri, diagnostics, version)
                         .await
                 },
+                AuxiliaryEvent::Close => {
```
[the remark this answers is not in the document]
jmcphers: OK, I've removed the unnecessary close.

crates/ark/src/interface.rs
```diff
@@ -1733,9 +1733,14 @@ impl RMain {
...
-    fn send_lsp_notification(&self, event: KernelNotification) {
+    fn send_lsp_notification(&mut self, event: KernelNotification) {
         if let Some(ref tx) = self.lsp_events_tx {
-            tx.send(Event::Kernel(event)).unwrap();
+            if let Err(err) = tx.send(Event::Kernel(event)) {
+                log::error!("Failed to send LSP notification: {err:?}");
+                self.lsp_events_tx = None;
+            }
+        } else {
+            log::warn!("Not sending LSP notification (no connection): {event:?}");
```
lionel-, line 1743: This would warn in Jupyter apps. We should probably keep treating the absence of connected LSPs as a normal state.

crates/ark/src/interface.rs
lionel-: > Ideally we'd change this to a a broadcast channel so that multiple clients could be notified.

Actually it's not possible for multiple clients to be connected at the same time, the LSP protocol does not support it:

> The protocol currently assumes that one server serves one tool. There is currently no support in the protocol to share one server between different tools. Such a sharing would require additional protocol e.g. to lock a document to support concurrent editing.

So ideally we'd shut down the currently running LSP when we get a request to create a new one.

lionel-: Well we can't really shut down a server from the server.

> The current protocol specification defines that the lifecycle of a server is managed by the client (e.g. a tool like VS Code or Emacs). It is up to the client to decide when to start (process-wise) and when to shutdown a server.

So I guess we just let the zombie server run 🤷

lionel-: This `AUXILIARY_EVENT_TX` global is a bit tricky. It's a global so that LSP methods can log messages from free functions/macros. That's purely for ergonomics of development so we don't have to pass around a logger handle everywhere we want to log something.

And we can't make it thread local since the LSP runs on a thread pool.

lionel- (approved):
Let's see if this improves the situation. Because of what I noted in this thread https://github.com/posit-dev/ark/pull/617/files#r1832404648, preventing Positron from creating a zombie extension host and creating a duplicate LSP server running in Ark would be great.

If you keep seeing weirdness, we can move the auxiliary loop handle from global to local state. I hope the current changes will fix the crashes though!

crates/ark/src/interface.rs
jmcphers: that's fair!

jmcphers:
OK, thanks so much for the review and discussion!

PR #625, opened 2024-11-13
Expose package version info

jennybc:
Related to an incoming Positron PR to address https://github.com/posit-dev/positron/issues/1957

There are times when positron-r needs multiple pieces of related info about a package installation. I know it looks funny to get the package version _and_ to pass back info about whether that version satisfies a minimum version requirement, but it's impractical to do that comparison in positron-r (i.e. R package versions don't mesh with semver).

And to message about a package that is installed, but at insufficient version, you need all 3 of these pieces of info: installed version, required version, if that requirement is met.

### DIFF
--- a/crates/ark/src/modules/positron/package.R
+++ b/crates/ark/src/modules/positron/package.R
@@ -21,6 +21,32 @@
 #' @export
 .ps.rpc.is_installed <- .ps.is_installed
 
+# Returns a list containing:
+#   * the version string if the package is installed and NULL otherwise
+#   * a minimum version to check for, with '0.0.0' as a dummy default
+#   * a logical indicating if package is installed at or above the minimum version
+#  This may seem weird, but it's impractical for positron-r to do version
+#  comparisons.
+#' @export
+.ps.rpc.packageVersion <- function(pkg, checkAgainst = "0.0.0") {
+    installed <- system.file(package = pkg) != ""
+
+    if (installed) {
+        version <- utils::packageVersion(pkg)
+        list(
+          version = as.character(version),
+          checkAgainst = as.character(checkAgainst),
+          checkResult = version >= checkAgainst
+        )
+    } else {
+        list(
+          version = NULL,
+          checkAgainst = as.character(checkAgainst),
+          checkResult = FALSE
+        )
+    }
+}
+
 #' @export
 .ps.rpc.install_packages <- function(packages) {
     for (pkg in packages) {


crates/ark/src/modules/positron/package.R
```diff
@@ -21,6 +21,32 @@
...
+#  comparisons.
+#' @export
+.ps.rpc.packageVersion <- function(pkg, checkAgainst = "0.0.0") {
+    installed <- system.file(package = pkg) != ""
+
+    if (installed) {
+        version <- utils::packageVersion(pkg)
+        list(
+          version = as.character(version),
+          checkAgainst = as.character(checkAgainst),
```
lionel-, line 38: why is this argument returned to the caller?

lionel- (approved):
LG!

crates/ark/src/modules/positron/package.R
jennybc: I guess it just felt convenient to package the version checked against and the result together? No more, no less. Maybe you're saying that either the returned info should be completist (include the package name) or be minimal (exclude the version checked against)?

DavisVaughan: I think the idea is that the caller knows `checkAgainst` already, so no need to duplicate the info

crates/ark/src/modules/positron/package.R
```diff
@@ -21,6 +21,32 @@
...
 .ps.rpc.is_installed <- .ps.is_installed
 
+# Returns a list containing:
+#   * the version string if the package is installed and NULL otherwise
+#   * a minimum version to check for, with '0.0.0' as a dummy default
+#   * a logical indicating if package is installed at or above the minimum version
+#  This may seem weird, but it's impractical for positron-r to do version
+#  comparisons.
+#' @export
+.ps.rpc.packageVersion <- function(pkg, checkAgainst = "0.0.0") {
```
DavisVaughan, line 31:
```suggestion
.ps.rpc.packageVersion <- function(pkg, minimumVersion = "0.0.0") {
```

I'd be tempted to call it this to be super clear that its:
- a version
- the minimum one to check against

I'd also be tempted to default it to `NULL`, implying "no minimum version check is done" - see also `rlang::is_installed()`

crates/ark/src/modules/positron/package.R
```diff
@@ -21,6 +21,32 @@
...
+    installed <- system.file(package = pkg) != ""
+
+    if (installed) {
+        version <- utils::packageVersion(pkg)
+        list(
+          version = as.character(version),
+          checkAgainst = as.character(checkAgainst),
+          checkResult = version >= checkAgainst
```
DavisVaughan, line 39: If you move to `minimumVersion` and drop `checkAgainst` from the returned value, it could just be

```r
list(version = version/NULL, compatible = TRUE/FALSE)
```

So you end up with 3 states

```r
# Not installed (caller knows package name, for error reporting purposes)
list(version = NULL, compatible = FALSE)

# Not compatible (caller knows minimum version, for error reporting purposes)
list(version = version, compatible = FALSE)

# Good to go
list(version = version, compatible = TRUE)
```

DavisVaughan (approved):

crates/ark/src/modules/positron/package.R
jennybc: Well, I'm trying to make argument names and shape match what's on the other side in positron-r. But I don't feel strongly and can just code around whatever form y'all want here.

DavisVaughan: On quick glance it looks like `minimumVersion` and `compatible` would also be nice and expressive in https://github.com/posit-dev/positron/pull/5365/files too

In particular I find `minimumVersion` much easier to understand on first glance than `checkAgainst`

jennybc: Yeah I'm changing it, just haven't pushed yet. Working on both ends.

crates/ark/src/modules/positron/package.R
lionel-: yep exactly

PR #626, opened 2024-11-14
Be defensive about global aux channel being dropped

lionel-:
Follow-up to https://github.com/posit-dev/ark/pull/617
Addresses https://github.com/posit-dev/positron/issues/5321

This will still cause weirdness as duplicate LSP sessions will keep using that global channel to communicate log messages and diagnostics, and these will end up being published by the newer LSP session.

To fix this we could take this opportunity to move from tower-lsp to lsp-server. The former has been unmaintained for a while and has multiple issues that require workarounds. The latter would give us much more control over the LSP event loop and allow us to initiate a session shutdown from the server to prevent duplicate LSP sessions.


jmcphers (approved):
LGTM, and fixed the crash in my testing. With my own local changes to fix reconnect on workbench, this crash becomes 100% reproducible, so I'm going to merge this around and bump versions so we can consider it for Workbench. Thanks so much for jumping on this!

PR #629, opened 2024-11-14
Honor `MAX_DISPLAY_ENTRIES` when inspecting variables

dfalbel:
This PR fixes some performance related issues that involve the variables pane.
Addressing: 
- https://github.com/posit-dev/positron/issues/4636 
- https://github.com/posit-dev/positron/issues/4573
- https://github.com/posit-dev/positron/issues/2223

The goal is to have the following scenarios working fine in positron:

```r
# create a single string that's very large
a <- paste0(rep(letters, length.out = 1e7), collapse="-")

# create a very large vector, and expand it in the variables pane
b <- 1:100000

# expand a data frame with many columns
d <- as.data.frame(lapply(1:10000, function(x) 1:100000))

# expand a data.frame with many rows
f <- as.data.frame(lapply(1:20, function(x) 1:5e6))

# visualizing large matrixes (many rows and columns)
g <- matrix(0, nrow = 100000, ncol = 10000)

# matrixes with large strings inside
h <- matrix(paste0(rep(letters, length.out = 1e7), collapse="-"), nrow = 10, ncol = 10)
```

Note: When testing it's important to use a Release version of ark. In my tests, with debug can be 20x slower in such cases - probably for the tail recursion optimizations important for the object size computation.

Not fixed in this PR: 

- https://github.com/posit-dev/positron/issues/2894
- If a package extends the variables pane and it's methods are slow
- R6 classes with many methods are not truncated yet
- S4 objects with many slots are not truncated
- Still expanding the `model` from https://github.com/posit-dev/positron/issues/4636 takes about 1s. I *think* a lot of this time is computing the objects size - because of the deeply nested object.
- https://github.com/posit-dev/positron/issues/3656 (This should be a quick fix as we are here)

### DIFF
--- a/crates/ark/src/variables/variable.rs
+++ b/crates/ark/src/variables/variable.rs
@@ -1195,45 +1195,22 @@ impl PositronVariable {
     }
 
     fn inspect_matrix_column(matrix: SEXP, index: isize) -> harp::error::Result<Vec<Variable>> {
-        unsafe {
-            let matrix = RObject::new(matrix);
-            let dim = IntegerVector::new(Rf_getAttrib(matrix.sexp, R_DimSymbol))?;
-
-            let n_row = dim.get_unchecked(0).unwrap();
-
-            let mut out: Vec<Variable> = vec![];
-            let formatted = FormattedVector::new(matrix.sexp)?;
-            let mut iter = formatted.column_iter(index);
-            let r_type = r_typeof(matrix.sexp);
-            let kind = if r_type == STRSXP {
-                VariableKind::String
-            } else if r_type == RAWSXP {
-                VariableKind::Bytes
-            } else if r_type == LGLSXP {
-                VariableKind::Boolean
-            } else {
-                VariableKind::Number
+        let column =
+            match harp::table::tbl_get_column(matrix, index as i32, harp::TableKind::Matrix) {
+                Ok(column) => column,
+                Err(err) => return Err(Error::Anyhow(err)),
             };
 
-            for i in 0..n_row {
-                out.push(Variable {
-                    access_key: format!("{}", i),
-                    display_name: format!("[{}, {}]", i + 1, index + 1),
-                    display_value: iter.next().unwrap(),
-                    display_type: String::from(""),
-                    type_info: String::from(""),
-                    kind: kind.clone(),
-                    length: 1,
-                    size: 0,
-                    has_children: false,
-                    is_truncated: false,
-                    has_viewer: false,
-                    updated_time: Self::update_timestamp(),
-                });
-            }
+        let variables: Vec<Variable> = Self::inspect_vector(column.sexp)?
+            .into_iter()
+            .enumerate()
+            .map(|(row, mut var)| {
+                var.display_name = format!("[{}, {}]", row + 1, index + 1);
+                var
+            })
+            .collect();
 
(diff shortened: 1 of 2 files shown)


crates/ark/src/variables/variable.rs
```diff
@@ -1537,6 +1565,17 @@ fn parse_index(x: &String) -> harp::Result<isize> {
     })
 }
 
+// We need to be careful when truncating the string, we don't want to return invalid
+// UTF8 sequences. `chars` makes sure we are not splitting a UTF8 character in half.
+// See also https://doc.rust-lang.org/book/ch08-02-strings.html#slicing-strings
+fn truncate_chars(value: String, len: usize) -> (bool, String) {
```
lionel-, line 1571: IIUC `String::truncate()`would be unsafe here because it takes a byte offset, not a number of chars? I see in the docs:

> Panics if new_len does not lie on a char boundary.

Tricky!

crates/ark/src/variables/variable.rs
```diff
@@ -1779,6 +1818,14 @@ mod tests {
         })
     }
 
+    fn inspect_from_expr(code: &str) -> Vec<Variable> {
+        let env = harp::parse_eval_global("new.env()").unwrap();
```
lionel-, line 1822: Probably worth adding an `Environment` method at some point.

crates/ark/src/variables/variable.rs
```diff
@@ -1779,6 +1818,14 @@ mod tests {
...
+        harp::parse_eval0(format!("x <- {code}").as_str(), env.clone()).unwrap();
```
lionel-, line 1823: `Environment::bind()` would be clearer.

crates/ark/src/variables/variable.rs
```diff
@@ -1810,4 +1857,124 @@ mod tests {
...
+            );
+            assert_eq!(vars[0].display_name.len(), MAX_DISPLAY_VALUE_LENGTH);
+        })
+    }
+
+    #[test]
+    fn test_truncation_on_matrices() {
+        r_task(|| {
+            let env = harp::parse_eval_global("new.env()").unwrap();
+            harp::parse_eval0(
```
lionel-, line 1910: Here as well (and below): `Environment::bind()`

lionel- (approved):
LGTM!

crates/ark/src/variables/variable.rs
dfalbel: Done in https://github.com/posit-dev/ark/pull/629/commits/15b03feaafde641ff64efa9b2097d6451882c55f

crates/ark/src/variables/variable.rs
dfalbel: Done in https://github.com/posit-dev/ark/pull/629/commits/ba5e35eedbdbd261f043f167ad9a6560b977180e

crates/ark/src/variables/variable.rs
dfalbel: Added in https://github.com/posit-dev/ark/pull/629/commits/840a877501e581ea0d3b93ae683b3e0513437dd9, let me know what you think?

lionel-: Seems reasonable

PR #630, opened 2024-11-15
Support formulas

dfalbel:
Build on top of #629 
Addresses https://github.com/posit-dev/positron/issues/4119


crates/ark/src/variables/variable.rs
```diff
@@ -90,8 +92,13 @@ impl WorkspaceVariableDisplayValue {
...
             CLOSXP => Self::from_closure(value),
             ENVSXP => Self::from_env(value),
+            LANGSXP => Self::from_language(value),
             _ if r_is_matrix(value) => Self::from_matrix(value),
-            _ => Self::from_default(value),
+            RAWSXP | LGLSXP | INTSXP | REALSXP | STRSXP | CPLXSXP => Self::from_default(value),
+            _ => Self::from_error(Error::Anyhow(anyhow!(
+                "Unexpected type {}",
+                r_type2char(r_typeof(value))
+            ))),
```
lionel-, line 101: why this new restriction? For explicitness?

crates/ark/src/variables/variable.rs
```diff
@@ -106,6 +113,35 @@ impl WorkspaceVariableDisplayValue {
...
+    fn from_language(value: SEXP) -> Self {
+        if r_inherits(value, "formula") {
+            return Self::from_formula(value);
+        }
+
+        return Self::from_error(Error::Anyhow(anyhow!("Unexpected language object type")));
+    }
+
+    fn from_formula(value: SEXP) -> Self {
+        let display_value: Result<RObject, Error> = RFunction::new("base", "format")
```
lionel-, line 125:
```suggestion
        let display_value: harp::Result<RObject> = RFunction::new("base", "format")
```

crates/ark/src/variables/variable.rs
```diff
@@ -106,6 +113,35 @@ impl WorkspaceVariableDisplayValue {
...
+            Ok(display_value) => match display_value.try_into() {
+                Ok(display_value) => display_value,
+                Err(err) => {
+                    return Self::from_error(err);
+                },
+            },
+            Err(err) => {
+                return Self::from_error(err);
+            },
+        };
```
lionel-, line 139: For clarity, consecutive matches instead of nested ones?

crates/ark/src/variables/variable.rs
lionel-, line 125: `~{}` will be multiline and return a character vector of length > 1. I think we could use `as.character(x)` and if not length 1 elide with `...`? In that case we'd mark it truncated?

crates/ark/src/variables/variable.rs
```diff
@@ -1903,6 +1939,14 @@ mod tests {
         })
     }
 
+    #[test]
+    fn test_support_formula() {
+        r_task(|| {
+            let vars = inspect_from_expr("list(x = x ~ y + z + a)");
```
lionel-, line 1945: Needs a test with multiline deparsing, e.g. with `{`.

crates/ark/src/variables/variable.rs
dfalbel: yes, I taught that capturing an unexpected type here would be more explicit.
Before that, for example, the formula support error was being captured at: https://github.com/posit-dev/ark/blob/92ce2f73fdf71c06a34b1041fa5f22cf09a5efdc/crates/harp/src/vector/formatted_vector.rs#L112-L115

Which is fine, but, AFAICT we should only use `FormattedVector` for vector like objects.

crates/ark/src/variables/variable.rs
dfalbel: Fixed and added a test for this!

crates/ark/src/variables/variable.rs
dfalbel: `as.character()` adds a some newline separators which might not be necessary. Still using `format` and then trimming/truncating when necessary.

crates/ark/src/variables/variable.rs
```diff
@@ -106,6 +113,38 @@ impl WorkspaceVariableDisplayValue {
...
+
+    fn from_formula(value: SEXP) -> anyhow::Result<Self> {
+        // `format` for formula will return a character vector, splitting the expressions within
+        // the formula, for instance `~{x + y}` will be split into `~` and `["~{", "x", "}"]`.
+        let formatted: Vec<String> = RFunction::new("base", "format")
+            .add(value)
+            .call_in(ARK_ENVS.positron_ns)?
+            .try_into()?;
+
+        let mut display_value = String::with_capacity(MAX_DISPLAY_VALUE_LENGTH);
```
lionel-, line 135: We might want to join lines with `; ` (with trailing space) and add a space after `{` and before `}`. But the deparser might add lines in other cases too, so I guess we can't easily do that.

I would have just ellipsised after the first line to avoid this issue, but your approach allows seeing more of the contents. I'll leave it up to you to decide what is best.

lionel- (approved):

crates/ark/src/variables/variable.rs
dfalbel: Ohh I missed that. I decided to just add an ellipsis after the first `{`: https://github.com/posit-dev/ark/pull/630/commits/c812b8ff6451a7f8a4f79aadba0869dfcef1cc23

PR #634, opened 2024-11-22
Add support for `haven_labelled` in the data explorer

dfalbel:
Addresses https://github.com/posit-dev/positron/issues/5327 by specializing some data explorer code paths in order to support the haven labelled data type.

At some point we might also want to implement an extension mechanism here too, allowing packages to implement a method that would create a proxy object - that's then used to compute the summary stats, histograms/freq tables, etc.

### DIFF
--- a/crates/ark/src/data_explorer/histogram.rs
+++ b/crates/ark/src/data_explorer/histogram.rs
@@ -161,6 +161,7 @@ mod tests {
     use stdext::assert_match;
 
     use super::*;
+    use crate::fixtures::package_is_installed;
     use crate::r_task;
 
     fn default_options() -> FormatOptions {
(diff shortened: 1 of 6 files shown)


juliasilge:
Does this also address posit-dev/positron#5010?

dfalbel:
No, I think for the variables pane, we'll probably want to implement a custom `ark_display_value` method for haven_labelled.
With the variables pane, there's no way to decide wether we really want the quotes or not.

crates/ark/src/data_explorer/histogram.rs
```diff
@@ -606,4 +607,29 @@ mod tests {
             );
         })
     }
+
+    #[test]
+    fn test_frequency_table_haven_labelled() {
+        r_task(|| {
+            if !package_is_installed("haven") {
```
lionel-, line 614: Also need to add haven to CI: https://github.com/posit-dev/ark/blob/5c3e1888ca1d864b761b8aaba573ebe99c3c411b/.github/workflows/test-linux.yml#L54

lionel- (approved):

PR #636, opened 2024-11-25
Refactor statement range tests

lionel-:
- Nest in top-level test module. Enables IDE gesture for running all tests in module.
- Split tests in multiple functions so they can be run and debugged individually.

Partially done with AI (more struggle than expected).


DavisVaughan (approved):
Didn't look too closely but seems nice. More easily debuggable is definitely a good thing

PR #639, opened 2024-11-25
Detect assignments in calls to avoid false positive lints

lionel-:
Addresses posit-dev/positron#3048.

#### New Features

- N/A

#### Bug Fixes

- Assignments in function calls (e.g. `list(x <- 1)`) are now detected by the missing symbol linter to avoid annoying false positive diagnostics (posit-dev/positron#3048). The downside is that this causes false negatives when the assignment happens in a call with local scope, e.g. in `local()` or `test_that()`. In these cases the nested assignments will incorrectly overlast the end of the call. We prefer to be overly permissive than overly cautious in these matters.

### QA Notes

Assignments in calls, e.g. `list(x <- 1)` should now be treated the same as at top-level. Any further references to `x` at top level should not be linted, e.g. `y` should cause a lint but not `x` in:

```r
list(x <- 1)
x; y
```

Note that calls can be nested `list(x <- 1, list(y, x, y <- 2), 1, y)`.

In that example, the first `y` should in principle be linted as it hasn't been defined yet but we don't support this sort of lints inside arguments yet.

### DIFF
--- a/crates/ark/src/lsp/diagnostics.rs
+++ b/crates/ark/src/lsp/diagnostics.rs
@@ -785,9 +785,24 @@ fn recurse_call_arguments_default(
     // TODO: Can we better handle NSE in things like `quote()` and
     // `dplyr::mutate()` so we don't have to turn off certain diagnostics when
     // we are inside a call's arguments?
-    let mut context = context.clone();
+
+    // We used to clone `context` here to prevent assignments from applying to
+    // the caller. We now purposely preserve the caller's context with its
+    // `document_symbols`. We explicitly want to index symbols assigned in the
+    // call arguments to avoid false positive lints about missing symbols, see
+    // https://github.com/posit-dev/positron/issues/3048.
+    //
+    // Because of the way we traverse the syntax tree, this makes the assumption
+    // that execution order is deterministic from left to right. This is not
+    // necessarily correct since arguments are lazily evaluated, and whether
+    // this is true depends on the function's implementation. For now we assume
+    // every function behaves like `list()`, which is our default model of
+    // strict evaluation.
+
+    // Save `in_call` to restore it on exit. Necessary to handle nested calls
+    // and maintain the state to `true` until we've left the outermost call.
+    let in_call = context.in_call;
     context.in_call = true;
-    let context = &mut context;
 
     // Recurse into arguments.
     if let Some(arguments) = node.child_by_field_name("arguments") {
@@ -804,6 +819,7 @@ fn recurse_call_arguments_default(
         }
     }
 
+    context.in_call = in_call;
     ().ok()
 }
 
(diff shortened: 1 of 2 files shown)


crates/ark/src/lsp/diagnostics.rs
```diff
@@ -1331,31 +1343,70 @@ foo
             let document = Document::new(code, None);
```
DavisVaughan, line 1343: Remove TODO at top of this test

crates/ark/src/lsp/diagnostics.rs
```diff
@@ -1331,31 +1343,70 @@ foo
...
 
-            let diagnostic = diagnostics.get(0).unwrap();
-            assert_eq!(diagnostic.range.start.line, 2);
-            insta::assert_snapshot!(diagnostic.message);
+            // Important to test nested case. We have a dynamic stack of state
+            // variable to keep track of whether we are in a call. The inner
+            // call should restore the outer state on exit.
+            let code = "
+                list(list(), x)
+            ";
```
DavisVaughan, line 1403: Maybe one more test of

```
list()
x
```

to make sure we get a diagnostic on `x`? that ensures we reset `in_call = false` on exit (similar to the existing test here, but a little different)

crates/ark/src/lsp/diagnostics.rs
```diff
@@ -804,6 +819,7 @@ fn recurse_call_arguments_default(
         }
     }
 
+    context.in_call = in_call;
```
DavisVaughan, line 822: The `?` within the `children` loop mean that we might not set `in_call` on the way out.

I recognize that if an error occurs we will basically refuse to do diagnostics all together here, so it doesn't matter _too_ much, but I really had to think about it to figure that out.

Is it possible to move the actual recursion into a sub helper that we call from here? We'd:
- Set `context.in_call` to new value
- Call the sub helper, save value as `out` (a `Result`)
- Set `context.in_call` to old value
- Return `out`

Or create a helper function that does this called `with_in_call(f: FnOnce)`

That way we can't "forget" to reset `in_call`, in case in the future we do some kind of error recovery when we hit an error rather than bailing completely on diagnostics as we do now

crates/ark/src/lsp/diagnostics.rs
```diff
@@ -785,9 +785,24 @@ fn recurse_call_arguments_default(
     // TODO: Can we better handle NSE in things like `quote()` and
     // `dplyr::mutate()` so we don't have to turn off certain diagnostics when
     // we are inside a call's arguments?
-    let mut context = context.clone();
+
+    // We used to clone `context` here to prevent assignments from applying to
+    // the caller. We now purposely preserve the caller's context with its
```
DavisVaughan, line 790: ```    // We used to clone `context` here to prevent assignments from applying to
    // the caller.```

Honestly i dont even think that was the original reason

we cloned just so we could set `in_call` for that scope and its children scopes, but we tried to avoid leaking `in_call = true` outside of this scope
https://github.com/posit-dev/ark/commit/17104159af642cd860fd0614765848e08cf68522

i think i was just copying some existing pattern being used in diagnostics, see other `context.clone()` calls in there

in retrospect your approach of setting it and making sure to reset the old value on the way out makes more sense in general, i think

DavisVaughan (approved):

crates/ark/src/lsp/diagnostics.rs
lionel-: oh good point, I thought explicitly about absence of exceptions, but somehow forgot about early returns -_-

PR #640, opened 2024-11-25
Define include, share, and doc environment variables

lionel-:
Addresses https://github.com/posit-dev/positron/issues/3637.

#### New Features

- N/A

#### Bug Fixes

- The following environment variables are now set in the same way that R does:

  - `R_SHARE_DIR`
  - `R_INCLUDE_DIR`
  - `R_DOC_DIR`

  This solves a number of problems in situations that depend on these variables being defined (https://github.com/posit-dev/positron/issues/3637).

### QA Notes

This should not fail:

```r
stopifnot(
  nzchar(Sys.getenv('R_SHARE_DIR')),
  nzchar(Sys.getenv('R_INCLUDE_DIR')),
  nzchar(Sys.getenv('R_DOC_DIR'))
)
```

We test for these on our side.

### DIFF
--- a/crates/harp/src/command.rs
+++ b/crates/harp/src/command.rs
@@ -9,7 +9,7 @@ use std::io;
 use std::process::Command;
 use std::process::Output;
 
-use crate::sys::command::COMMAND_R_LOCATIONS;
+use crate::sys::command::COMMAND_R_NAMES;
 
 /// Execute a `Command` for R, trying multiple locations where R might exist
 ///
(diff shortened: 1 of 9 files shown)


crates/harp/src/command.rs
```diff
@@ -22,11 +22,20 @@ pub fn r_command<F>(build: F) -> io::Result<Output>
 where
     F: Fn(&mut Command),
 {
-    assert!(COMMAND_R_LOCATIONS.len() > 0);
```
DavisVaughan, line 25: Split into `r_command()` and `r_command_from_path()` or `r_home_from_path()` depending on what makes sense

PR #641, opened 2024-11-25
Fix segfault in variables tests

dfalbel:
With current main, running:  

```
cargo test -p ark --lib variables::variable::tests::test_truncation
```

Will eventually fail with:

```
Error: VECTOR_ELT() can only be applied to a 'list', not a 'symbol'
Fatal error: unable to initialize the JIT
```

While I'm not entirely sure what's happenning, I believe this is a protection issue, and this PR seems to fix the problem.

### DIFF
--- a/crates/ark/src/variables/variable.rs
+++ b/crates/ark/src/variables/variable.rs
@@ -1864,7 +1864,8 @@ mod tests {
 
     fn inspect_from_expr(code: &str) -> Vec<Variable> {
         let env = Environment::new(harp::parse_eval_base("new.env(parent = emptyenv())").unwrap());
-        env.bind("x".into(), harp::parse_eval_base(code).unwrap());
+        let value = harp::parse_eval_base(code).unwrap();
+        env.bind("x".into(), value.sexp);
         // Inspect the S4 object
         let path = vec![String::from("x")];
         PositronVariable::inspect(env.into(), &path).unwrap()
@@ -1971,10 +1972,8 @@ mod tests {
     fn test_truncation_on_matrices() {
         r_task(|| {
             let env = Environment::new_empty().unwrap();
-            env.bind(
-                "x".into(),
-                harp::parse_eval_base("matrix(0, nrow = 10000, ncol = 10000)").unwrap(),
-            );
+            let value = harp::parse_eval_base("matrix(0, nrow = 10000, ncol = 10000)").unwrap();
+            env.bind("x".into(), value.sexp);
 
             // Inspect the matrix, we should see the list of columns truncated
             let path = vec![String::from("x")];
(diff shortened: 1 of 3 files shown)


crates/ark/src/variables/variable.rs
```diff
@@ -1864,7 +1864,8 @@ mod tests {
 
     fn inspect_from_expr(code: &str) -> Vec<Variable> {
         let env = Environment::new(harp::parse_eval_base("new.env(parent = emptyenv())").unwrap());
-        env.bind("x".into(), harp::parse_eval_base(code).unwrap());
+        let value = harp::parse_eval_base(code).unwrap();
+        env.bind("x".into(), value.sexp);
```
DavisVaughan, line 1868: I think what was happening is that with:

```
env.bind("x".into(), harp::parse_eval_base(code).unwrap());
```

- `harp::parse_eval_base(code).unwrap()` creates a `RObject` which does protection
- `env.bind()` consumes `value: impl Into<SEXP>`, and calls `value.into()` on `value`. This consumes the `RObject` (dropping and unprotecting it) and returns the unprotected `SEXP`
- The unprotected `SEXP` is then used in `Rf_defineVar()`

In theory it seems like `Rf_defineVar()` shouldn't do anything to trigger a reallocation, but in practice we see this crash.

---

More broadly I think this suggests a typing issue with `RObject`. I feel like its probably quite bad that we allow the underlying `SEXP` to escape the lifetime of `RObject` so easily. It's like we need to tie the lifetime of `RObject` to the lifetime of underlying `SEXP` unless you call something super explicit like `unsafe into_sexp()` which has a `SAFETY` note that you are taking protection ownership over the `SEXP` from here on out.

lionel- (approved):
Can you also remove the `impl Into<SEXP>` to reduce the footgun factor please?

crates/ark/src/variables/variable.rs
dfalbel: Removing `impl Into<SEXP>` seems to fix the problem as it seems `temporary()` is only deleted after the function execution. At least that what I can tell from something like:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=dca6608de1df2f3b16ba2c2a217aedf3

crates/harp/src/environment.rs
```diff
@@ -91,13 +91,13 @@ impl Environment {
         std::iter::successors(Some(self.clone()), |p| p.parent())
     }
 
-    pub fn bind(&self, name: RSymbol, value: impl Into<SEXP>) {
+    pub fn bind(&self, name: RSymbol, value: SEXP) {
```
lionel-, line 94: hmm or maybe make it take an `&RObject`?

dfalbel: Done in https://github.com/posit-dev/ark/pull/641/commits/94eba522c020eb6c75a91900382af85b9fbafee5

lionel- (approved):

PR #643, opened 2024-11-26
Data Explorer: Synchronize generated comms with upstream changes

wesm:


wesm:
I'll fix these compilation errors

PR #644, opened 2024-11-26
Use tree-sitter approach to expand across semicolons

DavisVaughan:
Alternative approach using tree-sitter rather than the R parser, so it will eventually be fairly easy to move to a standalone LSP

Joint work with @lionel-

### DIFF
--- a/crates/ark/src/lsp/statement_range.rs
+++ b/crates/ark/src/lsp/statement_range.rs
@@ -1,30 +1,27 @@
 //
 // statement_range.rs
 //
-// Copyright (C) 2023 Posit Software, PBC. All rights reserved.
+// Copyright (C) 2023-2024 Posit Software, PBC. All rights reserved.
 //
 //
 
 use anyhow::bail;
 use anyhow::Result;
-use harp::parse_exprs;
 use once_cell::sync::Lazy;
 use regex::Regex;
 use ropey::Rope;
 use serde::Deserialize;
 use serde::Serialize;
 use stdext::unwrap;
+use tower_lsp::lsp_types;
 use tower_lsp::lsp_types::Position;
-use tower_lsp::lsp_types::Range;
 use tower_lsp::lsp_types::VersionedTextDocumentIdentifier;
 use tree_sitter::Node;
 use tree_sitter::Point;
 
-use super::encoding::convert_character_from_utf8_to_utf16;
 use crate::lsp::encoding::convert_point_to_position;
 use crate::lsp::traits::cursor::TreeCursorExt;
 use crate::lsp::traits::rope::RopeExt;
-use crate::r_task;
 use crate::treesitter::NodeType;
 use crate::treesitter::NodeTypeExt;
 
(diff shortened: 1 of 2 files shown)


crates/ark/src/lsp/statement_range.rs
```diff
@@ -240,6 +186,49 @@ fn find_roxygen_comment_at_point<'tree>(
...
+        node = next;
+    }
+
+    tree_sitter::Range {
+        start_byte,
+        end_byte,
+        start_point,
+        end_point,
+    }
+}
```
DavisVaughan, line 230: Here is the magic we use to "expand" the range of the `node` we typically get back to also include support for trailing `;`

It's basically: "if there is a sibling node on the same line as us, expand the range to include that sibling node"

crates/ark/src/lsp/statement_range.rs
```diff
@@ -726,14 +712,15 @@ mod tests {
...
 
         assert_eq!(
-            node.start_position(),
+            range.start_point,
             sel_start.unwrap(),
             "Failed on test {original}"
         );
         assert_eq!(
-            node.end_position(),
+            range.end_point,
```
DavisVaughan, line 723: Tweaked the test function so we can use `statement_range_test()` even for the new tests

crates/ark/src/lsp/statement_range.rs
```diff
@@ -1099,13 +1086,14 @@ if (a > b) {
...
         statement_range_test(
             "
 <<if (@a > b)
     1 + 1>>",
         );
 
+        // TODO: This should run the whole if statement because there are no braces
         statement_range_test(
             "
 if (a > b)
```
DavisVaughan, line 1099: @lionel- I tweaked based on our discussions

crates/ark/src/lsp/statement_range.rs
```diff
@@ -1124,46 +1112,47 @@ if (a > b)
...
 
-        // TODO: I'm not exactly sure what this should run, but this seems strange
+        // TODO: This should run the whole if statement because there are no braces
         statement_range_test(
             "
 if (a > b)
-  <<1 + 1>> else if @(b > c)
-  2 + 2 else 4 + 4
+  <<1 + 1 else if @(b > c)
+  2 + 2 else 4 + 4>>
```
DavisVaughan, line 1129: It does make these _extremely_ weird edge cases a little "worse" because previously we at least selected a complete expression before, but as you can see from the previous TODO we didn't really know the right thing to return here.

We now thing the right thing to return is unambiguously the entire top level `if` statement, so I've updated the TODO and I'm not really caring too much about the current result.

I don't anticipate it affecting anyone in practice really

crates/ark/src/lsp/statement_range.rs
```diff
@@ -1436,119 +1428,96 @@ test_that('stuff', {
     #[test]
     fn test_multiple_expressions_on_one_line() {
         // https://github.com/posit-dev/positron/issues/4317
-
-        // Can't use `statement_range_test()` because it revolves around finding nodes not ranges
```
DavisVaughan, line 1440: Can now use this for all these new tests!

crates/ark/src/lsp/statement_range.rs
```diff
@@ -1436,119 +1428,96 @@ test_that('stuff', {
...
+        statement_range_test(
             "
-{
+@<<{
   1
-}; 2
-",
-            None,
+}; 2>>
+    ",
```
DavisVaughan, line 1483: Fixed!

lionel- (approved):

PR #646, opened 2024-11-27
Refactor `FormattedVector` for faster formatting of S3 objects

dfalbel:
The variables pane uses the `FormattedVector` API to format vector elements to be displayed. One important assumption that we do when using this API is that it's lazy, in the sense that we create an iterator over the elements of a vector and we only pay the formatting price for the values we actually iterated. 

This is true for atomic vectors, for which we iterate using R's C API eg, extracting values with `REAL_ELT` and formattinng them in Rust. However, this is not true for any vector that has a class (except factors), in this case we need to use the `format` function, and we were actually formatting the entire vector before being able to iterate over the formatted values.
This is very costly if we need for format very large vectors, as when we expand the data.frame in https://github.com/posit-dev/positron/issues/3628#issuecomment-2501089076

This PR refactors the `FormattedVector` API, with the main change being:

- We no longer format the entire vector upon construction of the `FormattedVector`
- Instead, we'll only format the vector when creating the `iter()`
- We provide a few different iterators: `iter()`, `iter_n()`, `column_iter()`, `column_iter_n()`. Those suffixed with `n` will only format the requested part of the vector.

As a consequence, now creating the `iter()` returns a `Result`, so we need to adapt in a few places. 
Now, most of the times in the variables pane we'll use `iter_n()` to only format the first `N` elements, that are necessary
to create the display value in the variables pane.


crates/harp/src/vector/formatted_vector.rs
```diff
@@ -4,233 +4,243 @@
...
+    vector: AtomicVector,
+    indices: Box<dyn Iterator<Item = i64>>,
+}
+
+impl FormattedVectorIter {
+    /// Creates a new iterator over the formatted elements of a vector.
+    /// If `indices` is `None`, the iterator will iterate over all elements of the vector.
+    /// If `indices` is `Some`, the iterator will iterate over the elements at the specified indices.
+    /// The indices must be valid and in bounds as we do no bounds checking.
+    fn new_unchecked(
```
lionel-, line 180: This method is missing a Safety comment to explain what is being unchecked exactly and what are the failure modes (panics, R errors?).

lionel- (approved):

crates/harp/src/vector/formatted_vector.rs
```diff
@@ -4,233 +4,243 @@
...
-            },
-            FormattedVector::Complex { vector } => vector.format_elt_unchecked(index, None),
-            FormattedVector::Factor { vector } => vector.format_elt_unchecked(index, None),
-            FormattedVector::FormattedVector { vector, options } => {
-                vector.format_elt_unchecked(index, Some(options))
-            },
+    /// Returns an iterator over the first `n` elements of a vector.
+    /// Should be used when the vector is potentially large and you won't need to
+    /// iterate over the entire vector.
+    pub fn iter_n(&self, n: usize) -> anyhow::Result<FormattedVectorIter> {
```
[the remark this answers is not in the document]
dfalbel: I like that!

crates/harp/src/vector/formatted_vector.rs
dfalbel: Added some notices here: https://github.com/posit-dev/ark/pull/646/commits/e862651607dbc04090722b4f691f8f99fbcf6a08

PR #650, opened 2024-12-03
Use `startup_file` to force a standardized `repos` option during testing

DavisVaughan:
Because https://github.com/r-lib/rig/issues/203 forces a non-`@CRAN@` default for my `CRAN` repo, meaning that the tests don't pass for me locally because `apply_repo_defaults()` sees that something is set and doesn't override it.

### DIFF
--- a/crates/ark/src/repos.rs
+++ b/crates/ark/src/repos.rs
@@ -111,7 +111,7 @@ fn apply_default_repos_auto() -> anyhow::Result<()> {
 }
 
 /// On Windows, we just use the RStudio CRAN mirror as the default.
-#[cfg(windows)]
+#[cfg(not(unix))]
 fn apply_default_repos_auto() -> anyhow::Result<()> {
     apply_default_repos(DefaultRepos::RStudio)
 }
--- a/crates/ark/tests/repos-conf-file.rs
+++ b/crates/ark/tests/repos-conf-file.rs
@@ -9,7 +9,8 @@ use std::io::Write;
 use amalthea::fixtures::dummy_frontend::ExecuteRequestOptions;
 use ark::fixtures::DummyArkFrontendDefaultRepos;
 
-/// Using a configuration file, set the default CRAN repo to a custom value.
+/// Using a configuration file, set the default CRAN repo to a custom value,
+/// and add an extra internal repo.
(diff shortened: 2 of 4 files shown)


crates/ark/src/repos.rs
```diff
@@ -111,7 +111,7 @@ fn apply_default_repos_auto() -> anyhow::Result<()> {
 }
 
 /// On Windows, we just use the RStudio CRAN mirror as the default.
-#[cfg(windows)]
+#[cfg(not(unix))]
```
DavisVaughan, line 114: I also tweaked this because we use `#[cfg(unix)]` above, so it seems cleanest to use `not(unix)` to ensure we dont miss anything

crates/ark/tests/repos-conf-file.rs
```diff
@@ -37,5 +47,20 @@ Internal=https://internal.cran.mirror/
...
+    let code = r#"getOption("repos")[["Internal"]]"#;
+    frontend.send_execute_request(code, ExecuteRequestOptions::default());
+    frontend.recv_iopub_busy();
+
+    let input = frontend.recv_iopub_execute_input();
+    assert_eq!(input.code, code);
+    assert_eq!(
+        frontend.recv_iopub_execute_result(),
+        r#"[1] "https://internal.cran.mirror/""#
+    );
```
DavisVaughan, line 61: Added testing for the `Internal` repo too to make sure it gets set

crates/ark/src/repos.rs
jmcphers: Makes sense!

jmcphers (approved):
LGTM, thanks for this!

Confirmed tests pass locally with this change, too.

```
running 1 test
test test_auto_repos ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.42s

     Running tests/repos-conf-file.rs (target/debug/deps/repos_conf_file-eca1d963ef415763)

running 1 test
test test_conf_file_repos ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.43s
```

PR #654, opened 2024-12-11
Add more logging for Help proxy errors

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/3543 by adding some more logging for us to see more detail in what is going on

This PR moves `return HttpResponse::BadGateway().finish()` after some initial logging we would do anyway, and then adds logging for the error case, including the status code and then just the whole body.

### DIFF
--- a/crates/ark/src/help_proxy.rs
+++ b/crates/ark/src/help_proxy.rs
@@ -144,11 +144,6 @@ async fn proxy_request(req: HttpRequest, app_state: web::Data<AppState>) -> Http
     match client.get(target_url.clone()).send().await {
         // OK.
         Ok(response) => {
-            // We only handle OK. Everything else is unexpected.
-            if response.status() != reqwest::StatusCode::OK {
-                return HttpResponse::BadGateway().finish();
-            }
-
             // Get the headers we need.
             let headers = response.headers().clone();
             let content_type = headers.get("content-type");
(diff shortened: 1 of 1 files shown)


crates/ark/src/help_proxy.rs
```diff
@@ -144,23 +144,29 @@ async fn proxy_request(req: HttpRequest, app_state: web::Data<AppState>) -> Http
...
-            }
-
             // Get the headers we need.
             let headers = response.headers().clone();
             let content_type = headers.get("content-type");
 
             // Log.
             log::info!(
-                "Proxying URL '{:?}' path '{}' content-type is '{:?}'",
+                "Proxying URL {:?} path '{}' content-type is '{:?}'",
```
juliasilge, line 153: See how these URLs look in the logs now (too many quotes):
https://github.com/posit-dev/positron/issues/3543#issuecomment-2524213456

crates/ark/src/help_proxy.rs
```diff
@@ -144,23 +144,29 @@ async fn proxy_request(req: HttpRequest, app_state: web::Data<AppState>) -> Http
...
                 content_type,
             );
 
+            // We only handle OK. Everything else is unexpected.
+            if response.status() != reqwest::StatusCode::OK {
+                log::error!(
+                    "Got status {} proxying {:?}: {:?}",
+                    response.status().clone().to_string(),
+                    target_url.to_string(),
+                    response.text().await.unwrap(),
```
lionel-, line 165: Since something is wrong in this code path, can we confidently say that `response` can be properly decoded? Might be better to condition this on `Ok()`?

lionel- (approved):

crates/ark/src/help_proxy.rs
DavisVaughan: Agreed about the `unwrap()` feeling unsafe here since we are in a `!= reqwest::StatusCode::OK` state here

DavisVaughan (approved):

crates/ark/src/help_proxy.rs
juliasilge: Thanks, both! Improved in a85e59227113d85ea09529c40e294a43c11fe3dc

Issue #661, opened 2025-01-08
Fix compilation warnings for shared reference to mutable static


lionel-:
With recent versions of stable Rust, we now get a bunch of compilation warnings like:

```
warning: creating a shared reference to mutable static is discouraged
    --> crates/ark/src/interface.rs:1953:31
     |
1953 |             unsafe { Rf_error(ERROR_BUF.as_ref().unwrap().as_ptr()) };
     |                               ^^^^^^^^^^^^^^^^^^ shared reference to mutable static
     |
     = note: for more information, see <https://doc.rust-lang.org/nightly/edition-guide/rust-2024/static-mut-references.html>
     = note: shared references to mutable statics are dangerous; it's undefined behavior if the static is mutated or if a mutable reference is created for it while the shared reference lives
```

PR #662, opened 2025-01-08
Make `R_MAIN` a thread-local variable

lionel-:
Progress towards #661.

This is the trickiest of the required changes. We access `R_MAIN` only from the R thread, which allows us to store non-Send/Sync fields in `RMain`. But all global variables must now be Sync. To work around that, we now store `R_MAIN` as thread-local. That's actually nice because this automatically takes care of providing a runtime check for accesses to `R_MAIN` from other threads. These will cause a panic from the lazy initialiser.

Unfortunately, because we now use `RefCell`, there is no longer a way to get a reference with `'static` lifetime, since `RefCell` can be set multiple times. `OnceCell` would likely be a better storage for our purposes but `thread_local!` doesn't have the same level of support for that type (see https://doc.rust-lang.org/stable/std/thread/struct.LocalKey.html). Consequently I had to remove the `get()` and `get_mut()` methods and replace all usages by `with()` and `with_mut()`.


crates/ark/src/interface.rs
```diff
@@ -438,46 +437,50 @@ impl RMain {
...
-            match modules::initialize() {
-                Err(err) => {
-                    log::error!("Can't load R modules: {err:?}");
-                },
-                Ok(namespace) => {
-                    r_main.positron_ns = Some(namespace);
-                },
-            }
+            R_MAIN.with(|cell| {
+                let mut r_main = cell.borrow_mut();
```
DavisVaughan, line 441: You should have access to `R_MAIN.with_borrow_mut()` here

DavisVaughan (approved):

Issue #663, opened 2025-01-10
The `R_MAIN` frontend state conundrum


lionel-:
The stable Rust compiler has started to warn about Ark taking references to static mut variables (see #661). The trickiest instance of this concerns `R_MAIN`, our global singleton for the state used by R's frontend callbacks: https://github.com/posit-dev/ark/blob/1366044e69062bc88a8d1bcc6f474c980f6f8166/crates/ark/src/interface.rs#L158. I attempted to fix it in #662 but drove myself into a dead end. This is a hard problem that we'll have to fix later if we can (and if it's worth the work, given that the current situation seemingly works without any obvious problem).

The current situation is kinda bad in terms of Rust safety. `R_MAIN` is a `static mut`, so a mutable global variable. It is accessed mainly by R's frontend callbacks such as `ReadConsole`, `WriteConsole`, or `PolledEvents`, as well as by some of our R-level functions, e.g. via `.ps.Call("ps_browse_url", url)`. Since these callbacks come from C and go through FFI, we must retrieve our `RMain` singleton from global state. R does not pass any frontend-provided state to these callbacks, and even if it did it would be through a raw pointer, which would be equivalent to what we have now (minus some undefined behaviour which I'll get into below). These callbacks access the global singleton via `RMain::get()` or `RMain::get_mut()`: https://github.com/posit-dev/ark/blob/1366044e69062bc88a8d1bcc6f474c980f6f8166/crates/ark/src/interface.rs#L598.

The problem here is that we're completely bypassing the borrow checker because Rust is not able to check how we take references to global `static mut` variables like `R_MAIN`. This is why this action is being deprecated/discouraged by the compiler. But not only are we on our own in terms of ensuring safety, we're also violating the terms of the language. As mentioned in https://doc.rust-lang.org/nightly/edition-guide/rust-2024/static-mut-references.html, taking a `&mut` in violation of Rust reference rules (there can't be any other reference to an object if there exists a `&mut` to it) is instant undefined behaviour. And that's exactly what we're currently doing without realising it.

[...]

PR #664, opened 2025-01-10
Wrap `R_MAIN` in an `UnsafeCell`

lionel-:
Progress towards #661.
Workaround for #663 (see this issue for context).
Supersedes and closes #662.

- Make `R_MAIN` a thread-local variable as this is a nicer way of controlling accesses from other threads.

- Wrap it in an `UnsafeCell` so we're able to create multiple `&mut` to it without compilation warning. This bypasses the borrow checker and causes undefined behaviour. No changes from the status quo in this regard. Alternatively we could pass raw pointers around but that would require lots of `unsafe` markers everywhere to call methods or set state. So accepting the UB is the easiest way to fix the warning until we can do better.

### DIFF
--- a/crates/ark/src/interface.rs
+++ b/crates/ark/src/interface.rs
@@ -10,6 +10,8 @@
 // The frontend methods called by R are forwarded to the corresponding
 // `RMain` methods via `R_MAIN`.
 
+use std::cell::RefCell;
+use std::cell::UnsafeCell;
 use std::collections::HashMap;
 use std::ffi::*;
 use std::os::raw::c_uchar;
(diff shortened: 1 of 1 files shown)


crates/ark/src/interface.rs
```diff
@@ -142,20 +144,23 @@ pub enum SessionMode {
...
 
-// We use the `once_cell` crate for init synchronisation because the stdlib
-// equivalent `std::sync::Once` does not have a `wait()` method.
-
 /// Used to wait for complete R startup in `RMain::wait_initialized()` or
 /// check for it in `RMain::is_initialized()`.
+///
+/// We use the `once_cell` crate for init synchronisation because the stdlib
+/// equivalent `std::sync::Once` does not have a `wait()` method.
 static R_INIT: once_cell::sync::OnceCell<()> = once_cell::sync::OnceCell::new();
```
DavisVaughan, line 152: Looks like `wait()` is a nightly method that may get stabilized soon ish https://doc.rust-lang.org/std/sync/struct.OnceLock.html#method.wait

crates/ark/src/interface.rs
```diff
@@ -585,33 +589,26 @@ impl RMain {
...
+            let main_ref = cell.get_mut();
 
-        unsafe {
-            R_MAIN
-                .as_mut()
-                .expect("`R_MAIN` can't be used before it is initialized!")
-        }
+            // We extend the lifetime to `'static` as `R_MAIN` is effectively static once initialized.
+            // This allows us to return a `&mut` from the unsafe cell to the caller.
+            unsafe { std::mem::transmute::<&mut RMain, &'static mut RMain>(main_ref) }
```
DavisVaughan, line 610: The alternative is to remove `get_mut()` in favor of the `with_mut()` everywhere right?

Since effectively what we are bypassing here is "the reference may outlive the object" kind of compiler errors?

DavisVaughan (approved):

crates/ark/src/interface.rs
lionel-: Correct. I went for minimal changes here.

lionel-: Also I don't think this transmute is problematic since `R_MAIN` is effectively `'static` after init.

PR #666, opened 2025-01-10
Remove `ARK_VDOCS` by sending vdocs over a Kernel -> LSP event

DavisVaughan:
Part of https://github.com/posit-dev/ark/issues/661

There was a TODO to do this anyways, and it seemed like a pretty reasonable alternative to the global.

Getting this right is a little tricky. It's common for the base R packages (and anything loaded in your `.Rprofile`) to have their namespace vdocs be generated _very quickly_ after startup before the LSP has even started up yet. This means we wouldn't have access to those namespace vdoc files, which means you would not be able to debug those packages.

We _could_ try and fix the timing so that we don't generate any vdocs until after the LSP starts up, but that is pretty brittle.

The alternative here is to maintain our own copy of the virtual documents on the kernel side. As soon as the LSP connects and we get `lsp_events_tx`, we send over any known virtual documents (anything we attempted to send over before this connection would have been dropped). This also has the benefit of being able to refresh the LSP if it restarts out from under us.

---

Side note: In the future I think it might be better if the kernel itself was the one registered as the `TextDocumentContentProvider` for these virtual documents. It's unlikely that the debugger will ever move out of ark, but the LSP probably will. It feels quite weird for Positron to be asking the LSP for the contents of one of these documents. I wonder if instead our `VirtualDocumentProvider` on the positron-r side can perform a Jupyter request to ark for the document contents?

### DIFF
--- a/crates/ark/src/interface.rs
+++ b/crates/ark/src/interface.rs
@@ -98,6 +98,7 @@ use crate::errors;
 use crate::help::message::HelpEvent;
 use crate::help::r_help::RHelp;
 use crate::lsp::events::EVENTS;
+use crate::lsp::main_loop::DidOpenVirtualDocumentParams;
 use crate::lsp::main_loop::Event;
 use crate::lsp::main_loop::KernelNotification;
 use crate::lsp::main_loop::TokioUnboundedSender;
@@ -220,6 +221,10 @@ pub struct RMain {
 
     /// Event channel for notifying the LSP. In principle, could be a Jupyter comm.
     lsp_events_tx: Option<TokioUnboundedSender<Event>>,
+
+    /// The kernel's copy of virtual documents to notify the LSP about when the LSP
+    /// initially connects and after an LSP restart.
+    lsp_virtual_documents: HashMap<String, String>,
 
     dap: RMainDap,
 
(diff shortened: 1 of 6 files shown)


lionel- (approved):
> The alternative here is to maintain our own copy of the virtual documents on the kernel side. As soon as the LSP connects and we get lsp_events_tx, we send over any known virtual documents (anything we attempted to send over before this connection would have been dropped). This also has the benefit of being able to refresh the LSP if it restarts out from under us.

Nice that's how I envisioned this to work.

> Side note: In the future I think it might be better if the kernel itself was the one registered as the TextDocumentContentProvider for these virtual documents. It's unlikely that the debugger will ever move out of ark, but the LSP probably will. It feels quite weird for Positron to be asking the LSP for the contents of one of these documents. I wonder if instead our VirtualDocumentProvider on the positron-r side can perform a Jupyter request to ark for the document contents?

I actually think the vdocs should be provided by the LSP. Ideally it would be based on downloaded sources. This will be useful for things like code navigation.

I'm not sure how this will interact with the srcref feature though. If we can't make it work, maybe we'll need two kinds of vdocs, one for analysis/navigation and one for debugging, but I'm hoping it will not come to that. I imagine that we can deparse the function and carefully match the text (taking into account formatting differences) to a package source and use the latter if there is only one occurrence. Such a procedure can't work reliably 100% of the time though, so we might end up needing two sorts of vdocs or rely on the srcref fallback.

PR #668, opened 2025-01-10
Use an `RwLock` for `AUXILIARY_EVENT_TX`

DavisVaughan:
Part of https://github.com/posit-dev/ark/issues/661

We expect it to be write locked very rarely, so it should still be pretty cheap to get access to this safely

As we saw in tracing-subscriber, this is the same kind of mechanism that the layer "handle" uses internally, so hopefully the overhead is quite low.

### DIFF
--- a/crates/ark/src/lsp/main_loop.rs
+++ b/crates/ark/src/lsp/main_loop.rs
@@ -8,6 +8,7 @@
 use std::collections::HashMap;
 use std::future;
 use std::pin::Pin;
+use std::sync::RwLock;
 
 use anyhow::anyhow;
 use futures::StreamExt;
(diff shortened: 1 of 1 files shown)


crates/ark/src/lsp/main_loop.rs
```diff
@@ -400,17 +404,18 @@ impl AuxiliaryState {
...
+        // auxiliary loop (logging messages or spawning a task) from free functions.
+        // Unfortunately this can theoretically be reset at any time, i.e. on reconnection
+        // after a refresh, which is why we need an RwLock. This is the only place we take
+        // a write lock though.
+        match AUXILIARY_EVENT_TX.write() {
+            Ok(mut tx) => {
+                *tx = Some(auxiliary_event_tx);
+            },
+            Err(error) => {
+                log::error!("Failed to set auxiliary event channel. {error}");
```
lionel-, line 417: I'd unwrap that since it means the auxiliary loop thread panicked, i.e. something very wrong happened?

lionel- (approved):

PR #672, opened 2025-01-23
Increase test timeout from 1s -> 10s

DavisVaughan:
I've been exploring nextest a bit for ark, and I would get some consistent false positive timeouts with a 1s timeout. I think having multiple processes running at once causes some kind of contention, so we need a longer timeout. With this one change (and after #669) I can run `cargo nextest run` on my mac without any issues.

I don't think having a big timeout here is that big a deal, we aren't trying to ensure its _fast_, we just dont want it to get hung.

I would really like to look into nextest as I think it will help us simplify our test infra quite a bit (no management of locks or global state or needing to worry about not cross contaminating the test R session, because each test gets its own R session). It looks like we can just switch to it if we want to with little extra effort, then start simplifying

### DIFF
--- a/crates/ark/tests/data_explorer.rs
+++ b/crates/ark/tests/data_explorer.rs
@@ -90,7 +90,7 @@ fn open_data_explorer(dataset: String) -> socket::comm::CommSocket {
 
     // Wait for the new comm to show up.
     let msg = comm_manager_rx
-        .recv_timeout(std::time::Duration::from_secs(1))
+        .recv_timeout(std::time::Duration::from_secs(10))
         .unwrap();
     match msg {
         CommManagerEvent::Opened(socket, _value) => {
@@ -123,7 +123,7 @@ fn open_data_explorer_from_expression(
 
     // Release the R lock and wait for the new comm to show up.
     let msg = comm_manager_rx
-        .recv_timeout(std::time::Duration::from_secs(1))
+        .recv_timeout(std::time::Duration::from_secs(10))
         .unwrap();
 
     match msg {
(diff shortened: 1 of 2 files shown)


DavisVaughan:
```
thread 'lsp::completions::sources::composite::call::tests::test_completions_after_user_types_part_of_an_argument_name' panicked at core\src\panicking.rs:223:5:
panic in a function that cannot unwind
```

this seems to reproduce on the windows runner (on main too, going back a few commits) but I have no idea why

DavisVaughan:
This one popped up again on the mac runner, but #673 should fix it

```
thread 'test_env_vars' panicked at crates/amalthea/src/fixtures/dummy_frontend.rs:195:9:
assertion failed: `Status(JupyterMessage { zmq_identities: [], header: JupyterHeader { msg_id: "24ea361b-b116-462e-a868-47d503ae01be", session: "5ccdaa85-4179-49b6-b8e2-97b484081e7e", username: "kernel", date: "2025-01-23T21:23:15.387272+00:00", msg_type: "status", version: "5.3" }, parent_header: None, content: KernelStatus { execution_state: Starting } })` does not match `Message::Welcome(data)`
```

crates/ark/tests/data_explorer.rs
```diff
@@ -90,7 +90,7 @@ fn open_data_explorer(dataset: String) -> socket::comm::CommSocket {
 
     // Wait for the new comm to show up.
     let msg = comm_manager_rx
-        .recv_timeout(std::time::Duration::from_secs(1))
+        .recv_timeout(std::time::Duration::from_secs(10))
```
lionel-, line 93: Probably worth using a named constant?

lionel- (approved):

PR #676, opened 2025-01-24
Set `RUST_BACKTRACE` for windows to try and get more info

DavisVaughan:
Fiddling around


DavisVaughan:
Figured it out in #718

PR #677, opened 2025-01-24
Fix `static mut` issues with routines

### LINKED ISSUE #661: Fix compilation warnings for shared reference to mutable static

lionel-:
Closes #661.

Ideally `R_ROUTINES` would be thread-local but that's currently inconvenient in unit tests.

### DIFF
--- a/crates/harp/src/routines.rs
+++ b/crates/harp/src/routines.rs
@@ -48,7 +48,7 @@ pub unsafe fn r_register_routines() {
     });
 
     // Now unwrap the definitions from our thread-safe type
-    let unwrapped: Vec<R_CallMethodDef> = R_ROUTINES
+    let routines: Vec<R_CallMethodDef> = R_ROUTINES
         .lock()
         .unwrap()
         .iter()
@@ -58,7 +58,7 @@ pub unsafe fn r_register_routines() {
     R_registerRoutines(
         info,
         std::ptr::null(),
-        unwrapped.as_ptr(),
+        routines.as_ptr(),
         std::ptr::null(),
         std::ptr::null(),
     );
--- a/crates/harp/src/routines.rs
+++ b/crates/harp/src/routines.rs
@@ -5,18 +5,32 @@
 //
 //
 
+use std::sync::Mutex;
+
 use libr::R_CallMethodDef;
 use libr::R_getEmbeddingDllInfo;
 use libr::R_registerRoutines;
 use log::error;
 
-static mut R_ROUTINES: Vec<R_CallMethodDef> = vec![];
+// Sync and Send wrapper that we can store in a global. Necessary because
+// `R_CallMethodDef` includes raw pointers.
+struct CallMethodDef {
+    inner: R_CallMethodDef,
+}
+
+unsafe impl Send for CallMethodDef {}
+unsafe impl Sync for CallMethodDef {}
+
+static R_ROUTINES: Mutex<Vec<CallMethodDef>> = Mutex::new(vec![]);
 
 // NOTE: This function is used via the #[harp::register] macro,
 // which ensures that routines are initialized and executed on
 // application startup.
 pub unsafe fn add(def: R_CallMethodDef) {
-    R_ROUTINES.push(def);
+    R_ROUTINES
+        .lock()
+        .unwrap()
+        .push(CallMethodDef { inner: def });
 }
 
 pub unsafe fn r_register_routines() {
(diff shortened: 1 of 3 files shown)


crates/harp/src/routines.rs
```diff
@@ -27,16 +41,24 @@ pub unsafe fn r_register_routines() {
...
         numArgs: 0,
     });
 
+    // Now unwrap the definitions from our thread-safe type
+    let unwrapped: Vec<R_CallMethodDef> = R_ROUTINES
+        .lock()
+        .unwrap()
+        .iter()
+        .map(|def| def.inner)
+        .collect();
```
DavisVaughan, line 56: I'm fairly surprised you dont get a warning about "dropping temporary value" since no one is holding the lock in a local variable here. I seem to get that all the time when doing stuff like this.

crates/harp/src/routines.rs
```diff
@@ -27,16 +41,24 @@ pub unsafe fn r_register_routines() {
...
     // Make sure we have an "empty" routine at the end.
-    R_ROUTINES.push(R_CallMethodDef {
+    add(R_CallMethodDef {
         name: std::ptr::null(),
         fun: None,
         numArgs: 0,
     });
 
+    // Now unwrap the definitions from our thread-safe type
+    let unwrapped: Vec<R_CallMethodDef> = R_ROUTINES
```
DavisVaughan, line 51:
```suggestion
    let routines: Vec<R_CallMethodDef> = R_ROUTINES
```

crates/harp/src/routines.rs
```diff
@@ -27,16 +41,24 @@ pub unsafe fn r_register_routines() {
...
+        .unwrap()
+        .iter()
+        .map(|def| def.inner)
+        .collect();
+
     R_registerRoutines(
         info,
         std::ptr::null(),
-        R_ROUTINES.as_ptr(),
+        unwrapped.as_ptr(),
```
DavisVaughan, line 61:
```suggestion
        routines.as_ptr(),
```

crates/harp/src/routines.rs
```diff
@@ -5,18 +5,32 @@
...
+// Sync and Send wrapper that we can store in a global. Necessary because
+// `R_CallMethodDef` includes raw pointers.
+struct CallMethodDef {
+    inner: R_CallMethodDef,
+}
+
+unsafe impl Send for CallMethodDef {}
+unsafe impl Sync for CallMethodDef {}
+
+static R_ROUTINES: Mutex<Vec<CallMethodDef>> = Mutex::new(vec![]);
```
DavisVaughan, line 24: Nice I also tried this when fiddling with this one last week but got stuck on the `R_CallMethodDef` not being send+sync problem

crates/harp/src/routines.rs
```diff
@@ -5,18 +5,32 @@
...
 
-static mut R_ROUTINES: Vec<R_CallMethodDef> = vec![];
+// Sync and Send wrapper that we can store in a global. Necessary because
+// `R_CallMethodDef` includes raw pointers.
+struct CallMethodDef {
+    inner: R_CallMethodDef,
+}
+
+unsafe impl Send for CallMethodDef {}
+unsafe impl Sync for CallMethodDef {}
```
DavisVaughan, line 22: I think my main comment is that `#[harp::register]` is pretty freaking magical. It generates some code that runs _at application startup time_ in an OS specific way. 

I guess we can assume that code is always run on a single thread? I have no idea how startup code gets run on a per OS basis. I just want to make sure this doesn't screw us up.

DavisVaughan (approved):

crates/harp/src/routines.rs
lionel-: We're just saying it is fine to send these raw pointers across threads, which it is as long as you don't do anything with them. We're not introducing any new area of unsafety AFAICT.

PR #679, opened 2025-01-28
Disable the LSP on crash

lionel-:
Part of https://github.com/posit-dev/positron/issues/5507

The goal of this PR is to prevent the LSP from ever crashing the R session and lose the user state. I've moved all LSP request handlers behind a [`catch_unwind()`](https://doc.rust-lang.org/std/panic/fn.catch_unwind.html), basically a `try` for panics. When an LSP handler panics, we detect it, report it, and flip our state to crashed. Once crashed, all LSP handlers respond with an error. This causes some chatter in the LSP logs but I made sure these errors do not carry a backtrace to avoid flooding the logs.

Ideally we'd shut down the LSP entirely and forcefully disconnect from the client. Unfortunately that scenario of a server-initiated shutdown is not supported by the LSP protocol and tower-lsp does not give us the tools to do this.

Alternatively we could send a notification to the client that the LSP has crashed. The client could then initiate a shutdown. I chose not to go that route because to avoid having to deal with synchronisation issues and having to make changes to both the client and the server.

For context, this is a temporary workaround. Once the LSP lives in Air a crash will never be a big deal for the user. Most of the time they will not be aware of it since VS Code / Positron silently restarts crashed servers (unless they crash too many times in a short period, in which case the user is notified and the server is no longer restarted).

Here is a screencast of what happens when the LSP crashes:

https://github.com/user-attachments/assets/61737d86-4018-47f9-8a1d-5ed4ad566291

The user is notified of the crash and requested to send a report with the logs.

Note that the relevant backtrace is sent by our panic hook to the _kernel_ logs rather than the LSP logs. The backtrace in the LSP logs is unlikely to be helpful.

### DIFF
--- a/crates/ark/src/lsp/backend.rs
+++ b/crates/ark/src/lsp/backend.rs
@@ -83,6 +83,7 @@ macro_rules! cast_response {
                     .await;
                 // The backtrace is reported via `err` and eventually shows up
                 // in the LSP logs on the client side
+                let _ = $self.shutdown_tx.send(()).await;
                 Err(new_jsonrpc_error(format!("{err:?}")))
             },
             RequestResponse::Disabled => Err(new_jsonrpc_error(String::from(
@@ -155,6 +156,8 @@ pub(crate) enum LspResponse {
 
 #[derive(Debug)]
 struct Backend {
+    shutdown_tx: tokio::sync::mpsc::Sender<()>,
+
     /// Channel for communication with the main loop.
     events_tx: TokioUnboundedSender<Event>,
 
@@ -436,6 +439,8 @@ pub fn start_lsp(runtime: Arc<Runtime>, address: String, conn_init_tx: Sender<bo
         log::trace!("Connected to LSP at '{}'", address);
         let (read, write) = tokio::io::split(stream);
 
+        let (shutdown_tx, mut shutdown_rx) = tokio::sync::mpsc::channel::<()>(1);
+
         let init = |client: Client| {
             let state = GlobalState::new(client.clone());
             let events_tx = state.events_tx();
@@ -458,6 +463,7 @@ pub fn start_lsp(runtime: Arc<Runtime>, address: String, conn_init_tx: Sender<bo
             });
 
             Backend {
+                shutdown_tx,
                 events_tx,
                 client,
                 _main_loop: main_loop,
@@ -480,12 +486,21 @@ pub fn start_lsp(runtime: Arc<Runtime>, address: String, conn_init_tx: Sender<bo
             .finish();
 
         let server = Server::new(read, write, socket);
-        server.serve(service).await;
 
-        log::trace!(
-            "LSP thread exiting gracefully after connection closed ({:?}).",
-            address
-        );
+        tokio::select! {
+            _ = server.serve(service) => {
+                log::trace!(
+                    "LSP thread exiting gracefully after connection closed ({:?}).",
+                    address
+                );
+            },
+            _ = shutdown_rx.recv() => {
+                log::trace!(
+                    "LSP thread exiting after receiving a shutdown request ({:?}).",
+                    address
+                );
+            }
+        }
     })
 }
 
--- a/crates/ark/src/lsp/backend.rs
+++ b/crates/ark/src/lsp/backend.rs
@@ -46,12 +46,23 @@ use crate::lsp::statement_range::StatementRangeParams;
 use crate::lsp::statement_range::StatementRangeResponse;
 use crate::r_task;
 
+// This enum mainly allows us to create a JSON RPC error when the LSP has
+// previously crashed without going through `anyhow` which always creates a
+// backtrace. Backtraces would flood the logs with irrelevant information.
+pub(crate) enum RequestResponse {
+    Crashed,
+    Result(anyhow::Result<LspResponse>),
+}
+
 // Based on https://stackoverflow.com/a/69324393/1725177
 macro_rules! cast_response {
     ($target:expr, $pat:path) => {{
         match $target {
-            Ok($pat(resp)) => Ok(resp),
-            Err(err) => Err(new_jsonrpc_error(format!("{err:?}"))),
+            RequestResponse::Result(Ok($pat(resp))) => Ok(resp),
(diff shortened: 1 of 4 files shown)


DavisVaughan (commented):
See slack comments, hopefully we can talk about it in the morning before merging

crates/ark/src/lsp/backend.rs
```diff
@@ -125,15 +158,21 @@ struct Backend {
...
     /// and drop all owned state.
     _main_loop: tokio::task::JoinSet<()>,
 }
 
 impl Backend {
-    async fn request(&self, request: LspRequest) -> anyhow::Result<LspResponse> {
-        let (response_tx, mut response_rx) =
-            tokio_unbounded_channel::<anyhow::Result<LspResponse>>();
+    async fn request(&self, request: LspRequest) -> RequestResponse {
+        if LSP_HAS_CRASHED.load(Ordering::Acquire) {
```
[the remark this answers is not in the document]
lionel-: I chose to touch as few things as possible and it seemed fine not to change notify

lionel-:
Davis pointed out that the `serve().await` is not really blocking (not sure how I missed that 😬) so we are able to fully shut down the LSP by waking up a `select`. That's nice because that removes a source of potential problems and that prevents any further log messages in the LSP output channel.

To be on the safe side I decided to keep the crash flag that disables request handlers because our internal notification races with incoming messages from the client so it's possible the main loop will tick again after we detect a crash.

Also I realised we already have the infrastructure to show notifications via Jupyter so I now do that. The downside is that this requires us to go through an `r_task()` to send the notification (it would be possible to avoid that but would require a non trivial amount of plumbing). The upside is that we leave the LSP out of this which seems safer since we are shutting down.

crates/ark/src/lsp/handlers.rs
```diff
@@ -334,6 +334,7 @@ pub(crate) fn handle_statement_range(
     params: StatementRangeParams,
     state: &WorldState,
 ) -> anyhow::Result<Option<StatementRangeResponse>> {
+    panic!("oh no");
```
DavisVaughan, line 337: Please remove this 😬

crates/ark/src/lsp/backend.rs
```diff
@@ -122,18 +168,26 @@ pub(crate) enum LspResponse {
 
 #[derive(Debug)]
 struct Backend {
+    shutdown_tx: tokio::sync::mpsc::Sender<()>,
+
     /// Channel for communication with the main loop.
     events_tx: TokioUnboundedSender<Event>,
 
+    /// Tower-LSP's client
+    client: Client,
```
DavisVaughan, line 177: Remove `client` from here again, no longer needed

crates/ark/src/lsp/backend.rs
```diff
@@ -44,24 +48,68 @@ use crate::lsp::statement_range::StatementRangeParams;
...
 
+// This enum is useful for two things. First it allows us to distinguish a
+// normal request failure from a crash. In the latter case we send a
+// notification to the client so the user knows the LSP has crashed.
+//
+// Once the LSP has crashed all requests respond with an error. This is a
+// workaround, ideally we'd properly shut down the LSP from the server. The
+// `Disabled` enum variant is an indicator of this state. We could have just
+// created an anyhow error passed through the `Result` variant but that would
+// flood the LSP logs with irrelevant backtraces.
```
DavisVaughan, line 59: Maybe update this message to mention that we do force a shutdown? AFAICT this is us being defensive in case there are any other things happening after we've shutdown that will no longer be able to communicate with the client

crates/ark/src/lsp/backend.rs
```diff
@@ -122,18 +168,26 @@ pub(crate) enum LspResponse {
 
 #[derive(Debug)]
 struct Backend {
+    shutdown_tx: tokio::sync::mpsc::Sender<()>,
```
DavisVaughan, line 171: Comment on what this is for would be nice

crates/ark/src/lsp/backend.rs
```diff
@@ -381,8 +451,10 @@ pub fn start_lsp(runtime: Arc<Runtime>, address: String, conn_init_tx: Sender<bo
         log::trace!("Connected to LSP at '{}'", address);
         let (read, write) = tokio::io::split(stream);
 
+        let (shutdown_tx, mut shutdown_rx) = tokio::sync::mpsc::channel::<()>(1);
+
         let init = |client: Client| {
-            let state = GlobalState::new(client);
+            let state = GlobalState::new(client.clone());
```
DavisVaughan, line 457:
```suggestion
            let state = GlobalState::new(client);
```

When you remove `client` from `Backend` can also go back to this

crates/ark/src/lsp/main_loop.rs
```diff
@@ -397,17 +399,47 @@ impl GlobalState {
 /// * - `response`: The response wrapped in a `anyhow::Result`. Errors are logged.
```
DavisVaughan, line 399: Update comment here

crates/ark/src/lsp/main_loop.rs
```diff
@@ -397,17 +399,47 @@ impl GlobalState {
...
 fn respond<T>(
-    response_tx: TokioUnboundedSender<anyhow::Result<LspResponse>>,
-    response: anyhow::Result<T>,
+    response_tx: TokioUnboundedSender<RequestResponse>,
+    response: impl FnOnce() -> anyhow::Result<T>,
     into_lsp_response: impl FnOnce(T) -> LspResponse,
 ) -> anyhow::Result<()> {
+    let mut crashed = false;
+
+    let response = std::panic::catch_unwind(std::panic::AssertUnwindSafe(response))
```
DavisVaughan, line 408: I'm not entirely certain why we can `AssertUnwindSafe` here, maybe a comment?

DavisVaughan (approved):

crates/ark/src/lsp/main_loop.rs
lionel-: I've not done a thorough analysis but the main source of unsafety is interior mutability which we a priori don't use in request handlers. We're not in a position to guarantee safety here though.

PR #680, opened 2025-01-29
Analyze completion site re: trailing parens

jennybc:
Addresses https://github.com/posit-dev/positron/issues/1818
Addresses https://github.com/posit-dev/positron/issues/2338
Closes #369 (I just came across this earlier effort that stalled out in draft from)

I think of this as a "big and dumb" solution to this problem. But it works! The discussion I want to have is basically hinted at in #1818. I suspect it's time to create some formal `CompletionOption` struct that gets formed (probably) in `provide_completions()` that contains some analysis of the completion site. This will then be passed down to all the lower level functions that marshal completions.

Summary on what we decided together:

* We will proceed with a ~big~ small and dumb solution along these lines, to enjoy immediate improvements in our completions for `?` and inside `debug()` and friends.
- [x] Open new issue for the more profound changes that are needed in the long-term to enable completions that are responsive to the completion site and how it's situated in the AST. The "don't add parentheses and parameter hints" feature from this PR is one (but not the only) capability that the new approach will facilitate. Done in #681.
- [x] Add tests to this PR.
- [x] Remove all/most of logging before merging. (The refactoring alluded to above will allow us to build in logging in a more streamlined way. It will be nice to make it possible to dump all completions *with their provenance*.)


crates/ark/src/lsp/completions/sources/composite.rs
```diff
@@ -31,33 +31,41 @@ use tower_lsp::lsp_types::CompletionItemKind;
...
     context: &DocumentContext,
     state: &WorldState,
+    no_trailing_parens: bool,
 ) -> Result<Vec<CompletionItem>> {
     log::info!("completions_from_composite_sources()");
 
     let mut completions: Vec<CompletionItem> = vec![];
 
+    // TO THINK: also related to doing analysis of the completion site
+    // would this also belong in a hypothetical CompletionOption?
```
jennybc, line 51: This analysis of the pipe situation feels related to the sort of activity and result that would belong in a `CompletionOption` struct.

crates/ark/src/lsp/completions/sources/composite/call.rs
```diff
@@ -67,6 +67,10 @@ pub(super) fn completions_from_call(
     // fn(name = value)
     //    ~~~~
     //
+
+    // JB: this caught my eye when doing the "no trailing parens" work
+    // it's a proper analysis of the completion site, e.g. investigating whether
+    // we know the site is CallNodePositionType::Value
```
jennybc, line 73: I found this other example of doing some formal analysis of the completion site to be interesting.

.vscode/launch.json
```diff
@@ -14,7 +14,8 @@
             },
             "windows": {
                 "program": "ark.exe"
-            }
+            },
+            "sourceLanguages": ["rust"]
```
jennybc, line 18: I think this is a neutral-to-positive move for us.

crates/ark/src/lsp/completions/completion_utils.rs
```diff
@@ -0,0 +1,123 @@
+use tree_sitter::Node;
+
+use crate::lsp::document_context::DocumentContext;
+use crate::lsp::traits::rope::RopeExt;
+use crate::treesitter::node_text;
+use crate::treesitter::NodeType;
+use crate::treesitter::NodeTypeExt;
+use crate::treesitter::UnaryOperatorType;
+
+#[allow(dead_code)]
```
jennybc, line 10: Several of these fields are actually used, but they were used along the way, i.e. when I had more logging in. I imagine this entire struct will go away when we do #681, so seems ok to have around for now.

crates/ark/src/lsp/completions/completion_utils.rs
```diff
@@ -0,0 +1,123 @@
...
+        parent_node,
+        parent_node_text,
+        grandparent_node,
+        grandparent_node_text,
+        greatgrandparent_node,
+        greatgrandparent_node_text,
+    }
+}
+
+pub fn check_for_function_value(context: &DocumentContext, node_context: &NodeContext) -> bool {
```
jennybc, line 71: I revised these to be less nested -- I liked that aspect of #369 once I discovered it.

crates/ark/src/lsp/completions/provide.rs
```diff
@@ -21,12 +24,138 @@ pub(crate) fn provide_completions(
...
+
+    fn find_completion(completions: &[CompletionItem], label: &str) -> CompletionItem {
+        completions
+            .iter()
+            .find(|item| item.label == label)
+            .unwrap()
+            .clone()
+    }
+
+    #[test]
```
jennybc, line 66: I decided to test this here after all, as opposed to testing individual completion sources. It feels like we just want to know that the completion is correct, regardless of which helper ultimately provided it. It didn't seem to make sense to test the helpers themselves, since there's nothing particularly tricky about them correctly using the `no_trailing_parens` boolean. Again #681 is going to present a big opportunity to revisit testing that relates to "is the `CompletionItem` correct and is it provided by the expected source(s)?"

jennybc: Used the basic structure of a test I found in #369.

jennybc:
@DavisVaughan I merged your proposals in here, tweaked tests a bit, and added a note re: binary help.

I remain intrigued by the connection to the `CallNodePositionType` enum, since it feels like `ParameterHints::Disabled` is morally equivalent to asserting that the completion site should be treated as `CallNodePositionType::Value`. So I end up wondering if we really need a completely new enum or could work with `CallNodePositionType`.

https://github.com/posit-dev/ark/blob/ca19105abad919d3d1392653100ff3fbe444752b/crates/ark/src/lsp/completions/sources/utils.rs#L94-L168

That being said, I'm happy to get this PR merged in and defer any further thought or refactoring on this matter to the work on #681.

.vscode/launch.json
lionel-: It looks like this enables these extra features, including formatters: https://github.com/vadimcn/codelldb/blob/b412cef3db6102591579d11c866ad60cae3fd65a/lang_support/rust.py#L10

crates/ark/src/treesitter.rs
```diff
@@ -460,6 +460,40 @@ pub(crate) fn node_is_call(node: &Node, name: &str, contents: &ropey::Rope) -> b
...
 }
 
+/// This function takes a [Node] that you suspect might be in a call argument position
+/// and walks up the tree, looking for the containing call node
+///
+/// If the [Node] is not a call argument, this returns `None`.
+///
+/// This works for `Call`, `Subset`, and `Subset2`, which all share the same tree-sitter
+/// structure.
+pub(crate) fn node_find_containing_call<'tree>(node: &Node<'tree>) -> Option<Node<'tree>> {
```
lionel-, line 470: I think the original name `node_find_parent_call()` was clearer?
Or were you worried "parent" implies immediate parent which is not technically the case? In that case, "ancestor" seems more conventional than "containing"?

lionel- (approved):
Looks good!

crates/ark/src/treesitter.rs
DavisVaughan: `node_find_parent_call()` is fine with me

DavisVaughan (approved):

crates/ark/src/treesitter.rs
jennybc: ok, done

Issue #692, opened 2024-03-28
R: Crash when saving CHARSXP to a binding


DavisVaughan:
Likely in the Variables pane

Just run:

```r
x <- rlang:::chr_get("foo", 0L)
```

Without the `x <-` it works fine, but saving to `x` causes a crash

Not likely for many users to do this, definitely expert only to even make one of these, so not super high priority

PR #701, opened 2025-02-06
Fix crash when inspecting a `CHRSXP`

dfalbel:
Adresses https://github.com/posit-dev/ark/issues/692

You should be able to create a binding containing a `CHARSXP` using eg:

```
x <- rlang:::chr_get("foo", 0L)
```

We could access the value of the CHARSXP, but it would require using unsafe `libr` in the variables pane, which we try to avoid, or export a harp function for this. Seems like it's not worth given how rare this happens, but I could implement this further if you think that's important.

### DIFF
--- a/crates/ark/src/variables/variable.rs
+++ b/crates/ark/src/variables/variable.rs
@@ -99,6 +99,7 @@ impl WorkspaceVariableDisplayValue {
             CLOSXP => Self::from_closure(value),
             ENVSXP => Self::from_env(value),
             LANGSXP => Self::from_language(value),
+            CHARSXP => Self::from_charsxp(value),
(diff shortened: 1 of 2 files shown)


crates/ark/src/variables/variable.rs
```diff
@@ -2074,4 +2084,18 @@ mod tests {
             assert!(vars[0].display_value.starts_with("<S4 class"),);
         })
     }
+
+    #[test]
+    fn test_charsxp() {
+        r_task(|| {
+            let env = Environment::new_empty().unwrap();
+            let value = harp::parse_eval_base(r#"rlang:::chr_get("foo", 0L)"#).unwrap();
```
lionel-, line 2092: skip if rlang is not installed?

PR #710, opened 2025-02-13
`row_labels.is_null()` actually refers to a method checking for a null pointer

dfalbel:
This fixes a regression caused by https://github.com/posit-dev/ark/pull/706. `is_null()` is actually checking for a NULL opinter. Now properly check using `r_is_null()`.


lionel- (approved):
oh wow `is_null()` is a method on `*T` and this is an `RObject`, how do we even dispatch there?

Should we add `RObject::is_null()` instead? Ideally the `r_` API would be low level.

dfalbel:
I have no idea how we dispatch to `*T`, quite unexpected.
I have implemented `is_null` for `RObject` in https://github.com/posit-dev/ark/pull/710/commits/e2f534dab77670668415d1fa6920c5c76ded3766

PR #711, opened 2025-02-13
Improved matrix representation in variables pane

dfalbel:
I agree with @EmilHvitfeldt  in https://github.com/posit-dev/positron/issues/3656 that diisplaying the number of rows and columns is more useful as displaying the actual values for matrices in the variables pane. One can still see the values by inspecting/ or opening in the data explorer.

Adresses https://github.com/posit-dev/positron/issues/3656

### DIFF
--- a/crates/ark/src/variables/variable.rs
+++ b/crates/ark/src/variables/variable.rs
@@ -304,43 +304,33 @@ impl WorkspaceVariableDisplayValue {
         Self::new(display_value, is_truncated)
     }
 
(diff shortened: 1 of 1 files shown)


lionel- (approved):
Maybe add a test to check for consistency of matrix and df methods so that if we change the display in one place we don't forget to update the other?

dfalbel:
Good idea, added it in https://github.com/posit-dev/ark/pull/711/commits/54ef6dbbc5a4e2e0481ff2104a6a137d12128630

PR #712, opened 2025-02-14
Fix `ark --install` on Linux

### LINKED ISSUE #648: Ark `--install` command panics
To repro, just run `ark --install`. This appears:

```
thread 'main' panicked at /home/jmcphers/git/ark/crates/harp/src/command.rs:33:42:
called `Result::unwrap()` on an `Err` value: NotPresent
```

Some debugging suggests the problem is that this command requires `R_HOME` to be set. So you can do this:

```bash
export R_HOME=$(R RHOME)
```

After this you can install, but you get a spurious error about needing a connection file.

```
./ark --install
Successfully installed Ark Jupyter kernelspec.

    Kernel: /home/jmcphers/.local/share/jupyter/kernels/ark/kernel.json
    
Error: A connection file must be specified. Use the `--connection_file` argument.
```

It's probably fine to require R_HOME to be set so that Ark is registered against the right version of R, but we should check and see if it's set and show something friendly if it isn't.

lionel-:
And fix `--connection_file` error.

Closes #648.

### DIFF
--- a/crates/ark/src/main.rs
+++ b/crates/ark/src/main.rs
@@ -128,7 +128,7 @@ fn main() -> anyhow::Result<()> {
             },
             "--install" => {
                 install_kernel_spec()?;
-                has_action = true;
+                return Ok(());
             },
             "--help" => {
                 print_usage();
--- a/crates/ark/src/version.rs
+++ b/crates/ark/src/version.rs
@@ -10,7 +10,7 @@ use std::env;
 use std::path::PathBuf;
 
 use anyhow::Context;
-use harp::command::r_command;
+use harp::command::r_command_from_path;
 use harp::object::RObject;
 use itertools::Itertools;
 use libr::SEXP;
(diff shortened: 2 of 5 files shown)


crates/ark/src/main.rs
```diff
@@ -128,7 +128,7 @@ fn main() -> anyhow::Result<()> {
             },
             "--install" => {
                 install_kernel_spec()?;
-                has_action = true;
+                return Ok(());
             },
```
DavisVaughan, line 132: Oh wow it was trying to start ark when you `--install`?

crates/ark/src/version.rs
```diff
@@ -30,19 +30,35 @@ pub struct RVersion {
...
-    // Convert the output to a string
-    let r_home = String::from_utf8(output.stdout)
-        .context("Failed to convert R_HOME output to string")?
-        .trim()
-        .to_string();
+pub fn detect_r() -> anyhow::Result<RVersion> {
+    let r_home: String = r_home_setup().to_string_lossy().to_string();
 
-    let output = r_command(|command| {
+    let output = r_command_from_path(|command| {
```
DavisVaughan, line 61: I think it is best to call `r_command()` here because you called `r_home_setup()` right above this, which sets you up to be able to call `r_command()` with the same `R_HOME` that `r_home_setup()` found.

crates/ark/src/version.rs
```diff
@@ -30,19 +30,35 @@ pub struct RVersion {
...
+            let r_home = String::from_utf8(result.stdout).unwrap();
+            let r_home = r_home.trim();
+
+            // Now set `R_HOME`. From now on, `r_command()` can be used to
+            // run exactly the same R as is running in Ark.
+            unsafe { std::env::set_var("R_HOME", r_home) };
+            PathBuf::from(r_home)
+        },
+    }
+}
```
DavisVaughan, line 56: Would you mind moving this to `command.rs` in harp instead? It is extremely related to `r_command()`, as you can see from this comment just above

```
            // Now set `R_HOME`. From now on, `r_command()` can be used to
            // run exactly the same R as is running in Ark.
```

So i think it makes sense to move it there and add a doc comment like

```
/// Use this before calling `r_command()` to ensure that `R_HOME` is set consistently
```

DavisVaughan (approved):

PR #713, opened 2025-02-14
Fix reticulate focus

dfalbel:
When there's already a comm open for reticulate, we need to send an event message to the front-end:

https://github.com/posit-dev/ark/blob/1eb4a8d9e9eab8211bc87f0356e611155662d435/crates/ark/src/reticulate.rs#L105-L118

I don't know of a direct way of doing this. But here, we send a message to the comm, and then forward it to the front-end
through the outgoing_tx.

This will help addressing https://github.com/posit-dev/positron/issues/3865, as in that case, the quarto extension will call `real_python(input=<chunk_code>)` and we'll forward that code to the front end so that we can execute within the reticulate Python session.

### DIFF
--- a/crates/ark/src/reticulate.rs
+++ b/crates/ark/src/reticulate.rs
@@ -19,7 +19,7 @@ use uuid::Uuid;
 
 use crate::interface::RMain;
 
-static RETICULATE_COMM_ID: LazyLock<Mutex<Option<String>>> = LazyLock::new(|| Mutex::new(None));
+static RETICULATE_COMM: LazyLock<Mutex<Option<CommSocket>>> = LazyLock::new(|| Mutex::new(None));
 
 pub struct ReticulateService {
     comm: CommSocket,
@@ -27,7 +27,10 @@ pub struct ReticulateService {
 }
 
 impl ReticulateService {
-    fn start(comm_id: String, comm_manager_tx: Sender<CommManagerEvent>) -> anyhow::Result<String> {
+    fn start(
+        comm_id: String,
+        comm_manager_tx: Sender<CommManagerEvent>,
+    ) -> anyhow::Result<CommSocket> {
         let comm = CommSocket::new(
             CommInitiator::BackEnd,
             comm_id.clone(),
@@ -45,13 +48,15 @@ impl ReticulateService {
             .send(event)
             .or_log_error("Reticulate: Could not open comm.");
 
+        let socket = service.comm.clone();
+
         spawn!(format!("ark-reticulate-{}", comm_id), move || {
             service
                 .handle_messages()
                 .or_log_error("Reticulate: Error handling messages");
         });
 
-        Ok(comm_id)
+        Ok(socket)
     }
 
     fn handle_messages(&self) -> Result<(), anyhow::Error> {
@@ -61,13 +66,7 @@ impl ReticulateService {
                 break;
             });
 
-            // Eventually, messages will be sent from R to this comm using
-            // main.get_comm_manager_tx().send()
-            // We want these to be forwarded to the front-end.
-            if let CommMsg::Data(_) = msg {
-                self.comm.outgoing_tx.send(msg)?;
-                continue;
-            }
+            log::trace!("Reticulate: Received message from front end: {msg:?}");
 
             if let CommMsg::Close = msg {
                 break;
@@ -81,9 +80,9 @@ impl ReticulateService {
             .or_log_error("Reticulate: Could not send close message to the front-end");
 
         // Reset the global comm_id before closing
-        let mut comm_id_guard = RETICULATE_COMM_ID.lock().unwrap();
-        log::info!("Reticulate Thread closing {:?}", (*comm_id_guard).clone());
-        *comm_id_guard = None;
+        let mut comm_guard = RETICULATE_COMM.lock().unwrap();
+        log::info!("Reticulate Thread closing {:?}", self.comm.comm_id);
+        *comm_guard = None;
 
         Ok(())
     }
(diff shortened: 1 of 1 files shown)


crates/ark/src/reticulate.rs
```diff
@@ -89,11 +98,13 @@ pub unsafe extern "C" fn ps_reticulate_open(input: SEXP) -> Result<SEXP, anyhow:
     let main = RMain::get();
 
     let input: RObject = input.try_into()?;
-    let input_code: Option<String> = input.try_into()?;
-
-    let mut comm_id_guard = RETICULATE_COMM_ID.lock().unwrap();
+    let input_code: Option<String> = match r_is_null(input.sexp) {
```
dfalbel, line 101: This is not the focus of this PR, but would be nice to sneak it here because reticulate by default will send a `NULL`.

crates/ark/src/reticulate.rs
```diff
@@ -89,11 +98,13 @@ pub unsafe extern "C" fn ps_reticulate_open(input: SEXP) -> Result<SEXP, anyhow:
...
+        true => None,
+        false => Some(input.try_into()?),
```
lionel-, line 103: Use `if` instead of `match`? Seems a bit strange with a boolean. (Feel free to ignore)

crates/ark/src/reticulate.rs
lionel-: Add a comment about this reticulate behaviour?

lionel- (commented):
Since these messages are supposed to be coming from the frontend, it seems a bit roundabout to forward them back to the frontend.

How about creating a global singleton representing the `ReticulateService` that you could dereference from the C callbacks? Maybe add a `send_event()` method on that service. And then you can send the event in a more straightforward manner.

What do you think?

In any case we should probably still add handling on these `CommMsg` from the frontend to at least log them when they come in.

dfalbel:
The message is now sent directly to the from-end with https://github.com/posit-dev/ark/pull/713/commits/4695ccf93c3fef633ca11fcdce966faf2f3e40a5 and https://github.com/posit-dev/ark/pull/713/commits/52895fc3eca8d73593d89e16c5efd671907cda81

It might still not be ideal, because we have to clone the service in order to send it to the execution thread. We could avoid cloning by introducing a different crossbean channel like we do in the UI channel:

https://github.com/posit-dev/ark/blob/52895fc3eca8d73593d89e16c5efd671907cda81/crates/ark/src/ui/ui.rs#L48

But this seemed an overkill, instead, we cleanup the global when the thread is closed:

https://github.com/posit-dev/ark/blob/52895fc3eca8d73593d89e16c5efd671907cda81/crates/ark/src/reticulate.rs#L83-L87

crates/ark/src/reticulate.rs
```diff
@@ -44,22 +50,25 @@ impl ReticulateService {
             .send(event)
             .or_log_error("Reticulate: Could not open comm.");
 
+        let serv = service.clone();
         spawn!(format!("ark-reticulate-{}", comm_id), move || {
-            service
+            serv.clone()
```
lionel-, line 55: It seems like this clone is unnecessary since`serv` was moved here?

lionel-: This is a more conventional way of moving a clone:

```rust
        spawn!(format!("ark-reticulate-{}", comm_id), {
            let service = service.clone();
            move || {
                service
                    .handle_messages()
                    .or_log_error("Reticulate: Error handling messages");
            }
        });
```

lionel- (commented):
oops, in hindsight it does not really make sense to make the whole service a singleton, since the service is performed from another thread. Having that global variable being a clone that could potentially run the service from another thread does not feel right.

How about storing the necessary channels in that global such as `comm.outgoing_tx`?

Since this will only be accessed from the R thread, it does not need to be in a Mutex. Instead you can store it in a `thread_local!` whose lazy intialiser panics in other threads. See https://github.com/posit-dev/ark/blob/7c2b7f6f530c5f03852f82bb1f18d0f801d7dd85/crates/ark/src/plots/graphics_device.rs#L65-L68 for an example. It'll have to be stored in a `RefCell`, which would cause a panic if double borrowed mutably (need to be careful), but that's better than a deadlock if double locked from the same thread.

dfalbel:
I had an intermediate approach where I stored to the `comm` socket (and thus the necessary channels) in the global: https://github.com/posit-dev/ark/pull/713/commits/4695ccf93c3fef633ca11fcdce966faf2f3e40a5

Does this sound right to you? Are there problems in storing a clone of the comm socket and calling it from the R thread (potentially at the same time as the service thread)? It seemed fine to me, but I don't know about the socket internals.

crates/ark/src/reticulate.rs
```diff
@@ -18,21 +19,27 @@ use uuid::Uuid;
 
 use crate::interface::RMain;
 
-static RETICULATE_COMM_ID: LazyLock<Mutex<Option<String>>> = LazyLock::new(|| Mutex::new(None));
+static RETICULATE_SOCKET: LazyLock<Mutex<Option<CommSocket>>> = LazyLock::new(|| Mutex::new(None));
```
lionel-, line 22: I'd have made that as minimal as possible so that only the service is able to handle messages but I guess we can also just be careful.

lionel- (approved):

crates/ark/src/reticulate.rs
dfalbel: Ok, I can probably just keep the channel the `outbound.tx` channel here.

lionel- (approved):

PR #723, opened 2025-02-25
Format with Air

DavisVaughan:
Going with 4 spaces as the indent size since we have this in `.editorconfig` and that is what most of the project (Rust) uses. Some files were fully committed to using 2 spaces so the diff is every line for those files, but I checked every file and it all looks good (except for some known if/else comment issues I manually tweaked)

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -4,17 +4,18 @@
         "editor.formatOnSaveMode": "file",
         "editor.defaultFormatter": "rust-lang.rust-analyzer"
     },
-
+    "[r]": {
+        "editor.formatOnSave": true,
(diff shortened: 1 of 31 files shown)


lionel- (approved):

.vscode/settings.json
```diff
@@ -4,17 +4,18 @@
         "editor.formatOnSaveMode": "file",
         "editor.defaultFormatter": "rust-lang.rust-analyzer"
     },
-
+    "[r]": {
+        "editor.formatOnSave": true,
```
lionel-, line 8: This is also needed in the Zed settings file

PR #732, opened 2025-02-28
Rework our plots infrastructure

DavisVaughan:
Addresses posit-dev/positron#5346
Addresses posit-dev/positron#5046
Addresses posit-dev/positron#5020
Addresses posit-dev/positron#5827
Addresses posit-dev/positron#4549
Addresses posit-dev/positron#3214
Addresses posit-dev/positron#6113
Addresses posit-dev/positron#4593
Addresses posit-dev/positron#3915
Addresses posit-dev/positron#3834
Addresses posit-dev/positron#3667
Addresses posit-dev/positron#3562
Addresses posit-dev/positron#727

This PR reworks our entire plots infrastructure. The goal here is to fix our timing related bugs, which involve things like:

- Dropping plots when multiple plot pages are generated inside a single expression (plotting in a loop, or running a single expression that makes >1 ggplots)
- Dropping plot updates when multiple updates happen in a single expression (`par()` related issues)
- Dropping plots when switching graphics devices within a single expression (`ggsave()` related issue)

It does not fix every bug, but I believe it fixes the most pressing and annoying ones.

I have added a new `graphics-devices.md` documentation page that describes how all of this works now. It's something I wanted when coming to this for the first time.

I've added some Ark tests that go through the Jupyter code paths, but we will want to add a number of similar tests on the Positron side to ensure we are going through the Positron side of things too. This is one place where not mimicking Jupyter exactly makes it a bit hard to test the Positron specific behavior on the Ark side.

TODOs in follow up PRs:
- Cleaning up `RECORDINGS` by getting "close" events from Positron (@lionel- said he could work on this)
- Sending Positron an "initial" plot using some default dimensions / graphics device so it can display something immediately (@lionel- said he could work on this)
- Add lots of tests upstream on the Positron side (Me)
- Test this PR on Windows extensively (Me)
- These still don't work https://github.com/posit-dev/positron/issues/2738
- Investigate using ragg devices in place of `png()` and friends where possible if installed

## Multiple plot pages in a single expression

If you run both of these lines in one chunk, then there was a good chance we'd drop the first plot.

```r
plot(1:10)
plot(2:20)
```

The problem was that while we were recording the display list of the first plot using the `before.plot.new` hook, we were not notifying Positron that we had a new plot available for it. We'd only notify positron about the most recent plot after finishing the whole top level expression chunk.

## Multiple plot updates in a single expression

If you run all of these lines in one chunk, you'd see the first plot show up on the left hand side, but you would not see the second plot show up on the right hand side. In fact, even if you run this line by line you would not see the update!

```r
par(mfrow = c(1, 2))
plot(1:10)
plot(2:20)
```

This again had to do with the `before.plot.new` hook. We would:
- Run `plot(1:10)`
- Before `plot(2:20)` was drawn, our `before.plot.new` hook would get called and we would record the display list for `plot(1:10)` with plot `id="foo"`
- After `plot(2:20)` was drawn (still on `id="foo"`), we'd recognize that we have changes and we'd tell Positron about the changes. When Positron responded we'd render the plot, but since we already had a recording corresponding to `id="foo"`, we'd respond with that recording, forgetting that there are extra changes in the display list we had not yet captured.

The solution here was to record more frequently. In this case, in the `process_changes()` hook when we detect some changes as well as in `before.plot.new`. This also allowed me to remove the idea of rendering from the current device altogether. Instead I've greatly simplified things by only rendering from pre-recorded snapshots.

An important note about the `before.plot.new` hook vs the `new_page` hook is that the `new_page` hook only gets called when we are about to fully wipe the screen. In the case of `par(mfrow = c(1, 2))`, when we _move on to the next plot_ by calling `plot.new()` that calls our `before.plot.new` hook, but does NOT necessarily call our `new_page` hook unless moving on to the next plot moves us past `prod(mfrow)` number of plots on screen.

## Switching graphics devices within a single expression

If you run the two chunks of code below, you'd find that Positron would try and render `p` (or `plot(1:10)`) but it would not be able to find the snapshot for it, nor would it be able to render it from the current graphics device's display list.

```r
# This example
library(ggplot2)
p <- ggplot(mtcars, aes(x = wt, y = mpg)) + geom_point()
p
ggsave("temp.png", p)

# Is roughly equivalent to this
plot(1:10)
grDevices::png()
plot(1:20)
dev.off()
```

What was happening here is that both `ggsave()` and `grDevices::png()` _change the graphics device_ mid expression! This results in our graphics device `deactivate()`-ing (the hook), and the png device comes only until we run `dev.off()`, at which time we `activate()` again.

But the process of deactivating and reactivating seems to clear out the display list for our graphics device in such a way that we can't record it anymore. The solution here was to also record the display list in our `deactivate()` hook when we notice there are changes, ensuring we capture them for this plot `id` before going offline temporarily.

## Figure margins are too damn large!

If you flip around the above example and run all of this at once, you'd likely get an error of

```r
par(mfrow = c(2, 1))
plot(1:10)
plot(2:20)
```

```r
#> Error in `plot.new()`:
#> ! figure margins too large
```

This somehow had to do with the fact that we used a resolution of 144 ppi in the `png()` device we shadow when recording the display list (assuming a high resolution screen), but then used a resolution of 72 ppi later on in the actual rendering code. I'm not entirely sure what was going on here, but using 72 ppi in the `png()` device we shadow seemed to get this to work correctly (even though at render time we often scale that up by a pixel ratio of 2 back to 144 ppi).


crates/ark/src/modules/positron/graphics.R
```diff
@@ -1,113 +1,69 @@
...
-        stop("This version of R wasn't built with plotting capabilities")
-    }
-}
-
-renderWithPlotDevice <- function(filepath, format, width, height, res, type) {
-    # width and height are in inches and use 72 DPI to create the requested size in pixels
-    dpi <- 72
+# A persistent list mapping plot `id`s to their display list recording.
+# Used for replaying recordings under a new device or new width/height/resolution.
+RECORDINGS <- list()
```
DavisVaughan, line 18: We now store `recordPlot()` recordings in a persistent list rather than writing them to disk. This should be a little faster. Once we gain the ability to know when Positron no longer needs a plot, we should also be able to remove recordings from the persistent list. Right now they persist forever.

crates/ark/src/modules/positron/graphics.R
```diff
@@ -1,113 +1,69 @@
...
 
 #' @export
-.ps.graphics.createDevice <- function(name, type, res) {
-    # Get path where plots will be generated.
-    plotsPath <- .ps.graphics.plotSnapshotRoot("current-plot.png")
-    ensure_parent_directory(plotsPath)
+.ps.graphics.createDevice <- function(name, type) {
+    # Get path where non-recorded plots will be generated.
+    root <- plotRecordingRoot()
+    filename <- file.path(root, "current-plot.png")
```
DavisVaughan, line 52: I don't fully understand when this gets written to. It is a side effect of the fact that we piggy back on a `png()` device. Possibly it writes to it anytime we change devices.

crates/ark/src/modules/positron/graphics.R
```diff
@@ -134,107 +90,178 @@ renderWithPlotDevice <- function(filepath, format, width, height, res, type) {
     # Replace bindings.
     env_bind_force(baseenv(), ".Devices", .Devices)
     env_bind_force(baseenv(), ".Device", newDevice)
+
+    # Also set ourselves as a known interactive device.
+    # Used by `dev.interactive()`, which is used in `stats:::plot.lm()`
+    # to determine if `devAskNewPage(TRUE)` should be set to prompt before
+    # each new plot is drawn.
+    grDevices::deviceIsInteractive(name)
```
DavisVaughan, line 98: This is new, without this `plot.lm()` doesn't stop and prompt the user to hit Enter

We still don't handle this quite right because we don't show any plots until the next top level expression idle time, but we can iterate on this.

crates/ark/src/modules/positron/graphics.R
```diff
@@ -134,107 +90,178 @@ renderWithPlotDevice <- function(filepath, format, width, height, res, type) {
...
 
 #' @export
-.ps.graphics.renderPlot <- function(id, width, height, dpr, format) {
-    # If we have an existing snapshot, render from that file.
-    snapshotPath <- .ps.graphics.plotSnapshotPath(id)
-    if (file.exists(snapshotPath)) {
-        .ps.graphics.renderPlotFromSnapshot(id, width, height, dpr, format)
-    } else {
-        .ps.graphics.renderPlotFromCurrentDevice(id, width, height, dpr, format)
-    }
```
DavisVaughan, line 170: No more reliance on this if/else. When we render we unconditionally pull it from a pre-recorded snapshot now.

crates/ark/src/modules/positron/graphics.R
```diff
@@ -134,107 +90,178 @@ renderWithPlotDevice <- function(filepath, format, width, height, res, type) {
...
+        ),
+        stop("Internal error: Unknown plot `format`.")
+    )
+
+    # Ensure we turn off the device on the way out, this:
+    # - Commits the plot to disk
+    # - Resets us back as being the current device
+    on.exit(utils::capture.output({
+        grDevices::dev.off()
+        if (old_dev > 1) {
```
DavisVaughan, line 213: Trying to add some extra safety here

crates/ark/src/modules/positron/graphics.R
```diff
@@ -134,107 +90,178 @@ renderWithPlotDevice <- function(filepath, format, width, height, res, type) {
...
+
+    if (format == "svg" || format == "pdf") {
+        # These devices require `width` and `height` in inches.
+        # We convert from pixels to inches here.
+        # There is no `type` or `res` argument for these devices.
+        return(list(
+            type = NULL,
+            res = NULL,
+            width = width * pixel_ratio / defaultResolutionInPixelsPerInch(),
+            height = height * pixel_ratio / defaultResolutionInPixelsPerInch()
```
DavisVaughan, line 242:
```suggestion
            width = width / defaultResolutionInPixelsPerInch(),
            height = height / defaultResolutionInPixelsPerInch()
```

I'm somewhat convinced this should just be this, but I think I need @thomasp85 to weigh in here. Compare with the png path above.

Right now I'm maintaining the previous calculation that was there before.

crates/ark/src/modules/positron/graphics.R
```diff
@@ -134,107 +90,178 @@ renderWithPlotDevice <- function(filepath, format, width, height, res, type) {
...
+    stop("Internal error: Unknown plot `format`.")
+}
+
+defaultResolutionInPixelsPerInch <- function() {
+    if (Sys.info()[["sysname"]] == "Darwin") {
+        96L
+    } else {
+        72L
+    }
+}
```
DavisVaughan, line 255: These were here before, but are they right?

crates/ark/src/r_task.rs
```diff
@@ -146,7 +146,9 @@ where
     T: 'env + Send,
 {
     // Escape hatch for unit tests
-    if stdext::IS_TESTING {
+    // In integration tests with dummy frontends, we have a "real" RMain and want to
+    // go through the standard r-task path
+    if stdext::IS_TESTING && !RMain::is_initialized() {
```
DavisVaughan, line 151: Our integration tests that start up a dummy frontend happened to never run any `r_task()`s. If they did, we would have seen `R has already been initialized` as an error.

Basically, in integration tests with our dummy frontend we _actually have RMain_ and starting that RMain already called `Rf_initialize_R()`, but we try and do it again here in `r_test_init()`, resulting in that above error. We want to bypass this particular usage of `stdext::IS_TESTING` in integration tests and actually treat it like production.

It is confusing and I'm hoping to move us over to nextest soon which may help simplify this

crates/ark/tests/plots.rs
DavisVaughan, line 1: These test Jupyter paths, not Positron paths, but are still very useful i think

crates/ark/src/plots/graphics_device.rs
```diff
@@ -70,147 +56,272 @@ thread_local! {
...
-        });
-
-        // Get the RPC request.
-        if socket.handle_request(message, |req| self.handle_rpc(req, plot_id)) {
-            return;
+        // Check for incoming plot render requests.
+        // Totally possible to have >1 requests pending, especially if we've plotted
+        // multiple things in a single chunk of R code. The `Err` case is likely just
+        // that no channels have any messages, so we don't log in that case.
+        while let Ok(selection) = select.try_select() {
```
DavisVaughan, line 279: This is an important change. Previously we only rendered 1 Positron render request per call of this function, which made rendering a batch of plots very slow. Now we render as many as we have in the pipeline.

crates/ark/src/plots/graphics_device.rs
```diff
@@ -229,41 +340,48 @@ impl DeviceContext {
...
+            self.process_new_plot_positron(id);
         } else {
-            self.process_new_plot_jupyter_protocol(id, iopub_tx);
+            self.process_new_plot_jupyter_protocol(id);
         }
     }
 
-    fn process_new_plot_positron(&self, id: &str, comm_manager_tx: Sender<CommManagerEvent>) {
+    #[tracing::instrument(level = "trace", skip_all, fields(id = %id))]
+    fn process_new_plot_positron(&self, id: &PlotId) {
```
DavisVaughan, line 382: I've added tracing instrumentation everywhere, which really helped me understand what the heck was going on

crates/ark/src/plots/graphics_device.rs
```diff
@@ -458,96 +603,142 @@ macro_rules! with_device {
...
+        cell.hook_new_page();
     });
 }
 
 unsafe fn ps_graphics_device_impl() -> anyhow::Result<SEXP> {
     // TODO: Don't allow creation of more than one graphics device.
     // TODO: Allow customization of the graphics device here?
 
-    // TODO: Infer appropriate resolution based on whether display is high DPI.
-    let res = 144;
```
DavisVaughan, line 541: Now set to `defaultResolutionInPixelsPerInch()` on the R side, which is 96 or 72 depending on the OS. I'm still not sure if that is exactly right but it fixes a bunch of "figure margins too big" issues.

crates/ark/src/modules/positron/graphics.R
thomasp85: They are more correct than any other numbers but that doesn't mean they are truly correct. macOS uses 96 DPI as it's internal scaling, whereas this is user definable on windows. However, this is a scaling factor that tries to make things that appear on screen be as close to the size they are printed in and this has always been a perilous proposition

crates/ark/src/plots/graphics_device.rs
thomasp85: also, the high dpi inference should come from the pixel_ratio value

crates/ark/src/modules/positron/graphics.R
thomasp85: Is it not better to store them in an environment? Or is it because `id` is an integer and not a string?

crates/ark/src/modules/positron/graphics.R
```diff
@@ -134,107 +90,178 @@ renderWithPlotDevice <- function(filepath, format, width, height, res, type) {
...
+
+defaultDeviceType <- function() {
+    if (has_aqua()) {
+        "quartz"
+    } else if (has_cairo()) {
```
thomasp85, line 260: For now (until we perhaps provide our own device), this should have the top priority over quartz, even on macOS. The Cairo implementation is what gets the most love so use that when available

crates/ark/src/modules/positron/graphics.R
DavisVaughan: `id` is a string, but remember when you put it in an environment that string is converted to a SYMSXP. @lionel- was worried about the fact that we'd be creating a lot of SYMSXPs that would never get cleaned up. R doesn't ever clean up the SYMSXP list.

DavisVaughan:
Documenting one peculiarity in particular. This should show the user a series of 5 plots, requiring the user to pause and hit Enter before being shown the next plot. It gives them a stdin prompt to hit Enter at each time.

```r
fit_lm <- lm(
  mpg ~ cyl + wt,
  data = mtcars
)

plot(fit_lm)
```

What actually happens is that you do get the prompt to hit Enter at, but all 5 plots are being shown at once at the end. You get 5 "blank" plots on the Positron side as Positron gets the notification that Ark has something that it is going to want to display, but Positron's message back to Ark that it is ready for us to render the plot doesn't makes it back to us until after we are back to top level.

This confused me, because the stdin prompt is considered "idle time", so we should be able to receive render requests at this time.

I've determined the problem is that Positron's render request comes back through over Shell as a `Message::CommMsg`, but Shell is currently completely blocked by the current `Message::ExecuteRequest`. I'm not sure if we've seen a problem like this before or not. It would be nice if Positron's render requests came through on a different channel that wasn't fully blocked (maybe Control or IOPub?). 
https://github.com/posit-dev/ark/blob/c72222136f4f61da31022c2d144e39c5f9345449/crates/amalthea/src/socket/shell.rs#L153

crates/ark/src/plots/graphics_device.rs
```diff
@@ -70,147 +56,272 @@ thread_local! {
...
+    ///   data
+    /// - AFTER the entire for loop has finished and we hit idle time, we drop into
+    ///   `process_rpc_requests()` and render all 5 plots at once
+    ///
+    /// Practically this seems okay, it is just something to keep in mind.
+    #[tracing::instrument(level = "trace", skip_all)]
+    fn process_rpc_requests(&self) {
         // Don't try to render a plot if we're currently drawing.
-        if self._mode.get() != 0 {
+        if self.is_drawing.get() {
```
lionel-, line 250: Can this ever be the case at idle time? I'm a bit fuzzy as to when `mode` changes exactly.

crates/ark/src/plots/graphics_device.rs
lionel-: Nice

crates/ark/src/plots/graphics_device.rs
```diff
@@ -70,147 +56,272 @@ thread_local! {
...
-        // Get the RPC request.
-        if socket.handle_request(message, |req| self.handle_rpc(req, plot_id)) {
-            return;
+        // Check for incoming plot render requests.
+        // Totally possible to have >1 requests pending, especially if we've plotted
+        // multiple things in a single chunk of R code. The `Err` case is likely just
+        // that no channels have any messages, so we don't log in that case.
+        while let Ok(selection) = select.try_select() {
+            let id = unsafe { &sockets.get_unchecked(selection.index()).0 };
+            let socket = unsafe { &sockets.get_unchecked(selection.index()).1 };
```
lionel-, line 281: Not that it matters much but I think get().unwrap() or expect() would be better than unsafe here?

crates/ark/src/plots/graphics_device.rs
```diff
@@ -70,147 +56,272 @@ thread_local! {
...
+
+            // Receive on the "selected" channel
+            let message = match selection.recv(&socket.incoming_rx) {
+                Ok(message) => message,
+                Err(error) => {
+                    log::error!("{error:?}");
```
lionel-, line 287: Should we try and clean up the socket if it's been closed?

crates/ark/src/plots/graphics_device.rs
lionel-: Cool

lionel- (approved):
Looks good and the new doc is of great value

crates/ark/src/plots/graphics_device.rs
thomasp85: Mode changes before and after each primitive is rendered. That means that it can change hundreds or thousands of times during a single R call depending on the complexity of what is rendered

lionel-: Right but when R is idle (all calls have returned) can this still be `1`?

DavisVaughan: Practically idle time is when all user executed calls have returned and the user's cursor is sitting at the command prompt.

I think I'm going to just leave this check in for now. If we determine later we really don't need it we can definitely remove it.

PR #734, opened 2025-03-07
WIP: Perform a two way handshake around `kernel_info_request`

DavisVaughan:
Addresses https://github.com/posit-dev/positron/issues/6344 particularly by looking closely at the message order in https://github.com/posit-dev/positron/issues/6344#issuecomment-2675438161

This addresses the fact that we could run `.Rprofile` before receiving an IOPub subscription message from Kallichore. In this case both sides of the socket have technically connected, but the Kallichore side hasn't yet sent over its IOPub subscription message, so Ark's IOPub socket ends up _dropping_ messages on the way out because no one is subscribed yet.

We already block responding to a `kernel_info_request` until R is "started up enough". What I've added is a two-way handshake where we also block the final steps of R's startup procedure (running `.Rprofile` and then entering the main repl) until we've actually received the `kernel_info_request` at all. This has the side effect of being late enough that by this point we've received the IOPub subscription message. It is a little hand wavy though because  `kernel_info_request` comes in on Shell and technically we should be waiting to hear something on IOPub (maybe we can block until we get the IOPub subscription instead?).

I've still got a good amount of thinking and clean up to do here, but...

https://github.com/user-attachments/assets/f6812522-f986-409b-a5fd-9d3c79dab962

### DIFF
--- a/crates/ark/src/interface.rs
+++ b/crates/ark/src/interface.rs
@@ -165,6 +165,8 @@ thread_local! {
 pub struct RMain {
     kernel_init_tx: Bus<KernelInfo>,
 
+    kernel_info_request_rx: Receiver<()>,
+
     kernel_request_rx: Receiver<KernelRequest>,
 
     /// Whether we are running in Console, Notebook, or Background mode.
@@ -313,6 +315,7 @@ impl RMain {
         stdin_reply_rx: Receiver<amalthea::Result<InputReply>>,
         iopub_tx: Sender<IOPubMessage>,
         kernel_init_tx: Bus<KernelInfo>,
+        kernel_info_request_rx: Receiver<()>,
         kernel_request_rx: Receiver<KernelRequest>,
         dap: Arc<Mutex<Dap>>,
         session_mode: SessionMode,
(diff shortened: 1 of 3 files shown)


crates/amalthea/src/fixtures/dummy_frontend.rs
```diff
@@ -190,6 +190,12 @@ impl DummyFrontend {
         assert_matches!(Self::recv(&iopub_socket), Message::Welcome(data) => {
             assert_eq!(data.content.subscription, String::from(""));
         });
+        // We also go ahead and handle the `ExecutionState::Starting` status that we know
+        // is coming from the kernel right after the `Welcome` message, so tests don't
+        // have to care about this.
+        assert_matches!(Self::recv(&iopub_socket), Message::Status(data) => {
+            assert_eq!(data.content.execution_state, ExecutionState::Starting);
+        });
```
DavisVaughan, line 198: The assert on `Message::Welcome` is what was failing on my local Windows machine. It should have been failing everywhere because we do try and send out `ExecutionState::Starting` first, but everywhere else we were just dropping `ExecutionState::Starting` because no one was connected yet!

PR #736, opened 2025-03-10
Expand patch renv support to renv 1.0.1 and earlier

DavisVaughan:
Addresses https://github.com/posit-dev/positron/issues/6681

This expands on https://github.com/posit-dev/ark/pull/383 where we automatically return `"n"` when we detect we are in a `readline()` in the renv autoloader. The code goes through a different code path in 1.0.1 so we need to do something different there.

- In renv 1.0.9 and up, we should never end up here
- In renv 1.0.2 to 1.0.8, we return `"n"` to escape the `renv:::ask()` call
- In renv 1.0.1 and earlier, we return `"Leave project library empty"` to escape the specific `renv:::menu()` call

I can't figure out any better way to do this due to the way renv bootstraps itself. Luckily this should be very static "patch" support since CRAN renv no longer needs anything like this.

Here is the rladies repo at least booting up now, you get no indication that you need to `restore()` but I think this is the best we can do right now.

Screenshot 2025-03-10 at 3 54 54 PM

### DIFF
--- a/crates/ark/src/interface.rs
+++ b/crates/ark/src/interface.rs
@@ -1083,15 +1083,23 @@ impl RMain {
     /// invalid `input_request` case we throw an R error and assume that it
     /// came from a `readline()` or `menu()` call during startup.
     ///
-    /// We make a single exception for preexisting renv `activate.R` scripts,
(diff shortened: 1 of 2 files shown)


lionel- (approved):
ew :)

PR #737, opened 2025-03-10
Improve bindings comparisons for the variables pane

dfalbel:
This PR addresses: https://github.com/posit-dev/positron/issues/6618

The main issue is that when comparing two S4 connections objects (that include environments as attributes) we're using all the derived methods and ultimately resolving to `impl PartialEq for BindingNestedEnvironment` which always says `false` if objects have nested environments.

However, the variables pane is not trying to compare by value, but only for identity. 
This PR removes the `has_nested_environment` attribute that doesn't seem to be used anywhere and proposes that equality of Bindings are:

1. Their symbols are identical
2. Their values are identical (ie, they are pointing to the same `SEXP`)

We might want to use a different naming for this compariso, instead of overloading `eq`, let me know what you think.

### DIFF
--- a/crates/harp/src/environment_iter.rs
+++ b/crates/harp/src/environment_iter.rs
@@ -17,9 +17,11 @@ pub struct Binding {
 }
 
 #[derive(Eq, PartialEq)]
-pub enum RObjectId {
-    Cons(SEXP, Box<RObjectId>),
-    Nil,
+pub enum RObjectValueId {
+    Active(SEXP),
+    Promise(SEXP),
+    Altrep(SEXP, SEXP, SEXP),
+    Standard(SEXP),
 }
 
 pub enum BindingValue {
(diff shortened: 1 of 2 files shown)


crates/harp/src/environment_iter.rs
```diff
@@ -13,20 +11,20 @@ pub struct EnvironmentIter {
...
-// What is this used for? Do we still need it?
-// https://github.com/posit-dev/amalthea/commit/04a9fe20a48cb01c4fe2d6fe1a4dfdc0f2a186fc
-#[derive(Eq)]
-pub struct BindingNestedEnvironment {
-    pub has_nested_environment: bool,
+// Bindings are equal if their names and values are exactly the same.
+impl PartialEq for Binding {
+    fn eq(&self, other: &Self) -> bool {
+        self.name == other.name && self.value == other.value
+    }
```
lionel-, line 24: Does this do something different than the automatic derivation? If not (it looks like it), probably best to keep deriving it to avoid suggesting something special is going on.

crates/harp/src/environment_iter.rs
```diff
@@ -38,14 +36,41 @@ pub enum BindingValue {
...
         data2: RObject,
-        has_nested_environment: BindingNestedEnvironment,
     },
     Standard {
         object: RObject,
-        has_nested_environment: BindingNestedEnvironment,
     },
 }
 
+// Two binding values are equal if they are identical (ie, their SEXP's)
```
lionel-, line 45: This is a bit misleading because this notion of equality is very different to R's notion of "identical"

lionel- (commented):
> We might want to use a different naming for this compariso, instead of overloading eq, let me know what you think.

I think that's a great idea because using pointer identity for Rust's equality is confusing. It should in principle do a deep comparison, e.g. it should not matter if `1:2` is an altrep object, an integer vector, or if the address is different.

How about we create an `id()` method that returns an enum like:

```rust
    enum RObjId {
        Cons(Sexp, RObjId),
        Nil,
    }
```

This recursive type could hold any number of `SEXP` references and have equality derived. The `id()` method would build a list of sexp pointers that you can then compare with `==`.

crates/harp/src/environment_iter.rs
```diff
@@ -13,20 +11,17 @@ pub struct EnvironmentIter {
...
 }
 
-// What is this used for? Do we still need it?
-// https://github.com/posit-dev/amalthea/commit/04a9fe20a48cb01c4fe2d6fe1a4dfdc0f2a186fc
-#[derive(Eq)]
-pub struct BindingNestedEnvironment {
-    pub has_nested_environment: bool,
+#[derive(Eq, PartialEq)]
+pub enum RObjectId {
+    Cons(SEXP, Box<RObjectId>),
```
lionel-, line 21: oh I did not realise the recursion would need to be indirected like that.

Now I wonder if a Vec would have been better so there's only one pointer dereference (to get the vector's array on the heap).

The other thing that now occurs to me is that the Id should be tagged by the original type to avoid potential ambiguity when two different wrappers of the same object have the same id structure (doesn't currently seem possible though).

I think it's fine as is though, just noting these potential issues. Maybe just add a comment that the id needs to be tagged if used in more complex situations.

lionel- (approved):

dfalbel:
@lionel- I have updated the PR to avoid using the recursive type and now avoid the ambiguity with an enum of tuples of SEXP's.

lionel- (approved):

PR #740, opened 2025-03-12
Better hygiene around S3 methods in tests

jennybc:
While working on something else, I noticed a flaky test:

```
failures:

---- variables::variable::tests::test_matrix_display stdout ----

thread 'variables::variable::tests::test_matrix_display' panicked at crates/ark/src/variables/variable.rs:2133:13:
assertion `left == right` failed
  left: "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
 right: "[1 row x 1 column] <foo>"


failures:
    variables::variable::tests::test_matrix_display
```

Discussed with @DavisVaughan who offered some very helpful analysis: this comes from two different tests being run in the same R session ?in parallel or in some random order?. Both have assertions that tickle the `ark_positron_variable_display_value` method for a toy `"foo"` class and they can crosstalk with each other.

Again at @DavisVaughan's suggestion, I am introducing and using a method to unregister an S3 method in ark's method table.

I have very low-tech empirical proof that it's working. Before and after this PR, I ran `cargo test --package ark --lib variables::variable::tests` 10 times. Before, I saw 2 failures (the exact one shown above) and 8 passes. After, I see 10 passes.

### DIFF
--- a/crates/ark/src/modules/positron/methods.R
+++ b/crates/ark/src/modules/positron/methods.R
@@ -1,7 +1,7 @@
 #
 # methods.R
 #
-# Copyright (C) 2024 Posit Software, PBC. All rights reserved.
+# Copyright (C) 2024-2025 Posit Software, PBC. All rights reserved.
 #
 #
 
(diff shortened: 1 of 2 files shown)


crates/ark/src/modules/positron/methods.R
```diff
@@ -26,38 +26,64 @@ lockEnvironment(ark_methods_table, TRUE)
...
 
-#' Register the methods with the Positron runtime
-#'
-#' @param generic Generic function name as a character to register
-#' @param class Class name as a character
-#' @param method A method to be registered. Should be a call object.
-#' @export
-.ark.register_method <- function(generic, class, method) {
-    # Check if the caller is an allowed package
+# check if the calling package is allowed to touch the methods table
```
jennybc, line 29: There was just so much duplication between the register/unregister that I think these helpers might be nice?

dfalbel (approved):
LGTM!

> ?in parallel or in some random order?

They don't run in parallel as only a single `r_task` can be running due to the [mutex](https://github.com/posit-dev/ark/blob/2a02ce062ab45793a966a17487433508a4113ea9/crates/ark/src/r_task.rs#L150-L151). But they can definitely run in a undeterministic order, depending on which `r_task` is able to acquire the lock first.

PR #745, opened 2025-03-13
Only force ALTREP compact row names (i.e. from duckplyr) if requested

DavisVaughan:
Addresses https://github.com/posit-dev/positron/issues/4158 at the request of @hadley (but not the part about errors during materialization crashing the kernel, that is a much deeper and harder to fix problem that requires us to adjust our assumptions about what a simple `INTEGER_ELT()` call can do - i.e. in an ALTREP world that can run arbitrary code that can error 😢)

Joint PR with https://github.com/tidyverse/duckplyr/pull/661

The problem is that duckplyr uses ALTREP compact row names as the `R_RowNamesSymbol` attribute. Calling `INTEGER_ELT()` or `INTEGER()` on this to get the number of rows will trigger the entire duckdb query to run, something we want to avoid. This is hard though, because that's how we determine the number of rows in the data frame. We've determined that rather than trying to carefully avoid ALTREP compact row names every time we look at the number of rows in a data frame, a better solution is to use @dfalbel's hooks for allowing S3 classes to provide custom Variables Pane implementations - avoiding these troublesome code paths entirely.

Most of the work that was _required_ for this is actually in https://github.com/tidyverse/duckplyr/pull/661. However, in `variable_length()` here on the Ark side we were calling `table_info()` which queried the number of rows (bad) and number of columns even though we only needed the number of columns. I've reworked some things to expose ways to compute _just_ the number of columns or _just_ the number of rows for a data frame. With that in place, https://github.com/tidyverse/duckplyr/pull/661 does the rest of the work.

This won't trigger the duckdb query:
- Creating a duckdb tibble and having it show up in the Variables pane

This will trigger the duckdb query, and we are ok with this, and this is similar to RStudio:
- "Expanding" the tibble in the Variables pane to look at the column values
- Viewing the tibble in the Data Explorer

Here are some videos of me toying around with this:

https://github.com/user-attachments/assets/2a82111b-5561-421e-b259-79c51fcfea11

https://github.com/user-attachments/assets/754a6ab4-8c94-4267-be4c-e46884da63bf

See also, RStudio's `rs_dim()` here:
https://github.com/rstudio/rstudio/blob/7a9ab7afe9ae006e60897596a189a904a716ec4f/src/cpp/session/modules/environment/SessionEnvironment.cpp#L536
https://github.com/rstudio/rstudio/blob/7a9ab7afe9ae006e60897596a189a904a716ec4f/src/cpp/session/modules/SessionEnvironment.R#L560

### DIFF
--- a/crates/ark/src/modules/positron/methods.R
+++ b/crates/ark/src/modules/positron/methods.R
@@ -24,7 +24,7 @@ ark_methods_table$ark_positron_variable_get_children <- new.env(
 )
 lockEnvironment(ark_methods_table, TRUE)
 
-ark_methods_allowed_packages <- c("torch", "reticulate")
+ark_methods_allowed_packages <- c("torch", "reticulate", "duckplyr")
 
 # check if the calling package is allowed to touch the methods table
 check_caller_allowed <- function() {
@@ -77,7 +77,9 @@ check_register_args <- function(generic, class) {
     check_register_args(generic, class)
 
     for (cls in class) {
-        if (exists(cls, envir = ark_methods_table[[generic]], inherits = FALSE)) {
+        if (
+            exists(cls, envir = ark_methods_table[[generic]], inherits = FALSE)
+        ) {
             remove(list = cls, envir = ark_methods_table[[generic]])
         }
     }
--- a/crates/harp/src/matrix.rs
+++ b/crates/harp/src/matrix.rs
@@ -1,0 +1,23 @@
+use libr::*;
+
+use crate::r_dim;
+use crate::r_int_get;
+use crate::r_length;
+use crate::utils::*;
+
+/// Compute the dimensions of a matrix
+pub fn mat_dim(x: SEXP) -> crate::Result<(i32, i32)> {
+    if !r_is_matrix(x) {
+        return Err(crate::anyhow!("`x` must be a matrix"));
+    }
+
+    let dim = r_dim(x);
+
+    if r_typeof(dim) != INTSXP || r_length(dim) != 2 {
+        return Err(crate::anyhow!(
+            "`dim` must be an integer vector of length 2"
+        ));
+    }
+
+    Ok((r_int_get(dim, 0), r_int_get(dim, 1)))
+}
--- a/crates/harp/src/matrix.rs
+++ b/crates/harp/src/matrix.rs
@@ -5,19 +5,30 @@ use crate::r_int_get;
 use crate::r_length;
 use crate::utils::*;
 
-/// Compute the dimensions of a matrix
-pub fn mat_dim(x: SEXP) -> crate::Result<(i32, i32)> {
-    if !r_is_matrix(x) {
-        return Err(crate::anyhow!("`x` must be a matrix"));
+/// Matrix support
+///
+/// # Notes
+///
+/// Currently we only utilize this for the static [Matrix::dim()] method,
+/// but we could actually wrap matrices here (with validation) and provide
+/// additional methods, like [harp::DataFrame].
+pub struct Matrix {}
+
+impl Matrix {
+    /// Compute the dimensions of a matrix
+    pub fn dim(x: SEXP) -> crate::Result<(i32, i32)> {
+        if !r_is_matrix(x) {
+            return Err(crate::anyhow!("`x` must be a matrix"));
+        }
+
+        let dim = r_dim(x);
+
+        if r_typeof(dim) != INTSXP || r_length(dim) != 2 {
+            return Err(crate::anyhow!(
+                "`dim` must be an integer vector of length 2"
+            ));
+        }
+
+        Ok((r_int_get(dim, 0), r_int_get(dim, 1)))
     }
-
-    let dim = r_dim(x);
-
-    if r_typeof(dim) != INTSXP || r_length(dim) != 2 {
-        return Err(crate::anyhow!(
-            "`dim` must be an integer vector of length 2"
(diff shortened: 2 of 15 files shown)


dfalbel:
I didn't look at the code still. I wonder if it would make sense to implement duckdb specifics using the variables pane extension mechanism. For instance by implementing custom `ark_positron_variable_display_value()` for `prudent_duckplyr_df` / `duckplyr_df`. This could potentially even live in duckplyr's codebase.

krlmlr:
Thanks for looking into it.

Calling `head()` on the duck frame (the same way pillar does) won't materialize and will be faster in many cases.

Why is it hard to handle ALTREP failures? Does this change mean that the kernel won't crash for a stingy duck frame unless the user clicks a button? Could this be worked around with calling `head()` ?

hadley:
@krlmlr why does an ALTREP access generate an error? That seems weird to be since you'd normally expect (e.g.) `INTEGER(x)` to always work.

krlmlr:
This may happen for any number of reasons, including out-of-memory conditions. The "prudence" feature tries to avoid these OOM conditions and makes the errors more visible.

Can you think of an alternative?

crates/ark/src/modules/positron/methods.R
```diff
@@ -24,7 +24,7 @@ ark_methods_table$ark_positron_variable_get_children <- new.env(
 )
 lockEnvironment(ark_methods_table, TRUE)
 
-ark_methods_allowed_packages <- c("torch", "reticulate")
+ark_methods_allowed_packages <- c("torch", "reticulate", "duckplyr")
```
krlmlr, line 27: I wonder if we also want to bless duckdb, this is where the problematic objects ultimately come from.

```suggestion
ark_methods_allowed_packages <- c("torch", "reticulate", "duckplyr", "duckdb")
```

DavisVaughan: But duckplyr owns `duckplyr_df` and that's what we'd register methods for.

Its still WIP but duckplyr would gain a `positron.R` file that looks somewhat like this

(see also https://github.com/mlverse/torch/blob/ca61df853546f6e2248f5de718d0f9d9c7163a8c/R/positron.R#L1)

```r
ARK_REGISTER_FUNCTION <- ".ark.register_method"

positron_is_enabled_to_register <- function() {
  exists(ARK_REGISTER_FUNCTION, mode = "function")
}

register_ark_method <- function(generic, class, method) {
  register_call <- call(
    ARK_REGISTER_FUNCTION,
    quote(generic),
    quote(class),
    quote(method)
  )
  eval(register_call, envir = environment())
}

register_positron_methods <- function() {
  if (!positron_is_enabled_to_register()) {
    return()
  }

  tryCatch(
    register_positron_methods_impl(),
    error = function(err) {
      cli::cli_warn("Failed to register Positron methods.", parent = err)
    }
  )
}

register_positron_methods_impl <- function() {
  register_ark_method(
    "ark_positron_variable_display_value",
    "duckplyr_df",
    function(x, ...) {
      duckplyr_df_variable_display_value(x)
    }
  )

  register_ark_method(
    "ark_positron_variable_display_type",
    "duckplyr_df",
    function(x, ...) {
      duckplyr_df_variable_display_type(x)
    }
  )
}

duckplyr_df_variable_display_value <- function(x) {
  n_col <- df_n_col(x)

  if (n_col == 1L) {
    col_word <- "column"
  } else {
    col_word <- "columns"
  }

  sprintf("[? rows x %s %s]", n_col, col_word)
}

duckplyr_df_variable_display_type <- function(x) {
  "duckplyr_df"
}
```

and in your onLoad:

```r
.onLoad <- function(lib, pkg) {
  run_on_load()
  register_positron_methods()
}
```

These methods are registered with Ark when duckplyr is loaded, and Ark uses them when communicating with Positron about `duckplyr_df` objects to avoid the code paths where we are otherwise forced to try and compute the number of rows.

crates/harp/src/matrix.rs
```diff
@@ -0,0 +1,23 @@
+use libr::*;
+
+use crate::r_dim;
+use crate::r_int_get;
+use crate::r_length;
+use crate::utils::*;
+
+/// Compute the dimensions of a matrix
+pub fn mat_dim(x: SEXP) -> crate::Result<(i32, i32)> {
```
lionel-, line 9: Might be worth making this `Matrix::dim()` for consistency

crates/harp/src/utils.rs
```diff
@@ -393,6 +393,28 @@ pub unsafe fn r_envir_remove(symbol: &str, envir: SEXP) {
...
+}
+
+/// Get the names attribute of an object
+///
+/// Raw access to the [R_NamesSymbol] attribute. Will return [R_NilValue] when no names
+/// are present.
+pub fn r_names(x: SEXP) -> SEXP {
+    unsafe { Rf_getAttrib(x, R_NamesSymbol) }
+}
+
```
lionel-, line 417: I'm a bit worried about returning `SEXP` here. `getAttrib()` may allocate so I think we should be returning `RObject` to avoid creating a footgun. And then this suggests this should be part of a safe class, e.g. `RObject`, so we'd have `RObject::get_names()` and `RObject::get_row_names()`. The `get` prefix would distinguish from e.g. `RObject::names()` which performs a conversion to a Rust object.

Probably worth implementing around `RObject::get_attribute()`.

lionel- (approved):

PR #746, opened 2025-03-13
Follow data explorer comm changes from Positron

wesm:
I made a small change in https://github.com/posit-dev/positron/pull/6758 to return ColumnValue in the frequency table response. There are a couple other additions that this brings Ark up to speed with.


wesm:
This change does not need to wait to be merged since the changes are backward compatible

crates/ark/tests/data_explorer.rs
```diff
@@ -1856,7 +1856,10 @@ fn test_frequency_table() {
             values: format_string(
```
dfalbel, line 1856: You could use `format_column` here instead, which would also report the Special codes if necessary.

dfalbel (approved):
LGTM!

Just a minor comment

PR #747, opened 2025-03-17
Add new OpenRPC method to ask for a password

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/2890

Goes together with https://github.com/posit-dev/positron/pull/6809

### DIFF
--- a/crates/ark/src/modules/rstudio/stubs.R
+++ b/crates/ark/src/modules/rstudio/stubs.R
@@ -17,6 +17,11 @@
 }
 
 #' @export
+.rs.api.askForPassword <- function(prompt = "Please enter your password") {
+    invisible(.ps.ui.askForPassword(prompt))
+}
+
+#' @export
 .rs.api.sendToConsole <- function(
     code,
     echo = TRUE,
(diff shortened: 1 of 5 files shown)


crates/ark/src/modules/rstudio/stubs.R
```diff
@@ -16,6 +16,11 @@
     invisible(.ps.ui.navigateToFile(file, line, column))
 }
 
+#' @export
+.rs.api.askForPassword <- function(prompt = "Please enter your password") {
+    invisible(.ps.ui.askForPassword(prompt))
```
juliasilge, line 21: I put `invisible()` here to at least keep it from automatically blurting out the password into the console, but note that this does store the password as a string in R, the same as `rstudioapi::askForPassword()`: <https://rstudio.github.io/rstudioapi/reference/askForPassword.html>

DavisVaughan: I'm tempted to request that we just match the rstudioapi exactly i.e. no `invisible()` call. Even though it probably isn't ideal, it may make it seem like the call to `rstudioapi::askForPassword()` didn't actually work otherwise if you're just testing it out

DavisVaughan (approved):
See https://github.com/posit-dev/ark/pull/748 but otherwise looks good!

PR #748, opened 2025-03-17
Set `askpass` option like RStudio

DavisVaughan:
CC @juliasilge 

See `?rstudioapi::askForPassword`, RStudio sets this option so packages can do `getOption("askpass")` and have it work in a tool agnostic way


juliasilge (approved):
Thank you!

Issue #681, opened 2025-01-30
Refactor some infrastructure for completions


jennybc:
In https://github.com/posit-dev/positron/issues/1818, @DavisVaughan says:

> It's probably time to introduce a `CompletionOption` struct to contain something like `in_help: bool` and `promise_strategy: Option<PromiseStrategy>` which gets fed through all of the completion functions as an immutable reference.

In discussions around #680, we've reached consensus that it is definitely time to do this. The details will likely be somewhat different than what's sketched above and we'll discuss that here.

**Why is this needed?** Many of the low-level helpers that provide completions from a certain source (from the search path, from the workspace, from a call, and so on) actually need some syntactic and/or semantic information about the completion site.

For example, #680 addresses https://github.com/posit-dev/positron/issues/1818 and https://github.com/posit-dev/positron/issues/2338, where we want to control the automatic addition of parentheses when completing a function. If we're helping the user write a function invocation, we want parentheses. We want to leave the cursor inside these parentheses and trigger the display of parameters hints. But if we're completing the function name in a function value position (e.g. it's the subject of `debug()` or `?`), we don't want parentheses.

The solution in #680 is (fondly) described as "dumb" because a specific assessment is made about the completion site very early, in `provide_completions()` and then threaded through many, many layers of helpers as the boolean `no_trailing_parens`. It would be nice to encapsulate this whole matter inside `completion_item_from_function()`, but there are (at least) two problems with that:

1. The document context is required to analyze the completion site and is not available to helpers like `completion_item_from_function()`. Such an approach would just require threading _that_ through to lots of new places. (See #369 for a sense of how that plays out.)
2. `completion_item_from_function()` gets called once per function-that-is-in-scope, which is >2300 even in a vanilla R session. But this assessment of the completion site only needs to happen **once**, at most.

Several more completion issues are coming down the pike (https://github.com/posit-dev/positron/issues/4441, https://github.com/posit-dev/positron/issues/2574, https://github.com/posit-dev/positron/issues/2417, for starters) which are likely to play out in a similar fashion. The existing logic for determining if we are inside a pipe, finding that pipe's root node, and behaving appropriately is also related.

**TL;DR the type of solution used in #680 is not going to scale gracefully as we continue to work through the long list of completion improvements (https://github.com/posit-dev/positron/issues/1603).**

One more note from discussion of #680:

[...]

PR #754, opened 2025-03-25
Refactor completions

jennybc:
Relates to #681

The main goal is to introduce a new struct to serve as headquarters when generating completions. This gives us a home for features that we compute about the completion site, that can then be used downstream in one or multiple concrete completion sources. Two existing examples of this, each previously implemented in different ways, are:

* "parameter hints": Distinguishes whether we (think) we're completing a function invocation or a function value. See #680.
* "pipe root": If we're dealing with a pipe chain that starts with, e.g., a data frame, you might want to expose its column names as completions.

Qualities we wanted in this refactoring:

* Completion site features are lazy: computed upon first need and are not re-computed unnecessarily.
* Features and/or document don't need to be threaded through as arguments through all layers of completion sources and completion item helpers as "loose parts".
* Make the completion tooling easier to work on and debug, e.g. through more intentional logging.
* General effort to remove incidental inconsistencies across individual completion sources.

I'll make a few comments on the source as well.

(I've drafted a 2nd PR #755 that builds on this one, that proposes changes to how we manage completion items as they roll in from various sources. That's in draft form until the dust settles here.)

### DIFF
--- a/crates/ark/src/lsp/completions/builder.rs
+++ b/crates/ark/src/lsp/completions/builder.rs
@@ -24,7 +24,7 @@ pub(crate) struct CompletionBuilder<'a> {
     pub(crate) context: &'a DocumentContext<'a>,
     pub(crate) state: &'a WorldState,
     parameter_hints_cell: OnceCell<ParameterHints>,
-    pipe_root: OnceCell<Option<PipeRoot>>,
+    pipe_root_cell: OnceCell<Option<PipeRoot>>,
 }
 
 impl<'a> CompletionBuilder<'a> {
@@ -33,7 +33,7 @@ impl<'a> CompletionBuilder<'a> {
             context,
             state,
             parameter_hints_cell: OnceCell::new(),
-            pipe_root: OnceCell::new(),
+            pipe_root_cell: OnceCell::new(),
         }
     }
 
@@ -43,17 +43,15 @@ impl<'a> CompletionBuilder<'a> {
         })
     }
 
-    pub fn get_pipe_root(&self) -> Result<Option<PipeRoot>> {
-        if let Some(root) = self.pipe_root.get() {
-            // Already computed, just clone and return
+    pub fn pipe_root(&self) -> Result<Option<PipeRoot>> {
+        if let Some(root) = self.pipe_root_cell.get() {
             return Ok(root.clone());
         }
 
-        // Not yet computed, find the pipe root
         let root = find_pipe_root(self.context)?;
 
         // Cache it for future calls (ignore failure if race condition, which shouldn't happen)
-        let _ = self.pipe_root.set(root.clone());
+        let _ = self.pipe_root_cell.set(root.clone());
 
         // Return the result
         Ok(root)
--- a/crates/ark/src/lsp/completions/builder.rs
+++ b/crates/ark/src/lsp/completions/builder.rs
@@ -16,7 +16,7 @@ use crate::lsp::completions::sources::composite::pipe::completions_from_pipe;
 use crate::lsp::completions::sources::composite::search_path::SearchPathCompletionProvider;
 use crate::lsp::completions::sources::composite::snippets::completions_from_snippets;
 use crate::lsp::completions::sources::composite::subset::completions_from_subset;
-use crate::lsp::completions::sources::composite::workspace::completions_from_workspace;
+use crate::lsp::completions::sources::composite::workspace::WorkspaceCompletionProvider;
 use crate::lsp::completions::sources::unique::colon::completions_from_single_colon;
 use crate::lsp::completions::sources::unique::comment::completions_from_comment;
 use crate::lsp::completions::sources::unique::custom::completions_from_custom_source;
@@ -54,12 +54,6 @@ impl<'a> CompletionBuilder<'a> {
         // set of reasonable completions based on loaded packages, the open
         // document, the current workspace, and any call related arguments
         self.completions_from_composite_sources()
-    }
-
-    // Temporary move: wrapper methods for functions that use parameter_hints
-
-    fn get_workspace_completions(&self) -> Result<Option<Vec<CompletionItem>>> {
-        completions_from_workspace(self.context, self.state, &self.parameter_hints)
     }
 
     pub fn completions_from_unique_sources(&self) -> Result<Option<Vec<CompletionItem>>> {
--- a/crates/ark/src/lsp/completions/builder.rs
+++ b/crates/ark/src/lsp/completions/builder.rs
@@ -1,3 +1,10 @@
+//
+// builder.rs
+//
+// Copyright (C) 2025 Posit Software, PBC. All rights reserved.
+//
+//
+
 use std::cell::OnceCell;
 use std::collections::HashSet;
 
@@ -24,8 +31,9 @@ use crate::lsp::completions::sources::unique::comment::completions_from_comment;
 use crate::lsp::completions::sources::unique::custom::completions_from_custom_source;
 use crate::lsp::completions::sources::unique::extractor::completions_from_at;
 use crate::lsp::completions::sources::unique::extractor::completions_from_dollar;
-use crate::lsp::completions::sources::unique::namespace::NamespaceCompletionProvider;
+use crate::lsp::completions::sources::unique::namespace::NamespaceSource;
 use crate::lsp::completions::sources::unique::string::completions_from_string;
+use crate::lsp::completions::sources::CompletionSource;
 use crate::lsp::document_context::DocumentContext;
 use crate::lsp::state::WorldState;
 
@@ -92,7 +100,7 @@ impl<'a> CompletionBuilder<'a> {
             return Ok(Some(completions));
         }
 
-        if let Some(completions) = self.get_namespace_completions()? {
+        if let Some(completions) = NamespaceSource::provide_completions(self)? {
             return Ok(Some(completions));
         }
 
--- a/crates/ark/src/lsp/completions/builder.rs
+++ b/crates/ark/src/lsp/completions/builder.rs
@@ -1,9 +1,22 @@
+use std::collections::HashSet;
+
 use anyhow::Result;
+use stdext::*;
 use tower_lsp::lsp_types::CompletionItem;
+use tower_lsp::lsp_types::CompletionItemKind;
 
 use crate::lsp::completions::parameter_hints::ParameterHints;
 use crate::lsp::completions::parameter_hints::{self};
-use crate::lsp::completions::sources::composite::completions_from_composite_sources;
+use crate::lsp::completions::sources::composite::call::completions_from_call;
+use crate::lsp::completions::sources::composite::document::completions_from_document;
+use crate::lsp::completions::sources::composite::find_pipe_root;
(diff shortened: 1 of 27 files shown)


crates/ark/src/lsp/completions/sources/unique.rs
```diff
@@ -15,61 +15,62 @@ mod string;
...
+            &AtSource,
+        ];
 
-    // Try `package::prefix` (or `:::`) namespace completions
-    if let Some(completions) = completions_from_namespace(context, parameter_hints)? {
-        return Ok(Some(completions));
-    }
+        for source in sources {
+            let source_name = source.name();
+            log::debug!("Trying completions from source: {}", source_name);
```
jennybc, line 61: Switched to a loop here in unique completions as well.

crates/ark/src/lsp/completions/sources/unique/string.rs
```diff
@@ -115,8 +131,13 @@ mod tests {
...
 
-            // Check one level up too
-            let res = completions_from_unique_sources(&context, ParameterHints::Enabled).unwrap();
+            // Check for same result when consulting (potentially all) unique sources
+            let state = WorldState::default();
+            let builder =
+                crate::lsp::completions::builder::CompletionBuilder::new(&context, &state);
+            let unique_sources = UniqueCompletionsSource;
+            let res = unique_sources.provide_completions(&builder).unwrap();
+
```
jennybc, line 140: I think this is the only test that required more than the most superficial of tweaks.

crates/ark/src/lsp/completions/sources/composite.rs
```diff
@@ -17,116 +17,122 @@ mod workspace;
...
+            let source_name = source.name();
+            log::debug!("Trying completions from source: {}", source_name);
+
+            if let Some(mut additional_completions) = source.provide_completions(builder)? {
+                log::debug!(
+                    "Found {} completions from source: {}",
+                    additional_completions.len(),
+                    source_name
+                );
+                completions.append(&mut additional_completions);
```
lionel-, line 63: I think this could be made more idiomatic and simpler by extracting the body of these for loops into a generic function, something like:

```rust
pub fn collect_completions<S>(
    source: S,
    builder: &CompletionBuilder,
) -> anyhow::Result<Option<Vec<CompletionItem>>>
where
    S: CompletionSource,
{
    let source_name = source.name();
    log::trace!("Trying completions from source: {}", source_name);

    if let Some(completions) = source.provide_completions(builder)? {
        log::info!(
            "Found {} completions from source: {}",
            completions.len(),
            source_name
        );
        Ok(Some(completions))
    } else {
        Ok(None)
    }
}
```

Then you call it like this for unique sources:

```rust
        if let Some(completions) = collect_completions(SingleColonSource, builder)? {
            return Ok(Some(completions));
        }
```

Or like this for composite sources:

```rust
        if let Some(mut additional_completions) = collect_completions(call::CallSource, builder)? {
                completions.append(&mut additional_completions);
        }
```

(For composite sources you could also write another another helper that takes a mut ref to the completion vector to make the calls one-liner.)

The benefits of doing this:

- Use of static dispatch instead of dynamic dispatch
- Less repetition in the way things are logged
- Perhaps a personal preference: A for loop implies some sort of dynamism whereas the control flow is fully static. An actual sequence of calls better reflects what's going on.

crates/ark/src/lsp/completions/sources/composite.rs
```diff
@@ -17,116 +17,122 @@ mod workspace;
...
-        if let Some(mut additional_completions) = completions_from_document(context)? {
-            completions.append(&mut additional_completions);
-        }
-
-        if let Some(mut additional_completions) =
-            completions_from_workspace(context, state, parameter_hints)?
-        {
-            completions.append(&mut additional_completions);
-        }
+impl CompletionSource for CompositeCompletionsSource {
```
lionel-, line 35: I think it's a bit unexpected for the composite and unique source routines to implement the `CompletionSource` trait. For instance the `name()` method is never used for those if I'm not missing anything, which I think reveals an asymmetry that makes the aggregators a weird fit for the source trait.

crates/ark/src/lsp/completions/builder.rs
```diff
@@ -0,0 +1,73 @@
...
+
+        let root = find_pipe_root(self.context)?;
+
+        // Cache it for future calls (ignore failure if race condition, which shouldn't happen)
+        let _ = self.pipe_root_cell.set(root.clone());
+
+        Ok(root)
+    }
+
+    pub fn build(self) -> Result<Vec<CompletionItem>> {
```
lionel-, line 59: It feels to me like something is wrong here. `build()` calls into the aggregators who then call into the sources with the builder passed in. So a source could potentially call `build()` again, which doesn't feel right?

I think we're incorrectly forcing the builder pattern here? It's more like we only have a context that is passed around. So I'd rename to "state" or "context" and make that `build()` method a free function that is not accessible from sources?

lionel- (commented):
Nice progress! I quite like the way sources are now structs that implement a common trait. However it feels like we could improve the way things are structured by:

- Removing the `build()` method from the builder, and renaming the builder to something more like "state" or "context".
- Not having the aggregators implement the source trait
- Using a generic function to statically dispatch on sources instead of for loops with dynamic dispatch.

See comments for justifications on these thoughts. Happy to discuss these points during our casual!

crates/ark/src/lsp/completions/builder.rs
```diff
@@ -0,0 +1,74 @@
...
+use crate::lsp::completions::parameter_hints::{self};
+use crate::lsp::completions::sources::composite::find_pipe_root;
+use crate::lsp::completions::sources::composite::pipe::PipeRoot;
+use crate::lsp::completions::sources::composite::CompositeCompletionsSource;
+use crate::lsp::completions::sources::unique::UniqueCompletionsSource;
+use crate::lsp::completions::sources::CompletionSource;
+use crate::lsp::document_context::DocumentContext;
+use crate::lsp::state::WorldState;
+
+pub(crate) struct CompletionBuilder<'a> {
```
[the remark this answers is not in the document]
jennybc: Discussed on slack. `CompletionBuilder` is now `CompletionContext`, which motivated other name changes related to the document context.

crates/ark/src/lsp/completions/sources/composite.rs
```diff
@@ -17,119 +17,122 @@ mod workspace;
...
-            // names. Particularly useful for `dev|` to get to `devtools::`,
-            // as that has a lot of base R functions with similar names.
-            item.kind == Some(CompletionItemKind::MODULE) => {
-                item.sort_text = Some(join!["3-", sort_text]);
+        // Call, pipe, and subset completions should show up no matter what when
+        // the user requests completions (this allows them to Tab their way through
+        // completions effectively without typing anything). For the rest of the
+        // general completions, we require an identifier to begin showing
+        // anything.
+        if is_identifier_like(builder.context.node) {
```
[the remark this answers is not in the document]
jennybc: We discussed this and it has pros and cons. Decided to leave alone.

crates/ark/src/lsp/completions/sources/composite/call.rs
```diff
@@ -16,21 +16,37 @@ use tower_lsp::lsp_types::CompletionItem;
...
+    }
+
+    fn provide_completions(
+        &self,
+        builder: &CompletionBuilder,
+    ) -> Result<Option<Vec<CompletionItem>>> {
+        let root = builder.pipe_root()?;
+        completions_from_call(builder.context, root)
+    }
+}
```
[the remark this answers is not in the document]
jennybc: We've decided to not inline the logic from, e.g. `completions_from_call()`, into, e.g., the `provide_completions()` method for `CallSource`, because seems like lots of churn for little payoff, especially in tests. But we could do it later, in its own PR.

crates/ark/src/lsp/completions/sources/unique.rs
jennybc: Removed these loops because they come with dynamic dispatch, which we'd rather avoid.

crates/ark/src/lsp/completions/sources/unique/colon.rs
```diff
@@ -1,25 +1,43 @@
 //
 // colon.rs
 //
-// Copyright (C) 2023 Posit Software, PBC. All rights reserved.
+// Copyright (C) 2023-2025 Posit Software, PBC. All rights reserved.
 //
 //
 
+use anyhow::Result;
```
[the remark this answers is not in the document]
jennybc: Whew 😩 a bunch of these were here before I came along (and some exist elsewhere in the repo), but I've now eliminated `use anyhow::Result;` from the neighborhood I'm working in (`crates/ark/src/lsp/completions/`).

crates/ark/src/lsp/completions/sources/composite.rs
jennybc: Done! Except the last suggestion re: additional helper for composite sources. Those matters will be handled in the PR that stacks on this one.

crates/ark/src/lsp/completions/sources.rs
```diff
@@ -1,14 +1,27 @@
...
 
-pub use composite::completions_from_composite_sources;
-pub use unique::completions_from_unique_sources;
+use anyhow::Result;
+use tower_lsp::lsp_types::CompletionItem;
+
+use crate::lsp::completions::builder::CompletionBuilder;
+
+/// Interface for any source we consult for completions
+pub trait CompletionSource {
```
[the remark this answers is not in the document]
jennybc: Backed out of its use for the aggregators.

crates/ark/src/lsp/completions/sources/composite.rs
jennybc: Done.

crates/ark/src/lsp/completions/builder.rs
jennybc: Done.

crates/ark/src/lsp/completions/sources.rs
```diff
@@ -1,14 +1,62 @@
...
+            completions.len(),
+            source_name
+        );
+        Ok(Some(completions))
+    } else {
+        Ok(None)
+    }
+}
+
+pub fn collect_and_append_completions<S>(
```
lionel-, line 50: `push_completions()`?

lionel- (approved):
Looks great!

crates/ark/src/lsp/completions/builder.rs
```diff
@@ -0,0 +1,73 @@
...
+
+    pub fn pipe_root(&self) -> Result<Option<PipeRoot>> {
+        if let Some(root) = self.pipe_root_cell.get() {
+            return Ok(root.clone());
+        }
+
+        let root = find_pipe_root(self.context)?;
+
+        // Cache it for future calls (ignore failure if race condition, which shouldn't happen)
+        let _ = self.pipe_root_cell.set(root.clone());
```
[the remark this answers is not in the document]
jennybc: I mostly implemented this and certainly think I've gotten at the spirit of it. If you have concerns with how I handled it, we can refine.

PR #755, opened 2025-03-25
More completion refactoring around how completions are marshalled

### LINKED ISSUE #681: Refactor some infrastructure for completions

jennybc:
Builds on #754 and closes #681. The second and final of piece of that refactoring effort.

This is about managing the accumulation of completion items from composite sources.

Main ideas:

* Instead of adding CompletionItems onto an ever-growing collection, manage the accumulation in a map of completion items + their ~source(s)~. *Update: we've decided to record the first-encountered source. If a completion item shows up again, we log this event, so we have a thread to pull on, if we choose to investigate this.*
* Update the map as each new completion source is consulted. Novel items are added. Pre-existing items ~get an update (addition) to their completion source~ *that are re-contributed get logged.*
* No more need for explicit deduplication.
* Keep adding / modifying the logging for future debugging happiness

### DIFF
--- a/crates/ark/src/lsp/completions/provide.rs
+++ b/crates/ark/src/lsp/completions/provide.rs
@@ -21,14 +21,11 @@ pub(crate) fn provide_completions(
     document_context: &DocumentContext,
     state: &WorldState,
 ) -> anyhow::Result<Vec<CompletionItem>> {
-    let node = document_context.node;
-    let node_text = node_text(&node, &document_context.document.contents).unwrap_or_default();
-    let node_type = format!("{:?}", node.node_type());
-
     log::info!(
-        "provide_completions() - Completion node text: '{}', Node type: '{}'",
-        node_text,
-        node_type
+        "provide_completions() - Completion node text: '{node_text}', Node type: '{node_type:?}'",
+        node_text = node_text(&document_context.node, &document_context.document.contents)
+            .unwrap_or_default(),
+        node_type = document_context.node.node_type()
     );
 
     let completion_context = CompletionContext::new(document_context, state);
--- a/crates/ark/src/lsp/completions/provide.rs
+++ b/crates/ark/src/lsp/completions/provide.rs
@@ -12,6 +12,8 @@ use crate::lsp::completions::sources::composite;
 use crate::lsp::completions::sources::unique;
 use crate::lsp::document_context::DocumentContext;
 use crate::lsp::state::WorldState;
+use crate::treesitter::node_text;
+use crate::treesitter::NodeTypeExt;
 
 // Entry point for completions.
 // Must be within an `r_task()`.
@@ -19,7 +21,15 @@ pub(crate) fn provide_completions(
     document_context: &DocumentContext,
     state: &WorldState,
 ) -> anyhow::Result<Vec<CompletionItem>> {
-    log::info!("provide_completions()");
+    let node = document_context.node;
+    let node_text = node_text(&node, &document_context.document.contents).unwrap_or_default();
+    let node_type = format!("{:?}", node.node_type());
+
+    log::info!(
+        "provide_completions() - Completion node text: '{}', Node type: '{}'",
+        node_text,
+        node_type
+    );
 
     let completion_context = CompletionContext::new(document_context, state);
 
--- a/crates/ark/src/lsp/completions/sources/composite.rs
+++ b/crates/ark/src/lsp/completions/sources/composite.rs
@@ -151,18 +151,6 @@ fn finalize_completions(
         .values()
         .map(|sourced_item| {
             let mut item = sourced_item.item.clone();
-
-            // Store source information in the `data` field as a simple string
-            if !sourced_item.sources.is_empty() {
-                let sources_string = sourced_item.sources.join(",");
-                item.data = Some(serde_json::Value::String(sources_string.clone()));
-                log::debug!(
-                    "Completion '{}' has sources: {}",
-                    item.label,
-                    sources_string
-                );
-            }
-
             item
         })
         .collect();
--- a/crates/ark/src/lsp/completions/sources/composite.rs
+++ b/crates/ark/src/lsp/completions/sources/composite.rs
@@ -105,7 +105,7 @@ where
     if let Some(source_completions) = collect_completions(source, completion_context)? {
         for item in source_completions {
             if let Some(existing) = completions.get(&item.label) {
-                log::debug!(
+                log::trace!(
                     "Completion with label '{}' already exists (first contributed by source: {}, now also from: {})",
                     item.label,
                     existing.source,
--- a/crates/ark/src/lsp/completions/sources/composite.rs
+++ b/crates/ark/src/lsp/completions/sources/composite.rs
@@ -40,7 +40,7 @@ pub(crate) fn get_completions(
 ) -> anyhow::Result<Option<Vec<CompletionItem>>> {
     log::info!("Getting completions from composite sources");
 
-    let mut completions: HashMap<String, CompletionItemWithSource> = HashMap::new();
+    let mut completions = HashMap::new();
 
     // Call, pipe, and subset completions should show up no matter what when
     // the user requests completions. This allows them to "tab" their way
--- a/crates/ark/src/lsp/completions/sources/composite.rs
+++ b/crates/ark/src/lsp/completions/sources/composite.rs
@@ -58,7 +58,6 @@ pub(crate) fn get_completions(
     // For the rest of the general completions, we require an identifier to
     // begin showing anything.
     if is_identifier_like(completion_context.document_context.node) {
-        // Consulted settings.json
         push_completions(keyword::KeywordSource, completion_context, &mut completions)?;
 
         push_completions(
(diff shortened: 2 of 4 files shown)


crates/ark/src/lsp/completions/provide.rs
```diff
@@ -12,14 +12,24 @@ use crate::lsp::completions::sources::composite;
...
+use crate::treesitter::NodeTypeExt;
 
 // Entry point for completions.
 // Must be within an `r_task()`.
 pub(crate) fn provide_completions(
     document_context: &DocumentContext,
     state: &WorldState,
 ) -> anyhow::Result<Vec<CompletionItem>> {
-    log::info!("provide_completions()");
+    let node = document_context.node;
```
jennybc, line 24: Here's how this looks (shows up in R 4.x.y: Kernel output channel):

```
[R]   2025-04-02T02:25:00.206078Z  INFO  provide_completions() - Completion node text: 'enc', Node type: 'Identifier'
[R]     at crates/ark/src/lsp/completions/provide.rs:28
[R] 
[R]   2025-04-02T02:25:00.206120Z  INFO  Getting completions from unique sources
[R]     at crates/ark/src/lsp/completions/sources/unique.rs:33
[R] 
[R]   2025-04-02T02:25:00.206134Z  INFO  Trying completions from source: single_colon
[R]     at crates/ark/src/lsp/completions/sources.rs:36

...
```

crates/ark/src/lsp/completions/sources/composite.rs
```diff
@@ -78,16 +101,81 @@ pub(crate) fn get_completions(
...
+    let mut items: Vec<CompletionItem> = completion_map
+        .values()
+        .map(|sourced_item| {
+            let mut item = sourced_item.item.clone();
+
+            // Store source information in the `data` field as a simple string
+            if !sourced_item.sources.is_empty() {
+                let sources_string = sourced_item.sources.join(",");
+                item.data = Some(serde_json::Value::String(sources_string.clone()));
+                log::debug!(
```
DavisVaughan, line 159:
```suggestion
                log::trace!(
```

we tend to avoid debug logs and go straight to trace

crates/ark/src/lsp/completions/sources/composite.rs
```diff
@@ -14,25 +14,48 @@ mod snippets;
...
+use crate::lsp::completions::sources::CompletionSource;
 use crate::treesitter::NodeType;
 use crate::treesitter::NodeTypeExt;
 
+// Locally useful data structures for tracking completions and their source(s)
+#[derive(Hash, Eq, PartialEq, Clone)]
+struct CompletionKey {
+    label: String,
+    insert_text: Option<String>,
+}
```
DavisVaughan, line 35: Previously the key was just `label`. Now its `label` and the `insert_text`. I wonder if this can cause duplication. In theory you can have the same `label` with two different `insert_text` and wed send both to the frontend, and i doubt it deduplicates those on that side. That doesn't feel quite right to me

crates/ark/src/lsp/completions/sources/composite.rs
```diff
@@ -78,16 +101,81 @@ pub(crate) fn get_completions(
...
+    completion_map: &mut HashMap<CompletionKey, CompletionWithSource>,
+) -> anyhow::Result<()>
+where
+    S: CompletionSource,
+{
+    let source_name = source.name();
+
+    if let Some(source_completions) = collect_completions(source, completion_context)? {
+        for item in source_completions {
+            let key = CompletionKey::from(&item);
```
DavisVaughan, line 122: This doesn't feel right because the `from()` method here is basically just cloning the underlying values so we end up with them in two places - one in `CompletionKey` and one in `CompletionItem`

Ah, maybe a better idea is to implement `Hash` for `CompletionWithSource` manually in such a way that it hashes the `label` and maybe the `insert_text` if we still think that is right, and just use a `HashSet` again.

jennybc:
This PR is updated based on this morning's discussion and I'm quite happy where it ended up. It feels pretty clean and efficient to me.

I did experiment with the `HashSet` approach, but it ended up making the code more complex, caused more cloning, and entailed an expensive search. You have implement `Hash`, `PartialEq`, etc. for `CompletionItemWithSource` and you have to create a temporary instance just for lookup. The lookup itself is also more expensive in the `HashSet` scenario. The ownership transfer of completion items feels pretty good to me now: Source completions → `HashMap` (as values) → final `Vec<CompletionItem>`.

crates/ark/src/lsp/completions/sources/composite.rs
```diff
@@ -14,25 +14,33 @@ mod snippets;
...
+}
+
 /// Gets completions from all composite sources, with deduplication and sorting
 pub(crate) fn get_completions(
     completion_context: &CompletionContext,
 ) -> anyhow::Result<Option<Vec<CompletionItem>>> {
     log::info!("Getting completions from composite sources");
 
-    let mut completions: Vec<CompletionItem> = vec![];
+    let mut completions: HashMap<String, CompletionItemWithSource> = HashMap::new();
```
DavisVaughan, line 43:
```suggestion
    let mut completions = HashMap::new();
```

I bet the compiler "just knows"

crates/ark/src/lsp/completions/sources/composite.rs
```diff
@@ -78,16 +86,64 @@ pub(crate) fn get_completions(
...
+) -> anyhow::Result<()>
+where
+    S: CompletionSource,
+{
+    let source_name = source.name();
+
+    if let Some(source_completions) = collect_completions(source, completion_context)? {
+        for item in source_completions {
+            if let Some(existing) = completions.get(&item.label) {
+                log::debug!(
```
DavisVaughan, line 108:
```suggestion
                log::trace!(
```

We almost unconditionally skip straight to trace! level

crates/ark/src/lsp/completions/sources/composite.rs
```diff
@@ -78,16 +86,64 @@ pub(crate) fn get_completions(
...
+
+    Ok(())
+}
+
+/// Produce plain old CompletionItems and sort them
+fn finalize_completions(
+    completions: HashMap<String, CompletionItemWithSource>,
+) -> Vec<CompletionItem> {
+    let mut items: Vec<CompletionItem> = completions
+        .into_values()
```
DavisVaughan, line 131: Nice - yay for ownership movement!

DavisVaughan (approved):
Nice job!

crates/ark/src/lsp/completions/sources/composite.rs
jennybc: Sorry this keeps happening intermittently! I do know about the preference for `log::trace!`. Sometimes it's me and sometimes I think it's co-pilot (and me not catching it).

PR #759, opened 2025-04-01
Add a method to query a unique session id

dfalbel:
This is part of https://github.com/posit-dev/positron/pull/7024
Each R session can only ever host a single reticulate session. With this PR we add a `reticulate_id()` method that allows us to link a Positron `sessionId` to the ark session it's running. This is necessary specially when a session is opened from Positron directly, (not calling `reticulate::repl_python()` ) - which currently doesn't trigger the reticulate service to be started.

### DIFF
--- a/crates/ark/src/reticulate.rs
+++ b/crates/ark/src/reticulate.rs
@@ -23,7 +23,7 @@ use crate::interface::RMain;
 static RETICULATE_OUTGOING_TX: LazyLock<Mutex<Option<Sender<CommMsg>>>> =
     LazyLock::new(|| Mutex::new(None));
 
-// Each ark session has it's unique session id which is used by Positron's front-end
+// Each ark session has a unique session id which is used by the Positron frontend
 // to associate a client id with a session.
 static RETICULATE_ID: Lazy<String> =
     Lazy::new(|| format!("reticulate-{}", Uuid::new_v4().to_string()));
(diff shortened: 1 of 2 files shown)


crates/ark/src/reticulate.rs
```diff
@@ -11,6 +11,7 @@ use harp::utils::r_is_null;
 use harp::RObject;
 use libr::R_NilValue;
 use libr::SEXP;
+use once_cell::sync::Lazy;
```
lionel-, line 14: This should now be `LazyCell` from the std lib

crates/ark/src/reticulate.rs
```diff
@@ -53,7 +59,8 @@ impl ReticulateService {
         let event = CommManagerEvent::Opened(
             service.comm.clone(),
             json!({
-                "input": input
+                "input": input,
+                "reticulate_id": (*RETICULATE_ID).clone(),
             }),
         );
```
lionel-, line 65: I'm guessing the comm is always opened from the backend, never the frontend?

lionel- (approved):

crates/ark/src/reticulate.rs
dfalbel: Yes, that's correct. We only open the comm from the backend when the user calls 'reticulate::repl_python()`.

crates/ark/src/reticulate.rs
dfalbel: I had to use `LazyLock` instead (https://github.com/posit-dev/ark/pull/759/commits/a616063a32b215fac5d079076892f344c0f96db2) because `LazyCell` is not safe to be shared as a global.

PR #767, opened 2025-04-08
Stop excluding `base::return()` from completions

jennybc:
I'm opening Completion Month with a real banger! Let's stop explicitly excluding `base::return()` from completions.

Relates to [#4842](https://github.com/posit-dev/positron/issues/4842) (the `function` part requires more thought -- will be separate PR)

I've

* Thought about this 🤔 
* Looked through relevant comments and `git blame` in ark
* Looked at what RStudio does

and I can't find any reason to _not_ just let `base::return()` show up in our search path completions.


jennybc:
I renamed this branch locally, so re-making the PR from that branch.

PR #768, opened 2025-04-08
Stop excluding base::return() and base::function from completions

jennybc:
I'm opening Completion Month with a real banger! Let's stop explicitly excluding `base::return()` and `base::function` from completions.

Relates to https://github.com/posit-dev/positron/issues/4842

I've

* Thought about this 🤔 
* Looked through relevant comments and `git blame` in ark
* Looked at what RStudio does

and I can't find any reason to _not_ just let `base::return()` and `base::function` show up in our completions. They will both be contributed by our search path completions, coming from `base`.

Both are represented by snippets, which I have some mixed feelings about. So I plan to discuss some additional improvements relating to the snippet treatment of `function` and `return()`. We could also discuss https://github.com/posit-dev/positron/issues/1850 at the same time. But perhaps this simple fix can happen first, especially for `return()`.

### DIFF
--- a/crates/ark/src/lsp/completions/sources/composite/search_path.rs
+++ b/crates/ark/src/lsp/completions/sources/composite/search_path.rs
@@ -53,7 +53,7 @@ fn completions_from_search_path(
     let mut completions = vec![];
 
     const R_CONTROL_FLOW_KEYWORDS: &[&str] = &[
-        "if", "else", "for", "in", "while", "repeat", "break", "next", "return", "function",
+        "if", "else", "for", "in", "while", "repeat", "break", "next", "function",
     ];
 
     unsafe {
--- a/crates/ark/src/lsp/completions/sources/composite/search_path.rs
+++ b/crates/ark/src/lsp/completions/sources/composite/search_path.rs
@@ -53,7 +53,7 @@ fn completions_from_search_path(
     let mut completions = vec![];
 
     const R_CONTROL_FLOW_KEYWORDS: &[&str] = &[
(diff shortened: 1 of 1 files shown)


crates/ark/src/lsp/completions/sources/composite/search_path.rs
```diff
@@ -53,7 +53,7 @@ fn completions_from_search_path(
     let mut completions = vec![];
 
     const R_CONTROL_FLOW_KEYWORDS: &[&str] = &[
-        "if", "else", "for", "in", "while", "repeat", "break", "next", "return", "function",
+        "if", "else", "for", "in", "while", "repeat", "break", "next",
```
lionel-, line 56: Do we need to exclude any of these actually? The shorter ones won't be triggered if they are below the character limit for completions but it presumably wouldn't hurt to include them.

lionel- (approved):

DavisVaughan (approved):

kevinushey:
Does Ark provide the appropriate escaping so that the completion from `base::` gives `` base::`function` ``?

jennybc:
@kevinushey No. That is already on my radar and will appear soon in its own issue or added to an existing issue. I definitely bumped into this while playing around for this PR.

crates/ark/src/lsp/completions/sources/composite/search_path.rs
jennybc: As we discussed in person, I suspect it's not good that we exclude these at all. I will open a new issue shortly with a systematic look at how reserved words currently fare in positron's R completions. And that will lead to some new proposals.

PR #772, opened 2025-04-10
Be willing to complete in the face of emptiness

### LINKED ISSUE #770: In a scenario where ark should provide all completions, it provides none
The most basic gesture to get completions in VS Code / Positron is Ctrl + Space. From the [IntelliSense docs](https://code.visualstudio.com/docs/editing/intellisense#_intellisense-features):

> You can trigger IntelliSense in any editor window by typing ⌃Space or by typing a trigger character ...

(BTW this is invoking the command `editor.action.triggerSuggest`.)

Now, more often, you type a few characters and *then* you get completions.
Either automatically after 3 characters or because you request with tab or because you trigger with a character, like `$` or `@`.

But what if you want to trigger completions *before* you've typed anything? You can use
Ctrl + Space to do it.

This should reveal "all completions", for some definition of "all".
But currently ark returns ZERO completions in this scenario.
This causes VS Code / Positron to fall back to this weird language agnostic word based completion list (quote is from Intellisense docs):

> VS Code supports word-based completions for any programming language but can also be configured to have richer IntelliSense by installing a language extension.

Empirically, this list seems to consist of tokens parsed out of ?all the files you have open in the current window?. Something like that.

Ways to experience this:

1. Ctrl + Space at the R prompt in the Console

    Image

1. Ctrl + Space in an empty R file (looks same as above)
1. Ctrl + Space on an empty line of a non-empty R file

    Image

Why does this happen?
Because the nominal completion node when the user hasn't typed anything yet does not meet our criteria for providing completions.
Mostly because our logic doesn't account for the possibility that there really is no completion node and, instead, always latches on to something "nearby", in `find_closest_node_to_point()`:

https://github.com/posit-dev/ark/blob/4074f5112b0868b1d6c14a8a42a16f0927ecf6ba/crates/ark/src/lsp/document_context.rs#L28

We then fail to pass the `is_identifier_like()` test when getting composite completions, which means we never consult keywords, snippets, the search path, the current document, or the current workspace.

https://github.com/posit-dev/ark/blob/4074f5112b0868b1d6c14a8a42a16f0927ecf6ba/crates/ark/src/lsp/completions/sources/composite.rs#L58-L60

And thus ark returns zero completions and the fallback list kicks in.
In the log, you can see the completion search just ends abruptly.

```
[R]   2025-04-10T15:47:27.133734Z  INFO  Getting completions from composite sources
[R]   2025-04-10T15:47:27.133751Z  INFO  Trying completions from source: call
[R]   2025-04-10T15:47:27.133796Z  INFO  Trying completions from source: pipe
[R]   2025-04-10T15:47:27.133816Z  INFO  Trying completions from source: subset
```

This happens in two different ways:

* The document really is empty, so the node type is 'Program' and the node text is the empty string. Cases 1 and 2 in the list above. How it looks in the log:

      [R]   2025-04-10T15:45:52.401668Z  INFO  provide_completions() - Completion node text: '', Node type: 'Program'

* The document is not empty, so `ast.root_node().find_closest_node_to_point(point)` latches on to a bit of code elsewhere in the document. The node text varies, of course, but generally it fails the `is_identifier_like()` test. Case 3 in the list above. Examples of how this looks in the log:

      [R]   2025-04-10T15:47:27.133082Z  INFO  provide_completions() - Completion node text: '10', Node type: 'Float'

      [R]   2025-04-10T16:40:39.468395Z  INFO  provide_completions() - Completion node text: ')', Node type: 'Anonymous(")")'

      [R]   2025-04-10T17:19:24.813969Z  INFO  provide_completions() - Completion node text: '
      [R]
      [R] toupper(month.abb)
      [R] ', Node type: 'Program'

Thoughts about how to solve this:

* Add the empty string or the 'Program' node as a special case that satisfies `is_identifier_like()`. See PR #772 for this. This feels like a bandaid. But it is a really good bandaid that fixes one route into this problem.
* Recognize it's possible that the completion node is actually undefined and account for this when constructing and processing a `DocumentContext`. This is a bigger undertaking.

I need to discuss with others before moving forward.

This analysis has brought up some other ideas:

* Store the the completion node's kind and text in the `CompletionContext`, alongside features like parameters hints and pipe root. You end up wanting to know these facts in various places downstream, for logging and branching, and it's a bit of a PITA to inline that repeatedly.
* Maybe we really should make "is identifier like" into another feature that we store in the completion context and can branch on. For example, I think we can probably skip all unique sources if that is false. Then, for composite sources, we could push that check down into the implementation of each composite source. We've discussed this before and already thought this might be a good idea.

jennybc:
Fixes #770

See the last "move" in https://github.com/posit-dev/ark/pull/805#discussion_r2089689821 for a concrete illustration of what this PR does. If the client asks for completions in any context where we aren't inside a node (i.e. we're just in the 'Program' node), we should basically send all composite completions, instead of no completions.

I've verified that the new tests fail on `main` and `bugfix/nearest-enclosing-node` for the reasons I expect.

### DIFF
--- a/crates/ark/src/lsp/completions/sources/composite.rs
+++ b/crates/ark/src/lsp/completions/sources/composite.rs
@@ -210,9 +210,12 @@ fn is_identifier_like(x: Node) -> bool {
 #[cfg(test)]
 mod tests {
     use crate::fixtures::point_from_cursor;
+    use crate::lsp::completions::completion_context::CompletionContext;
+    use crate::lsp::completions::sources::composite::get_completions;
     use crate::lsp::completions::sources::composite::is_identifier_like;
     use crate::lsp::document_context::DocumentContext;
     use crate::lsp::documents::Document;
+    use crate::lsp::state::WorldState;
     use crate::r_task;
     use crate::treesitter::NodeType;
     use crate::treesitter::NodeTypeExt;
@@ -235,5 +238,40 @@ mod tests {
                 );
             }
         })
+    }
+
+    #[test]
+    fn test_get_completions_on_empty_document() {
+        r_task(|| {
+            let (text, point) = point_from_cursor("@");
+            let document = Document::new(text.as_str(), None);
+            let document_context = DocumentContext::new(&document, point, None);
+            let state = WorldState::default();
+            let context = CompletionContext::new(&document_context, &state);
+
+            assert!(context.document_context.node.is_program());
+
+            let completions = get_completions(&context).unwrap();
+            assert!(completions.is_some());
+            assert!(!completions.unwrap().is_empty());
+        });
+    }
+
+    #[test]
+    fn test_get_completions_on_empty_line_in_non_empty_document() {
+        r_task(|| {
+            let code = "x <- 1:3\n@\nrnorm(3)";
+            let (text, point) = point_from_cursor(code);
+            let document = Document::new(text.as_str(), None);
+            let document_context = DocumentContext::new(&document, point, None);
+            let state = WorldState::default();
+            let context = CompletionContext::new(&document_context, &state);
+
+            assert!(context.document_context.node.is_program());
+
+            let completions = get_completions(&context).unwrap();
+            assert!(completions.is_some());
+            assert!(!completions.unwrap().is_empty());
+        });
     }
 }
--- a/crates/ark/src/lsp/completions/sources/composite.rs
+++ b/crates/ark/src/lsp/completions/sources/composite.rs
@@ -71,9 +71,12 @@ pub(crate) fn get_completions(
     // subset completions (`[` or `[[`)
     push_completions(subset::SubsetSource, completion_context, &mut completions)?;
 
-    // For the rest of the general completions, we require an identifier to
-    // begin showing anything.
-    if is_identifier_like(completion_context.document_context.node) {
+    // To offer the rest of the general completions, we should be completing:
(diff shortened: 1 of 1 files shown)


crates/ark/src/lsp/completions/sources/composite.rs
```diff
@@ -201,6 +201,20 @@ fn is_identifier_like(x: Node) -> bool {
...
+    //
+    // Gesture-wise, a Positron user could do this with Ctrl + Space, which
+    // invokes the command editor.action.triggerSuggest.
+    //
+    // The nominal completion node in these cases is just the root or 'Program'
+    // node of the AST. In this case, we should just provide "all" completions,
+    // for some reasonable definition of "all".
+    if x.node_type() == NodeType::Program {
+        return true;
+    }
```
DavisVaughan, line 216: I'm fine with the idea in general, but I don't think this belongs in `is_identifier_like()`, because the program node itself is not identifier-like.

I think at the call site it should be `is_identifier_like(node) | is_empty_line(node)` and `is_empty_line()` should contain this code.

The name `is_empty_line` is a WIP, if you have anything better to call this, feel free to use that.

DavisVaughan (approved):

crates/ark/src/lsp/completions/sources/composite.rs
jennybc: Yes, this is a definitely an improvement on the expressive-ness front. I'll do it.

PR #783, opened 2025-04-22
Register our console prompt listener before sending `CommManagerEvent::Opened`

### LINKED ISSUE #781: Flaky data explorer test
I see this intermittently locally while working on other things and running tests:

```        FAIL [  10.393s] ark::data_explorer test_live_updates
──── STDOUT:             ark::data_explorer test_live_updates

running 1 test
test test_live_updates ... FAILED

failures:

failures:
    test_live_updates

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 20 filtered out; finished in 10.36s

──── STDERR:             ark::data_explorer test_live_updates

thread 'test_live_updates' panicked at crates/ark/tests/data_explorer.rs:1081:65:
called `Result::unwrap()` on an `Err` value: Timeout

  Cancelling due to test failure
────────────
     Summary [  16.164s] 401 tests run: 400 passed, 1 failed, 0 skipped
        FAIL [  10.393s] ark::data_explorer test_live_updates
error: test run failed
```

The implicated line:
https://github.com/posit-dev/ark/blob/274c5a72541dba9ce9fb2cbdf616cc8e9c813029/crates/ark/tests/data_explorer.rs#L1081

DavisVaughan:
Closes #781 - see my comment there for full analysis

To try and avoid any potential race conditions here where a critical prompt update may come in right after we send out `Opened` but before we've registered a listener, resulting in us missing the prompt update.

I'm going to optimistically say this closes that issue, and we can revisit if it pops up again.

This feels like it would be a super rare race condition, but there are atomics involved, so who knows.

### DIFF
--- a/crates/ark/src/data_explorer/r_data_explorer.rs
+++ b/crates/ark/src/data_explorer/r_data_explorer.rs
@@ -231,6 +231,16 @@ impl RDataExplorer {
     }
 
     pub fn execution_thread(mut self) {
+        // Register a handler for console prompt events. We do this before sending
+        // `CommManagerEvent::Opened` to ensure we can't miss any console prompt events
+        // (#781).
+        let (prompt_signal_tx, prompt_signal_rx) = unbounded::<()>();
+        let listen_id = EVENTS.console_prompt.listen({
+            move |_| {
+                prompt_signal_tx.send(()).unwrap();
+            }
+        });
+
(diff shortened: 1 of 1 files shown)


dfalbel (approved):
LGTM!

Makes sense that this could be the race condition that's happening in the test.

PR #793, opened 2025-05-06
Implement SetMembership row filter and unit tests for data explorer

wesm:
Part of posit-dev/positron#2772.

### DIFF
--- a/crates/ark/src/modules/positron/r_data_explorer.R
+++ b/crates/ark/src/modules/positron/r_data_explorer.R
@@ -283,13 +283,12 @@ col_filter_indices <- function(col, idx = NULL) {
     # If inclusive is TRUE, include values in the set
     # If inclusive is FALSE, exclude values in the set
 
-    # Convert to character for comparison if needed
-    if (!is.character(col)) {
-        col <- as.character(col)
+    # Coerce values to numeric if the column is numeric
+    values <- if (is.numeric(col)) {
+        as.numeric(params$values)
+    } else {
+        params$values
     }
-
-    # Convert filter values to character for consistency
-    values <- as.character(params$values)
 
     # Return a logical vector indicating which elements match the filter
     if (params$inclusive) {
--- a/crates/ark/src/modules/positron/r_data_explorer.R
+++ b/crates/ark/src/modules/positron/r_data_explorer.R
@@ -278,6 +278,27 @@ col_filter_indices <- function(col, idx = NULL) {
     !.ps.filter_col.between(col, params)
 }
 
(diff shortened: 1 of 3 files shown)


crates/ark/src/modules/positron/r_data_explorer.R
```diff
@@ -278,6 +278,27 @@ col_filter_indices <- function(col, idx = NULL) {
...
+    # If inclusive is TRUE, include values in the set
+    # If inclusive is FALSE, exclude values in the set
+
+    # Convert to character for comparison if needed
+    if (!is.character(col)) {
+        col <- as.character(col)
+    }
+
+    # Convert filter values to character for consistency
+    values <- as.character(params$values)
```
DavisVaughan, line 292: I have a feeling this should be more like `.ps.filter_col.compare` and `.ps.filter_col.between` where the `params$values` are coerced to numeric if the `col` is numeric. `%in%` works with numeric types.

crates/ark/tests/data_explorer.rs
```diff
@@ -1686,6 +1686,105 @@ fn test_update_data_filters_reapplied() {
...
+    // Apply an inclusive set membership filter (include apple, banana, pear)
+    let inclusive_filter = RowFilter {
+        column_schema: schema.columns[0].clone(),
+        filter_type: RowFilterType::SetMembership,
+        filter_id: "3F984747-4667-40CB-9013-AA659AE37F1A".to_string(),
+        condition: RowFilterCondition::And,
+        is_valid: None,
+        params: Some(RowFilterParams::SetMembership(
+            amalthea::comm::data_explorer_comm::FilterSetMembership {
+                values: vec![
```
DavisVaughan, line 1732: Is it possible to represent `NA` in `values`? I have a feeling it is probably not possible, but one interesting thing about `%in%` is that `NA` is treated as a match, while with `==` it is not.

```
> NA_character_ == NA_character_
[1] NA
> NA_character_ %in% c("x", NA_character_)
[1] TRUE
```

DavisVaughan (approved):

crates/ark/src/modules/positron/r_data_explorer.R
wesm: That's right — I'll work on adding some tests for this

crates/ark/tests/data_explorer.rs
wesm: Currently we have a deficiency in the filter RPC protocol in that it isn't possible to distinguish between the value `NA`, so we're going to need to change that to use the same special value codes that we use to make that distinction when populating the data grid

PR #794, opened 2025-05-08
Add option to show `.Last.value` in the Variables pane

jmcphers:
Quick airplane feature to optionally show `.Last.value` in Positron's Variables pane when the option `positron.show_last_value` is set.

Note that the option can be turned on during a session, but turning it off requires a restart. It's likely most users who want this will set it in their `.Rprofile`; we could also add a Positron setting for this.

Partially addresses https://github.com/posit-dev/positron/issues/3034.

### DIFF
--- a/crates/ark/src/variables/r_variables.rs
+++ b/crates/ark/src/variables/r_variables.rs
@@ -433,7 +433,7 @@ impl RVariables {
                 Err(err) => {
                     // This isn't a critical error but would also be very
                     // unexpected.
-                    log::debug!("Environment: Could not evaluate .Last.value ({err:?})");
+                    log::error!("Environment: Could not evaluate .Last.value ({err:?})");
                     None
                 },
             }
--- a/crates/ark/src/variables/r_variables.rs
+++ b/crates/ark/src/variables/r_variables.rs
@@ -45,11 +45,13 @@ use crate::variables::variable::PositronVariable;
 
 /// Enumeration of treatments for the .Last.value variable
 pub enum LastValue {
-    /// Show the .Last.value variable in the Variables pane
-    Show,
+    /// Always show the .Last.value variable in the Variables pane. This is used
+    /// by tests to show the value without changing the global option.
+    Always,
 
-    /// Hide it
-    Hide,
+    /// Use the value of the global option `positron.show_last_value` to
+    /// determine whether to show the .Last.value variable
+    UseOption,
 }
 
 /**
@@ -80,6 +82,10 @@ pub struct RVariables {
     /// Whether to always show the .Last.value in the Variables pane, regardless
     /// of the value of positron.show_last_value
     show_last_value: LastValue,
+
+    /// Whether we are currently showing the .Last.value variable in the Variables
+    /// pane.
+    showing_last_value: bool,
 }
 
 impl RVariables {
@@ -91,7 +97,7 @@ impl RVariables {
      */
     pub fn start(env: RObject, comm: CommSocket, comm_manager_tx: Sender<CommManagerEvent>) {
         // Start with default settings
-        Self::start_with_config(env, comm, comm_manager_tx, LastValue::Hide);
+        Self::start_with_config(env, comm, comm_manager_tx, LastValue::UseOption);
     }
 
     /**
@@ -131,6 +137,7 @@ impl RVariables {
                 current_bindings,
                 version: 0,
                 show_last_value,
+                showing_last_value: false,
             };
             environment.execution_thread();
         });
@@ -232,6 +239,7 @@ impl RVariables {
             // list. This is a special R value that doesn't have its own
             // binding.
             if let Some(last_value) = self.last_value() {
+                self.showing_last_value = true;
                 variables.push(last_value.var());
             }
 
@@ -403,8 +411,8 @@ impl RVariables {
     fn last_value(&self) -> Option<PositronVariable> {
         // Check the cached value first
         let show_last_value = match self.show_last_value {
-            LastValue::Show => true,
-            LastValue::Hide => {
+            LastValue::Always => true,
+            LastValue::UseOption => {
                 // If we aren't always showing the last value, update from the
                 // global option
                 let use_last_value = get_option("positron.show_last_value");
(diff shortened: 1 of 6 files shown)


crates/ark/src/variables/variable.rs
```diff
@@ -1069,6 +1069,13 @@ impl PositronVariable {
         // Concrete nodes are objects that are treated as is. Accessing an element from them
         // might result in special node types.
 
+        // Check for the special access key `.Last.value`; we can only get the
+        // value for this object by evaluating it in R
+        if *access_key == String::from(".Last.value") {
```
lionel-, line 1074: It's a bit unidiomatic to create a `String` (potentially on the heap though the compiler is probably smarter) for comparison. I think `access_key == ".Last.value"` should work?

(Also the function should take `&str` instead of `String` for more idiomatic Rust)

crates/ark/src/variables/variable.rs
```diff
@@ -1069,6 +1069,13 @@ impl PositronVariable {
...
+            let last_robj = harp::parse_eval_global(".Last.value")?;
```
lionel-, line 1075: Same note here for last-value.

oh it might be worth creating `harp::last_value()`? We could use it in this file and in `interface.rs`.

crates/ark/tests/environment.rs
```diff
@@ -279,3 +289,234 @@ fn test_environment_list() {
...
+
+                    seen_test_var = params
+                        .assigned
+                        .iter()
+                        .any(|v| v.display_name == "test_var") || seen_test_var;
+                }
+            }
+        }
+    }
+
```
lionel-, line 417: oh I guess it would be hard to test the _contents_ of `.Last.value` here? We'd need a full integration test that actually instantiates a full kernel to which we could send `execute_request`, probably not worth it.

lionel- (approved):
LG!
Just a couple of nitpicks. It would be nice to have `harp::last_value()`.

crates/ark/src/variables/r_variables.rs
```diff
@@ -66,6 +67,9 @@ pub struct RVariables {
     /// thread. Tracked in https://github.com/posit-dev/positron/issues/1812
     current_bindings: RThreadSafe<Vec<Binding>>,
     version: u64,
+    /// Whether to always show the .Last.value in the Variables pane, regardless
+    /// of the value positron.show_last_value
+    show_last_value: bool,
```
DavisVaughan, line 72: I have a preference towards avoiding booleans where possible for call site readability

```rust
last_value: LastValue
```

```rust
enum LastValue {
  Show,
  Hide
}
```

In particular I think it really helps with

```rust
        // Start with default settings
        Self::start_with_config(env, comm, comm_manager_tx, false);
```

vs

```rust
        // Start with default settings
        Self::start_with_config(env, comm, comm_manager_tx, LastValue::Hide);
```

crates/ark/src/variables/r_variables.rs
```diff
@@ -357,6 +386,47 @@ impl RVariables {
...
+    /// last expression evaluated at the top level), if show_last_value is true.
+    ///
+    /// For backward compatibility, also checks the global option "positron.show_last_value"
+    /// if it exists, but the instance variable takes precedence.
+    ///
+    /// Returns None in all other cases.
+    fn last_value(&self) -> Option<PositronVariable> {
+        // First check our instance variable
+        let show_last_value = if self.show_last_value {
+            true
```
DavisVaughan, line 399: It isn't clear to me why we need `self.show_last_value`, or why the global option is mentioned here as a backwards compatible thing. It feels reasonable to me for that to be the "main feature" not part of backwards compat.

crates/ark/src/variables/r_variables.rs
```diff
@@ -357,6 +386,47 @@ impl RVariables {
...
+            match harp::parse_eval_global(".Last.value") {
+                Ok(last_robj) => Some(PositronVariable::from(
+                    String::from(".Last.value"),
+                    String::from(".Last.value"),
+                    last_robj.sexp,
+                )),
+                Err(err) => {
+                    // This isn't a critical error but would also be very
+                    // unexpected.
+                    log::debug!("Environment: Could not evaluate .Last.value ({err:?})");
```
DavisVaughan, line 420:
```suggestion
                    log::error!("Environment: Could not evaluate .Last.value ({err:?})");
```

We avoid `debug!` out of habit typically

crates/ark/src/variables/r_variables.rs
```diff
@@ -357,6 +386,47 @@ impl RVariables {
...
+    /// Gets the value of the special variable '.Last.value' (the value of the
+    /// last expression evaluated at the top level), if show_last_value is true.
+    ///
+    /// For backward compatibility, also checks the global option "positron.show_last_value"
+    /// if it exists, but the instance variable takes precedence.
+    ///
+    /// Returns None in all other cases.
+    fn last_value(&self) -> Option<PositronVariable> {
+        // First check our instance variable
+        let show_last_value = if self.show_last_value {
```
DavisVaughan, line 398: I see that you say

> Note that the option can be turned on during a session, but turning it off requires a restart. It's likely most users who want this will set it in their .Rprofile; we could also add a Positron setting for this.

I don't see where this occurs. During a session you can set `positron.show_last_value` and it looks like that will get used if `self.show_last_value` is `false`.

I don't see anywhere where `self.show_last_value` is persistently set to `true`, so I'm not sure how "turning it off requires a restart"

crates/ark/src/variables/r_variables.rs
```diff
@@ -357,6 +386,47 @@ impl RVariables {
...
+            // global option
+            let use_last_value = get_option("positron.show_last_value");
+            match use_last_value.get_bool(0) {
+                Ok(Some(true)) => true,
+                _ => false,
+            }
+        };
+
+        if show_last_value {
+            match harp::parse_eval_global(".Last.value") {
```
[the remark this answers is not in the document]
jmcphers: Done!

crates/ark/src/variables/variable.rs
jmcphers: Done!

crates/ark/src/variables/variable.rs
jmcphers: Done!

crates/ark/tests/environment.rs
jmcphers: Yeah, this would require a lot of overhead.

crates/ark/src/variables/r_variables.rs
jmcphers: Yes, I usually do this too but was lazy here. 😅 Updated.

crates/ark/src/variables/r_variables.rs
jmcphers: It is set to true during tests. Turning it off requires a restart because we don't have any way to clear the value even if we aren't updating it any more.

Just typing that out makes it sound dumb so I went ahead and made it so that you can change the option on the fly w/o a restart

PR #796, opened 2025-05-09
Preserve focus when user evaluates expression during debug session

lionel-:
Addresses https://github.com/posit-dev/positron/issues/3151

### DIFF
--- a/crates/ark/src/interface.rs
+++ b/crates/ark/src/interface.rs
@@ -92,6 +92,7 @@ use stdext::*;
 use uuid::Uuid;
 
 use crate::dap::dap::DapBackendEvent;
+use crate::dap::dap_r_main::FrameInfoId;
 use crate::dap::dap_r_main::RMainDap;
 use crate::dap::Dap;
 use crate::errors;
@@ -248,7 +249,11 @@ pub struct RMain {
     /// should only preserve focus if we're explicitly stepping through code as
     /// opposed to evaluating an expression in the debugger console.
     /// See https://github.com/posit-dev/positron/issues/3151.
-    preserve_debug_focus: bool,
+    debug_preserve_focus: bool,
+
+    /// The stack of frames we saw the last time we stopped. Used as a mostly
+    /// reliable indication of whether we moved since last time.
+    debug_last_stack: Vec<FrameInfoId>,
 }
 
 /// Represents the currently active execution request from the frontend. It
@@ -581,7 +586,8 @@ impl RMain {
             banner: None,
             r_error_buffer: None,
             captured_output: String::new(),
-            preserve_debug_focus: false,
+            debug_preserve_focus: false,
+            debug_last_stack: vec![],
         }
     }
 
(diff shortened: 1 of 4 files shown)


crates/ark/src/interface.rs
```diff
@@ -243,6 +244,16 @@ pub struct RMain {
...
+    /// Whether we should preserve focus when stopping in a debug session. We
+    /// should only preserve focus if we're explicitly stepping through code as
+    /// opposed to evaluating an expression in the debugger console.
+    /// See https://github.com/posit-dev/positron/issues/3151.
+    debug_preserve_focus: bool,
+
+    /// The stack of frames we saw the last time we stopped. Used as a mostly
+    /// reliable indication of whether we moved since last time.
+    debug_last_stack: Vec<FrameInfoId>,
 }
```
DavisVaughan, line 257: `RMainDap` already contains many "context related" pieces of information, like 

```rust
    /// The current call emitted by R as `debug: <call-text>`.
    call_text: DebugCallText,

    /// The last known `start_line` for the active context frame.
    last_start_line: Option<i64>,
```

I think I would really prefer that these move there as simply `preserve_focus` and `last_stack`.

I have a feeling it will clean some state management up too, as `self.debug_last_stack = vec![];` can move into `stop_debug()` where we already do cleanup like that.

crates/ark/src/interface.rs
```diff
@@ -761,13 +774,29 @@ impl RMain {
...
+                    // is not 100% reliable as this heuristic might have false
+                    // negatives, e.g. if the control flow exited the current
+                    // context via condition catching and jumped back in the
+                    // debugged function.
+                    let stack_id: Vec<FrameInfoId> = stack.iter().map(|f| f.into()).collect();
+                    let same_stack = stack_id == self.debug_last_stack;
+
+                    self.debug_last_stack = stack_id;
+                    self.dap
+                        .start_debug(stack, same_stack && self.debug_preserve_focus);
```
DavisVaughan, line 792: I'm optimistic that this whole thing can move into `RMainDap`!

crates/ark/src/interface.rs
```diff
@@ -1059,7 +1088,13 @@ impl RMain {
                 if self.dap.is_debugging() {
                     let continue_cmds = vec!["n", "f", "c", "cont"];
                     if continue_cmds.contains(&&code[..]) {
+                        // We're stepping so we want to focus the next location we stop at
+                        self.debug_preserve_focus = false;
                         self.dap.send_dap(DapBackendEvent::Continued);
+                    } else {
+                        // The user is evaluating some other expression so preserve current focus
+                        // https://github.com/posit-dev/positron/issues/3151
+                        self.debug_preserve_focus = true;
```
DavisVaughan, line 1097:
```suggestion
                        // We're stepping so we want to focus the next location we stop at
                        self.dap.set_preserve_focus(false);
                        self.dap.send_dap(DapBackendEvent::Continued);
                    } else {
                        // The user is evaluating some other expression so preserve current focus
                        // https://github.com/posit-dev/positron/issues/3151
                        self.dap.set_preserve_focus(true);
```

I'm imagining you'd then expose this like this

crates/ark/src/interface.rs
DavisVaughan: Actually it's possible you might be able to move them all the way down to `Dap` itself!

That actually feels like the right spot, and it's possible you won't even need to retain `last_stack` if you do that because in `Dap`'s `start_debug()` it looks like you can just compare the new `stack` to the old `self.stack` to finalize the `preserve_focus` value.

DavisVaughan (commented):
In addition to my comments, unfortunately it looks like this also removes the yellow highlighting that indicates where you are located if you just go down and run some arbitrary code while in the debugger.

This is rather confusing

https://github.com/user-attachments/assets/cdc020e0-2c9b-4191-af58-b3b29e4bbd27

lionel-:
@DavisVaughan I considered where to store that state and to me `RMain` felt like the right place because whether to focus or not is all about whether we're entering a nested debug repl. In the ideal setup, if we had our own repl, that state would be communicated by ReadConsole. The fact that the DAP can _in the current approach_ figure it out on its own by comparing the call stack is an implementation detail that shouldn't affect the proper boundaries of concerns. What do you think? Happy to discuss over a call.

DavisVaughan:
We decided to keep the current code where it is to match how we think about `preserve_focus` being an input to the DAP tooling, rather than something it computes itself

DavisVaughan:
@lionel- did you want to look further at the issue in the video though? https://github.com/posit-dev/ark/pull/796#pullrequestreview-2834259832

PR #797, opened 2025-05-10
Offset call stack by one in debugger

lionel-:
Addresses https://github.com/posit-dev/positron/issues/5155


DavisVaughan (approved):
Approved with the bug fix from https://github.com/posit-dev/ark/pull/799 related to the removal of `frame_name <- "<current>"`

PR #800, opened 2025-05-13
Switch to Cairo as the preferred fallback device

DavisVaughan:
Addresses https://github.com/posit-dev/positron/issues/2919

I think we can do this with fairly little fanfare, because we now use ragg as the preferred device when it is installed for png, jpeg, and tiff, which are the only devices that take the `type` argument. The pdf and svg devices are left unchanged by this PR.


thomasp85 (approved):
LGTM except for the whole "Cairo on Mac without XQuartz" thingy

lionel- (approved):

PR #805, opened 2025-05-14
Use smallest spanning node as the "starter" node in completions

### LINKED ISSUE #778: Newly constructed document context may get the wrong node
We're teasing this buglet out of https://github.com/posit-dev/ark/issues/770

When completions are triggered on, for example, an empty line, the current point isn't inside a syntax node. Or, rather, it's only inside the enclosing 'Program' node, which is sort of a degenerate case.

`find_closest_node_to_point()` will then latch on to a nearby node in some cases when it should not. (Which then interacts poorly with our completion logic, which is what #770 about.)

Image

Above, I've triggered completions on an empty line, but according to the document context, the completion node is the closing paren `)` of the line above.

We suspect this may be an edge case that hasn't been accounted for, i.e. that the current behaviour is unintentional. I'm going to see if we can change/fix this without breaking other things.

jennybc:
Fixes #778 

When we create a new `DocumentContext`, one of the pieces of data reflects some notion of the current node in the AST. Previously this was constructed as `ast.root_node().find_closest_node_to_point()` and I propose that we (mostly) switch to `ast.root_node().find_smallest_spanning_node()` instead. In this PR, I:

* Retain the previous notion of the current node under the name `closest_node`
* Recast the `node` field to mean "the node that point is actually in". (Frankly this is still somewhat aspirational, because so much existing code expects to start from something like a `"="` node, but this is a start. See #808 for more.)

This new definition of `node` is more favorable for completions (next I'll take #772 out of draft form, which will address #770), which is certainly the biggest user of `DocumentContext`, in terms of downstream code.

(The fact that redefining `node` has such a small impact on *anything else* below `crates/ark/src/lsp/` is sort of telling. It feels like there's a lot of bespoke fiddling around with the AST and nodes tucked away in various corners that could potentially be centralized/deduplicated over time. But not today.)

### DIFF
--- a/crates/ark/src/lsp/completions/completion_context.rs
+++ b/crates/ark/src/lsp/completions/completion_context.rs
@@ -7,6 +7,8 @@
 
 use std::cell::OnceCell;
 
+use tree_sitter::Node;
+
 use crate::lsp::completions::parameter_hints::ParameterHints;
 use crate::lsp::completions::parameter_hints::{self};
 use crate::lsp::completions::sources::composite::pipe::find_pipe_root;
@@ -20,7 +22,7 @@ pub(crate) struct CompletionContext<'a> {
     pub(crate) state: &'a WorldState,
     parameter_hints_cell: OnceCell<ParameterHints>,
     pipe_root_cell: OnceCell<Option<PipeRoot>>,
-    is_in_call_cell: OnceCell<bool>,
+    containing_call_cell: OnceCell<Option<Node<'a>>>,
 }
 
 impl<'a> CompletionContext<'a> {
@@ -30,7 +32,7 @@ impl<'a> CompletionContext<'a> {
             state,
             parameter_hints_cell: OnceCell::new(),
             pipe_root_cell: OnceCell::new(),
-            is_in_call_cell: OnceCell::new(),
+            containing_call_cell: OnceCell::new(),
         }
     }
 
@@ -44,8 +46,10 @@ impl<'a> CompletionContext<'a> {
     }
 
     pub fn pipe_root(&self) -> Option<PipeRoot> {
+        let call_node = self.containing_call_node();
+
         self.pipe_root_cell
-            .get_or_init(|| match find_pipe_root(self.document_context) {
+            .get_or_init(|| match find_pipe_root(self.document_context, call_node) {
                 Ok(root) => root,
                 Err(e) => {
                     log::error!("Error trying to find pipe root: {e}");
@@ -55,8 +59,9 @@ impl<'a> CompletionContext<'a> {
             .clone()
     }
 
-    pub fn is_in_call(&self) -> &bool {
-        self.is_in_call_cell
-            .get_or_init(|| node_find_containing_call(self.document_context.node).is_some())
+    pub fn containing_call_node(&self) -> Option<Node<'a>> {
+        *self
+            .containing_call_cell
+            .get_or_init(|| node_find_containing_call(self.document_context.node))
     }
 }
--- a/crates/ark/src/lsp/completions/completion_context.rs
+++ b/crates/ark/src/lsp/completions/completion_context.rs
@@ -13,7 +13,7 @@ use crate::lsp::completions::sources::composite::pipe::find_pipe_root;
 use crate::lsp::completions::sources::composite::pipe::PipeRoot;
 use crate::lsp::document_context::DocumentContext;
 use crate::lsp::state::WorldState;
-use crate::treesitter::NodeTypeExt;
+use crate::treesitter::node_find_containing_call;
 
 pub(crate) struct CompletionContext<'a> {
     pub(crate) document_context: &'a DocumentContext<'a>,
@@ -56,29 +56,7 @@ impl<'a> CompletionContext<'a> {
     }
 
     pub fn is_in_call(&self) -> &bool {
-        self.is_in_call_cell.get_or_init(|| {
-            let mut node = self.document_context.node;
-            let mut found_call = false;
-
-            loop {
-                if node.is_call() {
-                    found_call = true;
-                    break;
-                }
-
-                // If we reach a brace list, stop searching
-                if node.is_braced_expression() {
-                    break;
-                }
-
-                // Update the node
-                match node.parent() {
-                    Some(parent) => node = parent,
-                    None => break,
-                };
-            }
-
-            found_call
-        })
+        self.is_in_call_cell
+            .get_or_init(|| node_find_containing_call(self.document_context.node).is_some())
     }
 }
--- a/crates/ark/src/lsp/completions/completion_context.rs
+++ b/crates/ark/src/lsp/completions/completion_context.rs
@@ -13,12 +13,14 @@ use crate::lsp::completions::sources::composite::pipe::find_pipe_root;
 use crate::lsp::completions::sources::composite::pipe::PipeRoot;
 use crate::lsp::document_context::DocumentContext;
 use crate::lsp::state::WorldState;
+use crate::treesitter::NodeTypeExt;
 
 pub(crate) struct CompletionContext<'a> {
     pub(crate) document_context: &'a DocumentContext<'a>,
     pub(crate) state: &'a WorldState,
     parameter_hints_cell: OnceCell<ParameterHints>,
     pipe_root_cell: OnceCell<Option<PipeRoot>>,
+    is_in_call_cell: OnceCell<bool>,
 }
 
 impl<'a> CompletionContext<'a> {
@@ -28,6 +30,7 @@ impl<'a> CompletionContext<'a> {
             state,
             parameter_hints_cell: OnceCell::new(),
             pipe_root_cell: OnceCell::new(),
+            is_in_call_cell: OnceCell::new(),
         }
     }
 
@@ -45,10 +48,37 @@ impl<'a> CompletionContext<'a> {
             .get_or_init(|| match find_pipe_root(self.document_context) {
                 Ok(root) => root,
                 Err(e) => {
-                    log::error!("Error trying to find pipe root: {}", e);
+                    log::error!("Error trying to find pipe root: {e}");
                     None
                 },
             })
             .clone()
+    }
+
+    pub fn is_in_call(&self) -> &bool {
+        self.is_in_call_cell.get_or_init(|| {
+            let mut node = self.document_context.node;
+            let mut found_call = false;
+
+            loop {
+                if node.is_call() {
+                    found_call = true;
+                    break;
+                }
+
+                // If we reach a brace list, stop searching
+                if node.is_braced_expression() {
+                    break;
+                }
+
+                // Update the node
+                match node.parent() {
+                    Some(parent) => node = parent,
+                    None => break,
+                };
+            }
+
+            found_call
+        })
     }
 }
--- a/crates/ark/src/lsp/completions/sources/composite/call.rs
+++ b/crates/ark/src/lsp/completions/sources/composite/call.rs
@@ -23,7 +23,7 @@ use crate::lsp::completions::sources::CompletionSource;
 use crate::lsp::document_context::DocumentContext;
 use crate::lsp::indexer;
 use crate::lsp::traits::rope::RopeExt;
-use crate::treesitter::NodeTypeExt;
+use crate::treesitter::node_find_containing_call;
 
 pub(super) struct CallSource;
 
@@ -51,14 +51,10 @@ fn completions_from_call(
     let document_context = context.document_context;
     let point = document_context.point;
 
-    // Find the call node - we know we're in a call so we can walk up the tree
-    let mut node = document_context.node;
-    while !node.is_call() {
-        node = match node.parent() {
-            Some(parent) => parent,
-            None => return Ok(None), // This shouldn't happen since is_in_call() returned true
-        };
-    }
+    // Find the call node
+    let Some(node) = node_find_containing_call(document_context.node) else {
+        return Ok(None); // This shouldn't happen since is_in_call() returned true
+    };
 
     // Now that we know we are in a call, detect if we are in a location where
     // we should provide argument completions, i.e. if we are in the `name`
--- a/crates/ark/src/lsp/completions/sources/composite/call.rs
+++ b/crates/ark/src/lsp/completions/sources/composite/call.rs
@@ -23,7 +23,6 @@ use crate::lsp::completions::sources::CompletionSource;
 use crate::lsp::document_context::DocumentContext;
 use crate::lsp::indexer;
 use crate::lsp::traits::rope::RopeExt;
-use crate::treesitter::node_find_containing_call;
 
 pub(super) struct CallSource;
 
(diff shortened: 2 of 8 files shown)


crates/ark/src/lsp/document_context.rs
```diff
@@ -77,4 +93,36 @@ mod tests {
...
+        assert_eq!(
+            context
+                .document
+                .contents
+                .node_slice(&context.closest_node)
+                .unwrap()
+                .to_string(),
+            ")"
+        );
+    }
```
jennybc, line 127: To make this test more vivid, here's a screenshot of being in this scenario and requesting completions on the empty 2nd line:

Screenshot 2025-05-14 at 9 55 55 AM

That shows behaviour before this PR, where we think we're completing the `)` node and, therefore, don't provide any completions.

Here's the same, after this PR. The completions don't change (yet), but from the output channel you can see that now we realize we're not in a node, i.e. the smallest spanning node is the 'Program' node:

Screenshot 2025-05-14 at 9 57 37 AM

Here's a peek at what will be possible once this PR gets merged, once I add some version of #772. Now we will get real completions from R when completing on, e.g., an empty line.
Screenshot 2025-05-14 at 1 25 34 PM

crates/ark/src/lsp/document_context.rs
```diff
@@ -24,8 +30,15 @@ impl<'a> DocumentContext<'a> {
         // get reference to AST
         let ast = &document.ast;
 
-        // find node at point
-        let Some(node) = ast.root_node().find_closest_node_to_point(point) else {
+        let Some(node) = ast.root_node().find_smallest_spanning_node(point) else {
```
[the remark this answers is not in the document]
jennybc: I'm just following the existing patterns that were already used in this function.

crates/ark/src/lsp/document_context.rs
```diff
@@ -24,8 +30,15 @@ impl<'a> DocumentContext<'a> {
...
+            let contents = document.contents.to_string();
+            panic!(
+                "Failed to find spanning node containing point: {point} with contents '{contents}'"
```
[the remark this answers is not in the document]
jennybc: I'm just following the existing patterns that were already used in this function.

DavisVaughan (approved):
I think I have mostly made peace with this. I'm a little nervous that we are going to find edge cases in completions that expected the old behavior (I'll list one below), but I tried this out for a good while and most of the usual completion moves work (namespace, function args, string file paths, special completions for `Sys.getenv()` and friends, etc).

I am comforted by the fact that we do have quite a lot of tests for the usual cases, all of which go through a "real" `DocumentContext`.

If we do get reports of broken behavior, we can probably fix them up as required.

## Test cases we need to add

Here are a few tests we should add, related to a fix we need to do for `call_node_position_type()`, because it was written with the old form in mind.

Pressing ctrl+space here brings up the abort argument list

```r
rlang::abort(@)
```

Pressing ctrl+space here brings up nothing

```r
rlang::abort(
  @
)
```

Pressing ctrl+space here brings up the special env completion list

```r
Sys.setenv(@)
```

but does not here

```r
Sys.setenv(
  @
)
```

These are two separate completion paths, so we should add separate tests for these, but both have a root case of `call_node_position_type()` being broken, here is a test we can add for that as well in `test_call_node_position_type`:

```rust
        // After `(`, and on own line
        let (text, point) = point_from_cursor("fn(\n  @\n)");
        let document = Document::new(&text, None);
        let context = DocumentContext::new(&document, point, None);

        // On `main`, this passes
        assert_eq!(
            context.node.node_type(),
            NodeType::Anonymous(String::from("("))
        );

        // On this pr, `(`, doesn't contain the user's cursor, so instead it selects
        // the surrounding `Arguments` node
        // assert_eq!(context.node.node_type(), NodeType::Arguments);

        // On `main`, this passes, and this should also still pass on this PR, but currently does not
        assert_eq!(
            call_node_position_type(&context.node, context.point),
            CallNodePositionType::Name
        );
```

## Some more historical background

When we started out, I was convinced that `find_closest_node_to_point()` would only ever find a node that "enclosed" the user's cursor, and I thought that if it did not do that, it must have been a bug.

I have now convinced myself that I was wrong, and `find_closest_node_to_point()` CAN choose a node that does not enclose the user's cursor by design. The actual guarantees of this function are more like:

> Find the smallest node that is closest to the point in question. The node must be at or before the point in question. The node may be anonymous (i.e. an individual `{` token may be selected, rather than the entire `{ -> }` named braced expression node).

The `find_smallest_spanning_node()` helper is more like:

> Find the smallest node that encloses the point in question. For containment, bounds of `[]` are used, meaning that `{ 1 + 1 }@` will select the `}` token, `@{ 1 + 1 }` will select the `{` token, and `{ 1 + 1 @ }` will select the whole `{ -> }` node.

I think for many cases we really do want to include anonymous (i.e. non-named) nodes like the `}` and `{` tokens.

---

A particularly interesting PR to look at is:
https://github.com/posit-dev/ark/pull/321

Where I added this test, confirming that `find_closest_node_to_point()` does not have to "enclose" the point
https://github.com/posit-dev/ark/pull/321/commits/79f66e2452522b820386e4345f906528725d20f5

That PR was about the selection range feature, and ultimately I did not use `find_closest_node_to_point()` for it, because it was the wrong tool for the job.

I did end up using a native tree-sitter helper, `named_descendant_for_point_range()` instead. This is very similar to our `find_smallest_spanning_node()`, except for 2 key differences:

- The tree-sitter version finds only _named_ nodes, you'd have to use `descendant_for_point_range()` instead to also find anonymous ones
- The tree-sitter version has bounds of `[)` rather than `[]` like we use, which I do think is an important difference for things like `dplyr::@`. I wrote some extra comments about this here https://github.com/posit-dev/ark/blob/5d504f37e73c8e7db55e8f88d614ab69965e308c/crates/ark/src/lsp/selection_range.rs#L35-L47.

It would be nice to switch to something like `descendant_for_point_range()`, but I think we probably cannot because of these differences, but it is good to write them down somewhere for future reference.

crates/ark/src/lsp/completions/sources/utils.rs
```diff
@@ -434,17 +435,6 @@ mod tests {
...
 
-        // On `main`, this passes
-        // assert_eq!(
-        //     context.node.node_type(),
-        //     NodeType::Anonymous(String::from("("))
-        // );
-
-        // On this pr, `(`, doesn't contain the user's cursor, so instead it selects
-        // the surrounding `Arguments` node
-        assert_eq!(context.node.node_type(), NodeType::Arguments);
```
DavisVaughan: Can you keep this assertion?

crates/ark/src/lsp/completions/sources/composite/call.rs
```diff
@@ -302,8 +302,8 @@ fn completions_from_workspace_arguments(
 #[cfg(test)]
 mod tests {
     use harp::eval::RParseEvalOptions;
-    use tree_sitter::Point;
 
+    use crate::fixtures::point_from_cursor;
```
jennybc, line 306: I find tests that use this fixture much easier to read, so I've updated anything I touched to use this.

crates/ark/src/lsp/completions/sources/composite/call.rs
```diff
@@ -413,4 +413,39 @@ mod tests {
             harp::parse_eval("remove(my_fun)", options.clone()).unwrap();
         })
     }
+
```
jennybc, line 416: Here are some tests re: "the `rlang::abort()` case (using a base R function)".

crates/ark/src/lsp/completions/sources/unique/custom.rs
```diff
@@ -302,18 +302,34 @@ mod tests {
             let (text, point) = point_from_cursor("Sys.getenv(@)");
             assert_has_ark_test_envvar_completion(text.as_str(), point);
 
+            // Inside the parentheses, multiline
```
jennybc, line 305: Here I just interleaved the multiline scenarios inside the existing test. This is the `Sys.getenv()` custom completion case.

crates/ark/src/lsp/completions/sources/utils.rs
```diff
@@ -101,6 +101,7 @@ pub(super) enum CallNodePositionType {
 
 pub(super) fn call_node_position_type(node: &Node, point: Point) -> CallNodePositionType {
     match node.node_type() {
+        NodeType::Arguments => return CallNodePositionType::Name,
```
jennybc, line 104: The fixup for argument completions inside a call and custom completions such as `Sys.getenv()`.

crates/ark/src/lsp/completions/completion_context.rs
```diff
@@ -13,12 +13,14 @@ use crate::lsp::completions::sources::composite::pipe::find_pipe_root;
 use crate::lsp::completions::sources::composite::pipe::PipeRoot;
 use crate::lsp::document_context::DocumentContext;
 use crate::lsp::state::WorldState;
+use crate::treesitter::node_find_containing_call;
```
jennybc, line 16: New utility that allows me to eliminate 3 instances of same logic in the call, custom, and pipe completion sources.

crates/ark/src/lsp/completions/completion_context.rs
```diff
@@ -13,12 +13,14 @@ use crate::lsp::completions::sources::composite::pipe::find_pipe_root;
...
 
 pub(crate) struct CompletionContext<'a> {
     pub(crate) document_context: &'a DocumentContext<'a>,
     pub(crate) state: &'a WorldState,
     parameter_hints_cell: OnceCell<ParameterHints>,
     pipe_root_cell: OnceCell<Option<PipeRoot>>,
+    is_in_call_cell: OnceCell<bool>,
```
jennybc, line 23: New field that is useful to multiple completion sources.

*I think I'm going to add the call node itself as a field, so perhaps not quite done here.*

crates/ark/src/lsp/completions/sources/composite/call.rs
```diff
@@ -37,54 +36,34 @@ impl CompletionSource for CallSource {
...
-        if node.is_braced_expression() {
-            break;
-        }
-
-        // Update the node.
-        node = match node.parent() {
-            Some(node) => node,
-            None => break,
-        };
-    }
```
jennybc, line 72: This is the logic that existed in multiple places, sometimes with a comment reminding us to keep it in sync across those places. Now that we have completion context, we have some nice options for refactoring this.

crates/ark/src/lsp/completions/sources/composite/call.rs
```diff
@@ -37,54 +36,34 @@ impl CompletionSource for CallSource {
...
-        // let other sources add their own candidates instead
+    if !context.is_in_call() {
+        // Not in a call, let other sources add their own candidates
         return Ok(None);
     }
 
+    let document_context = context.document_context;
+    let point = document_context.point;
+
+    // Find the call node
```
jennybc, line 54: I think I'm actually going to store this as a field of `CompletionContext`. Coming soon.

crates/ark/src/lsp/completions/sources/composite/call.rs
```diff
@@ -302,32 +286,38 @@ fn completions_from_workspace_arguments(
...
-            let point = Point { row: 0, column: 9 };
-            let document = Document::new("match(tab)", None);
-            let context = DocumentContext::new(&document, point, None);
-            let completions = completions_from_call(&context, None).unwrap().unwrap();
+            let (text, point) = point_from_cursor("match(tab@)");
+            let document = Document::new(text.as_str(), None);
+            let document_context = DocumentContext::new(&document, point, None);
+            let state = WorldState::default();
+            let context = CompletionContext::new(&document_context, &state);
+            let completions = completions_from_call(&context).unwrap().unwrap();
```
jennybc, line 307: Using `CompletionContext` in lower-level completion source functions causes more boilerplate in the tests. I've taken any easy opportunity to use test helpers to combat this. But can't do it everywhere.

crates/ark/src/lsp/completions/sources/composite/call.rs
```diff
@@ -403,14 +401,86 @@ mod tests {
...
+
+            assert_eq!(completions.len(), 4);
+            assert_eq!(completions.get(0).unwrap().label, "x = ");
+            assert_eq!(completions.get(1).unwrap().label, "table = ");
+        })
+    }
+
+    #[test]
+    fn test_completions_in_value_position() {
```
jennybc, line 459: I think it's good to add tests of what happens in the "value" position of these completion sources.

I'm not totally sure we should provide 'No suggestions', but let's at least commit to that behaviour and test for it.

crates/ark/src/lsp/completions/completion_context.rs
DavisVaughan: Yea I'm expecting to see something like `OnceCell<Option<Node>>` here in the end

crates/ark/src/lsp/completions/sources/composite/call.rs
```diff
@@ -37,54 +36,34 @@ impl CompletionSource for CallSource {
...
-        };
-    }
-
-    if !has_call {
-        // Didn't detect anything worth completing in this context,
-        // let other sources add their own candidates instead
+    if !context.is_in_call() {
+        // Not in a call, let other sources add their own candidates
         return Ok(None);
     }
```
DavisVaughan, line 49: I don't even think you need this, just let this (from below) bail for you

```
let Some(node) = node_find_containing_call(document_context.node) else {
        return Ok(None); // This shouldn't happen since is_in_call() returned true
    };
```

crates/ark/src/lsp/completions/sources/composite/call.rs
DavisVaughan: Yes this is probably a related case to https://github.com/posit-dev/ark/pull/772/

DavisVaughan (approved):

crates/ark/src/lsp/completions/completion_context.rs
jennybc: Done!

crates/ark/src/lsp/completions/sources/composite/call.rs
jennybc: Done!

crates/ark/src/lsp/completions/sources/composite/call.rs
jennybc: Done.

jennybc:
OK @DavisVaughan I think this is ready if you want to give it one last look. These are the highlights of what's going on:

* Take a *partial* step in the direction of starting completions at the "node that contains the cursor", not necessarily the "closest node". For reasons discussed elsewhere and in #808, this is a complicated issue and you can't make changes in one place w/o making changes in other places. Hence the partial step.
  - In particular, when asking for completions on an empty line, the associated node is now the root 'Program' node. This is how this whole effort got started 😅 
* New field in `CompletionContext` that lets call, custom, and pipe completions access a shared notion of the "call node that contains point". Field is `containing_call_cell` and accessor is `containing_call_node()`.
* Adopt `point_from_cursor()` in any test module that I touched. Introduce other test helpers when it's easy to do so.
* Add many tests around completions in:
  - multiline contexts
  - the "value" position, in the sense of "name = value"

PR #807, opened 2025-05-21
Inject debug environment in search path completions

lionel-:
Addresses https://github.com/posit-dev/positron/issues/3001

### DIFF
--- a/crates/ark/src/interface.rs
+++ b/crates/ark/src/interface.rs
@@ -259,7 +259,7 @@ pub struct RMain {
     /// Current topmost environment on the stack while waiting for input in the
     /// debugger. This is `Some()` only when R is idle and in a `browser()`
     /// prompt.
-    pub(crate) debug_env: Option<RObject>,
+    debug_env: Option<RObject>,
 }
 
 /// Represents the currently active execution request from the frontend. It
@@ -2071,6 +2071,11 @@ impl RMain {
             // `harp::register`) will trigger an interrupt jump right away.
             StdInRpcReply::Interrupt => Ok(RObject::null()),
         }
+    }
+
+    #[cfg(not(test))] // Avoid warnings in unit test
+    pub(crate) fn debug_env(&self) -> Option<RObject> {
+        self.debug_env.clone()
     }
 }
 
--- a/crates/ark/src/lsp/completions/sources/composite/search_path.rs
+++ b/crates/ark/src/lsp/completions/sources/composite/search_path.rs
@@ -65,7 +65,7 @@ fn completions_from_search_path(
         #[cfg(not(test))] // Unit tests do not have an `RMain`
         {
             use crate::interface::RMain;
-            if let Some(debug_env) = &RMain::get().debug_env {
+            if let Some(debug_env) = &RMain::get().debug_env() {
                 // Mem-Safety: Object protected by `RMain` for the duration of the `r_task()`
                 env = debug_env.sexp;
             }


crates/ark/src/interface.rs
```diff
@@ -255,6 +255,11 @@ pub struct RMain {
     /// The stack of frames we saw the last time we stopped. Used as a mostly
     /// reliable indication of whether we moved since last time.
     debug_last_stack: Vec<FrameInfoId>,
+
+    /// Current topmost environment on the stack while waiting for input in the
+    /// debugger. This is `Some()` only when R is idle and in a `browser()`
+    /// prompt.
+    pub(crate) debug_env: Option<RObject>,
```
DavisVaughan, line 262: I still somewhat strongly prefer removing `pub(crate)` in favor of a helper `fn debug_env(&self) -> Option<&RObject>`. External consumers of `RMain` should not be able to modify this.

crates/ark/src/interface.rs
```diff
@@ -588,6 +593,7 @@ impl RMain {
             captured_output: String::new(),
             debug_preserve_focus: false,
             debug_last_stack: vec![],
+            debug_env: None,
```
DavisVaughan, line 596: Is it worth combining these into a `Debug` struct at some point?

crates/ark/src/interface.rs
```diff
@@ -774,6 +780,13 @@ impl RMain {
         if info.browser {
             match self.dap.stack_info() {
                 Ok(stack) => {
+                    if let Some(frame) = stack.first() {
+                        if let Some(ref env) = frame.environment {
```
DavisVaughan, line 784:
```suggestion
                    if let Some(ref env) = stack.first().map(|frame| frame.environment) {
```

Could probably drop a level of nesting with something like this? (Will need to be tweaked)

crates/ark/src/lsp/completions/sources/composite/search_path.rs
```diff
@@ -58,14 +58,25 @@ fn completions_from_search_path(
...
+        // If we're waiting for input in `read_console()` with a debugger
+        // prompt, start from current environment
+        #[cfg(not(test))] // Unit tests do not have an `RMain`
+        {
+            use crate::interface::RMain;
+            if let Some(debug_env) = &RMain::get().debug_env {
+                // Mem-Safety: Object protected by `RMain` for the duration of the `r_task()`
+                env = debug_env.sexp;
+            }
+        }
```
DavisVaughan, line 72: **Update: Actually, see other comment about going through `refresh_lsp()`**

How would you feel about moving `&RMain::get().debug_env` up to `provide_completions()` in `provide.rs`, so that you can provide the `debug_env` to `CompletionContext::new()`?

- The debug environment feels very relevant as part of the overall "completion context"
- Every completion source gets the `CompletionContext`, so if we need this `debug_env` in other places in the future, this makes it easily accessible
- `provide_completions()` already has a nice note about needing to be in an `r_task()`, which would be immediately apparent by us accessing `RMain` from there

Then here you'd just pass `CompletionContext` through to `completions_from_search_path()` rather than destructuring it as is currently done, and you'd be able to just reference `completion_context.env` or something like that from here (which would be an `Option<&RObject>` probably).

crates/ark/src/lsp/completions/sources/composite/search_path.rs
```diff
@@ -58,14 +58,25 @@ fn completions_from_search_path(
 
     unsafe {
         // Iterate through environments starting from the global environment.
-        let mut envir = R_GlobalEnv;
+        let mut env = R_GlobalEnv;
+
+        // If we're waiting for input in `read_console()` with a debugger
+        // prompt, start from current environment
+        #[cfg(not(test))] // Unit tests do not have an `RMain`
```
DavisVaughan, line 65: Okay, I have another thought I think we should strongly consider to avoid this and set us up better for the future.

In RMain's `console_inputs()` we already have a note to send the debug env through there:

https://github.com/posit-dev/ark/blob/2013ca15704284d36ea0908a3b497832dfdff49f/crates/ark/src/interface.rs#L2097-L2098

And a note here where we call `refresh_lsp()` to also update when we are in a browser state

https://github.com/posit-dev/ark/blob/2013ca15704284d36ea0908a3b497832dfdff49f/crates/ark/src/interface.rs#L759-L769

At the very least, I don't feel great about leaving these comments in as-is if we are also doing this new `debug_env` approach. They should at least mention `debug_env` as something we'd eventually remove.

But I think it would be nice if we could use `console_inputs()` and `refresh_lsp()` as the mechanism of getting this information over to the LSP, rather than accessing `RMain` from the LSP (which we currently never do anywhere else). We'd wrap the debug environment in an `RThreadSafe<>` and send it over in `DidChangeConsoleInputs`.

One key difference would be that the `RThreadSafe` debug environment would have to get stored in `LspState` rather than `WorldState`. An `RThreadSafe` purposefully cannot be `Clone`d, so it cannot be part of the "pure value" `WorldState`.

If this is possible, it would be nice because:
- We would not need `#[cfg(not(test))]`. `LspState.debug_env` would already either be `None` or `Some()` by the time we get here, and we would not need to talk to `RMain`.
- It would make it easier to analyze the requirements of the LSP in the future when we move it to Air, because we'd be able to look at `console_inputs()` to see what the LSP mostly requires from the kernel.

If this is not possible, it would be worth:
- Updating the above mentioned comments to at least mention our new `debug_env` approach
- Adding an ark issue about all this somewhere, documenting exactly what the current issue with going through `DidChangeConsoleInputs` is

crates/ark/src/interface.rs
```diff
@@ -774,6 +780,13 @@ impl RMain {
         if info.browser {
             match self.dap.stack_info() {
                 Ok(stack) => {
+                    if let Some(frame) = stack.first() {
```
DavisVaughan, line 783: I just want to point out that if you click on a different call stack frame in the UI, then you obviously only still get completions for the _first_ frame in the stack, because that's the position R is actually sitting at, even though you've walked up the stack in the UI.

I think that is totally expected and fine, it's just a possible source of confusion

DavisVaughan (changes requested):
It works pretty well for me in practice! It's nice to see contextual completions working as you'd expect.

But I do have some concerns about directly accessing `RMain` from the LSP (which we currently never do) vs trying hard to go through `DidChangeConsoleInputs` instead, which should set us up a little better for the future if it is possible (see comment below).

Using "Request changes" just so we can at least talk about it together Tuesday before merging

crates/ark/src/interface.rs
lionel-: I tried this before and tried again and could not get it to compile 😅 
Please feel free to tweak after I merge.

PR #815, opened 2025-05-28
Add folding ranges

kv9898:
Moving back from https://github.com/posit-dev/air/pull/146.

Old description:

It seems that the call for folding ranges has been quite a while: https://github.com/posit-dev/positron/issues/18, https://github.com/posit-dev/positron/issues/2924, https://github.com/posit-dev/positron/issues/3822.

I was initially only thinking about adding foldable comments, but it seems that doing this will disable the existing folding support for things like regions and brackets. Therefore, I rewrote these functionalities also.

The PR already supports the folding-range-handling of brackets, regions, code cells, indentations and *nested* comment sections:

(The screenshot is new, though)
image

Note that, compared to the [old PR](https://github.com/posit-dev/ark/pull/615), this PR no longer relies on a naive search for brackets. Instead it uses the new `tree_sitter` of air to walk down the AST for node handling.

### DIFF
--- a/crates/ark/src/lsp/backend.rs
+++ b/crates/ark/src/lsp/backend.rs
@@ -24,6 +24,7 @@ use tower_lsp::jsonrpc;
 use tower_lsp::jsonrpc::Result;
 use tower_lsp::lsp_types::request::GotoImplementationParams;
 use tower_lsp::lsp_types::request::GotoImplementationResponse;
+use tower_lsp::lsp_types::FoldingRange;
 use tower_lsp::lsp_types::SelectionRange;
 use tower_lsp::lsp_types::*;
 use tower_lsp::Client;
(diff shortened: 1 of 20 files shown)


kv9898:
@DavisVaughan @lionel- I think I need your help here. I'm struggling to register air as a folding range service provider - I have provided the functionalities but they are never invoked. Even my old PR doesn't work now. Have there been any breaking changes?

lionel-:
Let's try this out.

When this is ported to Rowan we should also consider a simple loop over an iterator that walks each node in DFS, like this: https://github.com/rust-lang/rust-analyzer/blob/789d9153e4615c63027305299f3a33d933e96464/crates/ide/src/folding_ranges.rs#L52

For nodes you can easily find the ending, and for sections we'd just match the next closing section within the current node. This way we avoid recursing (slower due to function calls, potential for overflow) and we simplify the logic (for instance no need to maintain stacks).

Thanks for the contribution!

PR #818, opened 2025-05-30
Don't normalize HTTP URLs for the Viewer

jmcphers:
This change addresses an issue that keeps URLs from being loaded into the Viewer on Windows. The issue is that we call `normalizePath` on everything that gets loaded into the Viewer, which is great for file paths but unsurprisingly doesn't work on URLs.

The fix here is to make our own wrapper for `normalizePath` that passes URLs through safely. This new function isn't vectorized (unlike the original `normalizePath`) but shouldn't need to be since the viewers accept only a single URL or path.

Addresses https://github.com/posit-dev/positron/issues/4843

### DIFF
--- a/crates/ark/src/modules/positron/utils.R
+++ b/crates/ark/src/modules/positron/utils.R
@@ -142,19 +142,6 @@ is_string <- function(x) {
     is.character(x) && length(x) == 1 && !is.na(x)
 }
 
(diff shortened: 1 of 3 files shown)


DavisVaughan (commented):
Let me know what you think about https://github.com/posit-dev/ark/pull/821 as a variant of this fix. Mostly just a little uncomfortable with having various forms of `normalizePath()` and `normalize_path()` floating around in the codebase

PR #819, opened 2025-06-02
Use a "function context" to create more responsive completions

jennybc:
Addresses https://github.com/posit-dev/positron/issues/4441
Addresses https://github.com/posit-dev/positron/issues/2574
Addresses https://github.com/posit-dev/positron/issues/2417

The common theme of this trio of issues is that folks want the completion to be more responsive to text that is already there. We are only talking about function completions here. The overall goal is to make it easy to accept completion help when tinkering with existing text (editing the function name or retroactively adding a `pkg::` prefix), without getting annoyances like two function names or two sets of parentheses.

We had already dipped our toe in using information about the completion site with the notion of `parameter_hints`. That story started in #680 (but got revised in later refactoring #754).

This PR creates a new `function_context` field in `CompletionContext` that contains rich information on the completion site, for the purposes of function completion. It supersedes `parameter_hints`.

Main points of interest:

* New `FunctionContext` struct is "everything you ever wanted to know about the AST, when forming a function completion". New `function_context` field in  `CompletionContext`. New accessor/initializer `function_context()`.
  - It feels good to frame the problem in terms of `FunctionRefUsage` (call vs. value) and `ArgumentsStatus` (absent, empty, nonempty).
* New standalone test module in `lsp/completions/tests/function_completions.rs`. I did this because these function completion tests don't really belong in `completion_item.rs` (and also that will not scale gracefully if we start to test other types of completions similarly). Function completions come via (at least) three sources (search path, namespace, document), so the tests don't belong with the sources either.
  - The test utilities in `completions/tests/utils.rs` remove a lot of boilerplate and I'd like to do similar elsewhere.
* I use tricks to make it feel like you are accepting text that's already present. Consider the problem of adding a `pkg::` prefix after the fact:
    ```
    # this fails because build_articles() can't be found
    build_articles()

    # go back and add `pkgdown::`
    # completions pop-up when cursor is at @
    pkgdown::@build_articles()

    # Trick #1: hoist build_articles() to the top of the completion list
    # Trick #2: use a text edit to replace "build_articles(" with "build_articles(",
    #   which has the effect of just moving the cursor
    # Tab or return to accept existing text and put the cursor inside `()` and
    #   enjoy automatic parameter hints
    pkgdown::build_articles(@)
    ```

✨ I definitely used copilot for assistance, but I ruled it with a very heavy hand. The overall design is 100% human-crafted, but certain implementation details or using this syntax vs. that might come from copilot. I have certainly tried to implement in a way that is consistent with what I see elsewhere, but I'm happy to make changes.


crates/ark/src/lsp/completions/completion_item.rs
```diff
@@ -119,6 +121,8 @@ pub(super) fn completion_item_from_assignment(
     item.documentation = Some(Documentation::MarkupContent(markup));
     item.kind = Some(CompletionItemKind::VARIABLE);
 
+    // TODO: This ad hoc completion item construction for a function does not
+    // benefit from the logic in completion_item_from_function() :(
```
jennybc, line 125: I actually found this problem when writing tests, because my first crack at it was with a user-defined function in the global environment. I wondered why things weren't working! We should probably look around for *ad hoc* construction of completion items and try to get rid of that in favor of using a `completion_item_from_*()` function.

crates/ark/src/lsp/completions/completion_item.rs
```diff
@@ -167,36 +171,117 @@ pub(super) unsafe fn completion_item_from_package(
...
     let label_details = item_details(package);
     item.label_details = Some(label_details);
 
-    let insert_text = sym_quote_invalid(name);
+    // Are we forming a completion item that's an exact match for the existing
+    // function identifier?
+    let item_name_is_match = name == function_context.name;
+
+    if item_name_is_match {
+        item.sort_text = Some(format!("0-{name}"));
```
jennybc, line 198: Putting existing text at the top of the list is part 1 of the key trick to make it easy to accept and move past existing text.

crates/ark/src/lsp/completions/completion_item.rs
```diff
@@ -167,36 +171,117 @@ pub(super) unsafe fn completion_item_from_package(
...
+        }
+
+        return Ok(item);
+    }
+
+    // from here on out, we use a text edit for all cases
+    item.insert_text = None;
+
+    if function_context.usage == FunctionUsage::Call &&
+        function_context.arguments_status == ArgumentsStatus::Empty
```
jennybc: Addresses an edit like this, where `@` indicates the cursor:

```r
some_thing()
# getting changed to
some_@()
# which becomes
some_other_thing(@)
```

When there are existing, empty parens, we put the cursor inside and trigger parameter hints, after editing the function name.

crates/ark/src/lsp/completions/completion_item.rs
```diff
@@ -167,36 +171,117 @@ pub(super) unsafe fn completion_item_from_package(
...
+            title: "Trigger Parameter Hints".to_string(),
+            command: "editor.action.triggerParameterHints".to_string(),
+            ..Default::default()
+        });
+
+        return Ok(item);
+    }
+
+    if function_context.usage == FunctionUsage::Call &&
+        function_context.arguments_status == ArgumentsStatus::Nonempty
```
jennybc: Addresses an edit like this, where `@` indicates the cursor:

```r
some_thing("stuff")
# getting changed to
some_@("stuff")
# which becomes
some_other_thing@("stuff")
```

When there are existing argument(s), we put the cursor at the end of the function identifier. It's not totally clear where to leave the cursor and, even if it were clear, there are many situations where that would be quite difficult/tricky to do in practice. Therefore, also no automatic parameter hints in this case.

crates/ark/src/lsp/completions/function_context.rs
```diff
@@ -0,0 +1,190 @@
...
+    /// (e.g. `foo(x = "hi")` or `foo(\n  \n)`)
+    Nonempty,
+}
+
+impl FunctionContext {
+    pub(crate) fn new(document_context: &DocumentContext) -> Self {
+        let completion_node = document_context.node;
+        let function_container_node = skip_namespace_operator(completion_node);
+
+        let function_identifier_node = if function_container_node.is_namespace_operator() {
```
jennybc: Getting these 3 node roles clear in my head was super useful.

crates/ark/src/lsp/completions/tests/function_completions.rs
```diff
@@ -0,0 +1,245 @@
...
+#[cfg(test)]
+use crate::lsp::completions::tests::utils::assert_no_command;
+#[cfg(test)]
+use crate::lsp::completions::tests::utils::assert_text_edit;
+#[cfg(test)]
+use crate::lsp::completions::tests::utils::find_completion_by_label;
+#[cfg(test)]
+use crate::lsp::completions::tests::utils::get_completions_at_cursor;
+#[cfg(test)]
+use crate::r_task;
```
jennybc: This looks kind of gross but the alternative syntax copilot showed me didn't look like a pattern used elsewhere here. My main goal, which this achieves, is to not repeat these `use` statements in all of the submodules below.

crates/ark/src/lsp/completions/tests/function_completions.rs
```diff
@@ -0,0 +1,245 @@
...
+
+#[cfg(test)]
+mod function_call_tests {
```
jennybc, line 8: I haven't struck on the perfect organizing principle for grouping these tests. If you have a preference, see please share! But I do plan to take another pass at organizing these and making sure I've covered everything I mean to.

The test cases to probe arise from combinations of these factors:
* Usage: call vs. reference
* Namespace: `fcn` vs. `pkg::fcn`
* Target scenario is completing/editing the function or the package
* Arguments: absent, empty, nonempty
* Matches existing function identifier or not

crates/ark/src/lsp/completions/completion_item.rs
jennybc: Well, I guess this *does* go through `completion_item_from_assignment()` but for the function case, that should presumably delegate to `completion_item_from_function()`.

crates/ark/src/lsp/encoding.rs
```diff
@@ -176,3 +176,15 @@ fn convert_character_from_utf8_to_utf16(x: &str, character: usize) -> usize {
...
+/// This is useful when we need to perform text edits on specific nodes.
+pub fn convert_node_to_range(x: &Rope, node: &tree_sitter::Node) -> tower_lsp::lsp_types::Range {
+    let start_point = node.start_position();
+    let end_point = node.end_position();
+
+    let start = convert_point_to_position(x, start_point);
+    let end = convert_point_to_position(x, end_point);
+
+    tower_lsp::lsp_types::Range::new(start, end)
+}
```
DavisVaughan: I'd like to avoid a proliferation of these helpers, I'm hoping to remove the Rope one day anyways. I think this is an existing tool that would work and is very expressive

```r
convert_tree_sitter_range_to_lsp_range (rope, node.range())
```

crates/ark/src/lsp/completions/tests/function_completions.rs
DavisVaughan, line 1: Gut reaction to this is that we should try hard to avoid integration tests like this, they are slower to compile (each one is its own separate binary) and sometimes tricky and annoying to work with (you're already seeing that with the amount of `#[cfg(test)]` used).

Instead, I think these are unit tests. It seems like they either all go in `function_context.rs`, or they get split up into individual files that are specific to the context (i.e. `"utils::a@"` is namespace related and could be a unit test in the namespace file).

If you really think things like `find_completion_by_label()` are going to be useful across >2 files of tests, then you can include them by adding `lsp/completions/test_utils.rs` and then have

```
#[cfg(test)]
mod test_utils;
```

and then you can use those test utils in any test blocks in `lsp/completions/`

But if you keep all of these tests together in `function_context.rs` then I think you should just inline these helpers right in the `mod tests {` block that you'd add there.

In general I prefer repeating myself in tests if it reduces the amount of abstraction I have to think about (i.e. when I'm debugging a single test I want everything right in front of me, not spread across files).

crates/ark/src/lsp/completions/tests/function_completions.rs
```diff
@@ -0,0 +1,245 @@
+//
+// completions/tests/function_completions.rs
```
DavisVaughan:
```suggestion
// function_completions.rs
```

And elsewhere

crates/ark/src/lsp/completions/function_context.rs
```diff
@@ -0,0 +1,190 @@
...
+                &document_context.document.contents,
+                &function_identifier_node,
+            ),
+            usage,
+            arguments_status,
+        }
+    }
+}
+
+pub(crate) fn skip_namespace_operator(node: Node) -> Node {
```
jennybc: I might rename this, because it's currently confusing. We're skipping past an anonymous node with string value `::` or `:::` (so, a namespace operator, in the colloquial sense) to get to a named `namespace_operator` node (so, a `namespace_operator` in the treesitter sense).

Consider this code:

```r
debug(testthat::expect_true)
```

Here's the tree:

```
program [0, 0] - [1, 0]
  call [0, 0] - [0, 28]
    function: identifier [0, 0] - [0, 5]
    arguments: arguments [0, 5] - [0, 28]
      open: ( [0, 5] - [0, 6]
      argument: argument [0, 6] - [0, 27]     # <------- HERE ---------------------┐
        value: namespace_operator [0, 6] - [0, 27]                                 |
          lhs: identifier [0, 6] - [0, 14]                                         |
          operator: :: [0, 14] - [0, 16]      # we often start here and skip up to ┘
          rhs: identifier [0, 16] - [0, 27]
      close: ) [0, 27] - [0, 28]
```

crates/ark/src/lsp/completions/tests/function_completions.rs
DavisVaughan: I was under the impression that these were _integration tests_ due to the `tests/` folder name

But these are actually _unit tests_ because they are at `ark/src/lsp/completions/tests/`

To be integration tests, they'd have to be at `ark/tests/`, i.e. https://doc.rust-lang.org/rust-by-example/testing/integration_testing.html

In light of this new information, and due to the fact that they aren't tied to any particular completion source, I think it is fine as is

crates/ark/src/lsp/completions/sources/composite/search_path.rs
```diff
@@ -116,8 +110,8 @@ fn completions_from_search_path(
                     symbol,
                     env,
                     name,
-                    promise_strategy,
-                    parameter_hints,
+                    promise_strategy.clone(),
```
DavisVaughan:
```suggestion
                    promise_strategy,
```

idk if you or claude did this, but `PromiseStrategy` is a `Copy` type (i.e. cloning happens automatically because we declared that this type is _ridiculously cheap to clone_), so you don't need this `clone()` and we prefer not to add them for Copy types

(i think we also removed this fairly recently, so possibly you just accidentally added it back in some rebase)

crates/ark/src/lsp/completions/function_context.rs
DavisVaughan: If you're on the lhs or rhs, doesn't this also skip up to the `namespace_operator` node there too? i.e. at `dplyr::a@`

DavisVaughan: I agree this name is confusing, I do think we should rename before merging. Possibly more like `find_function_container()`

crates/ark/src/lsp/completions/function_context.rs
```diff
@@ -0,0 +1,190 @@
...
+
+impl FunctionContext {
+    pub(crate) fn new(document_context: &DocumentContext) -> Self {
+        let completion_node = document_context.node;
+        let function_container_node = skip_namespace_operator(completion_node);
+
+        let function_identifier_node = if function_container_node.is_namespace_operator() {
+            match function_container_node.child_by_field_name("rhs") {
+                Some(node) => node,
+                None => completion_node, // should be impossible
```
DavisVaughan: Hmm, what about `dplyr::`? No rhs there.

crates/ark/src/lsp/completions/function_context.rs
```diff
@@ -0,0 +1,190 @@
...
+                &function_identifier_node,
+            ),
+            usage,
+            arguments_status,
+        }
+    }
+}
+
+pub(crate) fn skip_namespace_operator(node: Node) -> Node {
+    let Some(parent) = node.parent() else {
```
DavisVaughan, line 143: `skip_namespace_operator()` should not be `pub(crate)`

crates/ark/src/lsp/completions/function_context.rs
```diff
@@ -0,0 +1,190 @@
...
+        return false;
+    };
+
+    let Some(caller_node) = call_node.child_by_field_name("function") else {
+        return false;
+    };
+
+    let caller_name = node_text(&caller_node, contents).unwrap_or_default();
+
+    FUNCTIONS_EXPECTING_A_FUNCTION_REFERENCE.contains(&caller_name.as_str())
```
DavisVaughan:
```suggestion
fn is_inside_special_function(node: &Node, contents: &ropey::Rope) -> bool {
    let Some(call_node) = node_find_parent_call(node) else {
        return false;
    };

    let Some(call_name_node) = call_node.child_by_field_name("function") else {
        return false;
    };

    let call_name = node_text(&call_name_node, contents).unwrap_or_default();

    FUNCTIONS_EXPECTING_A_FUNCTION_REFERENCE.contains(&call_name.as_str())
```

The use of "caller" confused me a little

crates/ark/src/lsp/completions/function_context.rs
```diff
@@ -0,0 +1,190 @@
...
+    Empty,
+    /// Arguments node exists and has content, even if it's just whitespace
+    /// (e.g. `foo(x = "hi")` or `foo(\n  \n)`)
+    Nonempty,
+}
+
+impl FunctionContext {
+    pub(crate) fn new(document_context: &DocumentContext) -> Self {
+        let completion_node = document_context.node;
+        let function_container_node = skip_namespace_operator(completion_node);
```
DavisVaughan:
```suggestion
        let function_qualified_name_node = skip_namespace_operator(completion_node);
```

I find the name "container" to be a little confusing, I think because it is not precise? 

IIUC it either points to `dplyr::across` or just `across` or `acr` (depending on how much the user typed), so an informative name might be "qualified name"

And then you'd have

```rust
let function_name_node = if function_qualified_name_node.is_namespace_operator() {
```

for the unqualified version

crates/ark/src/lsp/completions/function_context.rs
```diff
@@ -0,0 +1,190 @@
...
+    Absent,
+    /// Arguments node exists, but is empty (e.g. `foo()`)
+    Empty,
+    /// Arguments node exists and has content, even if it's just whitespace
+    /// (e.g. `foo(x = "hi")` or `foo(\n  \n)`)
+    Nonempty,
+}
+
+impl FunctionContext {
+    pub(crate) fn new(document_context: &DocumentContext) -> Self {
```
DavisVaughan, line 57: Kind of a meta comment, I find this function a little scary because it can run on any arbitrary node, but it feels like it makes some assumptions that it is running on either:
- An `identifier`
- A `namespace_operator` (or one of it's children, which we walk back up to get to the `namespace_operator`)

In practice I think if this got called while the user was on, say, a `function_definition`, then everything would probably be okay, but I do wonder if we'd be doing too much work. Like, after `skip_namespace_operator()` runs, should we try and check if we are on an `identifier` or a `namespace_operator` and just bail if we aren't on one of those? Because its impossible for us to have a function without one of those?

crates/ark/src/lsp/completions/function_context.rs
```diff
@@ -0,0 +1,190 @@
...
+
+    let open_paren = match arguments_node.child_by_field_name("open") {
+        Some(node) => node,
+        None => return ArgumentsStatus::Absent,
+    };
+
+    let close_paren = match arguments_node.child_by_field_name("close") {
+        Some(node) => node,
+        None => return ArgumentsStatus::Absent,
+    };
```
DavisVaughan: This let-else pattern would work well for these 3 branches too to make the early return a little clearer

```rust
let Some(arguments_node) = parent.child_by_field_name("arguments") else {
    return ArgumentsStatus::Absent
};
```

crates/ark/src/lsp/completions/completion_item.rs
```diff
@@ -167,36 +171,115 @@ pub(super) unsafe fn completion_item_from_package(
...
     item.label_details = Some(label_details);
 
-    let insert_text = sym_quote_invalid(name);
+    // Are we forming a completion item that's an exact match for the existing
+    // function identifier?
+    let item_name_is_match = name == function_context.name;
+
+    if item_name_is_match {
+        item.sort_text = Some(format!("0-{name}"));
+        item.preselect = Some(true);
```
DavisVaughan, line 199: What exactly does preselect do?

crates/ark/src/lsp/completions/completion_item.rs
```diff
@@ -167,36 +171,115 @@ pub(super) unsafe fn completion_item_from_package(
...
+        } else {
+            item.insert_text_format = Some(InsertTextFormat::PLAIN_TEXT);
+            item.insert_text = Some(insert_text);
+        }
+
+        return Ok(item);
     }
 
+    // from here on out, we use a text edit for all cases
+    item.insert_text = None;
```
DavisVaughan: do you need this line?

crates/ark/src/lsp/completions.rs
```diff
@@ -7,11 +7,14 @@
```
DavisVaughan, line 7: Here is one case to fix and add a test for

https://github.com/user-attachments/assets/f00ede5d-1164-41e2-a82d-c16a53a8d4ed

crates/ark/src/lsp/completions/completion_item.rs
jennybc: It literally preselects this item. The UI highlights the item, with a grey background and black icon (vs. purple), below.

Screenshot 2025-06-17 at 9 01 38 AM

It's an interesting question. I think hoisting an item to the top, via something like the current approach of `item.sort_text = Some(format!("0-{name}"))`, and pre-selecting it are *almost* redundant, since the top item is pre-selected by default.

In the screenshot above, I temporarily *stopped* setting the sort text to cause this item to appear at the top of the list, but I kept `item.preselect = Some(true);`. You can (sort of) see that `build_articles()` appears in its normal location, with items above and below, but it's preselected and focused, so to speak.

If other completion sources / items started to manipulate the sort text, I guess there could be some little battle at the top for the very first item. In that case, preselection would matter.

I suppose some other source / item could also set `preselect` to `true`, and then I don't know how the client decides who wins.

I think setting sort text and preselect is probably the best fit for our goals here. But we could stop setting the sort text, I suppose.

crates/ark/src/lsp/completions.rs
jennybc: Done in 4b99a1f25

crates/ark/src/lsp/completions/completion_item.rs
jennybc: Technically no, but somehow it feels nicer to not set `insert_text` when we know we're going to apply a text edit instead.

crates/ark/src/lsp/completions/sources/utils.rs
```diff
@@ -57,11 +57,20 @@ pub(super) fn set_sort_text_by_words_first(completions: &mut Vec<CompletionItem>
...
         let text = match &item.sort_text {
-            Some(sort_text) => sort_text,
+            Some(sort_text) => {
+                // If it looks like the `sort_text` already has a prefix meant
+                // to bring this item to the top, don't prepend another prefix.
+                if priority_pattern.is_match(sort_text) {
+                    continue;
+                }
+                sort_text
+            },
```
jennybc, line 84: This is in reaction to a test failure I saw that looked like this:

```
lsp::completions::tests::function_completions::namespace_post_hoc_tests::test_function_call_with_post_hoc_namespace stdout ----

thread 'lsp::completions::tests::function_completions::namespace_post_hoc_tests::test_function_call_with_post_hoc_namespace' panicked at crates/ark/src/lsp/completions/tests/function_completions.rs:143:13:
sort_text should start with "0-", got: Some("1-0-adist")
```

which tipped me off that `set_sort_text_by_words_first()` was unconditionally adding a prefix intended to influence sort order. This PR appears to be the first time we've set `sort_text` in a `completion_item_from_*()` function. I think in this case `set_sort_text_by_words_first()` should defer to whomever set the `sort_text` intentionally and decline to add another prefix.

crates/ark/src/lsp/completions/tests/function_completions.rs
jennybc: FWIW I had done this to disambiguate `crates/ark/src/lsp/completions/tests/utils.rs` from `crates/ark/src/lsp/completions/sources/utils.rs`, but I've gone back to just a naked filename.

crates/ark/src/lsp/completions/function_context.rs
jennybc: I went with `effective_function_node`.

crates/ark/src/lsp/completions/function_context.rs
jennybc: Went with `get_effective_function_node()` to match `effective_function_node`.

crates/ark/src/lsp/completions/completion_item.rs
lionel-: Perhaps worth a comment?

crates/ark/src/lsp/completions/completion_item.rs
```diff
@@ -167,36 +171,130 @@ pub(super) unsafe fn completion_item_from_package(
...
+            item.insert_text = None;
+            let text_edit = TextEdit {
+                range: function_context.range,
+                new_text: insert_text,
+            };
+            item.text_edit = Some(CompletionTextEdit::Edit(text_edit));
+        } else {
+            item.insert_text_format = Some(InsertTextFormat::SNIPPET);
+            item.insert_text = Some(format!("{insert_text}($0)"));
+            item.command = Some(Command {
```
lionel-: It's cool that we can do this via completions

crates/ark/src/lsp/completions/completion_item.rs
```diff
@@ -167,36 +171,130 @@ pub(super) unsafe fn completion_item_from_package(
...
+            });
+        }
+
+        return Ok(item);
+    }
+
+    if function_context.usage == FunctionUsage::Reference {
+        // Create a text edit that replaces the existing function name
+        // This prevents duplication when the completion is accepted in the
+        // case of adding `pkg::` after the fact
```
lionel-: But we plan to not even trigger completions right? We'll consider the RHS "done" and not in need of further editing. Might be worth noting it.

crates/ark/src/lsp/completions/completion_item.rs
```diff
@@ -167,36 +171,130 @@ pub(super) unsafe fn completion_item_from_package(
...
+            item.insert_text = Some(insert_text);
+        }
+
+        return Ok(item);
+    }
+
+    // from here on out, we use a text edit for all cases
+    item.insert_text = None;
+
+    if function_context.usage == FunctionUsage::Call &&
```
lionel-: Might be worth a comment about why we condition this on parentheses being empty

crates/ark/src/lsp/completions/completion_item.rs
lionel-: Maybe add this example inline in comments? I find them very helpful

crates/ark/src/lsp/completions/completion_item.rs
lionel-: Same, might be a useful example to include in a comment

lionel-: I'm just wondering whether it would feel more consistent for users to always put the cursor right after the opening parenthesis, but no strong opinion about this.

crates/ark/src/lsp/completions/function_context.rs
```diff
@@ -0,0 +1,228 @@
...
+use crate::lsp::encoding::convert_tree_sitter_range_to_lsp_range;
+use crate::treesitter::node_find_parent_call;
+use crate::treesitter::node_text;
+use crate::treesitter::BinaryOperatorType;
+use crate::treesitter::NodeType;
+use crate::treesitter::NodeTypeExt;
+use crate::treesitter::UnaryOperatorType;
+
+/// Represents how a function is being used in an expression
+#[derive(Clone, Debug, PartialEq)]
```
lionel-, line 22: These new types felt quite natural while reading the code! I do have a few suggestions.

lionel-: I just wish it'd be clearer from the name that all these types are about _references_ to functions, rather than function declarations/bindings or function objects. So you'd have e.g.

```rs
pub(crate) enum FunctionRefUsage {
    /// Function is being called, e.g., `foo()`
    Call,
    /// Function is being referenced without calling, e.g., `foo` in `debug(foo)`
    Value,
}
```

This reflects the fact that _function references_ can be in both _value_ and _call_ positions.

lionel-: I am somewhat convinced we should make it clearer in our LSP what constructs represent "refs" by consistently using such a naming scheme.

crates/ark/src/lsp/completions/function_context.rs
```diff
@@ -0,0 +1,228 @@
...
+#[derive(Clone, Debug)]
+pub(crate) struct FunctionContext {
+    /// The name of the function (could be, and often is, a fragment)
+    pub name: String,
+    /// The LSP range of the function identifier
+    pub range: Range,
+    /// How the function is being used (call vs reference)
+    pub usage: FunctionUsage,
+    /// The status of the function's arguments
+    pub arguments_status: ArgumentsStatus,
```
lionel-, line 39: No need to change now, but to me it feels like this should be part of a "call" struct.
Maybe that Call struct would be an argument to `FunctionRefUsage::Call()`.

This way we keep a clean distinction between identifiers and calls. This would avoid the somewhat awkward `Absent` argument status.

crates/ark/src/lsp/completions/function_context.rs
```diff
@@ -0,0 +1,228 @@
...
+    "args",
+    "debug",
+    "debugonce",
+    "formals",
+    "help",
+    "trace",
+    "str",
+];
+
+fn is_inside_special_function(node: &Node, contents: &ropey::Rope) -> bool {
```
lionel-, line 168: Can we find a more explicit name?

`is_expected_function_value()` or something like that?

crates/ark/src/lsp/completions/function_context.rs
```diff
@@ -0,0 +1,228 @@
...
+#[derive(Clone, Debug, PartialEq)]
+pub(crate) enum FunctionUsage {
+    /// Function is being called, e.g., `foo()`
+    Call,
+    /// Function is being referenced without calling, e.g., `foo` in `debug(foo)`
+    Reference,
+}
+
+#[derive(Clone, Debug)]
+pub(crate) struct FunctionContext {
```
lionel-, line 31: I wonder if there should be a `namespace: Option<String>` field?

lionel- (approved):
Looks good!

I just had a few comments about the new API.

crates/ark/src/lsp/completions/completion_item.rs
jennybc: I've taken this suggestion to always put the cursor right after the opening parenthesis in this case ✅ (about to push this + more).

crates/ark/src/lsp/completions/completion_item.rs
jennybc: I'm about to push a commit that give the comments lots of love.

crates/ark/src/lsp/completions/completion_item.rs
jennybc: Comment added.

crates/ark/src/lsp/completions/function_context.rs
jennybc: Perhaps? But since there's no current need and it would be easy to add, I'm going to proceed as things are.

crates/ark/src/lsp/completions/function_context.rs
jennybc: I implemented the suggestion to rename the enum to `FunctionRefUsage ` and changed the `Reference` variant to `Value`.

crates/ark/src/lsp/completions/function_context.rs
jennybc: I've taken measures to exit early with a dummy function context if we can't possibly be completing a function call or reference.

crates/ark/src/lsp/completions/function_context.rs
lionel-: No worries but just to be clear, I thought about this to make it clear from the type definition that this struct handles reference paths, i.e. for `foo::bar` you can expect `name` to be `bar` if you see that there is an optional namespace field.

crates/ark/src/lsp/completions/completion_item.rs
jennybc: Opened #849 to track this

PR #825, opened 2025-06-04
Fixes for folding ranges

lionel-:
Follow up to #815.

- Fix detection of code chunks. `# %%` are now supported, `# +` is no longer recognised as chunk.
- Remove unwraps for safety.
- Fix end range of comment sections. They now extend up to the next section, including empty lines. Important because chunks do the same and they need to be nested in the comment sections. Otherwise the frontend is confused and removes the ranges that don't match its expectations.

### DIFF
--- a/crates/ark/src/lsp/folding_range.rs
+++ b/crates/ark/src/lsp/folding_range.rs
@@ -90,13 +90,9 @@ fn parse_ts_node(
             // Nested comment section handling
             let comment_line = get_line_text(document, start.row, None, None);
 
-            if let Err(err) = nested_processor(
-                document,
-                comment_stack,
-                folding_ranges,
-                start.row,
-                &comment_line,
-            ) {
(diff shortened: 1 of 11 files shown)


kv9898:
I couldn't comment where you did not make any changes so I will just type it here.

1. There remain other usage of `find_last_non_empty_line`. Should we drop the use of it altogether? Otherwise a cell like this in the end would still leave an empty line in the end:

```r
# %% Something
foo()


```

2. I suggest we change lines 127-134 to:
```rust
        end_node_handler(
            document,
            folding_ranges,
            end.row + 1, // added +1
            &mut child_comment_stack,
            &mut child_region_marker,
            &mut child_cell_marker,
        );
```
as currently a document ending like the example below would not fold properly with our current implementation.

```r
# some header ####
a <- 1
b <- 2
```

PR #826, opened 2025-06-04
Move `use` calls into `mod` blocks to avoid `#[cfg(test)]` proliferation

DavisVaughan:
@jennybc this solves the weirdness around having to specify `#[cfg(test)]` everywhere

In general, you put any `use` declarations that your test needs _inside the `mod tests {` block_, this ensures they are already inside the `#[cfg(test)]` of the `mod` block itself, avoiding the proliferation of `#[cfg(test)]`

### DIFF
--- a/crates/ark/src/lsp/completions.rs
+++ b/crates/ark/src/lsp/completions.rs
@@ -13,8 +13,8 @@ mod resolve;
 mod sources;
 mod types;
 
-pub(crate) use provide::provide_completions;
-pub(crate) use resolve::resolve_completion;
-
 #[cfg(test)]
 mod tests;
+
+pub(crate) use provide::provide_completions;
+pub(crate) use resolve::resolve_completion;
--- a/crates/ark/src/lsp/completions/tests.rs
+++ b/crates/ark/src/lsp/completions/tests.rs
@@ -1,5 +1,5 @@
 //
-// tests/mod.rs - Tests for completions functionality
+// tests.rs
 //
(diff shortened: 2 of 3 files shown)


crates/ark/src/lsp/completions.rs
```diff
@@ -13,8 +13,8 @@ mod resolve;
 mod sources;
 mod types;
 
-pub(crate) use provide::provide_completions;
-pub(crate) use resolve::resolve_completion;
-
 #[cfg(test)]
 mod tests;
```
DavisVaughan, line 17: We try to keep all mod calls up top

crates/ark/src/lsp/completions/tests.rs
```diff
@@ -1,5 +1,5 @@
 //
```
DavisVaughan, line 1: This was `completions/tests/mod.rs`, I've moved it to `completions/tests.rs`

The `mod.rs` is the "old style" that we try not to use anymore

jennybc:
Oh, I know this is possible, I was just trying to de-duplicate this stuff. This is how I started out. But I really don't have a strong opinion on this one, if you prefer to handle as in this PR.

jennybc:
I'll just merge this, since I don't have a strong opinion about the repeated `use` declarations and the rest of this I definitely want.

DavisVaughan:
I also think it fights rust-analyzer too much to do it the other way, because the "please add a `use` directive for this" code action will put it under `mod` by default

PR #830, opened 2025-06-05
Shut down gracefully if parent exits

jmcphers:
This change causes `ark` to gracefully exit if its parent process does, with the goal of preventing ark from becoming a zombie in cases where Kallichore exits abruptly (without shutting down kernels). 

It works by asking the OS to send it `SIGUSR1` (via `PR_SET_PDEATHSIG`) if the parent exits. When we get `SIGUSR1`, we ask the kernel to shut down. 

I thought about making this optional because maybe there are some cases where you might want to leave the kernel running and reconnect to it later? But I can't think of any.

This change is related to https://github.com/posit-dev/positron/issues/7991 and https://github.com/posit-dev/positron/issues/6692, but is not a complete fix for either one.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -12,6 +12,9 @@
             "osx": {
                 "program": "ark"
             },
+            "linux": {
+                "program": "ark"
+            },
             "windows": {
                 "program": "ark.exe"
             },
--- a/crates/ark/src/lib.rs
+++ b/crates/ark/src/lib.rs
@@ -25,8 +25,6 @@ pub mod lsp;
 pub mod methods;
 pub mod modules;
 pub mod modules_utils;
-#[cfg(target_os = "linux")]
-pub mod parent_monitor;
 pub mod plots;
 pub mod r_task;
 pub mod repos;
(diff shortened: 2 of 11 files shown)


DavisVaughan (approved):
If you like https://github.com/posit-dev/ark/pull/831, feel free to merge that in and then merge this

jmcphers:
@DavisVaughan thanks, that's much tidier!

lionel- (commented):
@jmcphers Interesting technique!

I have a few comments:

- In the interest of keeping memory usage low, I'd prefer to avoid spawning a thread just for this. I think we could get the heartbeat thread watch for the atomic? oh hmm that's an Amalthea thread.

  Then we could check from the R thread with an interrupt-time task. See https://github.com/posit-dev/ark/blob/67e3c25e907be3b0c732606dd185adf774f4b410/crates/ark/src/plots/graphics_device.rs#L76-L100 for how to set it up, it's basically the same setup but with an async infinite loop that waits on a timer.

- We might as well check for `getppid() == 1` after setting the pdeath signal, just in case the parent died during startup? If true set the atomic sentinel. (Better check _after_ setting the pdeath signal to prevent a tiny race condition between getppid/pdeathsig.)

- Should this mechanism be controllable by user? I could imagine cases where Ark is intended to be launched like a permanent daemon. Probably fine to wait and see whether any users actually need to disable this though.

PR #832, opened 2025-06-09
Sync with tree-sitter-r

DavisVaughan:
See also https://github.com/posit-dev/air/pull/357

Here are the meaningful changes we are pulling in:

- https://github.com/r-lib/tree-sitter-r/pull/171
- https://github.com/r-lib/tree-sitter-r/pull/172
- https://github.com/r-lib/tree-sitter-r/pull/174
- https://github.com/r-lib/tree-sitter-r/pull/175
- https://github.com/r-lib/tree-sitter-r/pull/157
- https://github.com/r-lib/tree-sitter-r/pull/154
- https://github.com/r-lib/tree-sitter-r/pull/153

All changes to our code are contained in the diagnostic module. I've added new tests for a number of these (in case we switch out the parser in the future, we don't want to regress). Each change is also its own commit.

The most annoying are https://github.com/r-lib/tree-sitter-r/pull/153 and https://github.com/r-lib/tree-sitter-r/pull/154. These both make tree-sitter-r _more_ accurate in terms of how it represents the R grammar, which is a good thing. `()` and `(1; 2)` are now syntax errors, and `fn(a b)` is also now a syntax error. This is great! But it means that tree-sitter just reports an `ERROR` at some unpredictable spot that _it_ gets to decide, and then it doesn't tell us anything about what kind of error it is (it can't do that and be general purpose at the same time). I've had to give up on giving "precise" errors in a few cases due to this, but I think that is okay. With a custom R Pratt parser, I think we should be able to do a lot better, because as we record parse errors we can report a _precise_ location and reason for the parse issue (i.e. missing parenthesis, or missing delimiter between arguments)

### DIFF
--- a/crates/ark/src/lsp/diagnostics_syntax.rs
+++ b/crates/ark/src/lsp/diagnostics_syntax.rs
@@ -335,21 +335,21 @@ mod tests {
         assert_eq!(diagnostics.len(), 1);
         let diagnostic = diagnostics.get(0).unwrap();
         assert_eq!(diagnostic.range.start, Position::new(0, 5));
-        assert_eq!(diagnostic.range.end, Position::new(0, 10));
+        assert_eq!(diagnostic.range.end, Position::new(0, 6));
         insta::assert_snapshot!(diagnostic.message);
 
         let diagnostics = text_diagnostics("foo[a, b");
         assert_eq!(diagnostics.len(), 1);
         let diagnostic = diagnostics.get(0).unwrap();
         assert_eq!(diagnostic.range.start, Position::new(0, 3));
-        assert_eq!(diagnostic.range.end, Position::new(0, 8));
+        assert_eq!(diagnostic.range.end, Position::new(0, 4));
         insta::assert_snapshot!(diagnostic.message);
 
         let diagnostics = text_diagnostics("foo[[a, b");
         assert_eq!(diagnostics.len(), 1);
         let diagnostic = diagnostics.get(0).unwrap();
         assert_eq!(diagnostic.range.start, Position::new(0, 3));
-        assert_eq!(diagnostic.range.end, Position::new(0, 9));
+        assert_eq!(diagnostic.range.end, Position::new(0, 5));
         insta::assert_snapshot!(diagnostic.message);
     }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3247,8 +3247,8 @@ checksum = "2545046bd1473dac6c626659cc2567c6c0ff302fc8b84a56c4243378276f7f57"
 
 [[package]]
 name = "tree-sitter-r"
-version = "1.1.0"
-source = "git+https://github.com/r-lib/tree-sitter-r?rev=bc8b4771bc4ac6c012f79877dbf44ba7cf87634f#bc8b4771bc4ac6c012f79877dbf44ba7cf87634f"
+version = "1.2.0"
+source = "git+https://github.com/r-lib/tree-sitter-r?rev=95aff097aa927a66bb357f715b58cde821be8867#95aff097aa927a66bb357f715b58cde821be8867"
 dependencies = [
  "cc",
  "tree-sitter-language",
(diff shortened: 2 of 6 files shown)


crates/ark/src/lsp/diagnostics.rs
```diff
@@ -690,80 +690,11 @@ fn recurse_parenthesized_expression(
...
-
-    if n > 1 {
-        // The tree-sitter grammar allows multiple `body` statements, but we warn
-        // the user about this as it is not allowed by the R parser.
-        let range = node.range();
-        let range = convert_tree_sitter_range_to_lsp_range(context.contents, range);
-        let message = format!("Expected at most 1 statement within parentheses, found {n}.");
-        let diagnostic = Diagnostic::new_simple(range, message);
-        diagnostics.push(diagnostic);
-    }
```
DavisVaughan, line 709: tree-sitter-r doesn't allow this anymore. This is now a syntax error.

crates/ark/src/lsp/diagnostics.rs
```diff
@@ -690,80 +690,11 @@ fn recurse_parenthesized_expression(
...
-
-    ().ok()
-}
-
-// TODO: This should be a syntax check, as the grammar should not allow
-// two `Argument` nodes side by side
-fn check_call_like_next_sibling(
-    child: Node,
```
DavisVaughan, line 717: tree-sitter-r doesn't allow this anymore. This is now a syntax error.

crates/ark/src/lsp/diagnostics.rs
```diff
@@ -1111,23 +1037,6 @@ foo
         })
     }
 
-    #[test]
-    fn test_expression_after_call_argument() {
```
DavisVaughan, line 1115: Moved to `diagnostics_syntax.rs`

crates/ark/src/lsp/diagnostics_syntax.rs
```diff
@@ -73,39 +73,17 @@ fn recurse_children(
...
-    } else {
-        // Not an unmatched closing token
-        return Ok(None);
-    };
-
-    let range = node.range();
-
-    Ok(Some(new_missing_open_diagnostic(
-        open_token, range, context,
-    )))
```
DavisVaughan, line 106: In some specific cases, when we saw an unmatched `)` that was also the `ERROR` node, we would report `Unmatched closing delimiter. Missing an opening '{open_token}'.`.

This is quite hard to do correctly, especially since `()` is a syntax error now, and tree-sitter reports the `ERROR` node as being `)` (so it "looks" like there isn't a closing delimiter based on our heuristic).

To make this more complex, sometimes tree-sitter will "recover" a `parameterized_expression`, and sometimes it won't when it should, so we can't use the presence of a `parameterized_expression` parent as a signal to not show this precise error.

```r
> treesitter::text_parse("()", treesitter.r::language())
<tree_sitter_node>

── Text ──────────────────────────────────────────────────────────────────────────────────────────────────
()

── S-Expression ──────────────────────────────────────────────────────────────────────────────────────────
(program [(0, 0), (0, 2)]
  (ERROR [(0, 0), (0, 2)]
    "(" [(0, 0), (0, 1)]
    (ERROR [(0, 1), (0, 2)])
  )
)
> treesitter::text_parse("(); 1 + 1", treesitter.r::language())
<tree_sitter_node>

── Text ──────────────────────────────────────────────────────────────────────────────────────────────────
(); 1 + 1

── S-Expression ──────────────────────────────────────────────────────────────────────────────────────────
(program [(0, 0), (0, 9)]
  (parenthesized_expression [(0, 0), (0, 9)]
    open: "(" [(0, 0), (0, 1)]
    (ERROR [(0, 1), (0, 3)]
      (ERROR [(0, 1), (0, 3)])
    )
    body: (binary_operator [(0, 4), (0, 9)]
      lhs: (float [(0, 4), (0, 5)])
      operator: "+" [(0, 6), (0, 7)]
      rhs: (float [(0, 8), (0, 9)])
    )
    close: ")" MISSING [(0, 9), (0, 9)]
  )
)
```

Sigh - this continues my belief that if you get an `ERROR` from tree-sitter, then all bets are off to be able to provide "precise" error messages. You have no guarantees about how it will recover from that error. A custom pratt R parser would be much better here.

So I've decided we just shouldn't do this for missing opening delimiters.

We do still show missing closing delimiters though, since tree-sitter tends to more reliably detect these as `MISSING`, and that is the more common case anyways (I think).

lionel- (approved):

Issue #791, opened 2025-05-06
Rewrite `on_process_idle_events()` as an async idle task


lionel-:
Right now plot requests are handled at idle time when no user executions are running in the console:

https://github.com/posit-dev/ark/blob/40908a3b2ba4c54bba05940ffd536444abf598a0/crates/ark/src/interface.rs#L842

The current way we poll for the messages is timeout-based, with a freq of 200ms at the time of writing. This approach translates to visible delays in the frontend when a plot needs to be rerendered. I'm about to reduce the frequency to 50ms but ideally we wouldn't wait at all when a request is ready.

To fix this, we could spawn an async idle task with `r_task::spawn_idle()` that runs an infinite loop selecting on the plot comm channels. This would involve:

- Switching to tokio async channels instead of sync crossbeam channels.
- Setting things up so that `is_drawing` and `should_render` causes the task to wait until the graphics device is ready to render.

This approach will improve overall performance of the communication between the frontend and Ark for plots (part of https://github.com/posit-dev/positron/issues/5184).

PR #836, opened 2025-06-11
Add a note about polled events and the later package

DavisVaughan:
Documenting that switching from 200ms to 50ms helped with https://github.com/posit-dev/positron/issues/7235

We did the switch in this commit https://github.com/posit-dev/ark/commit/452b84f459bf8777da3f03c1c6cd2ce0aec71fb2

### DIFF
--- a/crates/ark/src/interface.rs
+++ b/crates/ark/src/interface.rs
@@ -825,7 +825,8 @@ impl RMain {
 
         // Process R's polled events regularly while waiting for console input.
         // We used to poll every 200ms but that lead to visible delays for the
-        // processing of plot events.
+        // processing of plot events, it also slowed down callbacks from the later
+        // package. 50ms seems to be more in line with RStudio (posit-dev/positron#7235).
         let polled_events_rx = crossbeam::channel::tick(Duration::from_millis(50));
 
         let r_request_index = select.recv(&r_request_rx);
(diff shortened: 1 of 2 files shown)


DavisVaughan:
@lionel- I think I know why. Winston added `later::run_now()` to his original example, which messes things up if you're trying to rely on the `R_InputHandlers` being flushed. Try this

```r
# Put these in a block to ensure they're run without delay in between.
{
  later::later(\() { finish <<- Sys.time(); cat("Finished!")}, 0)
  start <- Sys.time()
}

# Wait for a moment after running block above.
# Then run:
finish - start
```

This respects the 1250ms delay, and gets faster with the 50ms delay.

crates/ark/src/interface.rs
```diff
@@ -825,7 +825,8 @@ impl RMain {
 
         // Process R's polled events regularly while waiting for console input.
         // We used to poll every 200ms but that lead to visible delays for the
-        // processing of plot events.
+        // processing of plot events, it also slowed down callbacks from the later
+        // package. 50ms seems to be more in line with RStudio (posit-dev/positron#7235).
```
lionel-, line 829: Also add link to https://github.com/posit-dev/ark/issues/791?

lionel-: No need, this is unrelated to the {later} issue.

@DavisVaughan Should we use an even smaller timeout?

DavisVaughan: We decided to keep 50ms for now

PR #837, opened 2025-06-12
Refuse to provide a statement range in the face of parse errors

DavisVaughan:
Part of https://github.com/posit-dev/positron/issues/5023

Instead, the frontend will send code to the console one line at a time, which is the best we can probably offer, and is somewhat similar to how RStudio works


lionel- (approved):
Looks good with the caveat that we might still want to return a result once we have better error recovery in our parser. I envision this custom request to be standardised in the future so it can be used in other IDEs, and having something that you can send at once to the console is easier to work with (no special case to handle).

I also think it would provide a better user experience than just evaluating one line all of a sudden. It will be non obvious at first why that happens when there are parse errors.

PR #843, opened 2025-06-17
Use safer `harp::parse_expr()` to decode file path strings

DavisVaughan:
Addresses https://github.com/posit-dev/positron/issues/6584

Minimal reprex was to put `source(".\R\utils.R")` in the console and try and request completions with tab after the `utils.R`.

This string actually doesn't parse. It sees `\R` and complains that this is an unrecognized escape sequence.

This crashed ark on both mac and windows, but it comes up on Windows because this is the string you get when you copy out of the windows file explorer (because they hate you).

Fixed by switching out the very old `r_string_decode()` for our more modern tooling of `parse_expr()`, which is safely wrapped in a `try_catch()`. Related to but unaffected by https://github.com/posit-dev/ark/pull/840. That PR will turn the actual error we get from an `harp::Error` into `ParseResult::SyntaxError`, but `parse_expr()` is insulated from that difference.


crates/harp/src/string.rs
```diff
@@ -1,48 +0,0 @@
...
-use libr::STRSXP;
-use libr::VECTOR_ELT;
-
-use crate::object::RObject;
-use crate::protect::RProtect;
-use crate::r_string;
-use crate::utils::r_typeof;
-
-// Given a quoted R string, decode it to get the string value.
-pub unsafe fn r_string_decode(code: &str) -> Option<String> {
```
DavisVaughan, line 24: Not needed, this is `harp::parse_expr()` with an extra check for STRSXP at the end.

juliasilge (approved):
Working great! 🚀 

- Before this change, I could definitely see the crash myself
- With this new build, I no longer see a crash (and no completions of course)

PR #848, opened 2025-06-24
Support functions in `View()`

lionel-:
Addresses posit-dev/positron#2945.

This PR adds support for functions in `View()`.

- For functions with valid source references, `View()` jumps to the function location in the source file.

- For functions in packages that don't have source references, `View()` first generates a virtual namespace document (the same as for debugging) and jumps into that file.

  Note we only generate the virtual namespace source if `View()` is called at top-level. The generation process causes global mutation of the namespace objects to attach the source refs, and we can't do that if code is already running as it might make assumptions about the srcrefs.

- As a fallback, we deparse the function and generate a virtual document for it.

Currently the virtual documents in the fallback approach are never cleaned up.  I'm not too worried about the memory leaks that this causes but we would ideally do better. The `TextDocumentContentProvider` API doesn't have any facilities for clean up but we could in principle open an editor for the vdoc, then watch over editors and trigger clean up when the editor is closed. However that would require lots of communication between the frontend and the backend and being careful about synchronisation. So I think this would be better handled entirely on the frontend side. This can be done as follow-up work.

I also made a small change to the Ark URI scheme to disambiguate multiple sessions. The URIs now contain the process ID of the Ark that generated them. If for some reason the frontend tries to open a vdoc for a session that's now in the background, you'll get an invalid vdoc editor. With the frontend-side PR, this sort of errors are now more user friendly. The error is mentioned in the virtual editor and the error message is logged.

All this is mostly defensive, it would be hard to trigger this error. But since the vdocs are currently provided by the LSP, and since the LSP is now transient, I just wanted to make sure we dealt with this more gracefully. In principle the vdocs should be provided via Jupyter instead of LSP and we'd dispatch to the relevant session based on an URI session parameter. It's not a priority to do it that way for now though.

### QA Notes

I've added some backend tests but there's some subtle behaviour and interactions on the frontend side that's not tested.

When viewing a function from a package, a virtual document containing the entire package should be generated and `View()` will take you at the appropriate location in that document. For instance with `lm()` from the stats package:

```r
View(lm)
```

Note that this is the same file that is used when you `debug()` a function in that package. The two features must not interfere with each other, i.e. you can do the following before and after `View(lm)` and everything should still work:

```r
debug(lm)
lm()
undebug(lm)
```

For functions with valid source references, `View()` should take you to the source file. This should be a regular editor that can be edited, unlike the virtual namespace file. Here's an example with a tempfile:

```r
# Source file with a function definition for `foo`
file <- tempfile(fileext = "R")
cat("foo <- function() 1", file = file)
source(file)

# Now view it
View(foo)
```

For functions that don't have source references, for instance script functions evaluated in the global environment, a virtual document is opened with the deparsed source:

```r
foo <- function() 1
View(foo)
```

You can see the URI for this document by hovering over the editor tab. The URI contains the Ark PID for disambiguation, as well as the environment name if it has one (in this case, "global"). The filename is based on the identifier provided to `View()` (in this case, "foo.R").

When the environment in which `View()` is called doesn't have a name, the URI contains the environment's address instead of a name:

```r
# Local function -> URL includes environment address
local({
    foo <- function() 1
    View(foo)
})
```

When `View()` is called with an expression instead of an identifier, the file is "unknown.R". Note that the environment address is shown instead of the name, due to a limitation of the current implementation:

```r
# Non-named function -> unknown.R
View(identity(foo))

# Same
xs <- list(foo = foo)
View(xs$foo)
```

A related change made to this PR is that trying to open an Ark URI that doesn't exists is now more user friendly. We no longer open an error dialog, instead an editor is opened with an error message inside and an error is logged in the extension output.

```r
# No longer opens an error dialog
.rs.api.navigateToFile("ark:hello")
```

Finally, we've also made functions viewable from the variable pane. After running the examples above you should see `foo` and `xs` defined. You can view the source by clicking the view button that has now appeared in the Variables pane:

Screenshot 2025-06-25 at 10 36 05

I don't have any backend tests for this feature.

### DIFF
--- a/crates/ark/src/modules/positron/debug.R
+++ b/crates/ark/src/modules/positron/debug.R
@@ -247,41 +247,21 @@ frame_info_from_srcref <- function(
     srcref,
     environment
 ) {
-    srcfile <- attr(srcref, "srcfile")
-    if (is.null(srcfile)) {
+    info <- srcref_info(srcref)
+    if (is.null(info)) {
         return(NULL)
     }
-
-    # If the file name is missing but there is a `srcref`, then we can try to use
-    # the `lines` to reconstruct a fake source file that `srcref` can point into.
-    # This is used when debugging user functions that are entered directly into the console,
-    # and for functions parsed with `parse(text = <text>, keep.source = TRUE)`.
-    file <- srcfile$filename
-    lines <- srcfile$lines
-
-    if (!identical(file, "") && !identical(file, "<text>")) {
-        # TODO: Handle absolute paths by using `wd`
-        file <- normalizePath(file, mustWork = FALSE)
-        content <- NULL
-    } else if (!is.null(lines)) {
-        file <- NULL
-        content <- paste0(lines, collapse = "\n")
-    } else {
-        return(NULL)
-    }
-
-    range <- srcref_to_range(srcref)
 
     new_frame_info(
         source_name = source_name,
         frame_name = frame_name,
-        file = file,
-        contents = content,
+        file = info$file,
+        contents = info$content,
         environment = environment,
-        start_line = range$start_line,
-        start_column = range$start_column,
-        end_line = range$end_line,
-        end_column = range$end_column
+        start_line = info$range$start_line,
+        start_column = info$range$start_column,
+        end_line = info$range$end_line,
+        end_column = info$range$end_column
     )
 }
 
@@ -361,35 +341,6 @@ frame_info_unknown_range <- function(
         start_column = 0L,
         end_line = 0L,
         end_column = 0L
-    )
-}
-
-srcref_to_range <- function(x) {
-    n <- length(x)
-
-    # The first and third fields are sensitive to #line directives if they exist,
-    # which we want to honour in order to jump to original files
-    # rather than generated files.
-    loc_start_line <- 1L
-    loc_end_line <- 3L
-
-    # We need the `column` value rather than the `byte` value, so we
-    # can index into a character. However the srcref documentation
-    # allows a 4 elements vector when the bytes and column values are
-    # the same. We account for this here.
-    if (n >= 6) {
-        loc_start_column <- 5L
-        loc_end_column <- 6L
-    } else {
-        loc_start_column <- 2L
-        loc_end_column <- 4L
-    }
-
-    list(
-        start_line = x[[loc_start_line]],
-        start_column = x[[loc_start_column]],
-        end_line = x[[loc_end_line]],
-        end_column = x[[loc_end_column]]
     )
 }
 
--- a/crates/ark/src/modules/positron/r_data_explorer.R
+++ b/crates/ark/src/modules/positron/r_data_explorer.R
@@ -5,35 +5,13 @@
 #
 #
 
-#' @export
-.ps.view_data_frame <- function(x, title) {
-    # Derive the name of the object from the expression passed to View()
-    object_name <- .ps.as_label(substitute(x))
-
-    # Create a title from the name of the object if one is not provided
-    if (missing(title)) {
-        title <- object_name
-    }
-
-    stopifnot(
-        is.data.frame(x) || is.matrix(x),
-        is.character(title) && length(title) == 1L && !is.na(title)
-    )
-
-    # If the variable is defined in the parent frame using the same name as was
-    # passed to View(), we can watch it for updates.
-    #
-    # Note that this means that (for example) View(foo) will watch the variable
-    # foo in the parent frame, but Viewing temporary variables like
-    # View(cbind(foo, bar)) does not create something that can be watched.
-    var <- ""
-    env <- NULL
-    if (isTRUE(exists(object_name, envir = parent.frame(), inherits = FALSE))) {
-        var <- object_name
-        env <- parent.frame()
-    }
-
+view_data_frame <- function(x, title, var, env) {
+    stopifnot(is_viewable_data_frame(x))
     invisible(.ps.Call("ps_view_data_frame", x, title, var, env))
+}
+
+is_viewable_data_frame <- function(x) {
+    is.data.frame(x) || is.matrix(x)
 }
 
 .ps.null_count <- function(column) {
--- a/crates/ark/src/modules/positron/srcref.R
+++ b/crates/ark/src/modules/positron/srcref.R
@@ -83,3 +83,65 @@ do_resource_namespaces <- function(default) {
 resource_namespaces <- function(pkgs) {
     .ps.Call("ps_resource_namespaces", pkgs)
 }
+
+srcref_info <- function(srcref) {
+    srcfile <- attr(srcref, "srcfile")
+    if (is.null(srcfile)) {
+        return(NULL)
+    }
+
+    # If the file name is missing but there is a `srcref`, then we can try to use
+    # the `lines` to reconstruct a fake source file that `srcref` can point into.
+    # This is used when debugging user functions that are entered directly into the console,
+    # and for functions parsed with `parse(text = <text>, keep.source = TRUE)`.
+    file <- srcfile$filename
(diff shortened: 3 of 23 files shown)


crates/ark/src/modules/positron/debug.R
```diff
@@ -247,41 +247,21 @@ frame_info_from_srcref <- function(
     srcref,
     environment
 ) {
-    srcfile <- attr(srcref, "srcfile")
-    if (is.null(srcfile)) {
+    info <- srcref_info(srcref)
```
lionel-, line 250: I've extracted `srcref_info()` to reuse in the view routines.

crates/ark/src/modules/positron/r_data_explorer.R
```diff
@@ -5,37 +5,15 @@
...
-    # foo in the parent frame, but Viewing temporary variables like
-    # View(cbind(foo, bar)) does not create something that can be watched.
-    var <- ""
-    env <- NULL
-    if (isTRUE(exists(object_name, envir = parent.frame(), inherits = FALSE))) {
-        var <- object_name
-        env <- parent.frame()
-    }
-
+view_data_frame <- function(x, title, var, env) {
```
lionel-, line 8: Now a data frame specific handler rather than the main `View()` hook.

crates/ark/src/modules/positron/srcref.R
```diff
@@ -1,9 +1,29 @@
...
     loadNamespace(ns_name)
     .ps.Call("ps_ns_populate_srcref", ns_name)
 }
 
+ns_populate_srcref_without_vdoc_insertion <- function(ns_name) {
+    loadNamespace(ns_name)
+    .ps.Call("ps_ns_populate_srcref_without_vdoc_insertion", ns_name)
+}
+
+fn_populate_srcref_without_vdoc_insertion <- function(fn) {
```
lionel-, line 16: This helper is not free of side effects (it generates and attaches srcrefs by mutation) but it does not insert the document to our list of vdocs. This is mostly for unit tests where we don't have acces to `RMain` yet, as it's `RMain` that manages the list of vdocs.

crates/ark/src/modules/positron/srcref.R
```diff
@@ -83,3 +103,65 @@ do_resource_namespaces <- function(default) {
 resource_namespaces <- function(pkgs) {
     .ps.Call("ps_resource_namespaces", pkgs)
 }
+
+srcref_info <- function(srcref) {
```
lionel-, line 107: These two helpers come from existing code for the debugger and have been extracted/moved here for reuse in `view_function()`.

crates/ark/src/modules/positron/view.R
```diff
@@ -0,0 +1,195 @@
...
+#
+# Copyright (C) 2023-2025 Posit Software, PBC. All rights reserved.
+#
+#
+
+# Dispatches to object handlers. The handlers take `var` and `env` arguments.
+# These are only passed if we could infer a variable name from the input and if
+# that variable exists in the calling environment. This is used for live
+# updating the objects, if supported (e.g. data frames in the data viewer).
+view <- function(x, title) {
```
lionel-: This new dispatching function is now hooked into `View()`.

It deals with `env` and `var` passes these to handlers, which are free to ignore them. We could use them in `view_function()` to get a "live  updating" effect in the vdoc editor as you update functions in an environment. While it would be cool, it's probably not worth the effort.

crates/ark/src/modules/positron/view.R
```diff
@@ -0,0 +1,195 @@
...
+
+    info <- view_function_info(
+        x,
+        title,
+        var = var,
+        env = env,
+        top_level = top_level
+    )
+
+    switch(
```
lionel-, line 68: My R code is starting to show that I've been spending most of my time writing Rust. I now find matching on tagged unions very natural.

jmcphers: 😂

jmcphers (approved):
This looks great and works well with the Positron changes in my testing!

My only feedback on the ark side is that it'd be helpful to be able to double-click function objects in the Variables pane and have them open from there (instead of typing `View(...)`). It should be very straightforward to do this on the ark side since you just need to tell the frontend that your object supports a view method, and then handle the viewer request. 

Up to you whether you want to add that feature, though, realize it's pretty close to the fork and we can add it later!

jennybc:
I have looked over the source changes, but I'm going to focus on reporting my experience using the PR. I'm not sure why I'm having a different experience than @jmcphers or if we're exercising it differently. I've got a dev build of Positron up, from https://github.com/posit-dev/positron/pull/8262 and a local ark build from this PR. To confirm:

```
> .ps.ark.version()
                                 branch                                  commit                                    date 
                "feature/view-function"                              "df8a918f"               "2025-06-24 17:01:56 PDT" 
                                 flavor                                    path                                 version 
                                "debug" "/Users/jenny/rrr/ark/target/debug/ark"                               "0.1.192"
```

I'm going to edit this comment or make new ones as I walk through the suggested usage.

lionel-:
> These URIs ☝️ contain the environment's address. I think I expected these to be "global" 🤔 in this case?

That's a reasonable expectation! The thing is that we get `env` and `var` from the dispatching function, and it only provides those if `var` is a valid object in `env`. That's because these are used for live-updating of data frames in the data viewer (and could be used for live updating of functions in a vdoc editor).

I could have the dispatching function pass those even when partially invalid, along with a boolean indicating whether the binding is valid. I'd prefer not to touch the data frame code path at this time though.

rodrigosf672:
### Verified Fixed
    Positron Version(s) : 2025.08.0 (Universal) build 11
    OS Version          : Sequoia 15.5

### Test scenario(s) 
This is a series of a bit confusing and extensive tests. I think it could be relevant to automate this. I verified the behavior by running each of the scripts from the Setup Code into the R console, and then observing what happens. Please note that behavior may be dependent on the virtual or global env. Please find cases I tested below.

| Description                                | Setup Code                                                      | View() Call               | Expected Behavior                                                                 |
|--------------------------------------------|------------------------------------------------------------------|---------------------------|------------------------------------------------------------------------------------|
| View function from installed package       | None                                                             | `View(lm)`                | Opens virtual namespace doc with `lm` at correct location                         |
| View after `debug()`                       | `debug(lm)`                                                      | `View(lm)`                | Debug and View work without interfering                                           |
| View after `undebug()`                     | `undebug(lm)`                                                    | `View(lm)`                | Debug cleared, View works                                                         |
| View global env function (no srcref)       | `foo <- function() 1`                                            | `View(foo)`               | Opens deparsed `foo` in `foo.R`, URI includes "global"                            |
| View local anonymous function              | `local({ foo <- function() 1; View(foo) })`                      | inside local block        | Opens `foo.R` with env address in URI                                             |
| View expression                             | `foo <- function() 1`                                            | `View(identity(foo))`     | Opens `unknown.R`, env address in URI                                             |
| View from list                             | `xs <- list(foo = foo)`                                          | `View(xs$foo)`            | Same as above (`unknown.R`)                                                      |
| View from unnamed env                      | `localenv <- new.env(); assign("foo", function() 1, envir = localenv)` | `with(localenv, View(foo))` | Opens `foo.R` with env address                                                    |
| Invalid Ark URI                            | N/A                                                              | `.rs.api.navigateToFile("ark:non-existing-file")` | Opens virtual doc with error message                                |
| View from Variables pane                   | `foo <- function() 1; xs <- list(foo = foo)`                     | Click "View" in pane       | Shows source in tab (file or virtual doc)                                         |
| View script-defined function (no srcref)   | `foo <- function(x) x^2`                                         | `View(foo)`               | Opens `foo.R` with deparsed source                                                |
| View from stale LSP session                | Open + close session, then run View()                            | `View(foo)`               | Shows error in tab if session is invalid                                          |

Issue #846, opened 2025-06-20
Problem with breadcrumbs and special-comment based sections


juliasilge:
_Originally posted by @SamT123 in https://github.com/posit-dev/positron/discussions/8205_:

Hi!

I think there may be a bug with breadcrumbs when using comment-defined sections in R (i.e. "# title ------") – but want to check here before opening an issue.

When not using comment-defined sections, `breadcrumbs.focus` starts with the "highlighting" focussed on the part of the code you are currently in, which is very useful for navigation:

i.e. if I do `breadcrumbs.focus` when my cursor is here, in `my_fun_2.2`:

<img width="646" alt="Screenshot 2025-06-19 at 16 36 33" src="https://github.com/user-attachments/assets/91c8dcb3-bb96-467f-8be9-d58864d2093d" />

breadcrumbs starts with my_fun_2.2 highlighted, and I can navigate from there with the arrow keys:

<img width="646" alt="Screenshot 2025-06-19 at 16 36 50" src="https://github.com/user-attachments/assets/d6ef965b-7c77-4ec5-8c27-3bab2d405dac" />

But, if there is are comment-defined sections above it, breadcrumbs seemingly doesn't know where I am in the code:

<img width="643" alt="Screenshot 2025-06-19 at 16 37 08" src="https://github.com/user-attachments/assets/0131e1d9-84c3-4110-bf32-f6fc1db72230" />

**EDIT:**

The hierarchy / structure of the document in breadcrumbs is correct, its just that the highlighting doesn't start in the right place. Here I've have expanded out the folded sections:

<img width="647" alt="Screenshot 2025-06-19 at 16 48 58" src="https://github.com/user-attachments/assets/07b824c2-dd3f-485e-8948-6830f241ce79" />

Positron Version: 2025.07.0 (Universal) build 134
Code - OSS Version: 1.100.0
Commit: 572a37b9b646af0b813dbe4d424bc9bfa1629646
Date: 2025-06-18T03:33:46.852Z
Electron: 34.5.1
Chromium: 132.0.6834.210
Node.js: 20.19.0
V8: 13.2.152.41-electron.0
OS: Darwin arm64 24.4.0

Any help much appreciated!
Sam

lionel-:
_Possibly_ related: if you select the section in the breadcrumbs popup, only the comment is highlighted, instead of the whole section. This does not happen with functions which are highlighted as a whole.

[...]

PR #854, opened 2025-06-25
Fix the range of sections in document symbols (#846)

kylebutts:
This solution fixes https://github.com/posit-dev/ark/issues/846 and https://github.com/posit-dev/positron/issues/6137.

The problem comes from here: 
https://github.com/posit-dev/ark/blob/fac7824615157269de80a0bb0fe37b5c75af1e23/crates/ark/src/lsp/symbols.rs#L231-L271

- `node` is the single comment line, so the "section" is defined as the single line.

The solution I settled on was to take the created nested symbols and then "correct" the the end positions of the sections
  - I did this to avoid a O(n^2) loop where you loop through the lines to find section starts and then loop through the lines again to find where the next header starts.

This also fixes the sticky scroll which was also breaking from the sections having "1-line height"
Example of headers working in outline and sticky header


lionel-:
Thanks @kylebutts! I went with a different approach in #855

kylebutts:
Great! The new logic looks a lot nicer :->

PR #855, opened 2025-06-26
Fix ranges of comment sections

lionel-:
Addresses #846.
Addresses posit-dev/positron#6137
Supersedes and closes #854.

Pair-programmed with an agent ✨.

The first commit refactors things to avoid the use of explicit stacks. The second commit is the fix.

Both linked issues stem from section ranges not extending to the section end. The section now ends at either the next non-nested section, or at the closing delimiter of the enclosing node (when the section is nested in a function for instance

The section end needs to be on the previous line. In principle range ends are exclusive, but in practice we need to go back a line for vscode breadcrumbs coverages to behave properly.

### QA Notes

The fixes in this PR are tested on the backend side. On the frontend side you should see the following when testing:

With: 

```r
# https://github.com/posit-dev/ark/issues/846

{
    # level 1 ####
    1
    # another level 1 ####
    2
}

# Level ----

foo <- function() {
    1
    2
}
```

The breadcrumbs coverage (revealed when a breadcrumb is selected) should be consistent with the sections. Nested symbols such as function definitions should work as well.

https://github.com/user-attachments/assets/e0bffd43-e9d6-48b3-88a6-682e21dfdc03

With:

```r
# https://github.com/posit-dev/positron/issues/6137

test_fun <- function(...) {












    return(invisible())
  }





  #### Section Title ####
  test_fun2 <- function(...) {







    fun3 <- function() {











    }







    # long function
    return(invisible())
  }
```

The breadcrumb sticky scrolls should incorporate sections. Nested breadcrumb should still work:

https://github.com/user-attachments/assets/8f5f35a9-b955-4a13-b3cf-5afd97af6bce

### DIFF
--- a/crates/ark/src/treesitter.rs
+++ b/crates/ark/src/treesitter.rs
@@ -590,3 +590,53 @@ pub(crate) fn point_end_of_previous_row(
         point
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use ropey::Rope;
+    use tree_sitter::Point;
+
+    use super::*;
+
+    #[test]
+    fn test_point_end_of_previous_row() {
+        let contents = Rope::from_str("hello world\nfoo bar\nbaz");
+        let point = Point { row: 2, column: 1 };
+        let result = point_end_of_previous_row(point, &contents);
+        assert_eq!(result, Point { row: 1, column: 7 });
+    }
+
+    #[test]
+    fn test_point_end_of_previous_row_first_row() {
+        let contents = Rope::from_str("hello world\nfoo bar\nbaz");
+        let point = Point { row: 0, column: 5 };
+        let result = point_end_of_previous_row(point, &contents);
+        assert_eq!(result, Point { row: 0, column: 0 });
+    }
+
+    #[test]
+    fn test_point_end_of_previous_row_empty_previous_line() {
+        let contents = Rope::from_str("hello\n\nworld");
+
+        let point = Point { row: 2, column: 1 };
+        let result = point_end_of_previous_row(point, &contents);
+        assert_eq!(result, Point { row: 1, column: 0 });
+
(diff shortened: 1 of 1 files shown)


crates/ark/src/treesitter.rs
```diff
@@ -571,3 +571,22 @@ pub(crate) fn node_find_containing_call<'tree>(node: Node<'tree>) -> Option<Node
...
+    contents: &ropey::Rope,
+) -> tree_sitter::Point {
+    if point.row > 0 {
+        let prev_row = point.row - 1;
+        let line = contents.line(prev_row as usize);
+        let line_len = line.len_chars().saturating_sub(1); // Subtract 1 for newline
+        tree_sitter::Point {
+            row: prev_row,
+            column: line_len,
+        }
```
DavisVaughan, line 586: Is there a reason you don't include the newline? I feel like the newline is technically just the end of the current line, right? So it feels like that position would be valid

crates/ark/src/treesitter.rs
```diff
@@ -571,3 +571,22 @@ pub(crate) fn node_find_containing_call<'tree>(node: Node<'tree>) -> Option<Node
 
     None
 }
+
+pub(crate) fn point_end_of_previous_row(
```
DavisVaughan, line 575: _Maybe_ a test or two for this? Particularly around row 0 behavior and newline behavior

DavisVaughan (approved):

crates/ark/src/treesitter.rs
lionel-: Because after the newline you are no longer on that line, so that position actually represents column 0 of the line we started with. The n + 1 position on the line doesn't really exist in fact.

Side note, I find it confusing that Rope's line iterator includes trailing newlines. I think a "split by line" viewpoint is more intuitive.

PR #858, opened 2025-06-27
Emit functions passed as named arguments as document symbols

lionel-:
Branched from #856 
Addresses https://github.com/posit-dev/positron/issues/6546
Addresses https://github.com/posit-dev/positron/issues/6107 and https://github.com/posit-dev/positron/issues/5241

The main functionality this PR aims to implement is inclusion of R6 symbols in the outline.

```r
class <- R6::R6Class(
  'class',
  public = list(
    initialize = function() 'initialize',
    foo = function() 'foo'
  ),
  private = list(
    bar = function() 'bar'
  )
)
```

I think it makes sense to include _all_ functions passed as named arguments, to any calls and not just `R6Class`. This is in contrast to _workspace symbols_ reachable across files. In this case I think only R6 methods should be registered as workspace symbols (see https://github.com/posit-dev/positron/issues/6549). In general document symbols are more exhaustive than workspace one, offering access to local objects.

Changes:

- We now collect symbols in the RHS of assigned objects. In particular this means we get a chance to collect methods in the `r6class()` call above. This also means we'll collect symbols in cases like this, which I think makes sense:

  ```r
  foo <- {
    bar <- function() {}
  }
  ```

- We now collect symbols across call arguments. This allows us to collect methods nested deeply inside the `r6class()` call. This also means we also collect symbols in cases like the following, which I also think makes sense:

  ```r
  local({
    a <- function() {
      1
    }
  })
  ```

  Addresses https://github.com/posit-dev/positron/issues/6107 and https://github.com/posit-dev/positron/issues/5241

- Finally, functions passed as named arguments are collected as Method symbols. Previously only functions assigned with `<-` at top level or in `{}` would be collected.

### QA Notes:

All these changes are tested in the backend. On the frontend you should now see the outline/breadcrumbs filled with `initialize`, `foo`, and `bar` methods.

```r
# section ----
class <- R6::R6Class(
  'class',
  public = list(
    initialize = function() 'initialize',
    foo = function() {
      1
      # section2 ----
      nested <- function() {}
      2
    }
  ),
  private = list(
    bar = function() {
      3
    }
  )
)

list(
  foo = function() {
    1
  }, # matched
  function() {
    nested <- function () {
        # `nested` is a symbol even if the unnamed method is not
    }
  }, # not matched
  bar = function() {
    3
  }, # matched
  baz = (function() {
    4
  }) # not matched
)

local({
  a <- function() {
    1
  }
})
```

https://github.com/user-attachments/assets/c369c5bc-2506-482b-a3f2-fa1a8146ba58

### DIFF
--- a/crates/ark/src/lsp/symbols.rs
+++ b/crates/ark/src/lsp/symbols.rs
@@ -265,16 +265,17 @@ fn collect_call(
     let Some(callee) = node.child_by_field_name("function") else {
         return Ok(());
     };
-    if !callee.is_identifier() {
-        return Ok(());
+
+    if callee.is_identifier() {
+        let fun_symbol = contents.node_slice(&callee)?.to_string();
+
+        match fun_symbol.as_str() {
+            "test_that" => return collect_call_test_that(node, contents, symbols),
+            _ => {}, // fallthrough
+        }
     }
 
-    let fun_symbol = contents.node_slice(&callee)?.to_string();
-
-    match fun_symbol.as_str() {
-        "test_that" => collect_call_test_that(node, contents, symbols)?,
-        _ => collect_call_arguments(node, contents, symbols)?,
-    }
+    collect_call_arguments(node, contents, symbols)?;
 
     Ok(())
 }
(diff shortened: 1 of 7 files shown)


jennybc (approved):
I've looked over the code and, with the caveat that I'm still a relative newcomer to ark, LGTM.

My main reaction is: I went to the tests to get a feel for what really happens here in the symbols area of the LSP and what is changing in this PR. But it's pretty hard for a human to parse the expectations that use the `insta::assert_debug_snapshot!()` mechanism.

Is there a reason the changes in this PR are tested only that way? Versus the more concrete assertions in some of the other tests. I'm definitely thinking about my recent foray into completions and tests, where it was possible to check that a certain completion was present (and of a specific form) or absent.

lionel-:
> Is there a reason the changes in this PR are tested only that way? Versus the more concrete assertions in some of the other tests. I'm definitely thinking about my recent foray into completions and tests, where it was possible to check that a certain completion was present (and of a specific form) or absent.

It's just for convenience, of both writing the tests and updating them (i.e. if you insert a line all ranges change).

crates/ark/src/lsp/symbols.rs
```diff
@@ -265,16 +265,90 @@ fn collect_call(
...
+    contents: &Rope,
+    symbols: &mut Vec<DocumentSymbol>,
+) -> anyhow::Result<()> {
+    let Some(arguments) = node.child_by_field_name("arguments") else {
         return Ok(());
+    };
+
+    let mut cursor = node.walk();
+    for arg in arguments.children(&mut cursor) {
+        if arg.kind() != "argument" {
```
DavisVaughan: Maybe walk this for clarity `arguments.children_by_field_name("argument", &mut cursor)`

crates/ark/src/lsp/symbols.rs
```diff
@@ -265,16 +265,90 @@ fn collect_call(
...
+        if arg_value.kind() == "function_definition" {
+            if let Some(arg_fun) = arg.child_by_field_name("name") {
+                // If this is a named function, collect it as a method
+                collect_method(&arg_fun, &arg_value, contents, symbols)?;
+            } else {
+                // Otherwise, just recurse into the function
+                let body = arg_value.child_by_field_name("body").into_result()?;
+                collect_symbols(&body, contents, 0, symbols)?;
+            };
+        }
```
DavisVaughan: It feels like maybe the `collect_symbols()` call up top should be mutually exclusive with the `function_definition` branch?

Like maybe you should first `match arg_value.kind()` with a variant for `"function_definition"` and then a fallthrough of `collect_symbols()`?

lionel-: yep that'd be clearer

PR #859, opened 2025-06-27
Don't emit nested objects as document symbols

lionel-:
Addresses https://github.com/posit-dev/positron/issues/8330

But I'd like to discuss what to do here with the team before merging.

### DIFF
--- a/crates/ark/src/lsp/symbols.rs
+++ b/crates/ark/src/lsp/symbols.rs
@@ -122,6 +122,8 @@ struct Section {
     children: Vec<DocumentSymbol>,
 }
 
+struct CollectContext;
+
 pub(crate) fn document_symbols(
     state: &WorldState,
     params: &DocumentSymbolParams,
@@ -136,7 +138,8 @@ pub(crate) fn document_symbols(
     let mut result = Vec::new();
 
     // Extract and process all symbols from the AST
-    if let Err(err) = collect_symbols(&root_node, contents, 0, &mut result) {
+    let mut ctx = CollectContext;
+    if let Err(err) = collect_symbols(&mut ctx, &root_node, contents, 0, &mut result) {
         log::error!("Failed to collect symbols: {err:?}");
         return Ok(Vec::new());
     }
@@ -146,6 +149,7 @@ pub(crate) fn document_symbols(
 
 /// Collect all document symbols from a node recursively
 fn collect_symbols(
+    ctx: &mut CollectContext,
     node: &Node,
     contents: &Rope,
     current_level: usize,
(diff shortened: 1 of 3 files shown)


crates/ark/src/lsp/symbols.rs
```diff
@@ -418,26 +441,34 @@ fn collect_assignment(
...
     let function = lhs.is_identifier_or_string() && rhs.is_function_definition();
     if function {
-        return collect_assignment_with_function(node, contents, symbols);
+        return collect_assignment_with_function(ctx, node, contents, symbols);
     }
 
-    // Otherwise, collect as generic object
-    let name = contents.node_slice(&lhs)?.to_string();
+    if ctx.top_level || ctx.include_assignments_in_blocks {
+        // Collect as generic object, but only if we're at top-level. Assigned
```
DavisVaughan:
```suggestion
        // Collect as generic object, but typically only if we're at top-level. Assigned
```

crates/ark/src/lsp/symbols.rs
```diff
@@ -614,33 +652,7 @@ mod tests {
...
-                line: 0,
-                character: 30,
-            },
-        };
-        let mut foo = new_symbol(String::from("foo"), SymbolKind::FUNCTION, range);
-        foo.children = Some(vec![bar]);
-        foo.detail = Some(String::from("function()"));
-
-        assert_eq!(test_symbol("foo <- function() { bar <- 1 }"), vec![foo]);
+        insta::assert_debug_snapshot!(test_symbol("foo <- function() { bar <- function() 1 }"));
```
DavisVaughan, line 658: I think it would be worth also having a snapshot test that sets `include_assignments_in_blocks` to `true` if you can easily expose that toggle

crates/ark/src/lsp/symbols.rs
```diff
@@ -834,4 +829,21 @@ class <- r6::r6class(
 "
         ));
     }
+
+    #[test]
+    // Assigned variables in nested contexts are not emitted as symbols
+    fn test_symbol_nested_assignments() {
+        insta::assert_debug_snapshot!(test_symbol(
+            "
```
DavisVaughan, line 840: And, to be super clear, at top level we do want assigned variables to be treated as symbols?

I honestly don't think that sounds super useful either

Like, a 300 line R script is bound to have a ton of assignments, and I'm not sure that seeing them in the outline feels that useful.

DavisVaughan (approved):

crates/ark/src/lsp/symbols.rs
lionel-: It's useful at least in packages with e.g. R6 classes.

In scripts top-level assignments allow to quickly look through things like assigned data frames. But you're right that lots of variables won't be very useful to have in there.

crates/ark/src/lsp/symbols.rs
lionel-: Good idea. (I'm assuming you mean testing nested assignments below, rather than making extra sure we always include functions here)

PR #860, opened 2025-06-27
Extend range of assignment symbols to end of node

lionel-:
Branched from #858 

This fix is similar to https://github.com/posit-dev/ark/pull/855

It makes sure the range of document symbol of assigned objects fully extends up to the end of the RHS. This allows the breadcrumbs to behave much better with the method symbols implemented in #858.

Before:

https://github.com/user-attachments/assets/737ab393-89c2-4471-a08a-1d50d7e78982

After:

https://github.com/user-attachments/assets/50d41990-2019-4f9b-9e32-8e699f31d6fe

### DIFF
--- a/crates/ark/src/lsp/symbols.rs
+++ b/crates/ark/src/lsp/symbols.rs
@@ -453,8 +453,8 @@ fn collect_assignment(
         // too busy.
         let name = contents.node_slice(&lhs)?.to_string();
 
-        let start = convert_point_to_position(contents, lhs.start_position());
-        let end = convert_point_to_position(contents, lhs.end_position());
+        let start = convert_point_to_position(contents, node.start_position());
+        let end = convert_point_to_position(contents, node.end_position());
 
         // Now recurse into RHS
         let mut children = Vec::new();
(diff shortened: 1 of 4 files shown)


crates/ark/src/lsp/symbols.rs
```diff
@@ -424,8 +424,8 @@ fn collect_assignment(
     // Otherwise, collect as generic object
     let name = contents.node_slice(&lhs)?.to_string();
 
-    let start = convert_point_to_position(contents, lhs.start_position());
-    let end = convert_point_to_position(contents, lhs.end_position());
+    let start = convert_point_to_position(contents, node.start_position());
+    let end = convert_point_to_position(contents, node.end_position());
 
     // Now recurse into RHS
```
DavisVaughan: I don't have super deep intuition about how this should work, but from what I can tell:

- If you have an assigned multi line function call, you get a breadcrumb
- If you have a free floating multi line function call, you do not

Is that expected / consistent? It _feels_ like maybe we should get a breadcrumb in both cases? Not totally sure. Probably worth comparing to other languages

https://github.com/user-attachments/assets/ab040749-a3c9-488f-91a8-e286b5c84e37

DavisVaughan (approved):
One comment about assigned vs unassigned function calls, but very nice little fix

crates/ark/src/lsp/symbols.rs
lionel-: I think that's expected, function calls are not emitted as document symbols. Otherwise every single function call would show up in the outline.

PR #861, opened 2025-06-30
Emit R6Class methods as workspace symbols

lionel-:
Addresses https://github.com/posit-dev/positron/issues/6549

- Add a bunch of tests for the workspace indexers which were not tested before.

- Add a new indexed type `Method`. This is used to include R6Class methods in workspace symbols but exclusing them from completions.

- Indexers now push symbols onto a list instead of returning a single symbol to their callers. This allows a single indexer to handle multiple symbols.

- The assignment handler now pushes R6Class method symbols in addition to the assigned object.

### QA Notes

Add the following to a file:

```r
class <- R6::R6Class(
    'class',
    public = list(
      initialize = function() 'initialize',
      foo = function() {
        1
      }
    ),
    private = list(
      bar = function() {
        2
      },
      not_indexed1 = NULL
    ),
    other = list(
      not_indexed2 = function() {
        3
      }
    )
  )
```

These symbols should now be available as Workspace symbols (`#` prefix in command palette): `initialize()`, `foo()`, `bar()` and you should be able to jump straight to the definition of these methods from any file.

The symbols starting with `not_indexed` should not be available.

All this is tested on the backend side.

### DIFF
--- a/crates/ark/src/lsp/indexer.rs
+++ b/crates/ark/src/lsp/indexer.rs
@@ -499,12 +499,14 @@ class <- R6Class(
         },
         public_method = function() {
             2
-        }
+        },
+        public_variable = NA
     ),
     private = list(
         private_method = function() {
             1
-        }
+        },
+        private_variable = NA
     ),
     other = list(
         other_method = function() {
--- a/crates/ark/src/lsp/snapshots/ark__lsp__indexer__tests__index_r6class.snap
+++ b/crates/ark/src/lsp/snapshots/ark__lsp__indexer__tests__index_r6class.snap
@@ -39,11 +39,11 @@ expression: entries
         key: "private_method",
         range: Range {
             start: Position {
-                line: 11,
+                line: 12,
                 character: 8,
             },
             end: Position {
-                line: 11,
+                line: 12,
                 character: 22,
             },
         },


crates/ark/src/lsp/completions/sources/composite/call.rs
```diff
@@ -267,6 +267,7 @@ fn completions_from_workspace_arguments(
             return Ok(None);
```
DavisVaughan, line 267: > `#` prefix in command palette

One of the best settings I've tweaked is

```
"search.quickOpen.includeSymbols": true
```

So that a simple `Cmd + P` brings up a search bar that merges both file names and workspace symbols

crates/ark/src/lsp/completions/sources/composite/call.rs
```diff
@@ -267,6 +267,7 @@ fn completions_from_workspace_arguments(
...
         },
         indexer::IndexEntryData::Variable { .. } => return Ok(None),
+        indexer::IndexEntryData::Method { .. } => return Ok(None),
     }
```
DavisVaughan, line 272: Would it be worth trying to give these a unique name? Like, trying to sniff out the R6 class name to include as well?

Imagine if I implemented two of three of these similar interfaces in the same file, then I'd have no way of telling them apart

What does Rust do for something like this?

https://github.com/DavisVaughan/almanac/blob/main/R/cache-rrule.R
https://github.com/DavisVaughan/almanac/blob/main/R/cache-radjusted.R

Image

crates/ark/src/treesitter.rs
```diff
@@ -571,3 +604,42 @@ pub(crate) fn node_find_containing_call<'tree>(node: Node<'tree>) -> Option<Node
...
+    ) -> Vec<tree_sitter::Node<'tree>> {
+        let mut result = Vec::new();
+
+        for m in self.cursor.matches(&self.query, node, contents) {
+            for cap in m.captures.iter() {
+                let cap_name = &self.query.capture_names()[cap.index as usize];
+                if *cap_name == capture_name {
+                    result.push(cap.node);
+                }
+            }
```
DavisVaughan, line 659: Do you want `cursor.matches()` or the flat `cursor.captures()`?

If I remember right, `captures()` is a flat list, but `matches()` has some structure that it doesn't look like you use

crates/ark/src/lsp/indexer.rs
```diff
@@ -198,127 +202,173 @@ fn index_document(document: &Document, path: &Path) {
...
+    };
+
+    let Some(rhs) = node.child_by_field_name("rhs") else {
+        return Ok(());
     };
 
+    if crate::treesitter::node_is_call(&rhs, "R6Class", contents) ||
+        crate::treesitter::node_is_namespaced_call(&rhs, "R6", "R6Class", contents)
+    {
+        index_r6_class(path, contents, &rhs, entries)?;
```
DavisVaughan: It should probably early return after this right? If not, maybe a comment about why fallthrough is good?

crates/ark/src/lsp/indexer.rs
```diff
@@ -198,127 +202,173 @@ fn index_document(document: &Document, path: &Path) {
...
+                let name = unwrap!(child.child_by_field_name("name"), None => continue);
+                if name.is_identifier() {
+                    let name = contents.node_slice(&name)?.to_string();
+                    arguments.push(name);
+                }
+            }
+        }
+
+        let start = convert_point_to_position(contents, lhs.start_position());
+        let end = convert_point_to_position(contents, lhs.end_position());
```
DavisVaughan, line 291: Just checking, do you use `lhs.start_position()` or `node.start_position()`?

Same question below in the `// Otherwise, emit variable` path

crates/ark/src/lsp/indexer.rs
```diff
@@ -198,127 +202,173 @@ fn index_document(document: &Document, path: &Path) {
...
+            name: (identifier) @method_name
+            value: (function_definition) @method_fn
+          )
+        )
+      )
+      (#match? @access "public|private")
+      (#eq? @_list_fn "list")
+    )
+    "#;
+    let mut ts_query = crate::treesitter::TSQuery::new(query_str)?;
```
DavisVaughan: One thing I will note is that tree-sitter recommends that you "compile" your static `Query` objects exactly once ahead of time, because they do take a non zero amount of overhead to parse and compile.

Would it be possible to rearrange things to end up with a `Lazy<Query>` that is initialized once and then fed to your `TSQuery` helper?

crates/ark/src/lsp/indexer.rs
```diff
@@ -198,127 +202,173 @@ fn index_document(document: &Document, path: &Path) {
...
+
+    // We'll switch from Rope to String in the near future so let's not
+    // worry about this conversion now
+    let contents_str = contents.to_string();
```
DavisVaughan, line 342: Agreed, should be nicer all around

DavisVaughan: If you wanted to be super fancy I believe you could implement `TextProvider` for `Rope`

https://github.com/tree-sitter/tree-sitter/blob/618b9dd66e1bd529d9757202cca5a99d333ce394/lib/binding_rust/lib.rs#L3092

https://github.com/tree-sitter/tree-sitter/blob/618b9dd66e1bd529d9757202cca5a99d333ce394/lib/binding_rust/lib.rs#L349-L355

It's _probably_ not worth it but i think the idea is that you can somehow provide some iterator when you don't have a complete contiguous buffer available

DavisVaughan: Actually it might be easier than that? It looks like they already provide an implementation of `TextProvider` for `FnMut(Node) -> <iterator over bytes>` 

https://github.com/tree-sitter/tree-sitter/blob/618b9dd66e1bd529d9757202cca5a99d333ce394/lib/binding_rust/lib.rs#L3561-L3580

So you could probably provide something like

```
mut |node: Node| {
  iter::once(contents.node_slice(&node).unwrap().to_string().as_bytes())
}
```

which would materialize a `node` worth of text in `to_string()` but not the whole document

It's not allowed to fail so I guess you have to `unwrap()`

May or may not be worth it, just a thought

crates/ark/src/lsp/indexer.rs
DavisVaughan: oh i see this in the test, maybe a comment here too?

```
    // Note that unlike document symbols whose ranges cover the whole entity
    // they represent, the range of workspace symbols only cover the identifers
```

crates/ark/src/lsp/indexer.rs
```diff
@@ -328,15 +378,152 @@ fn index_comment(_path: &Path, contents: &Rope, node: &Node) -> anyhow::Result<O
...
+        },
+        public_method = function() {
+            2
+        }
+    ),
+    private = list(
+        private_method = function() {
+            1
+        }
+    ),
```
DavisVaughan: Also include a private variable to ensure it isn't indexed?

Like
https://github.com/DavisVaughan/almanac/blob/7b14f6e8f1e685975231e5dadb40bb5bb8f2a9c8/R/cache-radjusted.R#L16

crates/ark/src/lsp/completions/sources/composite/call.rs
lionel-: Rust doesn't emit methods as Workspace symbols so this issue doesn't arise.

I think the file name is good enough to disambiguate most of the time? But we could probably make your suggestion work like this: `MyClass$get_events()`

crates/ark/src/treesitter.rs
lionel-: hmm that sounded plausible but it looks like I can't access the capture name with the `captures()` iterator

crates/ark/src/lsp/indexer.rs
lionel-: I've renamed to make it clear only methods are indexed, and added a comment explaining the fallthrough:

```rust
        index_r6_class_methods(path, contents, &rhs, entries)?;
        // Fallthrough to index the variable to which the R6 class is assigned
```

crates/ark/src/lsp/indexer.rs
lionel-: The problem is that we then need to split the query and the query cursor because the latter is stateful. I wasn't too worried about the overhead and thought I'd favour ergonomics.

crates/ark/src/lsp/indexer.rs
lionel-: Right, I didn't look into these more complex setups since we're going to switch to a flat document string soon that's straightforward to iterate over.

crates/ark/src/lsp/completions/sources/composite/call.rs
DavisVaughan: I think that would be kind of cool to also make it distinct as an R6 method? Maybe no parens, so `MyClass$get_events`

crates/ark/src/treesitter.rs
DavisVaughan: No matter what you use, I think you still get the name via `self.query.capture_names()[cap.index as usize]`, this just avoids the double loop?

crates/ark/src/lsp/indexer.rs
DavisVaughan: > we then need to split the query and the query cursor

tbh that sounds like a feature and is how tree-sitter is supposed to work

PR #865, opened 2025-07-05
Simplify LSP settings

lionel-:
I think you'll like this one @DavisVaughan.

Branched from #859. I was concerned at how hard it was to add a new setting and section. I think the refactor in this PR will make things much simpler and easier:

- Add a generic `Setting` type that allows us to flatten a representation of all our settings in a flat array `SETTINGS` that is easy to loop over. This drastically simplify the updating logic where we send all keys of interest to the client in a flat array, as we no longer need any bookkeeping.

- The setting objects in this array contain the conversion-from-json logic as simple methods assigned to a closure field. The default handling is still delegated to `Default` methods in our setting types.

- Remove the split between VS Code and LSP settings. If we want to add vscode agnostic settings in the future, we could add editor aliases in the `SETTINGS` array..

### DIFF
--- a/crates/ark/src/lsp/handlers.rs
+++ b/crates/ark/src/lsp/handlers.rs
@@ -9,7 +9,6 @@ use anyhow::anyhow;
 use serde_json::Value;
 use stdext::unwrap;
 use stdext::unwrap::IntoResult;
-use struct_field_names_as_array::FieldNamesAsArray;
 use tower_lsp::lsp_types::CodeActionParams;
 use tower_lsp::lsp_types::CodeActionResponse;
 use tower_lsp::lsp_types::CompletionItem;
@@ -46,9 +45,6 @@ use crate::lsp;
 use crate::lsp::code_action::code_actions;
 use crate::lsp::completions::provide_completions;
 use crate::lsp::completions::resolve_completion;
-use crate::lsp::config::VscDiagnosticsConfig;
-use crate::lsp::config::VscDocumentConfig;
-use crate::lsp::config::VscSymbolsConfig;
 use crate::lsp::definitions::goto_definition;
 use crate::lsp::document_context::DocumentContext;
 use crate::lsp::encoding::convert_lsp_range_to_tree_sitter_range;
@@ -106,22 +102,16 @@ pub(crate) async fn handle_initialized(
         // Note that some settings, such as editor indentation properties, may be
         // changed by extensions or by the user without changing the actual
         // underlying setting. Unfortunately we don't receive updates in that case.
-        let mut config_document_regs = collect_regs(
-            VscDocumentConfig::FIELD_NAMES_AS_ARRAY.to_vec(),
-            VscDocumentConfig::section_from_key,
-        );
-        let mut config_symbols_regs: Vec<Registration> = collect_regs(
-            VscSymbolsConfig::FIELD_NAMES_AS_ARRAY.to_vec(),
-            VscSymbolsConfig::section_from_key,
-        );
-        let mut config_diagnostics_regs: Vec<Registration> = collect_regs(
-            VscDiagnosticsConfig::FIELD_NAMES_AS_ARRAY.to_vec(),
-            VscDiagnosticsConfig::section_from_key,
-        );
 
-        regs.append(&mut config_document_regs);
-        regs.append(&mut config_symbols_regs);
-        regs.append(&mut config_diagnostics_regs);
+        use crate::lsp::config::SETTINGS;
+
+        for setting in SETTINGS {
+            regs.push(Registration {
+                id: uuid::Uuid::new_v4().to_string(),
+                method: String::from("workspace/didChangeConfiguration"),
+                register_options: Some(serde_json::json!({ "section": setting.key })),
+            });
+        }
     }
 
     client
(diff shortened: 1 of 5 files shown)


crates/ark/src/lsp/handlers.rs
```diff
@@ -106,22 +102,16 @@ pub(crate) async fn handle_initialized(
...
-            VscDiagnosticsConfig::section_from_key,
-        );
-
-        regs.append(&mut config_document_regs);
-        regs.append(&mut config_symbols_regs);
-        regs.append(&mut config_diagnostics_regs);
+
+        use crate::lsp::config::SETTINGS;
+
+        for setting in SETTINGS {
```
```suggestion
        for setting in crate::lsp::config::SETTINGS {
```

That feels very AI 😆

crates/ark/src/lsp/state_handlers.rs
```diff
@@ -284,121 +275,35 @@ pub(crate) fn did_change_formatting_options(
     // `insert_final_newline`
 }
 
+use crate::lsp::config::SETTINGS;
```
DavisVaughan: oh god please don't put a use declaration here

if this is ai please update the ai instructions md file

crates/ark/src/lsp/state_handlers.rs
```diff
@@ -284,121 +275,35 @@ pub(crate) fn did_change_formatting_options(
...
-    // interest in a flat vector
-    let document_keys = VscDocumentConfig::FIELD_NAMES_AS_ARRAY;
-    let mut document_items: Vec<ConfigurationItem> =
-        itertools::iproduct!(uris.iter(), document_keys.iter())
-            .map(|(uri, key)| ConfigurationItem {
-                scope_uri: Some(uri.clone()),
-                section: Some(VscDocumentConfig::section_from_key(key).into()),
-            })
-            .collect();
-    items.append(&mut document_items);
```
DavisVaughan: Hmm isn't this important? I'm pretty sure we collected document config on a _per open document basis_ and now we just collect it once with a `scope_uri: None` which doesn't seem right

crates/ark/src/lsp/state_handlers.rs
```diff
@@ -284,121 +275,35 @@ pub(crate) fn did_change_formatting_options(
...
-    // Deserialise the VS Code configuration
-    let config: VscDiagnosticsConfig = serde_json::from_value(serde_json::Value::Object(map))?;
-    let config: DiagnosticsConfig = config.into();
-
-    let changed = state.config.diagnostics != config;
-    state.config.diagnostics = config;
-
-    if changed {
-        lsp::spawn_diagnostics_refresh_all(state.clone());
-    }
```
DavisVaughan: We dropped a refresh here

crates/ark/src/lsp/state_handlers.rs
```diff
@@ -284,121 +275,35 @@ pub(crate) fn did_change_formatting_options(
...
-        });
-
-        // Deserialise the VS Code configuration
-        let config: VscDocumentConfig = serde_json::from_value(serde_json::Value::Object(map))?;
-
-        // Now convert the VS Code specific type into our own type
-        let config: DocumentConfig = config.into();
-
-        // Finally, update the document's config
-        state.get_document_mut(&uri)?.config = config;
```
DavisVaughan, line 401: Yea see we dropped the per document config update

crates/ark/src/lsp/handlers.rs
lionel-: yeah I keep having to edit these, I'll try the copilot instructions

crates/ark/src/lsp/state_handlers.rs
lionel-: I think I was actually fine with the declaration of this global close to its handler (and I think I've similarly pulled global decls down in other places) but I see you feel strongly about this

DavisVaughan: I do fairly strongly prefer managing all imports up top

lionel-:
Sorry I totally missed we lost the doc updates when the AI moved code around!

I've restored the logic. Note that Positron/Code do not support per-document settings so we can't actually test this with our IDE. This is mostly an overcomplication for now but this will be useful with other IDEs like vim or Emacs, perhaps Zed.

I've checked that global settings are updated as expected, diagnostics in particular (which do need the refresh):


https://github.com/user-attachments/assets/7186908d-b214-4ed0-b8d3-54c552142c77

crates/ark/src/lsp/state_handlers.rs
```diff
@@ -342,63 +330,32 @@ async fn update_config(
...
+        // split and updates the vector with the head
+        let tail = remaining.split_off(DOCUMENT_SETTINGS.len());
+        let head = std::mem::replace(&mut remaining, tail);
+
+        for (mapping, value) in DOCUMENT_SETTINGS.iter().zip(head) {
+            if let Ok(doc) = state.get_document_mut(&uri) {
+                (mapping.set)(&mut doc.config, value);
+            }
+        }
+    }
```
DavisVaughan, line 353: I was imagining this double loop, which I find easier to understand?

```rust
    let mut i = 0;

    // Iterate `n_uris * n_document_settings` total times, setting each document setting
    // on a per-uri basis
    for uri in uris.into_iter() {
        let uri = state.get_document_mut(&uri);

        let Ok(doc) = uri else {
            i += DOCUMENT_SETTINGS.len();
            continue;
        };

        for setting in DOCUMENT_SETTINGS.iter() {
            let value = std::mem::take(&mut document_configs[i]);
            (setting.set)(&mut doc.config, value);
            i += 1;
        }
    }
```

Requires `let mut document_configs` above as well

DavisVaughan (approved):

crates/ark/src/lsp/state_handlers.rs
lionel-: Hmm I think I prefer the iterator tools than a raw loop with an index

PR #866, opened 2025-07-07
Exclude comment sections from workspace symbols by default

lionel-:
Branched from #865.
Addresses https://github.com/posit-dev/positron/issues/4886.

When a comment section like `# name ---` clashes with an actual symbol like `name <- function() {}`, whichever symbols comes first (usually the section) is included in the workspace index, and the other is ignored. Ideally we'd track both but for now we only track one identifier per file.

This causes sections to win over function definitions in cases like this:

```r
# my_section ----

my_section <- function() {}

my_section # Reference
```

If you then jump to definition on a reference to that function, Positron jumps to the section instead of the function.

* To fix this, we now allow functions and variables to overwrite sections in the indexer.

* In addition, we no longer emit sections as workspace symbols by default. This can be turned back on with a new setting `positron.r.workspaceSymbols.includeCommentSections`. This is consistent with the fix in https://github.com/quarto-dev/quarto/pull/755 where we no longer export markdown headers as workspace symbols by default, to avoid flooding workspace symbol quickpicks (`#` prefix in command palette) with section headers.

### QA Notes

- You should be able to command+quick on the `my_section` reference and be taken to the function definition rather than the section. This should be the case no matter the value of `positron.r.workspaceSymbols.includeCommentSections`.

- When `positron.r.workspaceSymbols.includeCommentSections` is set to `true`, you should see comment sections in the workspace symbol quickpick. Otherwise they shouldn't be included.

### DIFF
--- a/crates/ark/src/lsp/symbols.rs
+++ b/crates/ark/src/lsp/symbols.rs
@@ -587,6 +587,7 @@ mod tests {
     use crate::lsp::config::WorkspaceSymbolsConfig;
     use crate::lsp::documents::Document;
     use crate::lsp::indexer::ResetIndexerGuard;
+    use crate::lsp::util::test_path;
 
     fn test_symbol(code: &str) -> Vec<DocumentSymbol> {
         let doc = Document::new(code, None);
@@ -914,7 +915,8 @@ a <- function() {
 
             // Index the document
             let doc = Document::new(code, None);
-            indexer::update(&doc, std::path::Path::new("/test.R")).unwrap();
+            let (path, _) = test_path();
+            indexer::update(&doc, &path).unwrap();
 
             // Query for all symbols
             let params = WorkspaceSymbolParams {
--- a/crates/ark/src/lsp/definitions.rs
+++ b/crates/ark/src/lsp/definitions.rs
@@ -78,14 +78,13 @@ pub fn goto_definition<'a>(
 
 #[cfg(test)]
 mod tests {
-    use std::path::PathBuf;
-
     use assert_matches::assert_matches;
     use tower_lsp::lsp_types;
 
     use super::*;
     use crate::lsp::documents::Document;
     use crate::lsp::indexer;
+    use crate::lsp::util::test_path;
 
     #[test]
     fn test_goto_definition() {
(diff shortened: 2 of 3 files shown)


crates/ark/src/lsp/indexer.rs
```diff
@@ -121,10 +121,17 @@ fn insert(path: &Path, entry: IndexEntry) -> anyhow::Result<()> {
...
+    // future we'll track every occurrences and their scopes but for now we
+    // only track the first definition of an object (in a way, its
     // declaration).
-    if !index.contains_key(&entry.key) {
+    if let Some(existing_entry) = index.get(&entry.key) {
+        // Give priority to non-section entries.
+        if matches!(existing_entry.data, IndexEntryData::Section { .. }) {
+            index.insert(entry.key.clone(), entry);
+        }
+        // Else, ignore.
```
DavisVaughan, line 143: I think a test for this behavior would be nice

crates/ark/src/lsp/symbols.rs
```diff
@@ -81,18 +84,21 @@ pub fn symbols(params: &WorkspaceSymbolParams) -> anyhow::Result<Vec<SymbolInfor
...
-                    kind: SymbolKind::STRING,
-                    location: Location {
-                        uri: Url::from_file_path(path).unwrap(),
-                        range: entry.range,
-                    },
-                    tags: None,
-                    deprecated: None,
-                    container_name: None,
-                });
+                if state.config.workspace_symbols.include_comment_sections {
```
DavisVaughan, line 87: Would enjoy seeing a test for this (on and off) as well if possible

crates/ark/src/lsp/symbols.rs
```diff
@@ -81,18 +84,21 @@ pub fn symbols(params: &WorkspaceSymbolParams) -> anyhow::Result<Vec<SymbolInfor
...
+                        tags: None,
+                        deprecated: None,
+                        container_name: None,
+                    });
+                }
             },
+
             IndexEntryData::Variable { name } => {
                 info.push(SymbolInformation {
                     name: name.clone(),
```
DavisVaughan: Could also add a nice test in `goto_definition` for

```
# my_section ----

my_section <- function() {}
```

to make sure that `my_section` is chosen

I know I'm also asking for a similar test in `indexer` itself, but it feels like having a duplicate test in both places is a good idea (so if we ever see the `indexer` test and think "why is this here?", then we'd also have the goto-definition test to show the real reason for it)

DavisVaughan (approved):
Looks good to me, mostly just want some test coverage since this is a bit tricky

lionel-:
We've now got the requested test coverage and a test-only RAII indexer guard to help clean up the index after tests have run.

PR #870, opened 2025-07-11
Static diagnostics for `library()` and `require()` calls

lionel-:
Just needs a few more tests and documenting the PR:

Screenshot 2025-07-11 at 18 30 27

### DIFF
--- a/crates/ark/src/lsp/diagnostics.rs
+++ b/crates/ark/src/lsp/diagnostics.rs
@@ -24,6 +24,7 @@ use crate::lsp::diagnostics_syntax::syntax_diagnostics;
 use crate::lsp::documents::Document;
 use crate::lsp::encoding::convert_tree_sitter_range_to_lsp_range;
 use crate::lsp::indexer;
+use crate::lsp::inputs::library::Library;
 use crate::lsp::state::WorldState;
 use crate::lsp::traits::rope::RopeExt;
 use crate::treesitter::node_has_error_or_missing;
@@ -55,6 +56,14 @@ pub struct DiagnosticContext<'a> {
     // The set of packages that are currently installed.
     pub installed_packages: HashSet<String>,
 
+    /// Reference to the library for looking up package exports.
+    pub library: &'a Library,
+
+    /// The symbols exported by packages loaded via `library()` calls in this
+    /// document. Currently global. TODO: Store individual exports in a BTreeMap
+    /// sorted by position in the source?
+    pub library_symbols: HashSet<String>,
+
     // Whether or not we're inside of a formula.
     pub in_formula: bool,
 
@@ -69,13 +78,15 @@ impl Default for DiagnosticsConfig {
 }
 
 impl<'a> DiagnosticContext<'a> {
-    pub fn new(contents: &'a Rope) -> Self {
+    pub fn new(contents: &'a Rope, library: &'a crate::lsp::inputs::library::Library) -> Self {
         Self {
             contents,
             document_symbols: Vec::new(),
             session_symbols: HashSet::new(),
             workspace_symbols: HashSet::new(),
             installed_packages: HashSet::new(),
+            library,
+            library_symbols: HashSet::new(),
             in_formula: false,
             in_call_like_arguments: false,
         }
@@ -86,16 +97,22 @@ impl<'a> DiagnosticContext<'a> {
         symbols.insert(name.to_string(), location);
     }
 
-    pub fn has_definition(&mut self, name: &str) -> bool {
+    pub fn has_definition(&self, name: &str) -> bool {
         // First, check document symbols.
-        for symbols in self.document_symbols.iter() {
+        for symbols in &self.document_symbols {
             if symbols.contains_key(name) {
                 return true;
             }
         }
 
-        // Next, check workspace symbols.
+        // Then, check workspace symbols.
         if self.workspace_symbols.contains(name) {
+            return true;
+        }
+
+        // Finally, check library symbols from library() calls.
+        // TODO: Take `Node` and check by position.
+        if self.library_symbols.contains(name) {
             return true;
         }
 
@@ -118,7 +135,7 @@ pub(crate) fn generate_diagnostics(doc: Document, state: WorldState) -> Vec<Diag
         return diagnostics;
     }
 
-    let mut context = DiagnosticContext::new(&doc.contents);
+    let mut context = DiagnosticContext::new(&doc.contents, &state.library);
 
     // Add a 'root' context for the document.
     context.document_symbols.push(HashMap::new());
@@ -780,11 +797,62 @@ fn recurse_call(
     let fun = fun.as_str();
 
     match fun {
+        "library" => {
+            // Track symbols exported by `library()` calls
+            handle_library_call(node, context, diagnostics)?;
+        },
         // default case: recurse into each argument
         _ => recurse_call_like_arguments_default(node, context, diagnostics)?,
     };
 
     ().ok()
+}
+
+fn handle_library_call(
+    node: Node,
+    context: &mut DiagnosticContext,
+    diagnostics: &mut Vec<Diagnostic>,
+) -> anyhow::Result<()> {
+    // Get the arguments node
+    let Some(arguments) = node.child_by_field_name("arguments") else {
+        return Ok(());
+    };
+
+    // Find the first argument (package name)
+    let mut cursor = arguments.walk();
+    let mut children = arguments.children_by_field_name("argument", &mut cursor);
+
+    let Some(first_arg) = children.next() else {
+        return Ok(());
+    };
+
+    // Get the package name from the argument value
+    let Some(value) = first_arg.child_by_field_name("value") else {
+        return Ok(());
+    };
+
+    let package_name = if value.is_identifier() {
+        context.contents.node_slice(&value)?.to_string()
+    } else if value.is_string() {
+        // Remove quotes from string literal
+        let raw = context.contents.node_slice(&value)?.to_string();
+        raw.trim_matches('"').trim_matches('\'').to_string()
+    } else {
+        return Ok(());
+    };
+
+    // Try to get the package from the library
+    if let Some(package) = context.library.get(&package_name) {
+        // Add all exported symbols to library_symbols
+        for symbol in &package.namespace.exports {
+            context.library_symbols.insert(symbol.clone());
+        }
+    }
+
+    // Continue with default recursion to handle any other arguments
+    recurse_call_like_arguments_default(node, context, diagnostics)?;
+
+    Ok(())
 }
 
 fn recurse_subset_or_subset2(
@@ -1454,5 +1522,93 @@ foo
                 0
             );
         })
+    }
+
+    #[test]
+    fn test_library_static_exports() {
+        r_task(|| {
+            use std::path::PathBuf;
+
+            use crate::lsp::inputs::library::Library;
+            use crate::lsp::inputs::package::Description;
+            use crate::lsp::inputs::package::Namespace;
+            use crate::lsp::inputs::package::Package;
+
+            // `mockpkg` exports `foo` and `bar`
+            let namespace = Namespace {
+                exports: vec!["foo".to_string(), "bar".to_string()],
+                imports: vec![],
+                bulk_imports: vec![],
+            };
+            let description = Description {
+                name: "mockpkg".to_string(),
+                version: "1.0.0".to_string(),
+                depends: vec![],
+            };
+            let package = Package {
+                path: PathBuf::from("/mock/path"),
+                description,
+                namespace,
+            };
+
+            // Create a library with `mockpkg` installed
+            let library = Library::new(vec![]).insert("mockpkg", package);
+
+            // Simulate a search path with `library` in scope
+            let console_scopes = vec![vec!["library".to_string()]];
+
+            // Whereas `DEFAULT_STATE` contains base package attached, this world state
+            // only contains `mockpkg` as installed package and `library()` on
+            // the search path.
+            let state = WorldState {
+                library,
+                console_scopes,
+                ..Default::default()
+            };
+
+            // Test that exported symbols are recognized
+            let code = "
+                library(mockpkg)
+                foo()
+                bar
+            ";
+            let document = Document::new(code, None);
+            let diagnostics = generate_diagnostics(document, state.clone());
+
+            assert_eq!(diagnostics.len(), 0);
+
+            // Test that non-exported symbols still generate diagnostics
+            let code = "
+                library(mockpkg)
+                undefined()
+                also_undefined
+            ";
+            let document = Document::new(code, None);
+
+            let diagnostics = generate_diagnostics(document, state.clone());
+            assert_eq!(diagnostics.len(), 2);
+
+            assert!(diagnostics
+                .get(0)
+                .unwrap()
+                .message
+                .contains("No symbol named 'undefined' in scope"));
+            assert!(diagnostics
+                .get(1)
+                .unwrap()
+                .message
+                .contains("No symbol named 'also_undefined' in scope"));
+
+            // Test duplicate call
+            let code = "
+                library(mockpkg)
+                library(mockpkg)  # duplicate is fine
+                foo()
+                bar
+            ";
+            let document = Document::new(code, None);
+            let diagnostics = generate_diagnostics(document, state);
+            assert_eq!(diagnostics.len(), 0);
+        });
     }
 }
--- a/crates/ark/src/lsp/diagnostics.rs
+++ b/crates/ark/src/lsp/diagnostics.rs
@@ -5,6 +5,7 @@
 //
 //
 
+use std::collections::BTreeMap;
 use std::collections::HashMap;
 use std::collections::HashSet;
 
@@ -17,6 +18,7 @@ use stdext::*;
 use tower_lsp::lsp_types::Diagnostic;
 use tower_lsp::lsp_types::DiagnosticSeverity;
 use tree_sitter::Node;
+use tree_sitter::Point;
 use tree_sitter::Range;
 
 use crate::lsp;
@@ -64,7 +66,7 @@ pub struct DiagnosticContext<'a> {
     /// The symbols exported by packages loaded via `library()` calls in this
     /// document. Currently global. TODO: Store individual exports in a BTreeMap
     /// sorted by position in the source?
-    pub library_symbols: HashSet<String>,
+    pub library_symbols: BTreeMap<Point, HashSet<String>>,
 
     // Whether or not we're inside of a formula.
     pub in_formula: bool,
@@ -88,7 +90,7 @@ impl<'a> DiagnosticContext<'a> {
             workspace_symbols: HashSet::new(),
             installed_packages: HashSet::new(),
             library,
-            library_symbols: HashSet::new(),
+            library_symbols: BTreeMap::new(),
             in_formula: false,
             in_call_like_arguments: false,
         }
@@ -99,8 +101,8 @@ impl<'a> DiagnosticContext<'a> {
         symbols.insert(name.to_string(), location);
     }
 
-    pub fn has_definition(&self, name: &str) -> bool {
-        // First, check document symbols.
+    // First, check document symbols.
+    pub fn has_definition(&self, name: &str, start_position: Point) -> bool {
         for symbols in &self.document_symbols {
             if symbols.contains_key(name) {
                 return true;
@@ -112,10 +114,15 @@ impl<'a> DiagnosticContext<'a> {
             return true;
         }
 
-        // Finally, check library symbols from library() calls.
-        // TODO: Take `Node` and check by position.
-        if self.library_symbols.contains(name) {
-            return true;
+        // Finally, check package symbols from `library()` calls.
+        // Check all symbols exported by `library()` before the given position.
+        for (library_position, exports) in self.library_symbols.iter() {
+            if *library_position > start_position {
+                break;
+            }
+            if exports.contains(name) {
+                return true;
+            }
         }
 
         // Finally, check session symbols.
@@ -823,7 +830,12 @@ fn handle_library_call(node: Node, context: &mut DiagnosticContext) -> anyhow::R
     // Insert exports globablly for now
     if let Some(package) = context.library.get(&package_name) {
         for symbol in &package.namespace.exports {
-            context.library_symbols.insert(symbol.clone());
+            let pos = node.end_position();
+            context
+                .library_symbols
+                .entry(pos)
+                .or_insert_with(HashSet::new)
+                .insert(symbol.clone());
         }
     } else {
         lsp::log_warn!("Can't get exports from package {package_name} because it is not installed.")
@@ -990,7 +1002,7 @@ fn check_symbol_in_scope(
 
     // Skip if a symbol with this name is in scope.
     let name = context.contents.node_slice(&node)?.to_string();
-    if context.has_definition(name.as_str()) {
+    if context.has_definition(name.as_str(), node.start_position()) {
         return false.ok();
     }
 
@@ -1008,6 +1020,8 @@ fn check_symbol_in_scope(
 
 #[cfg(test)]
 mod tests {
+    use std::path::PathBuf;
+
     use harp::eval::RParseEvalOptions;
     use once_cell::sync::Lazy;
     use tower_lsp::lsp_types::Position;
@@ -1015,6 +1029,10 @@ mod tests {
     use crate::interface::console_inputs;
     use crate::lsp::diagnostics::generate_diagnostics;
     use crate::lsp::documents::Document;
+    use crate::lsp::inputs::library::Library;
+    use crate::lsp::inputs::package::Description;
+    use crate::lsp::inputs::package::Namespace;
+    use crate::lsp::inputs::package::Package;
     use crate::lsp::state::WorldState;
     use crate::r_task;
 
@@ -1504,13 +1522,6 @@ foo
     #[test]
     fn test_library_static_exports() {
         r_task(|| {
-            use std::path::PathBuf;
-
-            use crate::lsp::inputs::library::Library;
-            use crate::lsp::inputs::package::Description;
-            use crate::lsp::inputs::package::Namespace;
-            use crate::lsp::inputs::package::Package;
-
             // `mockpkg` exports `foo` and `bar`
             let namespace = Namespace {
                 exports: vec!["foo".to_string(), "bar".to_string()],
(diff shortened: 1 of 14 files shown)


crates/ark/src/lsp/inputs/source_root.rs
```diff
@@ -0,0 +1,16 @@
+//
```
lionel-, line 1: I didn't end up using this here but will use it in the next PR for package imports.

crates/ark/src/lsp/traits/node.rs
```diff
@@ -90,6 +91,12 @@ pub trait NodeExt: Sized {
     fn bwd_leaf_iter(&self) -> BwdLeafIterator<'_>;
 
     fn ancestors(&self) -> impl Iterator<Item = Self>;
+    fn children_of(node: Self) -> impl Iterator<Item = Self>;
+    fn next_siblings(&self) -> impl Iterator<Item = Self>;
+    fn arguments(&self) -> impl Iterator<Item = (Option<Self>, Option<Self>)>;
+    fn arguments_values(&self) -> impl Iterator<Item = Self>;
+    fn arguments_names(&self) -> impl Iterator<Item = Self>;
+    fn arguments_names_as_string(&self, contents: &ropey::Rope) -> impl Iterator<Item = String>;
```
lionel-: Bunch of new helpers for iteration over TS nodes using the cursor API. I didn't end up using all of them but they could be useful later.

crates/ark/src/lsp/traits/node.rs
```diff
@@ -199,13 +206,96 @@ impl<'tree> NodeExt for Node<'tree> {
...
-    // From rowan. Note that until we switch to rowan, each `parent()` call
-    // causes a linear traversal of the whole tree to find the parent node.
-    // We could do better in the future:
-    // https://github.com/tree-sitter/tree-sitter/pull/3214
     fn ancestors(&self) -> impl Iterator<Item = Node<'tree>> {
+        // We'd ideally use the cursor API here too but
+        // `ts_tree_cursor_goto_parent()` doesn't behave like
+        // `ts_node_parent()`: the latter traverses `ERROR` nodes but not the
+        // former. So for now we accept the performance hit of tree traversal at
+        // each `parent()` call.
```
lionel-, line 217: That was a bummer.

lionel-:
Another thing to consider in the future is transitive imports.

File A:

```r
library(ggplot2)
```

File B:

```r
source("file_a.R")
ggplot() # In scope
```

The `ggplot` symbol is imported via `source()`.

For now this will have to be worked around by evaluating the `source()` call or the relevant `library()` calls.

crates/ark/src/lsp/main_loop.rs
```diff
@@ -174,13 +176,30 @@ impl GlobalState {
...
+                .try_into()?;
+
+            log::info!("Using library paths: {paths:#?}");
+            let paths: Vec<PathBuf> = paths.into_iter().map(PathBuf::from).collect();
+            state.world.library = Library::new(paths);
+
+            Ok(())
+        }) {
+            log::error!("Can't evaluate `libPaths()`: {err:?}");
+        };
```
DavisVaughan, line 200: What if you made this a `KernelNotification` that the kernel sends the LSP in `refresh_lsp()`?

Then you'd get updates to `.libPaths()` as well, like if a user changes it after startup, which is totally possible.

(This assumes you'd have a way to throw out and rebuild any information that utilized the libpaths when they change, but I figure we would have this)

crates/ark/src/lsp/inputs/package.rs
```diff
@@ -0,0 +1,59 @@
...
+use std::path::PathBuf;
+
+use crate::lsp::inputs::package_description::Description;
+use crate::lsp::inputs::package_namespace::Namespace;
+
+/// Represents an R package and its metadata relevant for static analysis.
+#[derive(Clone, Debug)]
+pub struct Package {
+    /// Path to the directory that contains `DESCRIPTION`. Could be an installed
+    /// package, or a package source.
```
DavisVaughan:
```suggestion
    /// Path to the directory that contains `DESCRIPTION` and `NAMESPACE`.
    /// Could be an installed package, or a package source.
```

It looks like you need both, from the `load()` impl

crates/ark/src/lsp/inputs/package.rs
```diff
@@ -0,0 +1,59 @@
...
+        let package_path = lib_path.join(name);
+
+        let description_path = package_path.join("DESCRIPTION");
+        let namespace_path = package_path.join("NAMESPACE");
+
+        // Only consider libraries that have a folder named after the
+        // requested package and that contains a description file
+        if !description_path.is_file() {
+            return Ok(None);
+        }
```
DavisVaughan, line 37: You seem to also require a NAMESPACE below, so bail early here too?

crates/ark/src/lsp/inputs/package.rs
```diff
@@ -0,0 +1,59 @@
...
+            return Err(anyhow::anyhow!(
+                "`Package` field in `DESCRIPTION` doesn't match folder name '{name}'"
+            ));
+        }
+
+        let namespace_contents = fs::read_to_string(&namespace_path)?;
+        let namespace = Namespace::parse(&namespace_contents)?;
+
+        Ok(Some(Package {
+            path: package_path.to_path_buf(),
```
DavisVaughan:
```suggestion
            path: package_path,
```

Already a PathBuf

crates/ark/src/lsp/traits/node.rs
DavisVaughan: Why does this matter to you? I do not trust tree-sitter's error recovered tree enough to try and do anything useful when there are ERRORs in the tree. i.e. if there are syntax errors in the file, bail entirely for right now.

crates/ark/src/lsp/traits/node.rs
```diff
@@ -199,13 +206,96 @@ impl<'tree> NodeExt for Node<'tree> {
...
+            None
+        };
+
+        std::iter::successors(first, move |_| {
+            if cursor.goto_next_sibling() {
+                Some(cursor.node())
+            } else {
+                None
+            }
+        })
```
DavisVaughan: It actually feels like `std::iter::from_fn()` would be way more ergonomic here. IIUC you won't even need to special case the first sibling

crates/ark/src/lsp/traits/node.rs
```diff
@@ -90,6 +91,12 @@ pub trait NodeExt: Sized {
     fn bwd_leaf_iter(&self) -> BwdLeafIterator<'_>;
 
     fn ancestors(&self) -> impl Iterator<Item = Self>;
+    fn children_of(node: Self) -> impl Iterator<Item = Self>;
+    fn next_siblings(&self) -> impl Iterator<Item = Self>;
+    fn arguments(&self) -> impl Iterator<Item = (Option<Self>, Option<Self>)>;
+    fn arguments_values(&self) -> impl Iterator<Item = Self>;
+    fn arguments_names(&self) -> impl Iterator<Item = Self>;
```
DavisVaughan:
```suggestion
    fn iter_children(node: Self) -> impl Iterator<Item = Self>;
    fn iter_following_siblings(&self) -> impl Iterator<Item = Self>;
    fn iter_arguments(&self) -> impl Iterator<Item = (Option<Self>, Option<Self>)>;
    fn iter_arguments_values(&self) -> impl Iterator<Item = Self>;
    fn iter_arguments_names(&self) -> impl Iterator<Item = Self>;
```

How do we feel about naming of iterators? There's a strong Rust convention of `.iter()` to get a non-mutable iterator over something, so maybe these could be renamed like this to be very self documenting? Just a thought.

I do like that it differentiates `iter_children()`, which clearly returns an iterator, from `children()`, which on first glance you can't tell if it returns an iterator or collects the actual `Vec<Node>` of children nodes.

crates/ark/src/lsp/traits/node.rs
```diff
@@ -199,13 +206,96 @@ impl<'tree> NodeExt for Node<'tree> {
...
+                cursor.goto_next_sibling()
+            };
+
+            if advanced {
+                Some(cursor.node())
+            } else {
+                None
+            }
+        })
+    }
```
DavisVaughan:
```suggestion
    fn children_of(node: Node<'tree>) -> impl Iterator<Item = Node<'tree>> {
        let mut cursor = node.walk();
        let mut done = cursor.goto_first_child();

        std::iter::from_fn(move || {
            if done {
                None
            } else {
                let item = Some(cursor.node());
                done = cursor.goto_next_sibling();
                item
            }
        })
    }
```

Not sure if you find this any clearer

crates/ark/src/lsp/traits/node.rs
```diff
@@ -199,13 +206,96 @@ impl<'tree> NodeExt for Node<'tree> {
...
+                    return None;
+                }
+
+                let name = node.child_by_field_name("name");
+                let value = node.child_by_field_name("value");
+
+                // Likely not possible but just in case
+                if value.is_none() && name.is_none() {
+                    return None;
+                }
```
DavisVaughan: The caller still has to handle the `(None, None)` case when doing exhaustive matching, so I don't mind if you just let this fall through to simplify things

crates/ark/src/lsp/traits/node.rs
```diff
@@ -199,13 +206,96 @@ impl<'tree> NodeExt for Node<'tree> {
...
+    }
+
+    /// Iterator over argument names and values. Either of `name` and `value`
+    /// may be absent, but not both.
+    fn arguments(&self) -> impl Iterator<Item = (Option<Node<'tree>>, Option<Node<'tree>>)> {
+        self.child_by_field_name("arguments")
+            // Create iterator that unpacks Option with `flat_map()`
+            .into_iter()
+            .flat_map(Self::children_of)
+            .filter_map(|node| {
```
DavisVaughan, line 255: Would it make sense to also use `children_by_field_name("argument")` in here? That would simplify the filter-map a bit and be more explicit, and it also returns an iterator which you can feed into your filter-map

crates/ark/src/lsp/traits/node.rs
```diff
@@ -199,13 +206,96 @@ impl<'tree> NodeExt for Node<'tree> {
...
+        self.arguments_names().filter_map(|node| -> Option<String> {
+            match contents.node_slice(&node) {
+                Err(err) => {
+                    tracing::error!("Can't convert argument name to text: {err:?}");
+                    None
+                },
+                Ok(text) => Some(text.to_string()),
+            }
+        })
+    }
```
DavisVaughan: I feel like it would be a footgun if you used `arguments_values()` alongside `arguments_names_as_string()` but they produced different _lengths_ due to a `node_slice()` failing when converting to string.

It would be nice if we could guarantee the lengths were the same, even if we have to return `Iterator<Item = Result<String>>` here (or `Option<String>`), which doesn't feel totally inappropriate.

crates/ark/src/treesitter.rs
```diff
@@ -324,6 +326,20 @@ impl NodeTypeExt for Node<'_> {
...
     }
 
+    fn get_identifier_or_string_text(&self, contents: &ropey::Rope) -> anyhow::Result<String> {
+        match self.node_type() {
+            NodeType::Identifier => return Ok(contents.node_slice(self)?.to_string()),
+            NodeType::String => {
+                // Remove quotes from string literal
+                let string = contents.node_slice(self)?.to_string();
+                Ok(string.trim_matches('"').trim_matches('\'').to_string())
+            },
```
DavisVaughan: You should be able to extract a child known as `"content"` that does this for you

But note that raw strings also go through `NodeType::String`, and they do not yet have a `"content"` field, so that won't work for them (but the current code doesn't work either)

Another random thing to note is that I think stripping off the surrounding `"` can have weird behavior if the string has escape sequences in it? Like `"this\12"` seen in a text file becomes `"this\n"` after going through the R parser. I imagine that's all fine unless you actually need to _evaluate_ the value inside the text for some reason.

crates/ark/src/lsp/inputs/package_namespace.rs
```diff
@@ -0,0 +1,159 @@
...
+
+/// Parsed NAMESPACE file
+#[derive(Clone, Debug)]
+pub struct Namespace {
+    /// Names of objects exported with `export()`
+    pub exports: Vec<String>,
+    /// Names of objects imported with `importFrom()`
+    pub imports: Vec<String>,
+    /// Names of packages bulk-imported with `import()`
+    pub bulk_imports: Vec<String>,
```
DavisVaughan:
```suggestion
    /// Names of objects imported with `importFrom()`
    pub import_froms: Vec<(String, String)>,
    /// Names of packages bulk-imported with `import()`
    pub imports: Vec<String>,
```

I was expecting to see a 1:1 mapping like this:
- The names match a little better, so you know what it refers to when you see it in code
- The importFrom retains the package it comes from, which feels like it should be useful?

I feel like our diagnostic code downstream is able to handle duplicate `import_froms` with no problems, it will just ignore the 2nd one if we are keying just off the value name and not the package name as well.

But it feels like a wasted opportunity to not try and go ahead and faithfully mock NAMESPACE here

crates/ark/src/lsp/inputs/package_namespace.rs
```diff
@@ -0,0 +1,159 @@
...
+        parser
+            .set_language(&tree_sitter_r::LANGUAGE.into())
+            .map_err(|err| anyhow::anyhow!("Failed to set tree-sitter language: {err:?}"))?;
+
+        let tree = parser
+            .parse(contents, None)
+            .ok_or_else(|| anyhow::anyhow!("Failed to parse NAMESPACE file"))?;
+        let root_node = tree.root_node();
+
+        let query_str = r#"
```
DavisVaughan: Again a note about preferring a static pre-generated `Query` object, which I noted in another PR too

crates/ark/src/lsp/inputs/package_namespace.rs
```diff
@@ -0,0 +1,159 @@
...
+            let symbol = capture
+                .utf8_text(contents.as_bytes())
+                .unwrap_or("")
+                .to_string();
+            imports.push(symbol);
+        }
+
+        let mut bulk_imports = Vec::new();
+        for capture in ts_query.captures_for(root_node, "bulk_imported", contents.as_bytes()) {
+            let symbol = capture
```
DavisVaughan: This runs the query 3 separate times, collecting everything relevant for all 3 queries each time and throwing out all but the capture name of interest, which I'd prefer to avoid.

Optimally you'd only run `cursor.captures()` exactly once to collect all of the captures across all 3 queries.

And then you'd use some helper that takes a `QueryCaptures` and filters out the capture name of interest, and that's what you'd call 3 times.

It probably implies some restructing of the `treesitter.rs` tooling.

crates/ark/src/lsp/inputs/library.rs
```diff
@@ -0,0 +1,148 @@
...
+use std::sync::RwLock;
+
+use super::package::Package;
+use crate::lsp;
+
+/// Lazily manages a list of known R packages by name
+#[derive(Default, Clone, Debug)]
+pub struct Library {
+    /// Paths to library directories, i.e. what `base::libPaths()` returns.
+    pub library_paths: Arc<Vec<PathBuf>>,
```
DavisVaughan, line 19:
```suggestion
    pub library_paths: Vec<PathBuf>,
```

Is it overkill to Arc this? Probably like 1 or 2 items right?

crates/ark/src/lsp/inputs/library.rs
```diff
@@ -0,0 +1,148 @@
...
+
+    packages: Arc<RwLock<HashMap<String, Option<Arc<Package>>>>>,
```
DavisVaughan, line 21:
```suggestion
    packages: Arc<RwLock<HashMap<String, Option<Arc<Package>>>>>,
```

Why is `Package` itself `Arc`ed?

I was assuming that `get()` would return an immutable reference to a `Package` in the package list

i.e. what `self.packages.read().unwrap().get(name)` already returns

And if it's not cached yet, you write it to the cache but still return a `get()` reference

Then it feels like you don't need to `Arc` the `Package` because `packages` would be the sole owner

crates/ark/src/lsp/inputs/library.rs
```diff
@@ -0,0 +1,148 @@
...
+        self
+    }
+
+    fn load_package(&self, name: &str) -> anyhow::Result<Option<Package>> {
+        for lib_path in self.library_paths.iter() {
+            match Package::load(&lib_path, name) {
+                Ok(Some(pkg)) => return Ok(Some(pkg)),
+                Ok(None) => (),
+                Err(err) => lsp::log_warn!("Can't load package: {err:?}"),
+            }
```
DavisVaughan: In the `Err` case, we should probably bail with the error right?

Right now I think this says "I found the package, but failed to load it"

And IMO we should _not_ continue on to the next `library_paths` element when that occurs

DavisVaughan: If you decide not to do that, note that you don't need the return value to be `anyhow::Result<>` here, as you never propagate an error right now

crates/ark/src/lsp/main_loop.rs
DavisVaughan: Yea ok after reviewing this in more detail, I feel like all we'd have to do after getting the kernel notif is update the `WorldState`'s `Library` with a fresh new empty one created with `Library::new(paths)` and we'd be good to go?

crates/ark/src/lsp/diagnostics.rs
```diff
@@ -780,13 +807,116 @@ fn recurse_call(
...
+        return Ok(());
+    }
+
+    let package_name = package_node.get_identifier_or_string_text(context.contents)?;
+    let attach_pos = node.end_position();
+
+    let package = insert_package_exports(&package_name, attach_pos, context)?;
+
+    // Also attach packages from `Depends` field
+    let mut attach_dependencies = package.description.depends.clone();
```
DavisVaughan:
```suggestion
    for package_name in package.description.depends.iter() {
        insert_package_exports(package_name, attach_pos, context)?;
    }
```

I'd try and fully avoid the clone here, you just need a reference.

Down below, you'd just add a second loop of 

```suggestion
    for package_name in attach_field.iter() {
        insert_package_exports(package_name, attach_pos, context)?;
    }
```

I think that should clean things up fairly nicely in terms of clones / ownership of `Package` data and I think it reads a little nicer too

crates/ark/src/lsp/diagnostics.rs
```diff
@@ -780,13 +807,116 @@ fn recurse_call(
...
+    }
+
+    Ok(())
+}
+
+fn insert_package_exports(
+    package_name: &str,
+    attach_pos: Point,
+    context: &mut DiagnosticContext,
+) -> anyhow::Result<Arc<Package>> {
```
DavisVaughan, line 904: Yea I'm fairly certain we can and should just return `anyhow::Result<&Package>`? And that `get()` method really should return a reference, not an `Arc`.

We'd still clone out the `package.namespace.exports`, which is fine and seems required, but otherwise `Package` is immutable

crates/ark/src/lsp/diagnostics.rs
```diff
@@ -55,6 +62,14 @@ pub struct DiagnosticContext<'a> {
     // The set of packages that are currently installed.
     pub installed_packages: HashSet<String>,
 
+    /// Reference to the library for looking up package exports.
+    pub library: &'a Library,
+
+    /// The symbols exported by packages loaded via `library()` calls in this
+    /// document. Currently global. TODO: Store individual exports in a BTreeMap
+    /// sorted by position in the source?
+    pub library_symbols: BTreeMap<Point, HashSet<String>>,
```
DavisVaughan: Update comment? You did this TODO?

crates/ark/src/lsp/diagnostics.rs
```diff
@@ -86,20 +103,30 @@ impl<'a> DiagnosticContext<'a> {
...
 
-        // Finally, check session symbols.
+        // Check all symbols exported by `library()` calls before the given position
+        for (library_position, exports) in self.library_symbols.iter() {
+            if *library_position > start_position {
+                break;
+            }
+            if exports.contains(name) {
+                return true;
+            }
```
DavisVaughan, line 125: NEAT

DavisVaughan (approved):
Did a full deep code review and it seems quite nice and quite easy to understand!

I haven't done much interactive playing around with this in scripts yet, but I will and will report back if there are issues

crates/ark/src/lsp/main_loop.rs
DavisVaughan: (I also really like the idea of having clear-ish boundaries about what information comes from the Kernel, rather than having the LSP try and query it itself)

crates/ark/src/lsp/diagnostics.rs
```diff
@@ -5,8 +5,10 @@
 //
```
DavisVaughan, line 5:
```r
library(terra)
rast(r)
```

This doesn't work right now due to the custom namespace in terra

https://github.com/rspatial/terra/blob/master/NAMESPACE

- `export()` is a comma separated list
- `exportClasses()` is also a comma separated list for S4

crates/ark/src/lsp/traits/node.rs
lionel-: It mattered to unit tests and I did not want to dig in and see what assumptions/behaviour were broken. I'm sure nothing of importance is actually broken though. But since this is all going to be replaced by rowan, I did not bother looking into this further.

crates/ark/src/lsp/main_loop.rs
lionel-: I initially had the same thought as you, but I'm no longer sure that would be the right place.

With an independent kernel, what should ideally happen is that the LSP, not the kernel, would invoke `R` to figure out the `.libPaths()`. It's true that we could potentially also make it a dynamic LSP input from the kernel to get updates when run inside Positron. But it's probably not worth the complication. I think we should first see how whether we can get away with a static libPath that is a constant for the duration of the session.

So the `r_task()` querying implemented here for convenience really stands for "invoke R to get `.libPaths()`". Squinting a bit you can compare this to rust-analyzer calling `cargo` to get information about the project.

Does that make sense?

crates/ark/src/lsp/traits/node.rs
lionel-: hmm that sounds tempting, though I will note that Rowan doesn't follow that convention. See for example: https://github.com/rust-analyzer/rowan/blob/b477a127e048b5d4d4663c2efba770a839e2a272/src/api.rs#L222-L239

crates/ark/src/lsp/traits/node.rs
lionel-: I like it!

crates/ark/src/lsp/traits/node.rs
lionel-: yep I didn't know about the `children_by_` iterators at the time

crates/ark/src/lsp/inputs/package_namespace.rs
lionel-: I'd say there are pros and cons as the naming here is much clearer. I'd almost be tempted to have `import_froms` and `bulk_imports`. I don't feel strongly about this and could go for the direct mapping of namespace fields though.

I agree that the singular imports should be a `(String, String)` and was expecting we'd get there eventually.

crates/ark/src/lsp/inputs/package_namespace.rs
lionel-: I can take another look, although I'm not convinced it would make any perceptible change to performance.

crates/ark/src/lsp/main_loop.rs
DavisVaughan: > With an independent kernel, what should ideally happen is that the LSP, not the kernel, would invoke R to figure out the .libPaths(). 

That seems impossible to me, as `.libPaths()` can be set dynamically by the user of the current R session at any time

IIUC you are suggesting that the LSP fire up an independent side car R session to get this info, which I think is not right, as it won't reflect the current state of the world.

crates/ark/src/lsp/inputs/library.rs
lionel-: Good catch

crates/ark/src/lsp/main_loop.rs
lionel-: > IIUC you are suggesting that the LSP fire up an independent side car R session to get this info, which I think is not right, as it won't reflect the current state of the world.

yep I think this is exactly what it should do. It _could_ also get updates from a running session, and that would be reasonable, but that shouldn't be required. If a project happens to have specific libPaths requirements that are not fulfilled by the `R` in the current environment, then this should be set in a way that can be statically figured out. Otherwise our static LSP will not be able to make sense of the project.

It sounds like that would be a good topic to discuss one on one.

crates/ark/src/lsp/inputs/package.rs
lionel-: hmm actually we probably should be robust to missing `NAMESPACE` files?

Edit: I went ahead and made `NAMESPACE` optional, with an info-level message when it's missing.

crates/ark/src/lsp/traits/node.rs
lionel-: With this approach we start moving the cursor even if the iterator is not used, but I guess that doesn't matter, it's part of the cursor init in a way.

crates/ark/src/lsp/traits/node.rs
lionel-: hmm then we have to maintain a `cursor` and find a place for it, i.e. switch to an `iter::from_fn()` approach. I think I prefer this composition of iterators instead.

crates/ark/src/lsp/traits/node.rs
lionel-: Great point. It's almost as easy to flatten at the call site, so let's do that.

crates/ark/src/lsp/inputs/package_namespace.rs
lionel-: Davis: `package_imports` and `imports`?

crates/ark/src/lsp/traits/node.rs
lionel-: We've decided to go with the current naming scheme in TS and Rowan for now.

crates/ark/src/lsp/inputs/package_namespace.rs
lionel-: I thought I'd just go for simple and straightforward here but I see that's only making eyebrows raise :)

I've now:

- Added `all_captures()`, which lazily iterates over matches and their captures.
- Added `captures_by()`, which takes a vector of capture names and uses `all_captures()` to build a HashMap keyed by capture names
- We use that here, with only one query/traversal. Well I do a second pass to extract strings but we could improve that too if we wanted to.

crates/ark/src/lsp/inputs/package_namespace.rs
lionel-: Following our discussion of yesterday, I've extracted `TsQuery::from_query()`. Our `TsQuery` struct now stores an enum, either a ref or an owned TS query.

I've changed all call sites to use `from_query()` with a static query. That's actually almost as simple, I was wrong to think that would complicate things. The `new()`/Owned path is no longer used. I've kept it for now but we could remove it.

crates/ark/src/lsp/inputs/library.rs
lionel-: yep I didn't Arc this to avoid cloning the data, but to express that this is a view into the worldstate. With an owned vector, the paths could have been modified along the way. Not sure it is that useful to make the distinction but that was the idea.

crates/ark/src/lsp/inputs/library.rs
lionel-: Packages are meant to be shared across threads so we can't use `&`. We could make the references `&'static` but I think in the future we'll want to react to updated packages, so that doesn't feel like the right move.

crates/ark/src/lsp/diagnostics.rs
lionel-: Can't do that for the reason mentioned in the other comment (sharing across threads).

Just to be clear, `Package` _is_ immutable. An `Arc` confers shared ownership in the sense that it's only deallocated when all instances go out of scope. But the ownership is not in terms of being able to mutate it. Essentially all instances are like non-mutable references that you can share across threads without lifetime issues.

crates/ark/src/lsp/diagnostics.rs
lionel-: Looks like multiple args in `export()`, `import()`, and `importFrom()` were already supported with the existing query!

I've added support for `exportClasses()` and `exportMethods()`. Surprisingly the latter also exports the generic!

crates/ark/src/treesitter.rs
```diff
@@ -624,45 +642,154 @@ pub(crate) fn point_end_of_previous_row(
...
+            cursor,
+        }
+    }
 
-        Ok(Self { query, cursor })
+    /// Run the query on `contents` and collect all captures as (capture_name, node) pairs
+    pub(crate) fn all_captures<'tree, 'query>(
+        &'query mut self,
+        node: tree_sitter::Node<'tree>,
+        contents: &'query [u8],
```
DavisVaughan: Do the contents not have their own `'contents` lifetime? It's tied to the `TSQuery` object?

PR #871, opened 2025-07-11
Add tool for querying multiple installed package versions

jmcphers:
I noticed that the LLM sometimes makes several calls to `Get Installed R Package Version` in a row. To save time & tokens, this adds a new RPC that can look up multiple package versions in one call. 

The Positron side of this change will be included in an upcoming PR.

### DIFF
--- a/crates/ark/src/modules/positron/llm_tools.R
+++ b/crates/ark/src/modules/positron/llm_tools.R
@@ -53,6 +53,48 @@
     res_list
 }
 
+#' Get the version of installed packages
+#'
+#' This function retrieves the versions of specified packages in R.
+#'
+#' It returns a named list where the names are the package names and the values
+#' are the corresponding package versions.
+#'
+#' @export
+.ps.rpc.get_package_versions <- function(package_names, ...) {
+    # Convert list to character vector if needed
+    if (is.list(package_names)) {
+        package_names <- unlist(package_names)
+    }
+
+    # Get package versions for each package individually
+    version_list <- list()
+
+    for (pkg in package_names) {
+        tryCatch(
+            {
+                version <- utils::packageVersion(pkg)
+                if (is.null(version)) {
+                    version <- 'Not installed'
+                }
+                version_list[[pkg]] <- as.character(version)
+            },
+            error = function(e) {
+                version_list[[pkg]] <<- 'Not installed'
+            }
+        )
+    }
+
+    if (length(version_list) == 0) {
+        return(paste(
+            "No packages found for names:",
+            paste(package_names, collapse = ", ")
+        ))
+    }
+
+    version_list
+}
+
 #' Get available vignettes for a package
 #'
 #' This function retrieves the vignettes available for a specified package in R.


jennybc:
We have existing package version tooling in `crates/ark/src/modules/positron/package.R`. I don't know if there's an argument for keeping everything exposed via tools in one place and self-contained, but thought I'd remind us of this other file:

https://github.com/posit-dev/ark/blob/main/crates/ark/src/modules/positron/package.R

For example, those functions take care to not (necessarily) load the package, if we care about that.

crates/ark/src/modules/positron/llm_tools.R
```diff
@@ -53,6 +53,48 @@
...
+#'
+#' It returns a named list where the names are the package names and the values
+#' are the corresponding package versions.
+#'
+#' @export
+.ps.rpc.get_package_versions <- function(package_names, ...) {
+    # Convert list to character vector if needed
+    if (is.list(package_names)) {
+        package_names <- unlist(package_names)
+    }
```
lionel-: nit: I'd keep the input type strict here (a character vector).

crates/ark/src/modules/positron/llm_tools.R
```diff
@@ -53,6 +53,48 @@
...
+
+    if (length(version_list) == 0) {
+        return(paste(
+            "No packages found for names:",
+            paste(package_names, collapse = ", ")
+        ))
+    }
+
+    version_list
```
lionel-: A few comments here:

- It doesn't look like the list can be length 0 (unless the input is length 0), so we don't need to handle this case.

- `packageVersion()` should never return `NULL`.

- Since we're looping over a vector and transforming it, it would be a little more idiomatic to use a map.

- Might be better to guard against expected errors to avoid the catch-all `tryCatch()`. Currently if there's an _unexpected_ error we'll incorrectly report "uninstalled". It would be better to let the original unexpected error bubble up throughout our RPC mechanism back to the frontend.

All in all I suggest:

```suggestion
    lapply(set_names(package_names), function(pkg) {
        if (is_on_disk(pkg)) {
            as.character(utils::packageVersion(pkg))
        } else {
            "Not installed"
        }
    })
```

With these new helpers in utils.R:

```r
set_names <- function(x, names = x) {
    names(x) <- x
    x
}

# From rlang
is_on_disk <- function(pkg) {
    system_path(pkg) != ""
}

system_path <- function(pkg) {
    # Important for this to be first because packages loaded with pkgload
    # will have a different path than the one in `.libPaths()` (if any).
    if (isNamespaceLoaded(pkg)) {
        return(.getNamespaceInfo(asNamespace(pkg), "path"))
    }

    for (path in file.path(.libPaths(), pkg)) {
        if (file.exists(path)) {
            return(path)
        }
    }

    ""
}
```

It would probably be helpful to rework the existing utils that Jenny mentioned around `is_on_disk()`.

lionel-: Is it worth it to return a result to disambiguate errors?

```r
# Success
list(ok = value, err = NULL)

# Failure
list(ok = NULL, err = "msg")
```

crates/ark/src/modules/positron/llm_tools.R
jmcphers: We need to handle lists here because this is invoked directly from `callMethod` which uses lists. (We could improve that too but it's beyond the scope of the PR!)

crates/ark/src/modules/positron/llm_tools.R
jmcphers: > Is it worth it to return a result to disambiguate errors?

Generally speaking, the RPC framework should handle this -- if an error is thrown then it'll be bubbled up as an RPC failure.

jmcphers:
@jennybc thank you for pointing that out! I'm putting this in a separate file b/c it is specifically for use by the LLM (the other script is called by other frontend functions). Stuff we want the LLM to use is going to have different semantics, e.g. here we have `Not installed` for a package that is not installed instead of something more composable like `NULL`. 

I don't feel strongly that these live in their own file; I could also put them in the main package script and annotate them re: their LLM use. Let me know if that seems better?

jmcphers:
@lionel- you've made this much better, thank you! The only thing I ran into when testing the new approach is that it does not work for the base package (which the LLM likes to check the version of as a way of checking R's version sometimes...).

```
> getNamespaceInfo(asNamespace("base"))
Error in `asNamespace()`:
! operation not allowed on base namespace
Hide Traceback
    ▆
 1. └─base::getNamespaceInfo(asNamespace("base"))
 2.   └─base::asNamespace(ns, base.OK = FALSE)
 3.
```

 I've made a small change to fall back to regular libPath searching for the base package.

crates/ark/src/modules/positron/llm_tools.R
lionel-: you're right! I actually meant something like an `Option`, not a `Result`, sorry. Returning `"Not installed"` as a string, the same type as versions, felt weird. But really it's like we'd return `NULL | version` in that case, not a structured list.

lionel- (approved):

PR #872, opened 2025-07-15
Add symbols imported in packages to workspace

lionel-:
Addresses https://github.com/posit-dev/positron/issues/2252.
Addresses https://github.com/posit-dev/positron/issues/8549.
Addresses https://github.com/posit-dev/positron/issues/8550.
Progress towards https://github.com/posit-dev/positron/issues/2321.

Branched from #870. It was rather easy to implement based on the infrastructure provided in that PR.

This fixes diagnostics for imported symbols but I was still seeing some weirdness with _local_ definitions because we didn't synchronise the indexer and the diagnostics properly:

https://github.com/posit-dev/ark/blob/7175d83f46357f2ecc57d820ef4677b3122eabba/crates/ark/src/lsp/state_handlers.rs#L414-L417

This is now fixed. I've also made a change to take into account objects assigned globally. We were detecting global functions but not other kinds of objects.

I've hacked in testthat imports inside `testthat/` files. Should be good enough for now. Will fail when people edit their `testthat.R` file with additional library loading.

### QA Notes

You should now be able to open a package like ellmer and not see diagnostics. This won't be 100% proof for all packages, but I've checked with rlang and ellmer.

See also https://github.com/posit-dev/positron/issues/8549 and https://github.com/posit-dev/positron/issues/8550 for reprexes for adjacent fixes.

### DIFF
--- a/crates/ark/src/lsp/main_loop.rs
+++ b/crates/ark/src/lsp/main_loop.rs
@@ -827,15 +827,11 @@ async fn process_diagnostics_batch(batch: Vec<RefreshDiagnosticsTask>) {
                 // stopgap approach that has some false positives (e.g. when we
                 // work on testthat itself the flag will always be true), but
                 // that shouldn't have much practical impact.
-                let mut doc = document.clone();
-                if Path::new(uri.path())
+                let testthat = Path::new(uri.path())
                     .components()
-                    .any(|c| c.as_os_str() == "testthat")
-                {
-                    doc.testthat = true;
-                };
+                    .any(|c| c.as_os_str() == "testthat");
 
-                let diagnostics = generate_diagnostics(doc, state.clone());
+                let diagnostics = generate_diagnostics(document.clone(), state.clone(), testthat);
                 Some(RefreshDiagnosticsResult {
                     uri,
                     diagnostics,
--- a/crates/ark/src/lsp/diagnostics.rs
+++ b/crates/ark/src/lsp/diagnostics.rs
@@ -135,7 +135,11 @@ impl<'a> DiagnosticContext<'a> {
     }
 }
 
-pub(crate) fn generate_diagnostics(doc: Document, state: WorldState) -> Vec<Diagnostic> {
+pub(crate) fn generate_diagnostics(
+    doc: Document,
+    state: WorldState,
+    testthat: bool,
+) -> Vec<Diagnostic> {
     let mut diagnostics = Vec::new();
 
     if !state.config.diagnostics.enable {
@@ -190,7 +194,7 @@ pub(crate) fn generate_diagnostics(doc: Document, state: WorldState) -> Vec<Diag
     // don't special-case how workspace inclusion works for packages). We might
     // want to provide a mechanism for test packages to declare this sort of
     // test files setup.
-    if doc.testthat {
+    if testthat {
         if let Some(pkg) = state.library.get("testthat") {
             for export in &pkg.namespace.exports {
                 context.workspace_symbols.insert(export.clone());
@@ -1138,10 +1142,10 @@ mod tests {
 
     use harp::eval::RParseEvalOptions;
     use once_cell::sync::Lazy;
+    use tower_lsp::lsp_types;
     use tower_lsp::lsp_types::Position;
 
     use crate::interface::console_inputs;
-    use crate::lsp::diagnostics::generate_diagnostics;
     use crate::lsp::documents::Document;
     use crate::lsp::inputs::library::Library;
     use crate::lsp::inputs::package::Package;
@@ -1153,6 +1157,10 @@ mod tests {
 
     // Default state that includes installed packages and default scopes.
     static DEFAULT_STATE: Lazy<WorldState> = Lazy::new(|| current_state());
+
+    fn generate_diagnostics(doc: Document, state: WorldState) -> Vec<lsp_types::Diagnostic> {
+        super::generate_diagnostics(doc, state, false)
+    }
 
     fn current_state() -> WorldState {
         let inputs = console_inputs().unwrap();
--- a/crates/ark/src/lsp/documents.rs
+++ b/crates/ark/src/lsp/documents.rs
@@ -48,10 +48,6 @@ pub struct Document {
 
     // Configuration of the document, such as indentation settings.
     pub config: DocumentConfig,
-
-    /// Whether the document is a testthat file. This property should not be
-    /// here, it's just a short term stopgap.
-    pub testthat: bool,
 }
 
 impl std::fmt::Debug for Document {
@@ -84,7 +80,6 @@ impl Document {
             version,
             ast,
             config: Default::default(),
-            testthat: false,
         }
     }
 


crates/ark/src/lsp/main_loop.rs
```diff
@@ -692,3 +676,191 @@ impl std::fmt::Debug for TraceKernelNotification<'_> {
...
+            },
+        }));
+    }
+
+    for handle in handles {
+        let _ = handle.await;
+    }
+}
+
+async fn process_diagnostics_batch(batch: Vec<RefreshDiagnosticsTask>) {
```
DavisVaughan, line 810: Reconsider batching? Based on zoom call

crates/ark/src/lsp/main_loop.rs
```diff
@@ -692,3 +676,191 @@ impl std::fmt::Debug for TraceKernelNotification<'_> {
...
+
+pub(crate) fn index_update(uri: Url, document: Document, state: WorldState) {
+    INDEXER_QUEUE
+        .send(IndexerQueueTask::Indexer(IndexerTask::Update {
+            document,
+            uri: uri.clone(),
+        }))
+        .unwrap_or_else(|err| lsp::log_error!("Failed to queue index update: {err}"));
+
+    diagnostics_refresh(uri, state.clone());
```
DavisVaughan: @lionel- thinks maybe this should be a refresh-all?

DavisVaughan: If I rename a function in opened file A, then diagnostics for opened file B should update

crates/ark/src/lsp/main_loop.rs
lionel-: I take it back, the interleaved indexer/diagnostics processing will arise only if the queue tasks are processed faster than they arrive so the current setup is fine. If we get a bunch of tasks very rapidly, we will collect them, split them by type, and process the indexer tasks first. So the batching is still useful.

This setup will be entirely replaced by Salsa dependencies. Diagnostics tasks will be cancelled automatically as document updates arrive. So we shouldn't worry about this temporary setup too much.

lionel-: I adapted the loop so that we check back for more indexer tasks once we have finished a round of indexing. We do that at most 10 times so diagnostics get refreshed once in a while if the user is writing too fast to keep up. I think I'm happy with the queue setup now!

crates/ark/src/lsp/main_loop.rs
lionel-: Agreed, now done!

crates/ark/src/lsp/main_loop.rs
```diff
@@ -692,3 +677,210 @@ impl std::fmt::Debug for TraceKernelNotification<'_> {
...
+                // that shouldn't have much practical impact.
+                let mut doc = document.clone();
+                if Path::new(uri.path())
+                    .components()
+                    .any(|c| c.as_os_str() == "testthat")
+                {
+                    doc.testthat = true;
+                };
+
+                let diagnostics = generate_diagnostics(doc, state.clone());
```
DavisVaughan: It feels pretty gross to leak `testthat` hacks out into `Document`

Is there any way we can avoid this? Maybe have

```rust
pub(crate) fn generate_diagnostics_opt(doc: Document, state: WorldState, testthat: bool) -> Vec<Diagnostic> {}

pub(crate) fn generate_diagnostics(doc: Document, state: WorldState) -> Vec<Diagnostic> {
  generate_diagnostics_opt(doc, state, false)
}
```

And you'd use `generate_diagnostics_opt()` here just to determine `testthat` but we wouldn't have to have it in `Document` itself, and we don't need `document.clone()`, which doesn't seem nice from a performance perspective (cloning the Document very often like this seems bad)

crates/ark/src/lsp/main_loop.rs
```diff
@@ -7,15 +7,20 @@
```
DavisVaughan, line 7: Somewhat reasonable place this fails - `vec_order_radix()` usage in dplyr, brought in from vctrs via

https://github.com/tidyverse/dplyr/blob/be3e3a05fd0081cb53168d6aedb417d62139b75d/R/zzz.R#L17-L20

crates/ark/src/lsp/main_loop.rs
DavisVaughan: This sounds particularly nice to me because it looks like `testthat: bool` would not have to extend past `generate_diagnostics()`, you take care of it right there, which really limits the leaking of this hack

crates/ark/src/lsp/inputs/package.rs
```diff
@@ -23,15 +23,12 @@ pub struct Package {
 }
 
 impl Package {
-    /// Attempts to load a package from the given path and name.
-    pub fn load(lib_path: &std::path::Path, name: &str) -> anyhow::Result<Option<Self>> {
-        let package_path = lib_path.join(name);
-
+    /// Load a package from a given path.
+    pub fn load(package_path: &std::path::Path) -> anyhow::Result<Option<Self>> {
```
DavisVaughan:
```suggestion
    pub fn load_from_folder(package_path: &std::path::Path) -> anyhow::Result<Option<Self>> {
```

?

For a second I thought there was a single file that `package_path` points to, but it looks like it is intended to point to a folder. So this plus a name of just `path: &Path` seems good imo

crates/ark/src/lsp/inputs/package.rs
```diff
@@ -41,12 +38,6 @@ impl Package {
         let description_contents = fs::read_to_string(&description_path)?;
         let description = Description::parse(&description_contents)?;
 
-        if description.name != name {
-            return Err(anyhow::anyhow!(
-                "`Package` field in `DESCRIPTION` doesn't match folder name '{name}'"
-            ));
-        }
```
DavisVaughan, line 48: What violates this?

DavisVaughan: Oh I see you moved this. Why would non-library-path packages be any different?

crates/ark/src/lsp/state_handlers.rs
```diff
@@ -88,8 +90,35 @@ pub(crate) fn initialize(
         for folder in workspace_folders.iter() {
             state.workspace.folders.push(folder.uri.clone());
             if let Ok(path) = folder.uri.to_file_path() {
-                if let Some(path) = path.to_str() {
-                    folders.push(path.to_string());
+                // Try to load package from this workspace folder and set as
+                // root if found. This means we're dealing with a package
+                // source.
+                if state.root.is_none() {
```
DavisVaughan, line 89: I guess implicitly this means we only work with the first `workspace`, in the case of multi-root workspaces

crates/ark/src/lsp/main_loop.rs
```diff
@@ -692,3 +677,210 @@ impl std::fmt::Debug for TraceKernelNotification<'_> {
...
+            } else {
+                None
+            }
+        }));
+    }
+
+    // Publish results as they complete
+    while let Some(Ok(Some(result))) = futures.next().await {
+        publish_diagnostics(result.uri, result.diagnostics, result.version);
+    }
```
DavisVaughan, line 849: Is it weird that we are already on the indexer thread, and we are calling `publish_diagnostics()`, which then sends to the auxiliary thread? I guess not?

crates/ark/src/lsp/main_loop.rs
```diff
@@ -692,3 +677,210 @@ impl std::fmt::Debug for TraceKernelNotification<'_> {
...
+}
+
+pub(crate) fn diagnostics_refresh_all(state: WorldState) {
+    for (uri, _document) in state.documents.iter() {
+        INDEXER_QUEUE
+            .send(IndexerQueueTask::Diagnostics(RefreshDiagnosticsTask {
+                uri: uri.clone(),
+                state: state.clone(),
+            }))
+            .unwrap_or_else(|err| lsp::log_error!("Failed to queue diagnostics refresh: {err}"));
```
DavisVaughan, line 880: I think some of this was there before, but there is more `clone()`ing of the `WorldState` than I expected in this code

It feels like we are cloning like crazy:
- Call site of `index_start()`
- Call site of `index_update()`, every change, really??
- Each document in `diagnostics_refresh_all()` gets its own copy
- Call site of `generate_diagnostics()`, so basically every change, I think this was here before

I know the intention of `WorldState` is that we can send it to other threads but it just seems like it's happening a lot here, is that fully intentional?

DavisVaughan (approved):
Seems to work fairly well with dplyr and vctrs

crates/ark/src/lsp/main_loop.rs
DavisVaughan: Similar with clock

```r
clock_init_weekday_utils <- function(env) {
  assign("clock_empty_weekday", weekday(integer()), envir = env)

  invisible(NULL)
}
```

called from onLoad

crates/ark/src/lsp/main_loop.rs
lionel-: Maybe, the thing is I'm not sure how we will express implicit preambles like `testthat.R` yet, but to me it feels like this will be a property of a document somehow.

crates/ark/src/lsp/main_loop.rs
lionel-: Yep packages need to declare their dynamic imports with a top-level stub like this:

```r
vec_order_radix <- function(x, ..., direction = "asc", na_value = "largest", nan_distinct = FALSE, chr_proxy_collate = NULL) {}

.onLoad <- function(libname, pkgname) {
  ns_dplyr <- ns_env(pkgname)

  env_bind(
    .env = ns_dplyr,
    vec_order_radix = import_vctrs("vec_order_radix")
  )
```

We could possibly do better in the future and infer the side effects in `.onLoad()`, but there are multiple challenges in that example, e.g. we'd need to understand what `ns_env()` is doing with dedicated declarations. The less we have to do stuff like this, the better, so I'm inclined to require more work from package authors. This has the nice benefit of documenting the expected signature of the dynamic import.

crates/ark/src/lsp/inputs/package.rs
lionel-: This is only a library invariant. My local purrr repository is still in a folder called "lowliner" 😅

crates/ark/src/lsp/main_loop.rs
lionel-: I don't think so? This thread (more precisely this task) owns the `Client`.

crates/ark/src/lsp/main_loop.rs
lionel-: It's intentional because once the state is managed by Salsa, there will be no cloning at all. If needed we could review things in the meantime by switching e.g. to `Arc` for clone-free sharing.

crates/ark/src/lsp/main_loop.rs
lionel-: I added `testthat: bool` to `generate_diagnostics()`. In the tests module we have this wrapper to keep things lean:

```rust
    fn generate_diagnostics(doc: Document, state: WorldState) -> Vec<lsp_types::Diagnostic> {
        super::generate_diagnostics(doc, state, false)
    }
```

Issue #588, opened 2024-10-15
Behavior of `options(browse =)` outside of Positron


DavisVaughan:
Right now we always override `options(browse =)` and add our special Positron specific behavior. This isn't great for Jupyter kernels

https://github.com/posit-dev/ark/blob/738afbaa7984b592f8ab78bd714e04efe8bbf09c/crates/ark/src/browser.rs#L37-L59

From @lionel- 

> The right thing to do outside Positron is to use the default action, i.e. open in a web browser. So we should not set up that option unless Positron is connected. Ideally Positron would inject the option via its init file. Or the UI comm could launch an idle task to do it, but there would a short time where the option is not set with this approach.

jennybc:
Specifically, in PR #877, here's the sort of error raised by calling `browseURL()` on a local folder's filepath:

```
~/tmp % jupyter console --kernel=ark
Jupyter console 6.6.3


R version 4.4.2 (2024-10-31) -- "Pile of Leaves"
Copyright (C) 2024 The R Foundation for Statistical Computing
Platform: aarch64-apple-darwin20

R is free software and comes with ABSOLUTELY NO WARRANTY.
You are welcome to redistribute it under certain conditions.
Type 'license()' or 'licence()' for distribution details.

  Natural language support but running in an English locale

R is a collaborative project with many contributors.
Type 'contributors()' for more information and
'citation()' on how to cite R or R packages in publications.

Type 'demo()' for some demos, 'help()' for on-line help, or
'help.start()' for an HTML browser interface to help.
Type 'q()' to quit R.


In [1]: browseURL("~/tmp")
  2025-07-21T21:35:02.025887Z ERROR  No help port is available to check if '~/tmp' is a help url. Is the help comm open?
    at crates/ark/src/interface.rs:2032

  2025-07-21T21:35:02.026935Z ERROR  Failed to browse url due to: UI comm not connected, can't run `open_with_system`.
    at crates/ark/src/browser.rs:22


In [2]: getOption("browser")
Out[2]:
function(url) {
    .ps.Call("ps_browse_url", as.character(url))
}
<environment: 0x11fd5bc70>
```

PR #877, opened 2025-07-19
Use new UI event to send a putative path to the system for opening

jennybc:
Addresses https://github.com/posit-dev/positron/issues/5486
Companion to https://github.com/posit-dev/positron/pull/8588
Surfaced by https://github.com/r-lib/usethis/issues/2126

### DIFF
--- a/crates/ark/src/browser.rs
+++ b/crates/ark/src/browser.rs
@@ -6,6 +6,7 @@
 //
 
 use harp::object::RObject;
+use harp::utils::r_normalize_path;
 use libr::Rf_ScalarLogical;
 use libr::SEXP;
 
@@ -55,8 +56,9 @@ unsafe fn ps_browse_url_impl(url: SEXP) -> anyhow::Result<SEXP> {
 
     // This is probably a file path? Send to the front end and ask for system
     // default opener.
-    log::trace!("Treating as file path");
-    send_open_with_system_default_event(&url_string)?;
+    log::trace!("Treating as file path and asking system to open");
+    let path = r_normalize_path(url.into())?;
+    send_open_with_system_default_event(&path)?;
     Ok(Rf_ScalarLogical(1))
 }
 
--- a/crates/ark/src/browser.rs
+++ b/crates/ark/src/browser.rs
@@ -5,8 +5,6 @@
 //
 //
 
-use amalthea::comm::ui_comm::ShowUrlParams;
-use amalthea::comm::ui_comm::UiFrontendEvent;
 use harp::object::RObject;
 use libr::Rf_ScalarLogical;
 use libr::SEXP;
@@ -14,6 +12,8 @@ use libr::SEXP;
 use crate::help::message::HelpEvent;
 use crate::help::message::ShowHelpUrlParams;
 use crate::interface::RMain;
+use crate::ui::events::send_open_with_system_default_event;
+use crate::ui::events::send_show_url_event;
 
 #[harp::register]
 pub unsafe extern "C-unwind" fn ps_browse_url(url: SEXP) -> anyhow::Result<SEXP> {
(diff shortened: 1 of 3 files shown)


crates/ark/src/browser.rs
```diff
@@ -35,33 +36,32 @@ fn handle_help_url(url: String) -> anyhow::Result<()> {
...
         log::trace!("Help is handling URL");
-        handle_help_url(url)?;
+        handle_help_url(url_string)?;
         return Ok(Rf_ScalarLogical(1));
-    } else {
-        log::trace!("Help is not handling URL");
     }
 
-    // TODO: What is the right thing to do outside of Positron when
-    // `options(browser =)` is called? Right now we error.
```
jennybc, line 52: Does anyone want me to hang onto this comment?

crates/ark/src/browser.rs
```diff
@@ -35,33 +36,32 @@ fn handle_help_url(url: String) -> anyhow::Result<()> {
...
-
-    // For all other URLs, create a ShowUrl event and send it to the main
-    // thread; Positron will handle it.
-    let params = ShowUrlParams { url };
-    let event = UiFrontendEvent::ShowUrl(params);
-
-    let main = RMain::get();
-    let ui_comm_tx = main
-        .get_ui_comm_tx()
-        .ok_or_else(|| anyhow::anyhow!("UI comm not connected."))?;
```
jennybc, line 62: This code was basically duplicated in `ps_ui_show_url()` so now they share logic via the `send_show_url_event()` helper.

crates/ark/src/browser.rs
```diff
@@ -35,33 +36,32 @@ fn handle_help_url(url: String) -> anyhow::Result<()> {
...
+    if is_web_url(&url_string) {
+        log::trace!("Handling web URL");
+        send_show_url_event(&url_string)?;
+        return Ok(Rf_ScalarLogical(1));
+    }
 
+    // This is probably a file path? Send to the front end and ask for system
+    // default opener.
+    log::trace!("Treating as file path and asking system to open");
+    let path = r_normalize_path(url.into())?;
```
jennybc, line 60: The opener service on the Positron side expects a resolved path, so it's important to normalize on the backend.

crates/ark/src/browser.rs
lionel-: I think so, we should only inject this option when launched from Positron. Probably an easy fix if you want to do it here?

crates/ark/src/ui/events.rs
```diff
@@ -96,23 +97,43 @@ pub unsafe extern "C-unwind" fn ps_ui_set_selection_ranges(ranges: SEXP) -> anyh
...
+}
+
+#[harp::register]
+pub unsafe extern "C-unwind" fn ps_ui_show_url(url: SEXP) -> anyhow::Result<SEXP> {
+    let url_string = RObject::view(url).to::<String>()?;
+    send_show_url_event(&url_string)?;
     Ok(R_NilValue)
 }
 
+pub fn send_open_with_system_default_event(path: &str) -> anyhow::Result<()> {
```
lionel-: Align with event name?

```suggestion
pub fn send_open_with_system_event(path: &str) -> anyhow::Result<()> {
```

crates/ark/src/browser.rs
jennybc: I don't really follow how ark would  hit `ps_browse_url_impl()` outside of Positron. Because I assume it works like this:

In Positron, ark takes charge of the `"browser"` option and that eventually routes through here.

But outside of Positron, ark would presumably have the default implementation of `"browser"`, yeah?

Is the claim that ark would error outside of Positron actually true or true anymore? 🤔

crates/ark/src/ui/events.rs
jennybc: I like it. I'm implementing in a local commit, because I need to use "Rename Symbol" to also hit the call sites.

crates/ark/src/browser.rs
jennybc: Apparently I am wrong. Why do we set an option like `"browser"` in ark, but outside of Positron? This feels like it might be out of scope for this PR (a bigger problem). In which case I should just retain the comment and move on.

```
~/tmp % jupyter console --kernel=ark
Jupyter console 6.6.3


R version 4.4.2 (2024-10-31) -- "Pile of Leaves"
Copyright (C) 2024 The R Foundation for Statistical Computing
Platform: aarch64-apple-darwin20

R is free software and comes with ABSOLUTELY NO WARRANTY.
You are welcome to redistribute it under certain conditions.
Type 'license()' or 'licence()' for distribution details.

  Natural language support but running in an English locale

R is a collaborative project with many contributors.
Type 'contributors()' for more information and
'citation()' on how to cite R or R packages in publications.

Type 'demo()' for some demos, 'help()' for on-line help, or
'help.start()' for an HTML browser interface to help.
Type 'q()' to quit R.


In [1]: browseURL("~/tmp")
  2025-07-21T21:35:02.025887Z ERROR  No help port is available to check if '~/tmp' is a help url. Is the help comm open?
    at crates/ark/src/interface.rs:2032

  2025-07-21T21:35:02.026935Z ERROR  Failed to browse url due to: UI comm not connected, can't run `open_with_system`.
    at crates/ark/src/browser.rs:22


In [2]: getOption("browser")
Out[2]:
function(url) {
    .ps.Call("ps_browse_url", as.character(url))
}
<environment: 0x11fd5bc70>
```

jennybc: We are already tracking this problem here: https://github.com/posit-dev/ark/issues/588.

So I'm going to merge this once the Positron side is approved.

PR #881, opened 2025-07-25
Clear index on `did_close`

lionel-:
Addresses https://github.com/posit-dev/positron/issues/8668

### QA Notes

Renaming files should not leave dangling workspace symbols (cmd/ctrl + t):

https://github.com/user-attachments/assets/1ede9524-5558-41a0-9567-622a74f4db06

### DIFF
--- a/crates/ark/src/lsp/main_loop.rs
+++ b/crates/ark/src/lsp/main_loop.rs
@@ -972,10 +972,8 @@ pub(crate) fn index_start(folders: Vec<String>, state: WorldState) {
 
 pub(crate) fn index_create(uris: Vec<Url>, state: WorldState) {
     for uri in uris {
-        crate::lsp::main_loop::INDEXER_QUEUE
-            .send(crate::lsp::main_loop::IndexerQueueTask::Indexer(
-                crate::lsp::main_loop::IndexerTask::Create { uri },
-            ))
+        INDEXER_QUEUE
+            .send(IndexerQueueTask::Indexer(IndexerTask::Create { uri }))
             .unwrap_or_else(|err| crate::lsp::log_error!("Failed to queue index create: {err}"));
     }
 
--- a/crates/ark/src/lsp/main_loop.rs
+++ b/crates/ark/src/lsp/main_loop.rs
@@ -295,6 +295,15 @@ impl GlobalState {
                         LspNotification::DidCloseTextDocument(params) => {
                             state_handlers::did_close(params, &mut self.lsp_state, &mut self.world)?;
                         },
+                        LspNotification::DidCreateFiles(params) => {
+                            state_handlers::did_create_files(params, &self.world)?;
+                        },
+                        LspNotification::DidDeleteFiles(params) => {
+                            state_handlers::did_delete_files(params, &mut self.world)?;
+                        },
+                        LspNotification::DidRenameFiles(params) => {
+                            state_handlers::did_rename_files(params, &mut self.world)?;
+                        },
                     }
                 },
 
@@ -686,8 +695,10 @@ pub(crate) enum IndexerQueueTask {
 
 #[derive(Debug)]
 pub enum IndexerTask {
-    Start { folders: Vec<String> },
-    Update { document: Document, uri: Url },
+    Create { uri: Url },
+    Delete { uri: Url },
+    Rename { uri: Url, new: Url },
+    Update { uri: Url, document: Document },
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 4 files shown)


lionel-:
Wait that's not right, it's not about opened files it's about workspace files, so we'll need to manually watch for file deletion.

DavisVaughan:
Surely there's some LSP notification we can get about file renames?

crates/ark/src/lsp/main_loop.rs
```diff
@@ -770,44 +802,94 @@ async fn process_indexer_queue(mut rx: mpsc::UnboundedReceiver<IndexerQueueTask>
...
+                    }
+                    let path = to_path_buf(uri)?;
+                    indexer::delete(&path)?;
+                },
+
+                IndexerTask::Rename {
+                    uri: old_uri,
+                    new: new_uri,
+                } => {
+                    let handles = flush_updates(&mut updates);
```
DavisVaughan: I'm very confused as to why we batch the `Update` tasks. I figure we would just process them in order. Batching them like this (and across different URIs too?) feels pretty complicated?

Is the idea to collapse multiple `IndexerTask::Update` calls on the same document together into just one run of `indexer::update()`?

Having to manage the timing of 3 separate calls to `flush_updates()` doesn't feel worth it for that. Is there evidence of a performance issue here?

crates/ark/src/lsp/main_loop.rs
```diff
@@ -850,27 +932,112 @@ async fn process_diagnostics_batch(batch: Vec<RefreshDiagnosticsTask>) {
...
+    index_create(uris, state);
+}
+
+pub(crate) fn index_create(uris: Vec<Url>, state: WorldState) {
+    for uri in uris {
+        crate::lsp::main_loop::INDEXER_QUEUE
+            .send(crate::lsp::main_loop::IndexerQueueTask::Indexer(
+                crate::lsp::main_loop::IndexerTask::Create { uri },
+            ))
+            .unwrap_or_else(|err| crate::lsp::log_error!("Failed to queue index create: {err}"));
```
DavisVaughan: Use some imports to reduce the number of `::`?

crates/ark/src/lsp/state_handlers.rs
```diff
@@ -166,7 +173,28 @@ pub(crate) fn initialize(
                     supported: Some(true),
                     change_notifications: Some(OneOf::Left(true)),
                 }),
-                file_operations: None,
+                file_operations: {
+                    let r_file_filter = FileOperationFilter {
+                        scheme: Some(String::from("file")),
+                        pattern: FileOperationPattern {
+                            glob: String::from("**/*.[Rr]"),
```
DavisVaughan: Are we sure `"**/*.[Rr]"` is right? Should it be `"**/*.{R,r}"`?

My reading of https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#fileOperationPattern suggests it is the latter

DavisVaughan (approved):
Everything looks fairly reasonable, but the loop in `process_indexer_batch()` could definitely use some comments because I'm pretty confused about why we have to batch updates and flush them. It does not feel necessary to me (see comment), so I think we should either simplify that part of justify it with a detailed comment

crates/ark/src/lsp/main_loop.rs
lionel-: This does two things:

- Coalesce multiple updates for the same file to discard earlier ones, since we don't do any incremental updating we can just discard them

- Run all updates for different files in parallel. Would only matter when changing branches though.

Worth noting that it was much simpler previously. I just preserved that approach in that now more complicated setup.

There's evidence of performance issues for the first round of indexing and diagnostics (visible delay), not for updates.

crates/ark/src/lsp/main_loop.rs
lionel-: It's in scope, no need to qualify! I really try to fix all these agent idiosyncrasies but after a long day I don't see them anymore

crates/ark/src/lsp/state_handlers.rs
lionel-: It's not clear from these docs but I think it's treated as a general character class so correct here (and works in my tests, I get notifications for the right files).

I can use the more explicit `{}` syntax though, it's generally used because that's what works best when you can't express with a character class, e.g. https://github.com/souma987/erg/blob/9a131ecc53b670b0835900736e8556a37e1b90e6/crates/els/file_cache.rs#L51

crates/ark/src/lsp/main_loop.rs
lionel-: Now a simpler queue without any batching, we'll revisit if performance is lacking.

PR #883, opened 2025-07-29
Treat all documented names as known symbols

lionel-:
Branched from #872.

Progress towards https://github.com/posit-dev/positron/issues/8521
Addresses https://github.com/posit-dev/positron/issues/8708

This adds all symbol-looking documentation topics to our set of known exported symbols to cover exported datasets. This information is installed in an INDEX file that we can inspect statically: https://cran.r-project.org/doc/manuals/r-devel/R-exts.html#The-INDEX-file-1. This INDEX file might be manually written by the package author (see link above), and so might throw off our attempts at parsing this unstructured file. That said I expect it's automatically generated in the vast majority of cases

This stopgap approach will potentially produce false negatives:

- Documented topics are not necessarily about exported symbols.
- This approach will indistinctly export datasets whether `LazyData` is true or false. In the false case, the dataset is only exported after a corresponding `data()` call.

For now though, we prefer to avoid false positives and spurious diagnostics.

### QA Notes

In a fresh session with:

```r
# `LazyData: true` case
penguins # Not in scope
library(palmerpenguins)
penguins

# `LazyData: false` case
liner # Not in scope
library(BCA1SG)
liner # FIXME: Should not be in scope
data(liner)
liner
```

You should only see diagnostics _before_ the `library()` calls:

Screenshot 2025-07-29 at 16 40 02


DavisVaughan:
Reviewed together over zed

PR #884, opened 2025-07-29
data explorer: set edges to value of single bin

isabelizimm:
R half of https://github.com/posit-dev/positron/pull/8611
Addressing https://github.com/posit-dev/positron/issues/8095

```
View(data.frame(values = rep(10, 10)))
```

Upper and lower values should be equal on hover.

Screenshot 2025-07-29 at 10 48 00 AM

### DIFF
--- a/crates/ark/tests/data_explorer.rs
+++ b/crates/ark/tests/data_explorer.rs
@@ -1987,6 +1987,51 @@ fn test_histogram() {
 }
 
 #[test]
+fn test_histogram_single_bin_same_values() {
+    let _lock = r_test_lock();
+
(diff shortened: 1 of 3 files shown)


crates/ark/tests/data_explorer.rs
```diff
@@ -1986,6 +1986,51 @@ fn test_histogram() {
     });
 }
 
+#[test]
```
isabelizimm, line 1989: disclosure that claude wrote initial draft, I did some ergonomic updates to confirm it matched precedents set in `test_histogram()`

PR #886, opened 2025-07-30
Prefer current file in jump to definition

lionel-:
Addresses https://github.com/posit-dev/positron/issues/8630

### QA Notes

Jump to definition should now prefer definitions in current files, and still fall back to other files if not present.

https://github.com/user-attachments/assets/ccfd57f9-4ed8-4696-b885-6347d1664728

### DIFF
--- a/crates/ark/src/lsp/indexer.rs
+++ b/crates/ark/src/lsp/indexer.rs
@@ -91,6 +91,7 @@ pub fn start(folders: Vec<String>) {
     );
 }
 
+/// Search the workspace files and return the first symbol match
 pub fn find(symbol: &str) -> Option<(String, IndexEntry)> {
     let index = WORKSPACE_INDEX.lock().unwrap();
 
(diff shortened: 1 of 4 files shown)


crates/ark/src/lsp/indexer.rs
```diff
@@ -103,6 +103,20 @@ pub fn find(symbol: &str) -> Option<(String, IndexEntry)> {
     None
 }
 
+pub fn find_in_file(symbol: &str, path: &std::path::Path) -> Option<(String, IndexEntry)> {
```
DavisVaughan: Should you add some docs above `pub fn find(` that clarify that that function finds the _first_ instance?

crates/ark/src/lsp/indexer.rs
```diff
@@ -194,7 +208,7 @@ impl Drop for ResetIndexerGuard {
 fn str_from_path(path: &Path) -> anyhow::Result<&str> {
     path.to_str().ok_or(anyhow!(
         "Couldn't convert path {} to string",
-        path.display()
+        path.to_string_lossy()
```
DavisVaughan, line 213: What's wrong with display here?

DavisVaughan (approved):
Looks like a reasonable fallback to me

crates/ark/src/lsp/indexer.rs
lionel-: Probably nothing, but `to_string_lossy()` is more explicit and you can look up docs.

PR #887, opened 2025-07-30
Fix navigate to file on Windows

lionel-:
Addresses https://github.com/posit-dev/positron/issues/8374

I think the scheme detection was getting tripped up by windows paths like `C:\`

### QA Notes

I'll add a positron-r extension test for this.

To test, open a file called `foo.R`, navigate to another tab, then run:

```r
.rs.api.navigateToFile("foo.R")
```

### DIFF
--- a/crates/ark/src/modules/positron/frontend-methods.R
+++ b/crates/ark/src/modules/positron/frontend-methods.R
@@ -38,8 +38,8 @@
     line = 0L,
     column = 0L
 ) {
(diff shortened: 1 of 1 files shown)


DavisVaughan:
Proof!

Screenshot 2025-07-30 at 11 54 44 AM

lionel-:
I've confirmed `View(data.frame)` still works as it should.

PR #889, opened 2025-08-03
Data Explorer: Implement SearchSchema RPC, refactor and improve data explorer test suite 

wesm:
Addresses posit-dev/positron#8805, follows backend API change in https://github.com/posit-dev/positron/pull/8810. Also pulls in variables comm changes which can be implemented in a follow up PR

### DIFF
--- a/crates/ark/src/data_explorer/r_data_explorer.rs
+++ b/crates/ark/src/data_explorer/r_data_explorer.rs
@@ -532,9 +532,7 @@ impl RDataExplorer {
                 return Err(anyhow!("Data Explorer: Not yet supported"));
             },
 
-            DataExplorerBackendRequest::SearchSchema(_) => {
-                return Err(anyhow!("Data Explorer: Not yet supported"));
-            },
+            DataExplorerBackendRequest::SearchSchema(params) => self.search_schema(params),
 
             DataExplorerBackendRequest::SetColumnFilters(_) => {
                 return Err(anyhow!("Data Explorer: Not yet supported"));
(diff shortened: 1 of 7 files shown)


crates/ark/src/lsp/completions/sources/unique/colon.rs
```diff
@@ -30,7 +30,9 @@ impl CompletionSource for SingleColonSource {
 // Don't provide completions if on a single `:`, which typically precedes
 // a `::` or `:::`. It means we don't provide completions for `1:` but we
 // accept that.
```
lionel-: Do you know what happened in these completion files? Seems like purely whitespace changes, is it from rustfmt?

crates/ark/tests/data_explorer.rs
```diff
@@ -150,6 +157,543 @@ fn socket_rpc(
...
+    ) -> DataExplorerBackendRequest {
+        DataExplorerBackendRequest::ExportDataSelection(ExportDataSelectionParams {
+            format,
+            selection,
+        })
+    }
+}
+
+/// Assertion helpers to reduce repetitive assertion patterns
+struct TestAssertions;
```
lionel-, line 284: I'm a bit worried that we're abstracting assertions with methods rather than macros, because the former won't reveal as much of actual values supplied in test outputs and CI logs.

With AI it's now easy to create these so it might be worth switching to macros.

crates/ark/tests/data_explorer.rs
```diff
@@ -150,6 +157,543 @@ fn socket_rpc(
...
+        })
+    }
+
+    fn socket(&self) -> &socket::comm::CommSocket {
+        &self.socket
+    }
+}
+
+/// Request builder helpers to reduce repetitive request construction
+struct RequestBuilder;
```
lionel-, line 185: Not a big deal but in Rust a `Builder` struct generally denotes a multi-step ctor where you call multiple methods to specify your object, ending with `finish()` or `create()`.

Another way of setting this up would have been a `request` submodule that isn't imported, then you call it like `request::get_schema()`.

Just a note, no need to change now.

crates/ark/src/data_explorer/r_data_explorer.rs
```diff
@@ -859,6 +859,110 @@ impl RDataExplorer {
         self.view_indices = Some(view_indices);
     }
 
+    /// Search the schema for columns matching the given filters and sort order.
+    ///
+    /// - `params`: The search parameters including filters and sort order.
+    fn search_schema(
+        &self,
+        params: amalthea::comm::data_explorer_comm::SearchSchemaParams,
```
lionel-, line 869: Here and below: These paths are rather long. Import these types instead?

crates/ark/src/data_explorer/r_data_explorer.rs
```diff
@@ -859,6 +859,110 @@ impl RDataExplorer {
...
+                matches: matching_indices,
+            },
+        ))
+    }
+
+    /// Check if a column matches a given column filter.
+    fn column_matches_filter(&self, column: &ColumnSchema, filter: &ColumnFilter) -> bool {
+        use amalthea::comm::data_explorer_comm::ColumnFilterParams;
+        use amalthea::comm::data_explorer_comm::ColumnFilterType;
+        use amalthea::comm::data_explorer_comm::TextSearchType;
```
lionel-: Imports at the top of the file

crates/ark/src/data_explorer/r_data_explorer.rs
```diff
@@ -859,6 +859,110 @@ impl RDataExplorer {
...
+
+        match filter.filter_type {
+            ColumnFilterType::TextSearch => {
+                if let ColumnFilterParams::TextSearch(text_search) = &filter.params {
+                    let column_name = if text_search.case_sensitive {
+                        column.column_name.clone()
```
lionel-: .to_owned() would better reflect the intent here

crates/ark/src/data_explorer/r_data_explorer.rs
```diff
@@ -859,6 +859,110 @@ impl RDataExplorer {
...
+        &self,
+        params: amalthea::comm::data_explorer_comm::SearchSchemaParams,
+    ) -> anyhow::Result<DataExplorerBackendReply> {
+        let mut matching_indices: Vec<i64> = Vec::new();
+
+        // Get all column schemas
+        let all_columns = &self.shape.columns;
+
+        // Apply column filters to find matching columns
+        for (index, column) in all_columns.iter().enumerate() {
```
lionel-: This is a case where iterator chaining is clearer than raw loops:

```rust
let matching_indices: Vec<i64> = &self.shape.columns
    .iter()
    .enumerate()
    .filter(|(_, column)| {
        params.filters.iter().all(|filter| self.column_matches_filter(column, filter))
    })
    .map(|(index, _)| index as i64)
    .collect();
```

crates/ark/src/data_explorer/r_data_explorer.rs
```diff
@@ -859,6 +859,110 @@ impl RDataExplorer {
...
+                    all_columns[a as usize]
+                        .column_name
+                        .cmp(&all_columns[b as usize].column_name)
+                });
+            },
+            amalthea::comm::data_explorer_comm::SearchSchemaSortOrder::Descending => {
+                matching_indices.sort_by(|&a, &b| {
+                    all_columns[b as usize]
+                        .column_name
+                        .cmp(&all_columns[a as usize].column_name)
```
lionel-: Slightly clearer? Since this is now the same as the previous branch except for the reverse()

```suggestion
                matching_indices.sort_by(|&a, &b| {
                    all_columns[a as usize]
                        .column_name
                        .cmp(&all_columns[b as usize].column_name)
                        .reverse()

```

Oh or maybe something like this:

```rust
match params.sort_order {
    amalthea::comm::data_explorer_comm::SearchSchemaSortOrder::Original => {
        // matching_indices is already in original order
    },
    order => {
        let ascending = matches!(
            order,
            amalthea::comm::data_explorer_comm::SearchSchemaSortOrder::Ascending
        );
        matching_indices.sort_by(|&a, &b| {
            let ord = all_columns[a as usize]
                .column_name
                .cmp(&all_columns[b as usize].column_name);
            if ascending { ord } else { ord.reverse() }
        });
    }
}
```

lionel- (approved):
LG!

The tests are now much clearer. I just worry about using functions rather than macros (cf comment inline).

crates/ark/src/lsp/completions/sources/unique/colon.rs
wesm: Yes, these are from `cargo +nightly fmt`, I reverted them for now

crates/ark/tests/data_explorer.rs
wesm: I asked Claude Code to work on this but it ran into issues with complex expressions / closures inside of macros, so there would be a bunch of work to get things to work everywhere with macros. Instead, it added more debugging output in the assertions for the sake of CI logs — I figure this can always be improved later

PR #890, opened 2025-08-04
Send `OpenFile` request with `kind` field

lionel-:
Companion PR for https://github.com/posit-dev/positron/pull/8816

### QA Notes

See linked PR.

### DIFF
--- a/crates/amalthea/src/comm/ui_comm.rs
+++ b/crates/amalthea/src/comm/ui_comm.rs
@@ -98,6 +98,18 @@ pub struct Range {
 	pub end: Position
 }
 
+/// Possible values for Kind in OpenEditor
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, strum_macros::Display, strum_macros::EnumString)]
+pub enum OpenEditorKind {
+	#[serde(rename = "path")]
+	#[strum(to_string = "path")]
+	Path,
+
+	#[serde(rename = "uri")]
+	#[strum(to_string = "uri")]
+	Uri
+}
+
 /// Parameters for the DidChangePlotsRenderSettings method.
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
 pub struct DidChangePlotsRenderSettingsParams {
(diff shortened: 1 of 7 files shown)


crates/amalthea/src/comm/ui_comm.rs
```diff
@@ -133,6 +145,10 @@ pub struct OpenEditorParams {
 
 	/// The column number to jump to
 	pub column: i64,
+
+	/// How to interpret the 'file' argument: as a file path or as a URI. If
+	/// omitted, defaults to 'path'.
+	pub kind: OpenEditorKind,
```
DavisVaughan, line 151: I feel like the argument names are a little mixed up, I feel like if we are generalizing, it should become

```
// The path of the file or URI to open
pub path: String

// How to interpret the `path` argument: as a file or a URI.
// If omitted, defaults to `File`
pub kind: OpenEditorPathKind
```

```
#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, strum_macros::Display, strum_macros::EnumString)]
pub enum OpenEditorPathKind {
	#[serde(rename = "file")]
	#[strum(to_string = "file")]
	File,

	#[serde(rename = "uri")]
	#[strum(to_string = "uri")]
	Uri
}
```

i.e. to me the ambiguous concept is a _path_, which is specialized to either a _file_ or a _uri_.

DavisVaughan: The positron side of this PR seems to agree with me, where you have

```
				if (ed.kind === 'uri') {
					file = URI.parse(ed.file);
				} else {
					file = URI.file(ed.file);
				}
```

i.e. you have an arbitrary `path`, if URI you `parse()` it, otherwise if File you `file()` it

DavisVaughan (approved):

crates/amalthea/src/comm/ui_comm.rs
lionel-: That makes sense

lionel-: Although I think I'd prefer not to change `path` as this would likely require updating the Python side. The current setup is backward compatible.

crates/ark/src/modules/positron/frontend-methods.R
```diff
@@ -39,7 +39,10 @@
     column = 0L
 ) {
     # Don't normalize if that's an `ark:` URI
-    if (!grepl("^ark:", file)) {
+    if (grepl("^ark:", file)) {
```
DavisVaughan:
```suggestion
    if (is_ark_uri(file)) {
```

PR #892, opened 2025-08-08
Recurse into special constructs to find document symbols

lionel-:
Addresses https://github.com/posit-dev/positron/issues/8881

We now consistently recurse inside these constructs to collect nested document symbols such as comment sections:

- if-else branches
- loops

The behaviour regarding nested assignments implemented in https://github.com/posit-dev/ark/pull/859 has been adapted so that `{` blocks in if-else branches and loops still counts as top-level. This way these assignments are all treated as consistently part of the outline:

```r
x <- 1

if (TRUE) {
  x <- 2
} else {
  x <- 3
}

for (i in xs) {
  x <- 4
}
```

In addition we also now collect comment sections in parameter lists of function definitions:

```r
x <- function(
  # section ----
  arg
) {
  ...
}
```

### QA Notes

Comment sections and other kinds of document symbols in if-else branches and loops are now part of the outline:

```r
# top ----

if (TRUE) {
  # if ----
  foo <- function() {}
} else {
  # else ----
  foo <- function() {}
}

for (i in xs) {
  # loop ----
  bar <- function() {}
}
```

Comment sections in parameter lists of function definitions are now included as well. We don't recurse through default arguments though.

```r
x <- function(
  # section ----
  arg = function() {
    # not a section ----
  }
) {
  ...
}
```

Screenshot 2025-08-08 at 13 54 15

### DIFF
--- a/crates/ark/src/lsp/symbols.rs
+++ b/crates/ark/src/lsp/symbols.rs
@@ -545,7 +545,7 @@ fn collect_function_parameters(
         contents,
         symbols,
         |_ctx, _child, _contents, _symbols| {
-            // Only collect sections
+            // We only collect sections and don't recurse inside parameters
             return Ok(());
         },
     )


crates/ark/src/lsp/symbols.rs
```diff
@@ -195,24 +195,142 @@ fn collect_symbols(
         },
 
         NodeType::BracedExpression => {
-            let old = ctx.top_level;
-            ctx.top_level = false;
             collect_list_sections(ctx, node, contents, current_level, symbols)?;
-            ctx.top_level = old;
         },
```
DavisVaughan: Does this change the behavior of general braced expressions? i.e.

```
# a section ----

{
  # another one ----
}
```

And do we have tests for this?

crates/ark/src/lsp/symbols.rs
```diff
@@ -414,26 +532,41 @@ fn collect_call_arguments(
...
+) -> anyhow::Result<()> {
+    collect_sections(
+        ctx,
+        &node,
+        contents,
+        0,
+        symbols,
+        |_ctx, _child, _contents, _symbols| {
+            // Only collect sections
+            return Ok(());
```
DavisVaughan: I'm not sure what this comment means. Does it mean no recursion? Can you provide more details?

crates/ark/src/lsp/symbols.rs
```diff
@@ -414,26 +532,41 @@ fn collect_call_arguments(
...
+    ctx: &mut CollectContext,
+    node: &Node,
+    contents: &Rope,
+    symbols: &mut Vec<DocumentSymbol>,
+) -> anyhow::Result<()> {
+    collect_sections(
+        ctx,
+        &node,
+        contents,
+        0,
```
DavisVaughan: Hmmmmmm `0`? Are you sure you shouldn't pass the `current_level` through to here? You do for the function body. Add a test either way about this behavior?

DavisVaughan: Or at least a comment here about why its 0

DavisVaughan (approved):
A few questions requesting clarifying behavior and a few more tests but I think this looks fairly straightforward?

crates/ark/src/lsp/symbols.rs
lionel-: I meant 0 here, but I think you're right to be confused. With the changes made over time, we no longer propagate the current level across syntax constructs. So I've just removed this argument.

crates/ark/src/lsp/symbols.rs
lionel-: Good point! We tested the brace-in-call and brace-in-function cases, but not the brace-in-program case.

I've tweaked a snapshot to include these (as well as the brace-in-brace case) and regenerated before all changes in this PR, then reran the tests with all the changes (including removed `current_level`), and there were no snapshot changes.

crates/ark/src/lsp/symbols.rs
lionel-: yep no recursion, the comment is now clearer about that

PR #898, opened 2025-08-18
data explorer: switch convert to code to be unsupported

isabelizimm:
In the effort to turn off the feature flag for the convert to code feature, I've flipped this to be unsupported for R. 

Originally, I had planned for the button to be greyed out, but this change will totally remove the button for R data explorer instances. This seems a bit more obvious that the feature is not available for this data type, rather than just inactive.


jennybc (approved):
And then I'll switch this back to `Supported` in the PR that adds tidyverse/dplyr capability, yes?

PR #904, opened 2025-08-21
Add support for table summary tool

lionel-:
For https://github.com/posit-dev/positron/issues/8343

Implemented by Claude ✨, including the tests. Everything looks legit and I've verified the summaries match those emitted on the Python side.

Claude had access to the Python side implementation which is more complicated as it opens up a temporary data explorer comm to make requests. It decided to use the internal profile summary utils, which is simpler here.

I've also cleaned up a bit:

- Environment → Variables renames
- Use a lock in tests to avoid them confusing in eacher.

### DIFF
--- a/crates/ark/src/variables/r_variables.rs
+++ b/crates/ark/src/variables/r_variables.rs
@@ -6,10 +6,12 @@
 //
 
 use amalthea::comm::comm_channel::CommMsg;
+use amalthea::comm::data_explorer_comm::ColumnSchema;
 use amalthea::comm::event::CommManagerEvent;
 use amalthea::comm::variables_comm::ClipboardFormatFormat;
 use amalthea::comm::variables_comm::FormattedVariable;
 use amalthea::comm::variables_comm::InspectedVariable;
+use amalthea::comm::variables_comm::QueryTableSummaryResult;
 use amalthea::comm::variables_comm::RefreshParams;
 use amalthea::comm::variables_comm::UpdateParams;
 use amalthea::comm::variables_comm::Variable;
@@ -33,6 +35,7 @@ use harp::utils::r_assert_type;
 use harp::utils::r_is_function;
 use harp::vector::CharacterVector;
 use harp::vector::Vector;
+use harp::ColumnNames;
 use libr::R_GlobalEnv;
 use libr::Rf_ScalarLogical;
 use libr::ENVSXP;
@@ -40,10 +43,13 @@ use stdext::spawn;
 
 use crate::data_explorer::r_data_explorer::DataObjectEnvInfo;
 use crate::data_explorer::r_data_explorer::RDataExplorer;
+use crate::data_explorer::summary_stats::summary_stats;
+use crate::data_explorer::utils::display_type;
 use crate::lsp::events::EVENTS;
 use crate::r_task;
 use crate::thread::RThreadSafe;
 use crate::variables::variable::PositronVariable;
+use crate::variables::variable::WorkspaceVariableDisplayType;
 use crate::view::view;
 
 /// Enumeration of treatments for the .Last.value variable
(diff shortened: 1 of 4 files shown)


crates/harp/src/column_names.rs
```diff
@@ -54,4 +54,15 @@ impl ColumnNames {
         }
         None
     }
+
+    pub fn get(&self, index: isize) -> anyhow::Result<Option<String>> {
```
lionel-, line 55: The AI code was using `get_unchecked()`. I think that was fine but I'm not comfortable letting AI code make assumptions that could panic.

crates/ark/src/variables/r_variables.rs
```diff
@@ -400,6 +407,126 @@ impl RVariables {
...
+            // Generate column profiles if requested
+            let mut column_profiles = Vec::new();
+            if query_types.contains(&"summary_stats".to_string()) {
+                let format_options = amalthea::comm::data_explorer_comm::FormatOptions {
+                    large_num_digits: 4,
+                    small_num_digits: 6,
+                    max_integral_digits: 7,
+                    max_value_length: 1000,
+                    thousands_sep: None,
+                };
```
lionel-: Same values as in https://github.com/posit-dev/positron/pull/8208/files#diff-9618192463e8ce6efcdbd4c9d5dc1bec8aa49e219adfe635e0625d3f49b2cc20R757-R764

crates/ark/src/variables/r_variables.rs
```diff
@@ -400,6 +407,126 @@ impl RVariables {
...
+    ///
+    /// Returns summary information about the table including schemas and profiles.
+    fn query_table_summary(
+        &mut self,
+        path: &Vec<String>,
+        query_types: &Vec<String>,
+    ) -> anyhow::Result<QueryTableSummaryResult> {
+        r_task(|| {
+            let env = self.env.get().clone();
+            let obj = PositronVariable::resolve_data_object(env.clone(), &path)?;
```
DavisVaughan: Extra clone?

```suggestion
            let obj = PositronVariable::resolve_data_object(env, &path)?;
```

crates/ark/src/variables/r_variables.rs
```diff
@@ -400,6 +407,126 @@ impl RVariables {
...
+
+            // Generate column schemas
+            let mut column_schemas = Vec::new();
+            for i in 0..num_cols {
+                let column_name = match column_names.get(i as isize) {
+                    Ok(Some(name)) => name,
+                    Ok(None) | Err(_) => format!("V{}", i + 1),
+                };
+
+                // Get column type information
```
DavisVaughan: Isn't all of this code basically just `r_get_shape()` from `RDataExplorer`?

Can we extract out a common utility instead of copy/pasting like this?

This is why I really don't like AI PRs 😞, the AI would never look at this and go "yea I see this is duplicated over there, lets make a common helper"

crates/harp/src/column_names.rs
```diff
@@ -54,4 +54,15 @@ impl ColumnNames {
...
+        if let Some(names) = &self.names {
+            if let Some(name) = names.get(index)? {
+                if name.len() > 0 {
+                    return Ok(Some(name));
+                }
```
DavisVaughan: Is that returning `None` on `""`? I think it should return `""` if so

crates/ark/src/variables/r_variables.rs
lionel-: I'm sure we'll get someday the option to use an anxious agent worried about good practices and that isn't afraid to interrupt the human with questions

crates/harp/src/column_names.rs
lionel-: This is from the `get_unchecked()` variant that you implemented for the data explorer. I see the only call site does:

```rust
                let column_name = match column_names.get_unchecked(i) {
                    Some(name) => name,
                    None => String::from(""),
                };
```

So it would be safe to change it. I don't have any particular opinion about this so I'll change as you suggested.

lionel-:
It feels like we should have some kind of safe `struct Table` API to make this stuff less verbose, easier to use, and safer.

PR #905, opened 2025-08-21
Set as interactive on startup, set `options(device =)` to a string

DavisVaughan:
My preferred alternative to https://github.com/posit-dev/ark/pull/895

In a fresh R session, `.Device` in RStudio returns `"null device"` and `getOption("device")` returns `"RStudioGD"`. i.e. R doesn't have a current device active, but it knows to use a function named `RStudioGD()` when it needs to make one (and indeed you can call `RStudioGD` at the console and you can find it yourself).

For these reasons, I do not think we should inject ourselves as the current active device on startup. I think we should also follow RStudio's lead.

To solve the problem a different way, we just recognize that the main issue is that we are calling `grDevices::deviceIsInteractive()` to register ourselves as an interactive device too late - we are calling it at first plot time, but should call it at startup time. I have made that change in this PR, and in theory that should have been enough to fix the issue.

In practice there is one other issue I had to deal with. We set `options(device = <fn>)` where RStudio sets `options(device = "RStudioGD")`. This is surprisingly important. In a fresh session, `grDevices::dev.interactive(orNone = TRUE)` (used by `demo(graphics)`) will look to see if the current device is in the set of known devices provided by `deviceIsInteractive()`. However, if that device is `"null device"`, then it will consult `getOption("device")` and see if that _name_ is in the set provided by `deviceIsInteractive()` instead. But if you set `options(device =)` to a _function_ rather than a _name_ then you don't get to take advantage of this nice feature. I've reworked a few things to take advantage of this now.

I've also updated docs and added tests to assert our beliefs about how these functions work on startup.

### DIFF
--- a/doc/graphics-devices.md
+++ b/doc/graphics-devices.md
@@ -50,19 +50,33 @@ For pure Jupyter settings outside of Positron, we don't have "dynamic" plots. In
 
 # Ark graphics device
 
-Ark creates a "shadow" graphics device that manages other graphics devices. We technically wrap a `grDevices::png()` and inherit all of the default hooks and behaviors that come with that, and then we inject our own hooks on top of specific png hooks. For example, when the `newPage` hook is called we call the png device's `newPage` hook manually and then also layer in our own `newPage` behavior. This is how we learn when changes have occurred, when to record a plot, when we are deactivated, etc.
+Ark creates a "shadow" graphics device that manages other graphics devices. We technically wrap `grDevices::png()` or `ragg::agg_record()` and inherit all of the default hooks and behaviors that come with that, and then we inject our own hooks on top of specific png hooks. For example, when the `newPage` hook is called we call the device's `newPage` hook manually and then also layer in our own `newPage` behavior. This is how we learn when changes have occurred, when to record a plot, when we are deactivated, etc.
+
+We prefer using the ragg device if it is available, as it is cheaper and has better behavior with some features that surprisingly do still use the underlying device, like rendering of Chinese text in a plot.
 
 # Device interactivity
 
-Certain plotting functions like `stats:::plot.lm()` will draw multiple plots in a row, pausing for the user to hit enter before advancing to the next page. This requires a few things to work properly:
+Certain plotting functions like `stats:::plot.lm()` or `demo(graphics)` will draw multiple plots in a row, pausing for the user to hit enter before advancing to the next page. This requires a few things to work properly:
 
--   `grDevices::deviceIsInteractive(name)` must be called during Ark graphics device initialization with `name` set to our Ark graphics device name to "register" ourselves as a known interactive graphics device.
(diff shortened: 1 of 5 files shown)


lionel-:
> For these reasons, I do not think we should inject ourselves as the current active device on startup as proposed by https://github.com/posit-dev/ark/pull/895. I think we should also follow RStudio's lead and let the device be "null device" until a plot is actually required.

Sorry I don't understand. What are the reasons?

lionel-:
I'm not sure about "standard", or whether either approach makes a material difference, but feel free to go with the approach you're more comfortable with.

doc/graphics-devices.md
```diff
@@ -50,19 +50,33 @@ For pure Jupyter settings outside of Positron, we don't have "dynamic" plots. In
...
+
+We prefer using the ragg device if it is available, as it is cheaper and has better behavior with some features that surprisingly do still use the underlying device, like rendering of Chinese text in a plot.
 
 # Device interactivity
 
-Certain plotting functions like `stats:::plot.lm()` will draw multiple plots in a row, pausing for the user to hit enter before advancing to the next page. This requires a few things to work properly:
+Certain plotting functions like `stats:::plot.lm()` or `demo(graphics)` will draw multiple plots in a row, pausing for the user to hit enter before advancing to the next page. This requires a few things to work properly:
+
+-   Internally, `ARK_GRAPHICS_DEVICE_NAME` is set to `"ark.graphics.device"`, which matches the name of a function we expose named `ark.graphics.device()`, which is in charge of creating a new Ark graphics device and is findable by `get0("ark.graphics.device", globalenv(), inherits = TRUE)`.
```
lionel-:
```suggestion
-   Internally, `ARK_GRAPHICS_DEVICE_NAME` is set to `".ark.graphics.device"`, which matches the name of a function we expose named `ark.graphics.device()`, which is in charge of creating a new Ark graphics device and is findable by `get0(".ark.graphics.device", globalenv(), inherits = TRUE)`.
```

doc/graphics-devices.md
```diff
@@ -50,19 +50,33 @@ For pure Jupyter settings outside of Positron, we don't have "dynamic" plots. In
...
+-   `grDevices::dev.interactive()` should return `TRUE` after you've created your first plot (before then, the device is `"null device"` and this returns `FALSE`, this matches RStudio).
 
-    -   One way we could possibly improve on this is to render an initial version of each plot in the opening notification we send to Positron using some default dimensions and device type. Positron could also supply us these defaults to use during the startup procedure.
+    -   This is used in the default value of the `ask` argument of `stats:::plot.lm()` and by the time `ask` is evaluated our device has been created so this returns `TRUE` there as intended.
```
lionel-, line 79: Perhaps mention the user is still able to set a different device in Rprofile like `options(device = "quartz")`, our setup won't interfere.

lionel-:
So  the material difference between the two approaches is that mine pushes a new device to the device ring on startup, whereas yours doesn't. I agree we should have an empty device ring on startup. (Furthermore I see that the started device in my PR is always Ark's, but we'd like the user to be able to override it e.g. to `"quartz"`.)

PR #906, opened 2025-08-25
Add support for configuring a custom Posit Package Manager URL

atheriel:
This commit builds on the existing support for installing packages from Posit Public Package Manager via `--default-repos=posit-ppm` by allowing users to pass a specific Package Manager repository, overriding the hardcoded https://packagemanager.posit.co/cran/latest URL.

This allows users of self-hosted Package Manager installations to benefit from our logic for determining the the appropriate Linux binaries automatically. It also opens the door to users pointing at a specific CRAN snapshot on Public Package Manager rather than `latest`.

Custom URLs are passed via a new `--ppm-repo` option, which is mutually exclusive with `--default-repos` and `--repos-conf`.

Unit tests for the repository URL construction (including for the existing but previously untested logic) are included.

### DIFF
--- a/crates/ark/src/main.rs
+++ b/crates/ark/src/main.rs
@@ -12,6 +12,7 @@ use std::env;
 
 use amalthea::kernel;
 use amalthea::kernel_spec::KernelSpec;
+use anyhow::Context;
 use ark::interface::SessionMode;
 use ark::logger;
 use ark::repos::DefaultRepos;
@@ -48,6 +49,7 @@ Available options:
                          "none" (do not alter the 'repos' option in any way)
 --repos-conf             Set the default repositories to use from a configuration file
                          containing a list of named repositories (`name = url`)
+--default-ppm-repo       Set the default repositories to a custom Posit Package Manager URL.
 --version                Print the version of Ark
 --log FILE               Log to the given file (if not specified, stdout/stderr
                          will be used)
@@ -139,12 +141,12 @@ fn main() -> anyhow::Result<()> {
                 if let Some(repos) = argv.next() {
                     if default_repos != DefaultRepos::Auto {
                         return Err(anyhow::anyhow!(
-                            "The default repository options can only be specified once; (found '{repos}, had {default_repos:?}')."
+                            "Only one of `--default-repos`, `--repos-conf`, or `--default-ppm-repo` can be specified."
                         ));
                     }
                     default_repos = match repos.as_str() {
                         "rstudio" => DefaultRepos::RStudio,
-                        "posit-ppm" => DefaultRepos::PositPPM,
+                        "posit-ppm" => DefaultRepos::PositPackageManager(None),
                         "none" => DefaultRepos::None,
                         _ => {
                             return Err(anyhow::anyhow!(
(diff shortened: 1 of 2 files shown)


crates/ark/src/main.rs
```diff
@@ -180,6 +182,34 @@ fn main() -> anyhow::Result<()> {
                     ));
                 }
             },
+            "--default-ppm-repo" => {
+                if let Some(url) = argv.next() {
+                    if default_repos != DefaultRepos::Auto {
+                        return Err(anyhow::anyhow!(
+                            "Only one of --default-repos, --repos-conf, or --default-ppm-repo can be specified."
```
lionel-: Perhaps update the other error messages for the other options to make them consistent? Might be worth extracting a little helper.

crates/ark/src/main.rs
```diff
@@ -180,6 +182,34 @@ fn main() -> anyhow::Result<()> {
...
+                    if segments.count() != 2 {
+                        return Err(anyhow::anyhow!(
+                            "Invalid Package Manager repository URL: {}",
+                            url
+                        ));
+                    }
+                    default_repos = DefaultRepos::PositPackageManager(Some(parsed))
+                } else {
+                    return Err(anyhow::anyhow!(
+                        "A repository URL must follow the --default-ppm-repo option."
```
lionel-: Somehow I find this clearer, I think because it's context → requirement.
```suggestion
                        "The `--default-ppm-repo` option must be followed by a repository URL."
```

crates/ark/src/repos.rs
```diff
@@ -211,26 +230,31 @@ fn get_p3m_linux_repo(linux_name: String) -> String {
...
+    if distro.is_empty() {
+        return generic_url.to_string();
+    }
+
+    let mut distro_url = generic_url.clone();
+    if let Some(segments) = distro_url.path_segments() {
+        let parts: Vec<&str> = segments.collect();
+        if parts.len() == 2 {
+            distro_url.set_path(&format!("{}/__linux__/{}/{}", parts[0], distro, parts[1]));
+        }
```
lionel-: Emit a warning or return an error when `parts` is not as expected?

crates/ark/src/repos.rs
```diff
@@ -296,3 +323,93 @@ fn get_p3m_binary_package_repo() -> String {
...
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    #[cfg(target_os = "linux")]
+    fn test_get_p3m_linux_repo() {
```
lionel-:
```suggestion
    fn test_get_ppm_linux_repo() {
```

lionel- (approved):

crates/ark/src/main.rs
atheriel: Done! I don't think a helper was worth it in this case.

crates/ark/src/repos.rs
atheriel: We kind of rely on the `DefaultRepos::PositPackageManager` variant holding valid data right now, because this function doesn't return a `Result`.

Do you want me to introduce an `anyhow`-style error?

lionel-: yep that'd be fine I think

atheriel: Done, and I added some more tests and logging to cover these error paths.

PR #909, opened 2025-08-26
Data Explorer: Export column labels to data explorer frontend

wesm:
R side of https://github.com/posit-dev/positron/pull/9179.

### DIFF
--- a/crates/ark/src/data_explorer/r_data_explorer.rs
+++ b/crates/ark/src/data_explorer/r_data_explorer.rs
@@ -626,14 +626,19 @@ impl RDataExplorer {
                     harp::TableKind::Dataframe => {
                         let col_obj = harp::RObject::view(col);
                         col_obj.get_attribute("label").and_then(|label_obj| {
-                            // Convert to string if it's a character vector
-                            if label_obj.kind() == STRSXP && harp::r_length(label_obj.sexp) > 0 {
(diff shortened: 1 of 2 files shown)


crates/ark/src/data_explorer/r_data_explorer.rs
```diff
@@ -619,8 +621,27 @@ impl RDataExplorer {
...
+                        col_obj.get_attribute("label").and_then(|label_obj| {
+                            // Convert to string if it's a character vector
+                            if label_obj.kind() == STRSXP && harp::r_length(label_obj.sexp) > 0 {
+                                CharacterVector::new(label_obj.sexp)
+                                    .ok()
+                                    .and_then(|cv| cv.get_unchecked(0))
+                            } else {
+                                None
+                            }
+                        })
```
lionel-: (Just a comment, feel free to ignore)

This looks correct but it also feels like such a simple operation should look simpler. I wish Rust had stabilised `try {}` already so you could use `?` to collapse options and keep a linear flow. You could use a self-called closure instead if you wanted to.

The `kind()` check seems a bit redundant with `CharacterVector::new()` which also does the check and returns `None()` after `ok()` if the type is not correct.

wesm: 👍

PR #910, opened 2025-08-27
Key by `Url` instead of `Path` in workspace indexer

lionel-:
Addresses https://github.com/posit-dev/positron/issues/8790

The warnings only appear on Windows because the conversion to `Path` actually succeeds on Unixes. So we were calling `create() with non-file URIs on Unixes but not on Windows. To fix the warnings and make things more consistent, this PR:

- Now uses `Url` (wrapped in a new `FileId` struct) as key in our map of file to indexer. This is more general and simplifies a bunch of `Path` to `Url` conversions, avoiding by the same token potential warnings. The `FileId` wrapper is used for stronger internal type checking / self-documentation, but not exposed in the public API.

- Declines to index non-file URIs. This makes the check for `[rR]` files stronger.

### DIFF
--- a/crates/ark/src/lsp/completions/sources/composite/workspace.rs
+++ b/crates/ark/src/lsp/completions/sources/composite/workspace.rs
@@ -71,7 +71,7 @@ fn completions_from_workspace(
     let token = token.as_str();
 
     // get entries from the index
-    indexer::map(|path, symbol, entry| {
+    indexer::map(|uri, symbol, entry| {
         if !symbol.fuzzy_matches(token) {
             return;
         }
@@ -87,15 +87,21 @@ fn completions_from_workspace(
                     return;
                 });
 
-                // add some metadata about where the completion was found
-                let mut path = path.to_str().unwrap_or_default();
-                for folder in &state.workspace.folders {
-                    if let Ok(folder) = folder.to_file_path() {
-                        if let Some(folder) = folder.to_str() {
-                            if path.starts_with(folder) {
-                                path = &path[folder.len() + 1..];
-                                break;
-                            }
+                // Add some metadata about where the completion was found
+                let mut path = uri.as_str().to_owned();
+
+                if uri.scheme() == "file" {
+                    let Ok(file_path) = uri.to_file_path() else {
+                        return;
+                    };
+
+                    for folder in &state.workspace.folders {
+                        let Ok(folder_path) = folder.to_file_path() else {
+                            continue;
+                        };
+                        if let Ok(relative_path) = file_path.strip_prefix(&folder_path) {
+                            path = relative_path.to_string_lossy().to_string();
+                            break;
                         }
                     }
                 }
--- a/crates/ark/src/lsp/completions/sources/composite/workspace.rs
+++ b/crates/ark/src/lsp/completions/sources/composite/workspace.rs
@@ -91,19 +91,17 @@ fn completions_from_workspace(
                 let mut path = uri.as_str().to_owned();
 
                 if uri.scheme() == "file" {
-                    let Ok(file_path) = uri.to_file_path() else {
-                        return;
-                    };
-
-                    for folder in &state.workspace.folders {
-                        let Ok(folder_path) = folder.to_file_path() else {
-                            continue;
-                        };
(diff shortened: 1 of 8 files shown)


crates/ark/src/lsp/completions/sources/composite/workspace.rs
```diff
@@ -87,15 +87,21 @@ fn completions_from_workspace(
...
-                                path = &path[folder.len() + 1..];
-                                break;
-                            }
+                // Add some metadata about where the completion was found
+                let mut path = uri.as_str().to_owned();
+
+                if uri.scheme() == "file" {
+                    let Ok(file_path) = uri.to_file_path() else {
+                        return;
+                    };
```
DavisVaughan: Isn't this a change in behavior? i.e. it's now saying we dont push the completion if we can't figure out where it came from, where previously i think we just pushed it anyways with that default `path`?

crates/ark/src/lsp/indexer.rs
```diff
@@ -32,6 +31,28 @@ use crate::treesitter::NodeType;
...
+    /// The URL representing the file
+    url: Url,
+}
+
+impl FileId {
+    pub fn from_uri(uri: &Url) -> Self {
+        Self { url: uri.clone() }
+    }
+
+    #[allow(unused)]
```
DavisVaughan: Is it actually still unused at this point?

crates/ark/src/lsp/indexer.rs
```diff
@@ -32,6 +31,28 @@ use crate::treesitter::NodeType;
...
+/// FileId represents a unique identifier for a file in the workspace index
+#[derive(Clone, Eq, PartialEq, Hash, Debug)]
+pub struct FileId {
+    /// The URL representing the file
+    url: Url,
+}
+
+impl FileId {
+    pub fn from_uri(uri: &Url) -> Self {
+        Self { url: uri.clone() }
```
DavisVaughan: Are you not typically owning it? I typically prefer `uri: Url` and then I like to see an explicit `clone()` at the call site.

This both hides and forces the clone, meaning:
- I think code is efficient, but its not, i.e. `FileId::from_uri(&this.uri)` clones but doesn't look like it does
- I can't actually be efficient when i want to, i.e. `FileId::from_uri(owned_uri)` is not possible

DavisVaughan: Even if it results in a lot of `clone()` calls at call sites, I'd argue that is a good thing because it forces you to think about whether or not this is actually the correct abstraction or not.

crates/ark/src/lsp/indexer.rs
```diff
@@ -238,8 +258,19 @@ pub fn filter_entry(entry: &DirEntry) -> bool {
     true
 }
 
-pub(crate) fn create(path: &Path) -> anyhow::Result<()> {
-    // Only index R files
+pub(crate) fn create(uri: &Url) -> anyhow::Result<()> {
+    // Only index R files for file URIs. This discards `inmemory` (Console) and
+    // `ark` schemes in particular.
+
+    log::error!("=== Creating index for URI: {uri:?}");
```
DavisVaughan:
```suggestion
    log::trace!("Creating index for URI: {uri:?}");
```

😞

DavisVaughan (approved):
I like the idea but have a few questions about behavior change (maybe it needs a test?) and some efficiency things

crates/ark/src/lsp/indexer.rs
lionel-: This is a debug printf of mine, not AI slop :P

DavisVaughan: 😬

DavisVaughan: i appreciate you keeping me on my toes 😆

crates/ark/src/lsp/indexer.rs
lionel-: yes but I had to go from `pub(crate)` to `pub` to solve a transitive visibility issue so we can now remove it

crates/ark/src/lsp/indexer.rs
lionel-: Thanks for catching things I no longer see in diffs after working on a PR for a while...

PR #916, opened 2025-09-04
Fix up formatting for S3 objects

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/8053

With this PR, you can see `Surv` objects in the Data Explorer:

Screenshot 2025-09-03 at 9 29 43 PM

And we no longer get this error I point out here, when you click to expand a `Surv` object in the Variables Pane:

Screenshot 2025-09-03 at 8 38 45 PM

Good news on both fronts! 🎉 

This manhandling of `dim()` was introduced for https://github.com/posit-dev/positron/issues/1862, also motivated by `Surv` objects. However, I think we must have updated other pieces that handle this, because those original examples are fine even after I remove this code:

Screenshot 2025-09-03 at 9 32 18 PM

No panics! More good news 🎉 

There is still a little bit of bad news. Now when you click to expand a vector of `Surv` objects in the Variables pane, there is a _new_ error:

Screenshot 2025-09-03 at 9 35 03 PM

Try with this code:

```r
library(tidyverse)
library(survival)

res <- modeldata::cat_adoption |>
    slice_sample(n = 5) |>
    mutate(
        event_time = Surv(time, event),
        .keep = "unused",
        .before = everything()
    )
```

And the error message has:

> Error expanding variable item: Failed to process positron.variables request: Error evaluating 'harp_subset_vec(structure(c(76, 95, 28, 104, 18, 1, 1, 1, 1, 1), dim = c(5L, 2L), dimnames = list(NULL, c("time", "status" )), type = "right", class = "Surv"), 6:10)': subscript out of bounds R backtrace: 0. harp_subset_vec(structure(c(76, 95, 28, 104, 18, 1, 1, 1, 1, 1), dim = c(5L, 2L), dimnames = list(NULL, c("time", "status" )), type = "right", class = "Surv"), 6:10)

Notice that it is trying to subset `6:10` here, which doesn't exist. I suspect this is related to https://github.com/posit-dev/ark/pull/646, and how we changed to subsetting _before_ formatting.

@dfalbel could you take a look at this and see if you have suggested changes we could make here to make this work in the Variables pane? It would be nice to just get it all fixed and write some tests to make sure it does not regress.


.github/workflows/test-linux.yml
```diff
@@ -65,6 +65,7 @@ jobs:
             tibble
             haven
             R6
+            survival
```
juliasilge, line 68: I added survival here, despite it being (because it is?) such a problem child so we can write tests for this especially weird/bad case.

crates/ark/src/data_explorer/format.rs
```diff
@@ -673,6 +674,26 @@ mod tests {
...
     }
 
+    #[test]
+    fn test_s3_formatting() {
+        r_task(|| {
+            if !package_is_installed("survival") {
+                return;
+            }
+
+            let data =
```
juliasilge, line 684: I am not sure if we can write a test for the Variables pane that is as directly helpful as this one, but it would be great if so!

dfalbel (commented):
We should likely dispatch on the S3 implementation of length here: 

https://github.com/posit-dev/ark/blob/1bed8ebeaa0010f727413766efe54510774dd1ad/crates/harp/src/vector/formatted_vector.rs#L63-L64

and 

https://github.com/posit-dev/ark/blob/1bed8ebeaa0010f727413766efe54510774dd1ad/crates/harp/src/vector/formatted_vector.rs#L72-L73

crates/ark/src/data_explorer/format.rs
dfalbel: We could similar tests in: 

https://github.com/posit-dev/ark/blob/1bed8ebeaa0010f727413766efe54510774dd1ad/crates/harp/src/vector/formatted_vector.rs#L273-L284

juliasilge:
Thanks, @dfalbel!

- In 539b8ae49bef26ba36199a954297ef459e546859 I am showing what I tried using the `.ark.register_method()` approach, but adding these keeps R from starting up ("R 4.5.0 failed to start up (exit code -1) The process exited abnormally (signal: 6 (SIGABRT))")
- In 2a132ddd2515335436bdd9e74a56a49f99963279 I am showing how I started to sketch out dispatching on `length()` but I could not get that to work because a `Surv` object is a _matrix_; I could not get the Variables pane handling for matrices untangled from how/when we format vs. subset.

I tend to think it might be better to use `.ark.register_method()` and make special handling for `Surv` objects? But open to what you think.

crates/ark/src/modules/positron/methods.R
```diff
@@ -48,7 +54,7 @@ check_caller_allowed <- function() {
 
 check_register_args <- function(generic, class) {
     stopifnot(
-        is_string(generic),
+        is.character(generic) && length(generic) == 1 && !is.na(generic),
```
dfalbel: `is_string` is not available when executing this during namespace load. So we just inline it here.

PR #919, opened 2025-09-12
Remove S3 overrides for `shiny.tag` and `shiny.tag.list`

juliasilge:
This PR addresses:

- https://github.com/posit-dev/positron/issues/6708
- https://github.com/posit-dev/positron/issues/5338

The specific set of S3 overrides were brought over from the RStudio notebook implementation, but I think we're finding that while rendering `shiny.tag` and `shiny.tag.list` may have made sense in a notebook context, it is turning out somewhat unexpected for most use cases in a Console/Plots/Viewer situation.

It is possible that there are some R packages that render widgets with these classes, but I spent a bit of time looking through what's out there, and I didn't find anything very compelling for this treatment:
https://github.com/search?q=%22shiny.tag.list%22&type=code

With this change, what we see is that...

- kableExtra renders without error, even after loading htmlwidgets

Screenshot 2025-09-11 at 7 02 51 PM

- a `shiny.tag` prints in the console as expected, even after loading devtools

Screenshot 2025-09-11 at 7 04 07 PM

- htmlwidgets that are plots continue to render in the Plots pane

Screenshot 2025-09-11 at 7 05 55 PM

- all htmlwidgets I checked are behaving as expected

Screenshot 2025-09-11 at 7 06 27 PM

We haven't saved this file since we turned air on this repo, so I just checked in the unrelated air formatting changes while I was here.

### DIFF
--- a/crates/ark/src/modules/positron/html_widgets.R
+++ b/crates/ark/src/modules/positron/html_widgets.R
@@ -36,15 +36,11 @@
 #' @export
 .ps.viewer.addOverrides <- function() {
     add_s3_override("print.htmlwidget", .ps.view_html_widget)
-    add_s3_override("print.shiny.tag", .ps.view_html_widget)
-    add_s3_override("print.shiny.tag.list", .ps.view_html_widget)
 }
 
 #' @export
 .ps.viewer.removeOverrides <- function() {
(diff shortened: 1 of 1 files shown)


jmcphers (approved):
LGTM, thanks for taking care of this!

PR #920, opened 2025-09-13
Add showPrompt API handlers

jmcphers:
Adds API handlers for `rstudioapi::showPrompt` and creates a UI event of the name name.

The only interesting thing here is that the RStudio implementation of this method implements the timeout by (theoretically) limiting the amount of time the backend waits for the frontend. In Positron, we implement the timeout entirely on the frontend. 

In practice, it doesn't look like the `timeout` parameter works at all on RStudio, so the difference is probably academic. 

Part of the fix for https://github.com/posit-dev/positron/issues/9099.

### DIFF
--- a/crates/ark/src/modules/rstudio/dialogs.R
+++ b/crates/ark/src/modules/rstudio/dialogs.R
@@ -14,6 +14,9 @@
 
 
 #' @export
-.rs.api.showPrompt <- function(title, message, default, timeout = 60L) {
+.rs.api.showPrompt <- function(title, message, default) {
+    # rstudioapi doesn't pass the timeout directly but sets it as
+    # `rstudioapi.remote.timeout` option
+    timeout <- getOption('rstudioapi.remote.timeout', 60L)
     .ps.ui.showPrompt(title, message, default, timeout)
 }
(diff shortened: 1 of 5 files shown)


crates/ark/src/modules/rstudio/dialogs.R
```diff
@@ -11,3 +11,12 @@
 .rs.api.showQuestion <- function(title, message, ok = NULL, cancel = NULL) {
     .ps.ui.showQuestion(title, message, ok, cancel)
 }
+
+
+#' @export
+.rs.api.showPrompt <- function(title, message, default) {
+    # rstudioapi doesn't pass the timeout directly but sets it as
+    # `rstudioapi.remote.timeout` option
```
juliasilge, line 19: The actual R function does pass it in as plain old `60`, so I think we need to do some kind of coercion here to integer.

If I run the function without specifying `timeout`, I see an error:

```
Unexpected vector type (expected integer; got double)
```

If I pass in something like `timeout = 40L`, then I see the new dialog working:

Screenshot 2025-09-12 at 8 08 42 PM

jmcphers: You're right, thanks for catching that! Addressed in https://github.com/posit-dev/ark/pull/920/commits/eded84af802f21e70a2d0cad526f97b1ddb80e61.

PR #924, opened 2025-09-23
Data Explorer: Fix exporting sorted columns in data explorer

wesm:
This is the R fix for https://github.com/posit-dev/positron/issues/9316. The sort view indices were not being propagated during the export process


dfalbel (approved):
LGTM!

Do we also need to propagate in?

https://github.com/posit-dev/ark/blob/202d84300aac488f7e2a9e0a1f7751ced844447d/crates/ark/src/data_explorer/export_selection.rs#L92-L101

Is the front-end sending the expected row ids already? 

Actually I see that there's a test for this already, so we don't need to propagate...

PR #925, opened 2025-09-24
Get rid of warnings about unnecessary transmutation

jennybc:
I'm about to do ark dev work on Windows, so I tried to make sure it was building and testing cleanly first. I discovered some warnings about unnecessary transmutation. I'm not sure if we want to handle this with changes like this PR or by turning off such warnings? I used Copilot to create these changes.

Also, I see that this file was originally created with a code generation process, but it seems like it's received various manual tweaks since then, so I assume it's fair game.

I paste the warnings below and also link to a recent CI run, where these are also showing up:

https://github.com/posit-dev/ark/actions/runs/17880174813/job/50846807788#step:8:22

```
   Compiling libr v0.1.0 (C:\Users\jenny\ark\crates\libr)
warning: unnecessary transmute
   --> crates\libr\src\sys\windows\types.rs:211:18
    |
211 |         unsafe { ::std::mem::transmute(self._bitfield_1.get(16usize, 16u8) as u32) }
    |                  ---------------------^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |                  |
    |                  help: replace this with: `u32::cast_signed`
    |
    = note: `#[warn(unnecessary_transmutes)]` on by default

warning: unnecessary transmute
   --> crates\libr\src\sys\windows\types.rs:216:28
    |
216 |             let val: u32 = ::std::mem::transmute(val);
    |                            ---------------------^^^^^
    |                            |
    |                            help: replace this with: `i32::cast_unsigned`

warning: unnecessary transmute
   --> crates\libr\src\sys\windows\types.rs:231:47
    |
231 |             let RstartVersion: u32 = unsafe { ::std::mem::transmute(RstartVersion) };
    |                                               ---------------------^^^^^^^^^^^^^^^
    |                                               |
    |                                               help: replace this with: `i32::cast_unsigned`
```

### DIFF
--- a/crates/libr/src/sys/windows/types.rs
+++ b/crates/libr/src/sys/windows/types.rs
@@ -213,8 +213,8 @@ impl structRstart {
     }
     #[inline]
     pub fn set_RstartVersion(&mut self, val: ::std::os::raw::c_int) {
-        let val: u32 = i32::cast_unsigned(val);
-        self._bitfield_1.set(16usize, 16u8, val as u64)
+        self._bitfield_1
+            .set(16usize, 16u8, i32::cast_unsigned(val) as u64)
     }
     #[inline]
     pub fn new_bitfield_1(
@@ -226,10 +226,7 @@ impl structRstart {
             let NoRenviron: u32 = unsafe { ::std::mem::transmute(NoRenviron) };
             NoRenviron as u64
         });
-        __bindgen_bitfield_unit.set(16usize, 16u8, {
-            let RstartVersion: u32 = i32::cast_unsigned(RstartVersion);
-            RstartVersion as u64
-        });
+        __bindgen_bitfield_unit.set(16usize, 16u8, i32::cast_unsigned(RstartVersion) as u64);
         __bindgen_bitfield_unit
     }
 }


jennybc:
Indeed the warnings are no longer there in CI: https://github.com/posit-dev/ark/actions/runs/17966840203/job/51100976919#step:8:21

crates/libr/src/sys/windows/types.rs
```diff
@@ -208,29 +208,20 @@ impl structRstart {
     }
     #[inline]
     pub fn RstartVersion(&self) -> ::std::os::raw::c_int {
-        unsafe { ::std::mem::transmute(self._bitfield_1.get(16usize, 16u8) as u32) }
+        self._bitfield_1.get(16usize, 16u8) as ::std::os::raw::c_int
```
DavisVaughan: I do not interpret "help: replace this with: `u32::cast_signed`" as identical to your change.

I would like to do _exactly_ what the compiler suggests in this scenario. I think what you have is slightly different.

I would expect to see this here

```rust
u32::cast_signed(self._bitfield_1.get(16usize, 16u8) as u32)
```

This seems to more exactly capture the spirit of this

This function was not available before 1.87.0 so it makes sense that we would not have had access to this then.

crates/libr/src/sys/windows/types.rs
```diff
@@ -208,29 +208,20 @@ impl structRstart {
...
     }
     #[inline]
     pub fn set_RstartVersion(&mut self, val: ::std::os::raw::c_int) {
-        unsafe {
-            let val: u32 = ::std::mem::transmute(val);
-            self._bitfield_1.set(16usize, 16u8, val as u64)
-        }
+        self._bitfield_1.set(16usize, 16u8, val as u64)
```
DavisVaughan:
```suggestion
        self._bitfield_1.set(16usize, 16u8, i32::cast_unsigned(val) as u64)
```

Again, reading the help message closely.

crates/libr/src/sys/windows/types.rs
```diff
@@ -208,29 +208,20 @@ impl structRstart {
...
-        __bindgen_bitfield_unit.set(0usize, 16u8, {
-            let NoRenviron: u32 = unsafe { ::std::mem::transmute(NoRenviron) };
-            NoRenviron as u64
-        });
-        __bindgen_bitfield_unit.set(16usize, 16u8, {
-            let RstartVersion: u32 = unsafe { ::std::mem::transmute(RstartVersion) };
-            RstartVersion as u64
-        });
+        __bindgen_bitfield_unit.set(0usize, 16u8, NoRenviron as u64);
+        __bindgen_bitfield_unit.set(16usize, 16u8, RstartVersion as u64);
```
DavisVaughan:
```suggestion
        __bindgen_bitfield_unit.set(0usize, 16u8, NoRenviron as u64);
        __bindgen_bitfield_unit.set(16usize, 16u8, i32::cast_unsigned(RstartVersion) as u64);
```

crates/libr/src/sys/windows/types.rs
```diff
@@ -208,29 +208,20 @@ impl structRstart {
...
         let mut __bindgen_bitfield_unit: __BindgenBitfieldUnit<[u8; 4usize]> = Default::default();
-        __bindgen_bitfield_unit.set(0usize, 16u8, {
-            let NoRenviron: u32 = unsafe { ::std::mem::transmute(NoRenviron) };
-            NoRenviron as u64
-        });
-        __bindgen_bitfield_unit.set(16usize, 16u8, {
-            let RstartVersion: u32 = unsafe { ::std::mem::transmute(RstartVersion) };
-            RstartVersion as u64
-        });
+        __bindgen_bitfield_unit.set(0usize, 16u8, NoRenviron as u64);
```
DavisVaughan: It does not look like the `NoRenviron` transmute shows up in your warnings.

And indeed this is an `Rboolean`, so it has no equivalent to `i32::cast_unsigned()` or `u32::cast_signed()`, so I think we should leave the old implementation in place for this one.

DavisVaughan (changes requested):
I believe the compiler is telling you the precise way to preserve the semantics of this operation by using `i32::cast_unsigned()` or `u32::cast_signed()`.

I do not believe the current changes implement this faithfully, and I believe the semantics have subtly changed.

- I would like us to use the functions suggested by the compiler, see inline comments
- I would like to remove the change to `NoRenviron`, as it does not show up in the warnings and does not have a compiler suggested equivalent

crates/libr/src/sys/windows/types.rs
```diff
@@ -208,12 +208,12 @@ impl structRstart {
     }
     #[inline]
     pub fn RstartVersion(&self) -> ::std::os::raw::c_int {
-        unsafe { ::std::mem::transmute(self._bitfield_1.get(16usize, 16u8) as u32) }
+        unsafe { u32::cast_signed(self._bitfield_1.get(16usize, 16u8) as u32) }
```
DavisVaughan: Are the sections all still unsafe?

DavisVaughan: If you don't get warnings about "remove this not needed unsafe" then I'm fine either way with what is here

crates/libr/src/sys/windows/types.rs
lionel-: (An `Rboolean` is documented as an enum with `i32` representation.)

crates/libr/src/sys/windows/types.rs
lionel-: Just for technical clarity, `Rstartversion` and `Rnoenviron` are meant to be small numbers, so the intricacies of integer casting do not really apply here.

lionel-: Kept thinking about this. First of all I agree with Davis' review that preserving current behaviour is the right approach so that we don't even have to discuss what is safe. I'm just geeking here.

My comment above about small numbers was confused. Casting intricacies also matter with small numbers in the signed widening case, because sign extension will fill the high bits with 1s. If the setter reinterprets to unsigned first, then casts, Rust will 0-extend instead of sign-extending, changing the meaning of the number. (What we do in the getter doesn't matter as both approaches lead to simple truncation and sign reinterpretation)

In our case the extended bits (in range 32-64) don't have any impact because we're storing in 16 bits, and the setter is explicitly discarding the high bits. But what if we were storing in more than 32 bits (i.e. more than our input type)? In that case, I think Jenny's approach is actually safer for round-tripping. Sign-extending the number before storing it in an unsigned type is exactly what we need to preserve the meaning of the signed integer when the getter truncates the input back to 32 bits.

So direct casting seems to be preferable as a matter of principle (_if_ my reasoning is correct).

jennybc: Let's be clear that "Jenny's approach" was mostly "Copilot's approach" 😄 

For the record, what I ended up doing (and merged) is what the Rust compiler is suggesting, so that seems like a good place to start in any case.

My goal was to get rid of the nuisance compiler warnings on Windows, which has been a success.

PR #926, opened 2025-09-24
Rationalize pandoc path building, especially on Windows

jennybc:
Addresses https://github.com/posit-dev/positron/issues/9574

I can successfully invoke the R code that backs the help-reading tool in the R Console (e.g. via `.ps.rpc.get_help_page("trim", "glue")`), but when I can convince Positron Assistant to call the tool, it just hangs indefinitely. I'm using a release build of Positron with a locally built ark (from this PR). I'm pushing so that @sharon-wang can brainstorm with me tomorrow.

### DIFF
--- a/crates/ark/src/modules/positron/pandoc.R
+++ b/crates/ark/src/modules/positron/pandoc.R
@@ -147,10 +147,11 @@ quoted <- function(args) {
 
 #' Get the path to the pandoc executable
 pandoc <- function() {
-    file.path(
-        find_pandoc()$dir,
-        "pandoc"
-    )
+    pandoc_info <- find_pandoc()
+    if (is.null(pandoc_info$dir)) {
+        return(NULL)
+    }
+    build_pandoc_path(pandoc_info$dir)
 }
 
 #' Test if a directory exists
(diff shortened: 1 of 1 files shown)


crates/ark/src/modules/positron/pandoc.R
```diff
@@ -197,6 +198,15 @@ find_pandoc <- function(dir = NULL, version = NULL) {
     )
 }
 
+# Build the full path to the pandoc executable, handling Windows .exe suffix
+build_pandoc_path <- function(pandoc_dir) {
+    path <- file.path(pandoc_dir, "pandoc")
+    if (identical(.Platform$OS.type, "windows")) {
+        path <- paste0(path, ".exe")
```
jennybc, line 205: This is the really important bit that explains why the tool previously could not access R help. But instead of fixing it in `pandoc()` I also rationalized things a bit, since the need to append `.exe` also affects `get_pandoc_version()`.

jennybc:
@sharon-wang kicked off a Windows release build incorporating ark from this branch and I can confirm that the "get package help tool" is successfully called on Windows. I can even use it with a local ark build successfully. Unclear why I could not test drive this PR with one of our dailies 🤔 

Screenshot 2025-09-25 133656

sharon-wang (approved):
Works great for me on Windows!

I was able to test in a dev build of positron + dev ark with this PR branch, but I needed to manually install Quarto and set `RSTUDIO_PANDOC` due to https://github.com/posit-dev/positron/issues/8449.

I was also able to test with the Positron `2025.10.0-131` daily, subbing in my local ark build from this branch. In the Settings UI, I had to set `positron.r.kernel.path` to `C:\Users\sharon\ark\target\debug\ark.exe` to point to the local ark.

image

sharon-wang:
I learned in this process that you can do:

https://github.com/posit-dev/positron/commit/11342d3c60988fde91d4ec58ab2904ea69c11da9

which is documented here: https://github.com/posit-dev/positron/blob/main/extensions/positron-r/scripts/README.md#ci-development-mode

To get a CI build going with a particular Ark ref!

PR #928, opened 2025-09-25
Disable proxies when connecting to local help server

jmcphers:
Ark's built-in help proxy doesn't work correctly when proxies are present, since the `reqwest` client, by default, uses the configured proxies. Of course, these proxies can't call back into localhost, so hilarity ensues.

Since we're always connecting to `localhost`, we don't ever want to use a proxy here; this change updates the reqwest client so that it doesn't use proxies.

Part of https://github.com/posit-dev/positron/issues/8426.

### DIFF
--- a/crates/ark/src/help_proxy.rs
+++ b/crates/ark/src/help_proxy.rs
@@ -148,8 +148,18 @@ async fn proxy_request(req: HttpRequest, app_state: web::Data<AppState>) -> Http
     };
 
     // Set up reqwest client with 3 retries (posit-dev/positron#3753).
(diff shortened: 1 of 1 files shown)


juliasilge (approved):
This is working well for me. When I make a request for help, in the kernel logs I see all the expected messages similar to what I outline at https://github.com/posit-dev/positron/issues/3543#issuecomment-2524213456.

Speaking of which, do we think this change will resolve the problems outlined in https://github.com/posit-dev/positron/issues/3543? They mostly seem to be about proxies, or otherwise a mysterious _something_ on the machine blocking the request without an error message.

PR #932, opened 2025-09-26
Data Explorer: Fix histogram edge cases for 0- and 1-row data frames

wesm:
Histograms were failing for data frames with 1 row. This patch fixes that and adds test cases for 0- and 1-row data frame cases.


dfalbel (approved):
LGTM!

PR #933, opened 2025-09-29
Convert to code: the first move, which is dplyr

jennybc:
Relates to https://github.com/posit-dev/positron/issues/9030

This was written with substantial use of Claude Code, so consider it a hybrid of my high-level guidance + lots of Claude implementation choices, when reviewing. In particular, I've done the least hands-on refining of tests.

This can be test driven with a release build of Positron, but using `positron.r.kernel.path` to point at a local ark executable.

Screenshot 2025-09-29 001558

### DIFF
--- a/crates/ark/src/data_explorer/convert_to_code.rs
+++ b/crates/ark/src/data_explorer/convert_to_code.rs
@@ -958,6 +958,18 @@ mod tests {
 
         // Empty string
         assert_eq!(escape_character_constant(""), r#""""#);
+
+        // String with single quote - not escaped, R accepts it in double-quoted strings
+        assert_eq!(
+            escape_character_constant("it's"),
+            r#""it's""#
+        );
+
+        // String with backtick - not escaped, R accepts it in double-quoted strings
+        assert_eq!(
+            escape_character_constant("code: `x`"),
+            r#""code: `x`""#
+        );
     }
 
     #[test]


crates/ark/src/data_explorer/convert_to_code.rs
```diff
@@ -0,0 +1,698 @@
...
+use amalthea::comm::data_explorer_comm::ConvertedCode;
+use amalthea::comm::data_explorer_comm::FilterComparisonOp;
+use amalthea::comm::data_explorer_comm::RowFilter;
+use amalthea::comm::data_explorer_comm::RowFilterParams;
+use amalthea::comm::data_explorer_comm::RowFilterType;
+use amalthea::comm::data_explorer_comm::TextSearchType;
+
+/// Sort key with resolved column name
+#[derive(Clone, Debug)]
+pub struct ResolvedSortKey {
```
jennybc, line 20: I am beginning to think there might be some room for improvement in the API we've given ourselves for convert to code. But as it stands, the columns to sort by are available by their position, not name, hence the need for this.

crates/ark/src/data_explorer/convert_to_code.rs
```diff
@@ -0,0 +1,698 @@
...
+        RowFilterType::NotNull => Some(format!("!is.na({})", column_name)),
+        RowFilterType::IsTrue => Some(format!("{} == TRUE", column_name)),
+        RowFilterType::IsFalse => Some(format!("{} == FALSE", column_name)),
+        RowFilterType::IsEmpty => Some(format!("{} == \"\"", column_name)),
+        RowFilterType::NotEmpty => Some(format!("{} != \"\"", column_name)),
+        RowFilterType::Search => {
+            if let Some(RowFilterParams::TextSearch(search)) = &filter.params {
+                match search.search_type {
+                    TextSearchType::Contains => Some(format!(
+                        "grepl({}, {}, fixed = TRUE)",
```
jennybc, line 305: It feels tempting to use stringr in these text search types, but then this becomes full-on tidyverse syntax vs. just dplyr. So for now, I'm taking the narrower route and sticking with base R inside `filter()`.

crates/ark/src/data_explorer/convert_to_code.rs
```diff
@@ -0,0 +1,854 @@
...
+
+            let full_code = code_lines.join("\n");
+            harp::parse_eval_global(&full_code)?;
+
+            Ok(())
+        })
+    }
+
+    #[test]
+    fn test_convert_to_code_execution_basic_filter() {
```
jennybc, line 1043: I don't think this test is terribly interesting as it is. I consider more of an MVP for a type of test where we actually execute the generated code and assert things about the result. I could imagine this being a useful tool as we build this feature out. Similar tests are used on the Python side.

crates/ark/src/data_explorer/convert_to_code.rs
```diff
@@ -0,0 +1,920 @@
...
+            // Check that result has 2 rows (Bob: 30, Charlie: 35)
+            let nrows =
+                DataFrame::n_row(harp::parse_eval_global("filtered_people").unwrap().sexp).unwrap();
+            assert_eq!(nrows, 2);
+
+            // Check that the filtered data contains exactly Bob and Charlie
+            let names_check =
+                harp::parse_eval_global("setequal(filtered_people$name, c('Bob', 'Charlie'))")
+                    .unwrap();
+            assert_eq!(harp::r_lgl_get(names_check.sexp, 0), 1);
```
jennybc, line 1123: Am I doing this right? I struggled somewhat to figure out the right way make assertions about the result of assigning the generated code (to `filtered_people` in this case).

PR #935, opened 2025-10-01
Data Explorer: Incorporate specific number types into Ark

wesm:
This is the companion Ark PR for upstream https://github.com/posit-dev/positron/pull/9715 which adds specific Integer, Floating display types.


crates/ark/src/data_explorer/utils.rs
```diff
@@ -67,16 +67,16 @@ pub fn display_type(x: SEXP) -> ColumnDisplayType {
...
-            return ColumnDisplayType::Number;
+            return ColumnDisplayType::Integer;
         }
         if r_inherits(x, "double") {
-            return ColumnDisplayType::Number;
+            return ColumnDisplayType::Floating;
         }
         if r_inherits(x, "complex") {
-            return ColumnDisplayType::Number;
+            return ColumnDisplayType::Floating;
```
jennybc, line 76: This seems pretty weird, but I honestly don't foresee lots of use of complex numbers in the data explorer. But perhaps they should stay as `ColumnDisplayType::Number`.

crates/ark/src/data_explorer/utils.rs
```diff
@@ -107,7 +107,9 @@ pub fn display_type(x: SEXP) -> ColumnDisplayType {
 
     match r_typeof(x) {
         LGLSXP => return ColumnDisplayType::Boolean,
-        INTSXP | REALSXP | CPLXSXP => return ColumnDisplayType::Number,
+        INTSXP => return ColumnDisplayType::Integer,
+        REALSXP => return ColumnDisplayType::Floating,
+        CPLXSXP => return ColumnDisplayType::Floating,
```
jennybc, line 112: Would also need to change back to `Number` if we do so above.

PR #937, opened 2025-10-02
Use vendor-agnostic AGENTS.md file

lionel-:
- Update notes about running tests. It's always better to use `just test`

- Merge `.github/copilot-instructions.md` into `AGENTS.md`.

- Rename `CLAUDE.md` to `AGENTS.md` from https://agents.md/

- Symlink copilot and claude files to `AGENTS.md`


juliasilge (approved):

AGENTS.md
```diff
@@ -0,0 +1,106 @@
+# CLAUDE.md
```
jennybc: Perhaps we should agnosticize this? Leftover from being `CLAUDE.md`.

lionel-: oops yes!

PR #940, opened 2025-10-06
Add CI tests and releases for Windows arm64

lionel-:
Closes https://github.com/posit-dev/positron/issues/2934
See https://github.com/posit-dev/positron/issues/2934#issuecomment-3371285146 for details.

- Release build tested at https://github.com/posit-dev/ark/actions/runs/18280487687/job/52042952373

- I've added Windows arm64 to the matrix of CI tests. Since things are still in flux on that front, it seems safer to explicitly monitor reliability on this platform.

### DIFF
--- a/.github/workflows/release-windows.yml
+++ b/.github/workflows/release-windows.yml
@@ -13,7 +13,7 @@ on:
 jobs:
     build_windows:
         name: Build Windows
-        runs-on: ${{ matrix.arch == 'arm64' && 'windows-11-arm' || 'windows-latest' }}
+        runs-on: ${{ matrix.runs_on }}
         timeout-minutes: 40
 
         env:
@@ -26,8 +26,10 @@ jobs:
                 flavor: [debug, release]
                 include:
                     - arch: x64
+                      runs_on: windows-latest
                       rust_target_prefix: x86_64
                     - arch: arm64
+                      runs_on: windows-11-arm
                       rust_target_prefix: aarch64
 
         steps:
--- a/.github/workflows/release-windows.yml
+++ b/.github/workflows/release-windows.yml
@@ -13,7 +13,7 @@ on:
 jobs:
     build_windows:
         name: Build Windows
-        runs-on: windows-latest
+        runs-on: ${{ matrix.arch == 'arm64' && 'windows-11-arm' || 'windows-latest' }}
         timeout-minutes: 40
 
         env:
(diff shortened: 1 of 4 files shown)


.github/workflows/release-windows.yml
```diff
@@ -22,11 +22,13 @@ jobs:
...
         strategy:
             matrix:
-                arch: [x64]
+                arch: [x64, arm64]
                 flavor: [debug, release]
                 include:
                     - arch: x64
                       rust_target_prefix: x86_64
+                    - arch: arm64
+                      rust_target_prefix: aarch64
```
DavisVaughan, line 33:
```suggestion
                    - arch: x64
                      runs_on: windows-latest
                      rust_target_prefix: x86_64
                    - arch: arm64
                      runs_on: windows-11-arm
                      rust_target_prefix: aarch64
```

Maybe this? Then use `${{ matrix.runs_on }}` above?

.github/workflows/test-windows.yml
```diff
@@ -6,11 +6,12 @@ on:
 
 jobs:
   windows:
-    runs-on: windows-latest
-    name: "Rust: ${{ matrix.config.rust }}, R: ${{ matrix.config.r }}"
+    runs-on: ${{ matrix.arch == 'arm64' && 'windows-11-arm' || 'windows-latest' }}
```
DavisVaughan: Same note about an `include:` with `runs_on:` here

crates/harp/src/sys/windows/library.rs
```diff
@@ -90,5 +90,13 @@ pub fn open_r_shared_library(path: &PathBuf) -> Result<libloading::Library, libl
...
-    path.join("bin").join("x64")
+    #[cfg(target_arch = "aarch64")]
+    {
+        // arm64 has a flatter structure
+        path.join("bin")
+    }
+    #[cfg(not(target_arch = "aarch64"))]
+    {
+        path.join("bin").join("x64")
+    }
```
DavisVaughan, line 101:
```suggestion
    cfg_if::cfg_if! {
        if #[cfg(target_arch = "aarch64")] {
            // arm64 has a flatter structure
            path.join("bin")
        } else {
            path.join("bin").join("x64")
        }
    }
```

you may appreciate this

DavisVaughan (approved):
Awesome, looks like we had prepped pretty well for this 👍

crates/harp/src/sys/windows/library.rs
lionel-: oh!

lionel-: hmm... I don't know which I prefer 😅

DavisVaughan: Oh really? I like cfg if so much more

lionel-: It's probably nicer in many cases. In that case I like the simplicity of a non-macro approach.

.github/workflows/release-windows.yml
lionel-: That's better, thanks!

PR #941, opened 2025-10-07
Continue listening when `poll()` is interrupted

lionel-:
The `poll()` syscall might be interrupted by a SIGCHLD (a `system()` call) or a SIGINT. It is safe to retry in that case.

### DIFF
--- a/crates/amalthea/src/sys/unix/stream_capture.rs
+++ b/crates/amalthea/src/sys/unix/stream_capture.rs
@@ -47,19 +47,28 @@ impl StreamCapture {
         let stderr_poll = nix::poll::PollFd::new(stderr_fd, nix::poll::PollFlags::POLLIN);
         let mut poll_fds = [stdout_poll, stderr_poll];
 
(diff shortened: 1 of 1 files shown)


crates/amalthea/src/sys/unix/stream_capture.rs
```diff
@@ -47,19 +47,28 @@ impl StreamCapture {
...
+                Err(err) => {
+                    // https://pubs.opengroup.org/onlinepubs/9699919799/functions/poll.html
+                    match err {
+                        // If the poll was interrupted, silently continue
+                        nix::errno::Errno::EINTR => continue,
+
+                        // Internal allocation has failed, but a retry might succeed
+                        nix::errno::Errno::EAGAIN => continue,
+
+                        _ => {
```
lionel-, line 67: The remaining error documented in https://pubs.opengroup.org/onlinepubs/9699919799/functions/poll.html represents an unexpected state, so quitting the thread is likely appropriate.

jmcphers (approved):
LGTM!

PR #943, opened 2025-10-13
Fix Reload Workspace problem in Windows

kv9898:
This PR addresses https://github.com/posit-dev/positron/issues/9927 for Windows. Since this problem doesn't occur in my WSL, I will only fix it for Windows.


crates/ark/src/sys/windows/interface.rs
```diff
@@ -67,6 +67,18 @@ pub fn setup_r(mut _args: Vec<*mut c_char>) {
         // R_DefParamsEx(params, bindings::RSTART_VERSION as i32);
         R_DefParamsEx(params, 0);
 
+        // Check if --no-restore-date is in args
```
lionel-:
```suggestion
        // Check if --no-restore-data is in args
```

PR #944, opened 2025-10-13
Add handler to allow for autosave

kv9898:
This PR goes in tandem with https://github.com/posit-dev/positron/pull/9936 to allow auto-saving the workspace to `.RData` when the R session is shut down. The reason we need this is that in Windows Ark is having problems with propagating arguments to the R binary: https://github.com/posit-dev/positron/issues/9927.

### DIFF
--- a/crates/ark/src/sys/windows/interface.rs
+++ b/crates/ark/src/sys/windows/interface.rs
@@ -75,11 +75,17 @@ pub fn setup_r(mut args: Vec<*mut c_char>) {
             })
             .collect();
 
-        if args_str.contains(&"--save".to_string()) {
-            (*params).SaveAction = libr::SA_TYPE_SA_SAVE;
-        } else {
-            (*params).SaveAction = libr::SA_TYPE_SA_NOSAVE;
+        // Determine save action by scanning args in order and letting the last
+        // occurrence of --save or --no-save win (mirrors R's last-one-wins behavior).
+        let mut save_action = libr::SA_TYPE_SA_NOSAVE;
+        for arg in &args_str {
+            match arg.as_str() {
+                "--save" => save_action = libr::SA_TYPE_SA_SAVE,
+                "--no-save" => save_action = libr::SA_TYPE_SA_NOSAVE,
+                _ => {},
+            }
         }
+        (*params).SaveAction = save_action;
 
         (*params).R_Interactive = 1;
         (*params).CharacterMode = libr::UImode_RGui;


crates/ark/src/sys/windows/interface.rs
```diff
@@ -67,6 +67,20 @@ pub fn setup_r(mut _args: Vec<*mut c_char>) {
...
+                let cstr = CStr::from_ptr(arg);
+                cstr.to_string_lossy().to_string()
+            })
+            .collect();
+
+        if args_str.contains(&"--save".to_string()) {
+            (*params).SaveAction = libr::SA_TYPE_SA_SAVE;
+        } else {
+            (*params).SaveAction = libr::SA_TYPE_SA_NOSAVE;
+        }
```
lionel-, line 87: (minor) It's fine to convert the C strings to Rust string to avoid `unsafe`, but can you please use `.any()` instead of `.map()` to create a bool instead of a vector? This is to avoid creating a vector of args that we won't use afterwards, and walking the list of arguments twice.

lionel-: Ah now I see you'll reuse the vector for https://github.com/posit-dev/ark/pull/943/files, nevermind!

kv9898:
> @DavisVaughan Another way to set these options would be to call `cmdlineoptions()` from https://github.com/r-devel/r-svn/blob/48256015cc3c3e7cc2235179adccc60a641f6603/src/gnuwin32/system.c#L1098 (it calls the `R_common_command_line()` function discussed above).
> 
> This is what Rgui does (https://github.com/r-devel/r-svn/blob/48256015cc3c3e7cc2235179adccc60a641f6603/src/gnuwin32/front-ends/rgui.c#L60), as well as RStudio: https://github.com/rstudio/rstudio/blob/1062bea412a4410c0761ba65ca7ba65f019ee9b8/src/cpp/r/session/REmbeddedWin32.cpp#L235-L242.
> 
> The RStudio call site also has a note regarding initialisation of translation domains. It seems like we should look into calling that routine to initialise CLI arguments and other stuff, what do you think? It does quite a lot, so maybe we could pair to carefully review it and discuss implications.

The RStudio code seems to set the options separate from cmdlinoptions, so more similar to my proposal in this PR?

crates/ark/src/sys/windows/interface.rs
```diff
@@ -67,6 +67,20 @@ pub fn setup_r(mut _args: Vec<*mut c_char>) {
...
 
+        let args_str: Vec<String> = args
+            .iter()
+            .map(|&arg| {
+                let cstr = CStr::from_ptr(arg);
+                cstr.to_string_lossy().to_string()
+            })
+            .collect();
+
+        if args_str.contains(&"--save".to_string()) {
```
lionel-: Should also handle `--no-save`.

I think that suggests you need to loop over the arguments and handle them one by one (last one wins) as in https://github.com/r-devel/r-svn/blob/48256015cc3c3e7cc2235179adccc60a641f6603/src/main/CommandLineArgs.c#L102.

(That is, if we're not going to rely on `cmdlineoptions()` to set these)

PR #946, opened 2025-10-24
Set macOS 11 deployment target

kv9898:
Addresses https://github.com/posit-dev/positron/issues/10192

### DIFF
--- a/.github/env/macos.env
+++ b/.github/env/macos.env
@@ -1,4 +1,10 @@
 # Update this to reflect our last supported version of macOS
 # (same as Positron's). This is used both for testing and
 # building releases.
+#
+# Note that even when Rust's minimum supported version is our
+# target version, we still need to set it explicitly because
+# there are parts of the build, such as tree-sitter or 0MQ,
+# that are not governed by Rust's minimum supported version.
+# https://github.com/posit-dev/positron/issues/10192#issuecomment-3455373010
 MACOSX_DEPLOYMENT_TARGET=11


lionel-:
Relevant comment from July 2024: https://github.com/posit-dev/positron/issues/3812#issuecomment-2205387863
At the time 10.15 was the target version of Positron.

I wonder if we should have a process to align the required versions of Ark with that of Positron, in our release build CI.

.github/workflows/release-macos.yml
```diff
@@ -45,6 +45,11 @@ jobs:
                 rustup update --no-self-update stable
                 rustup default stable
 
+            - name: Set macOS 11 deployment target
+              run: |
+                echo "MACOSX_DEPLOYMENT_TARGET=11" >> $GITHUB_ENV
+                echo "RUSTFLAGS=-C link-args=-mmacosx-version-min=11" >> $GITHUB_ENV
```
lionel-: According to https://doc.rust-lang.org/rustc/platform-support/apple-darwin.html? this line is redundant, as rustc obeys the `MACOSX_DEPLOYMENT_TARGET` flag


```suggestion
```

.github/env/macos.env
```diff
@@ -0,0 +1,4 @@
+# Update this to reflect our last supported version of macOS
+# (same as Positron's). This is used both for testing and
+# building releases.
+MACOSX_DEPLOYMENT_TARGET=11
```
DavisVaughan, line 10: Would you mind saying more about the "why" of why we need to set this at all?

Something about "Even though Rust's minimum supported macOS target is already 11.0, we must set this so zeromq is also built against 11.0. Without this, `otool` revealed that zeromq was being built with a different (newer) macOS target"

And maybe link to https://github.com/rust-lang/rust/blob/master/src/doc/rustc/src/platform-support/apple-darwin.md to see the Rust macOS target minimum version for future us.

DavisVaughan: Maybe even link directly to this comment or at least the issue number https://github.com/posit-dev/positron/issues/10192#issuecomment-3455373010

.github/workflows/test-macos.yml
```diff
@@ -6,19 +6,23 @@ on:
...
-          - { rust: 'stable', r: 'release' }
+          - { rust: 'nightly', r: 'release', runs_on: 'macos-latest' }
+          - { rust: 'stable', r: '4.2', runs_on: 'macos-14' }
     timeout-minutes: 30
     env:
       GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     steps:
       - uses: actions/checkout@v4
 
+      - name: Set macOS deployment target
```
DavisVaughan:
```suggestion
      - name: Configure environment
```

Just to be consistent with the other one

.github/env/macos.env
lionel-: > Something about "Even though Rust's minimum supported macOS target is already 11.0, we must set this so zeromq is also built against 11.

We're not necessarily tracking Rust's minimum supported OS though

DavisVaughan: I mainly want to recognize that this also affects zeromq in a very unexpected way

Issue #670, opened 2025-01-15
Set `R_set_command_line_arguments()` to reflect what actually happens


DavisVaughan:
Some packages like `startup::startup()` utilize `commandArgs()` to make decisions. For example, that package checks to see if `--no-site-file` is in the list of command args. If it is, then it refuses to try and load the "extended" site profiles that that package specially supports. Same with `--no-init-file`.

The problem is that we "manually" run both `.Rprofile.site` and `.Rprofile` for the user once ark has set up enough to be able to handle running arbitrary scripts. We have to forcibly tell R _not_ to run these by setting `--no-site-file` and `--no-init-file`, even if we eventually run them. This confuses the startup package.

We should probably retain a copy of `args` that reflect the "real" state of the world and call `R_set_command_line_arguments()` after calling `Rf_initialize_R()` with the "fake" `args`. This is what `commandArgs()` ends up pulling from.

A test plan for this would be an integration test that checks `commandArgs()` after startup with a test client that allows both the user and site level R profiles to run. The command args should not contain `--no-site-file` nor `--no-init-file`

PR #947, opened 2025-10-27
Use `R_common_command_line()` to process args on Windows

DavisVaughan:
Joint with @lionel- 

Closes https://github.com/posit-dev/ark/pull/944
Closes https://github.com/posit-dev/ark/pull/943

This takes a different approach than these PRs.

Addresses https://github.com/posit-dev/positron/issues/9927
Related to https://github.com/posit-dev/positron/pull/9936
Related to https://github.com/posit-dev/ark/issues/708

---

On Unix:
- We call `Rf_initialize_R()` and hand it all of the command line `args`
- We then tweak some one off options after that

On Windows:
- We call `cmdlineoptions()` but only gave it `R.exe`
- We then do a wholesale replacement of options with `R_SetParams()`
- *At no point along the way did we consider the user's command line `args`*

We can't really give `cmdlineoptions()` the `args`, because our call to `R_SetParams()` is just going to overwrite anything that is set in `cmdlineoptions()`. Instead, we need to adjust our set of `params` to account for any user supplied `args`. The easiest way to do this seems to be to call `R_common_command_line()`, which updates `params` in place, along with globals like `R_RestoreHistory`.

Notably `R_common_command_line()` is what both `Rf_initialize_R()` (Unix) and `cmdlineoptions()` (Windows) already were calling, so this is really just getting us closer to existing Unix behavior by passing them all through like this.

An easy way to test that we really were not passing any `args` through was to note that the UI option that sets `--quiet` wasn't working, but it does now!

Screenshot 2025-10-27 at 2 56 32 PM

Ark on Windows now supports the `R: Restore Workspace` Positron option because Ark now respects the `--no-restore-data` option passed in by Positron. But more importantly, Ark now _holistically_ supports the full family of `--restore`, `--no-restore`, `--no-restore-data`, and `--vanilla`, and uses `R_common_command_line()` to recreate R's "last wins" behavior when multiple are supplied.

Here's an image showing that the `.Rdata` is _not_ restored on restart

Screenshot 2025-10-27 at 2 58 12 PM

As @lionel- reminded me, this PR is orthogonal to what we should _expose_ on the Positron side for setting these options. I think we should be very careful about exposing UI that leads to good practices (i.e. do you ever really want `--save` without `--restore`?)


crates/ark/src/sys/windows/interface.rs
```diff
@@ -47,18 +48,20 @@ pub fn setup_r(mut _args: Vec<*mut c_char>) {
...
-        // note that R does a lot of initialization here that's not accessible
+        // Note that R does a lot of initialization here that's not accessible
         // in any other way; e.g. the default translation domain is set within
-        //
+        // via `BindDomain()`. We don't supply the `args` here because we do a
+        // wholesale replacement of the options they set via `R_SetParams()`
+        // later on, so setting them here would have no effect anyways.
         // https://github.com/rstudio/rstudio/issues/10308
-        let rargc: i32 = 1;
-        let mut rargv: Vec<*mut c_char> = cargs!["R.exe"];
```
DavisVaughan, line 56: Notably on the Unix side the `commandArgs()` that shows up on the R side says `ark`, not `R`.

To be consistent, we now also say `ark.exe`, not `R.exe`, on Windows.

Screenshot 2025-10-27 at 3 00 14 PM

Screenshot 2025-10-27 at 2 59 57 PM

There are still some issues here. As you can see we aren't getting the `args` passed through in `commandArgs()`. This is due to `cmdlineoptions()` being the one to call `R_set_command_line_arguments()`, but we just gave it an empty set of args. This is related to https://github.com/posit-dev/ark/issues/670 and is more appropriate to tackle alongside that.

crates/ark/src/sys/windows/interface.rs
```diff
@@ -67,6 +70,29 @@ pub fn setup_r(mut _args: Vec<*mut c_char>) {
...
+        // - `(*params).SaveAction` via `--save`, `--no-save`, `--vanilla`, `--no-echo`
+        // - `(*params).RestoreAction` via `--restore`, `--no-restore`, `--no-restore-data`, `--vanilla`
+        // - `R_RestoreHistory` (a global) via `--restore`, `--no-restore`, `--no-restore-history`, `--vanilla`
+        let mut c_args = RMain::build_ark_c_args(args);
+        let mut c_args_len = c_args.len() as std::ffi::c_int;
+        R_common_command_line(
+            &mut c_args_len,
+            c_args.as_mut_ptr() as *mut *mut c_char,
+            params,
+        );
```
DavisVaughan, line 94: We were not respecting any of these on the Windows side. We now respect all of them in a way that should exactly mimic what happens on the Unix side, because we call the same `R_common_command_line()` utility on both sides.

DavisVaughan: Remember that `--no-site-file` and `--no-init-file` (for `Rprofile` handling) are handled elsewhere.

For the purposes of `setup_r()`, the `args` _always_ contain both of these flags, and Ark itself is in charge of whether or not to load these, and at what time to do so. We don't rely on `R_common_command_line()` to set the options for those, we just hardcode them further below

```
        // Never load the user or site `.Rprofile`s during `setup_Rmainloop()`.
        // We do it for the user once ark is ready. We faithfully reimplement
        // R's behavior for finding these files in `startup.rs`.
        (*params).LoadInitFile = Rboolean_FALSE;
        (*params).LoadSiteFile = Rboolean_FALSE;
```

crates/ark/src/interface.rs
```diff
@@ -517,6 +509,29 @@ impl RMain {
...
+                let mut out = cargs!["ark"];
+            } else if #[cfg(windows)] {
+                let mut out = cargs!["ark.exe"];
+            } else {
+                unreachable!("Unsupported OS");
+            }
+        }
+
+        // The `into_raw()` call leaks each individual `arg` because we never call
+        // `CString::from_raw()`, but we are fine with that
```
lionel-: Is there an expectation on the R side that the array is static? If so it would be nice to document, otherwise we could just free the vector after usage?

DavisVaughan: Turns out that we can't actually free these at all

I tried, but R's initialization routines like `Rf_initialize_R()` actually modifies the array of C strings directly, invalidating our pointers, and I got a ton of segfaults in the tests when I tried to free them afterwards.

I've added some comments

PR #949, opened 2025-10-28
Recover with base graphics if ragg or svglite fail to load

### LINKED ISSUE #917: Ark tries to use ragg even if it is broken, leading to inscrutable errors
If `ragg` is installed but unusable, then attempting to plot anything will result in an unhelpful error, `no active device and default getOption("device") is invalid`:

Image

You can get a slightly better error by calling `.ps.graphics.create_device()` directly.

Image

There is no reasonable way for an end user to know to do this, and the fact that `ragg` is not working should not leave the user in a state where they can't plot at all. We should do one or more of the following:

- Test to see whether we can actually load `ragg` (in addition to checking its version) before enabling ragg graphics.
- If attempting to create the graphics device with `ragg` fails, fall back on the default (non-`ragg`) approach.

DavisVaughan:
Closes #917

### DIFF
--- a/crates/ark/src/modules.rs
+++ b/crates/ark/src/modules.rs
@@ -292,6 +292,21 @@ fn r_poke_option_ark_testing() {
     }
 }
 
+#[harp::register]
+pub extern "C-unwind" fn ark_log_trace(msg: SEXP) -> harp::error::Result<SEXP> {
+    Ok(harp::utils::harp_log_trace(msg))
+}
+
+#[harp::register]
+pub extern "C-unwind" fn ark_log_warning(msg: SEXP) -> harp::error::Result<SEXP> {
+    Ok(harp::utils::harp_log_warning(msg))
+}
+
+#[harp::register]
+pub extern "C-unwind" fn ark_log_error(msg: SEXP) -> harp::error::Result<SEXP> {
+    Ok(harp::utils::harp_log_error(msg))
+}
+
 #[cfg(test)]
 mod tests {
     use harp::environment::Environment;
--- a/crates/ark/tests/plots.rs
+++ b/crates/ark/tests/plots.rs
@@ -165,3 +165,66 @@ fn test_graphics_device_initialization() {
     frontend.recv_iopub_idle();
     assert_eq!(frontend.recv_shell_execute_reply(), input.execution_count);
 }
+
+#[test]
+fn test_ragg_is_used_by_default() {
+    let frontend = DummyArkFrontend::lock();
+
+    // We install ragg on CI and expect developers to have it locally
+    let code = ".ps.internal(use_ragg())";
+    frontend.send_execute_request(code, ExecuteRequestOptions::default());
+    frontend.recv_iopub_busy();
+    let input = frontend.recv_iopub_execute_input();
+    assert_eq!(input.code, code);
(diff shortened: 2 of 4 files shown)


.github/workflows/test-linux.yml
```diff
@@ -62,11 +62,12 @@ jobs:
           packages:
             data.table
             dplyr
-            rstudioapi
-            tibble
             haven
             R6
+            ragg
```
DavisVaughan, line 67: Just adding ragg to CI, but also sorting these

crates/ark/src/modules.rs
```diff
@@ -292,6 +292,21 @@ fn r_poke_option_ark_testing() {
     }
 }
 
+#[harp::register]
+pub extern "C-unwind" fn ark_log_trace(msg: SEXP) -> harp::error::Result<SEXP> {
+    Ok(harp::utils::harp_log_trace(msg))
+}
```
DavisVaughan, line 298: I've needed these before from the ark modules side

When you're in ark, you can't access the R level `log_trace()` and friends that exist in the R level harp modules

crates/ark/tests/plots.rs
```diff
@@ -165,3 +165,66 @@ fn test_graphics_device_initialization() {
...
+
+newLoadNamespace <- function(package, ...) {
+  if (identical(package, "ragg")) {
+    stop("Can't load ragg")
+  }
+  oldLoadNamespace(package, ...)
+}
+
+assign("loadNamespace", newLoadNamespace, envir = .BaseNamespaceEnv)
+lockBinding("loadNamespace", .BaseNamespaceEnv)
```
DavisVaughan, line 202: This took some effort to come up with. I was going to use `trace(loadNamespace, exit =)` but the way that works is that at the start of the function an `on.exit()` is set up with your `exit` function, but `loadNamespace()` itself _clears_ all exit expressions with an empty call to `on.exit()` at the end!

crates/ark/tests/plots.rs
```diff
@@ -165,3 +165,66 @@ fn test_graphics_device_initialization() {
...
+    frontend.send_execute_request(code, ExecuteRequestOptions::default());
+    frontend.recv_iopub_busy();
+    let input = frontend.recv_iopub_execute_input();
+    assert_eq!(input.code, code);
+    frontend.recv_iopub_idle();
+    assert_eq!(frontend.recv_shell_execute_reply(), input.execution_count);
+
+    // ragg is installed on CI, so our graphics code should try and use it,
+    // but should fail when loading the package and should fall back to base R graphics
+    let code = ".ps.internal(use_ragg())";
```
DavisVaughan, line 213: I don't have a good way to test that we are logging something intelligent, but anecdotally when I try this we are

Mainly I'm testing that `use_ragg()` does indeed return `FALSE`, and that we still get a plot below, which is way better than what happened before.

crates/ark/tests/plots.rs
DavisVaughan: Side note, once again the fact that each test gets its own R session means we can do crazy things like this without having to worry about cleaning up 😄

lionel-: Another option that doesn't involve mocking would be to create a libPath with a corrupted version of ragg.

Good enough though!

PR #950, opened 2025-10-28
Make protocol assertions macros

lionel-:
So that downstream assertions can expose file and line location in the test files, instead of `dummy_frontend.rs`. This way we're able to click on the reported locations to jump to the relevant test failure.


lionel-:
Before:


https://github.com/user-attachments/assets/5ac978aa-abbe-405e-8dbc-c31c11e9b86b


After:


https://github.com/user-attachments/assets/24e51f52-88ba-4160-97d1-e16b54854276

PR #951, opened 2025-10-28
Make ragg optional in tests

lionel-:
Not sure what R ark in unit tests is linking to, but it doesn't have ragg installed (the R on the path does have it, weird).

Our current convention is to opt out of the test gracefully if the package is not installed.

### DIFF
--- a/crates/ark/tests/plots.rs
+++ b/crates/ark/tests/plots.rs
@@ -170,10 +170,13 @@ fn test_graphics_device_initialization() {
 fn test_ragg_is_used_by_default() {
     let frontend = DummyArkFrontend::lock();
 
+    if !frontend.is_installed("ragg") {
+        report_skipped("test_ragg_is_used_by_default", "ragg");
+        return;
+    }
+
     // We install ragg on CI, but developers may not have it locally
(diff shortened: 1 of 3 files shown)


crates/ark/tests/plots.rs
```diff
@@ -170,8 +170,10 @@ fn test_graphics_device_initialization() {
 fn test_ragg_is_used_by_default() {
     let frontend = DummyArkFrontend::lock();
 
-    // We install ragg on CI and expect developers to have it locally
-    let code = ".ps.internal(use_ragg())";
+    // We install ragg on CI, but developers may not have it locally
+    let code = ".ps.internal(use_ragg()) ||
+        # Returns TRUE if ragg is available, FALSE otherwise
+        identical(system.file(package = 'ragg'), '')";
```
DavisVaughan: Can you please copy over this utility

```rust
    if !has_rstudioapi(&frontend) {
        report_skipped("test_get_mode");
        return;
    }
```

From `rstudioapi.rs` and make it `has_ragg()`?

Then you should use that in both this test and the `test_inability_to_load_ragg_falls_back_to_base_graphics` one to just not even run that test when ragg isn't installed.

PR #956, opened 2025-10-28
Remove error peeking in top level exec results

### LINKED ISSUE #953: Misleading error report in case of longjump
While reproducing https://github.com/posit-dev/positron/issues/9467 I got:

```
[R]   2025-10-28T15:41:53.803084Z ERROR  Panic! In file 'crates/ark/src/interface.rs' at line 2422: called `Result::unwrap()` on an `Err` value: Unexpected longjump
[R] Likely caused by:
[R] object 'writeConsole' not found
```

"Likely caused by" message is totally wrong. It's showing an older error message from earlier.

This is a remnant of our top-level-exec wrapper:

https://github.com/posit-dev/ark/blob/5e1d1adc78d9fe2979ce3a599a26e40b5d6b4124/crates/harp/src/exec.rs#L352-L355

It used to be that we didn't catch R errors, so if a longjump was detected it was likely caused by an error, and we would peek into R's buffer to report it to callers. Now that we do catch them, the only errors that could go through there are:

- Stack overflow errors that can't be caught by our calling condition handler
- `stop()` on a condition that doesn't inherit from error

Now that I think of it, we could make both of these more robust by inserting an `"abort"` restart. Probably not worth it though.

In any case we should now remove that "Likely caused by" addendum since it's now more likely to be misleading.

lionel-:
Closes #953

### DIFF
--- a/crates/harp/src/exec.rs
+++ b/crates/harp/src/exec.rs
@@ -11,6 +11,8 @@ use std::os::raw::c_void;
 
 use anyhow::anyhow;
 use libr::*;
+use once_cell::sync::Lazy;
+use regex::Regex;
 
 use crate::call::RCall;
 use crate::environment::R_ENVS;
@@ -21,6 +23,9 @@ use crate::object::r_null_or_try_into;
 use crate::object::RObject;
 use crate::r_symbol;
 use crate::utils::r_stringify;
+
+pub static RE_STACK_OVERFLOW: Lazy<Regex> =
+    Lazy::new(|| Regex::new(r"C stack usage [ 0-9]+ is too close to the limit\n").unwrap());
 
 pub struct RFunction {
     pub call: RCall,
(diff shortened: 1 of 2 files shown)


crates/harp/src/exec.rs
```diff
@@ -348,19 +348,11 @@ where
...
-                backtrace: std::backtrace::Backtrace::force_capture(),
-                span_trace: tracing_error::SpanTrace::capture(),
-            })
-        },
+        None => Err(Error::TopLevelExecError {
+            message: String::from("Unexpected longjump"),
+            backtrace: std::backtrace::Backtrace::force_capture(),
+            span_trace: tracing_error::SpanTrace::capture(),
+        }),
     }
```
DavisVaughan, line 372: There is a doc comment above that needs updating

```
/// If a longjump does occur for any reason (including but not limited to R
/// errors), the caller is notified, in this case by an `Err` return value
/// of kind `TopLevelExecError`. The error message contains the contents of
/// the C-level error buffer. It might or might not be related to the cause
/// of the longjump. The error also carries a Rust backtrace.
```

I'm not sure if it is worth mentioning something about stack overflows in that doc comment. Don't care either way.

PR #958, opened 2025-11-01
Add shim for `rstudioapi::readRStudioPreference()`

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/10325

With this change, you can do something like `rstudioapi::readRStudioPreference("num_spaces_for_tab", 4)`.

### DIFF
--- a/crates/ark/src/modules/rstudio/prefs.R
+++ b/crates/ark/src/modules/rstudio/prefs.R
@@ -1,4 +1,4 @@
 #' @export
-.rs.api.readRStudioPreference <- function(name, default) {
+.rs.api.readRStudioPreference <- function(name, default = NULL) {
     default
 }


crates/ark/src/modules/rstudio/prefs.R
```diff
@@ -0,0 +1,4 @@
+#' @export
+.rs.api.readRStudioPreference <- function(name, default) {
```
jennybc: I suspect this should be `default = NULL`, just to increase the quality of downstream error messages (?). I.e. someone would hear about a bad value vs. a missing argument with no default.

https://github.com/rstudio/rstudio/blob/423565d016e4d890fb13aeeaf53f86c62e97d6d3/src/cpp/r/R/Api.R#L563-L566

Issue #722, opened 2025-02-25
Weird error with `"\s"`


lionel-:
Reported by @hadley 

Evaluating `"\s"` in the console gives a weird error:

![Image](https://github.com/user-attachments/assets/85b467e6-d195-43c7-9e75-f8149ca928b3)

I traced it down to `R_ParseVector()` throwing an error instead of returning a syntax error code. This causes the control flow to end up in this safeguard path: https://github.com/posit-dev/ark/blob/452d00d689c381e51e66edf92ec62b2e13397c66/crates/ark/src/interface.rs#L1335.

lionel-:
You'll see the same thing with `1 |> {}`.

In general see https://github.com/r-devel/r-svn/blob/7da17dcbb20d1dfc901efd7dfee1623c21c925cf/src/main/gram.y#L4439 and https://github.com/r-devel/r-svn/blob/7da17dcbb20d1dfc901efd7dfee1623c21c925cf/src/main/gram.y#L4532 for the class of errors that can be thrown by the parser.

We should catch these classed R errors and convert them to proper syntax errors on the Rust side.

DavisVaughan:
Why doesn't `R_ParseVector()` return an error code? Should we report that?

lionel-:
I don't know but this seems intentional.

Issue #598, opened 2024-10-21
Sending a pair of empty backticks to the kernel results in an internal error showing up in the console


DavisVaughan:
![Image](https://github.com/user-attachments/assets/f1319dfc-f8bd-489f-b6c5-c189f7ab8b55)

DavisVaughan:
Similarly, `_;` triggers this invalid pipe placeholder parse error
https://github.com/wch/r-source/blob/988774e05497bcf2cfac47bfbec59d551432e3fb/src/main/gram.y#L1755

![Image](https://github.com/user-attachments/assets/5157c7c1-a0c6-4143-add7-f96ead280c26)

PR #960, opened 2025-11-06
Take charge of parsing and evaluating inputs

lionel-:
Progress towards https://github.com/posit-dev/positron/issues/1766
Progress towards https://github.com/posit-dev/positron/issues/3078
Progress towards https://github.com/posit-dev/positron/issues/9156
Addresses https://github.com/posit-dev/positron/issues/6553
Closes https://github.com/posit-dev/ark/issues/598
Closes https://github.com/posit-dev/ark/issues/722
Closes https://github.com/posit-dev/ark/pull/840

This PR does two things:

1. It takes charge of Reading/Parsing and Evaluation of R code. (First commit)
2. Refactors, simplifies, and fixes a bunch of issues. (All other commits)

Doing parsing and evaluation in Ark is a significant departure of how `read_console()`, the heart of the kernel, is currently structured. This handler is hooked to the R frontend method ReadConsole, which takes user inputs as _strings_. The R REPL gets these strings, parses them, evaluates them, prints the result, then loops back. With this PR, we take control of both parsing and evaluation.

There is no functional changes here except for a few bugfixes. In the future though, this will allow us to:

- Parse all inputs sent from editors with source references and injected breakpoints.
- When debugging, evaluate code in the frame selected in the IDE.

### New Features

- Selections of code are now parsed in a single pass. A minor consequence is that all functions evaluated as part of the same selection share a common virtual document when step-debugging through them.

- `recover()` is now an alias for `browser()`. The R version of `recover()` was buggy (https://github.com/posit-dev/positron/issues/9156) and it was hard to conciliate its behaviour with the requirements of the Jupyter protocol. By aliasing it to `browser()`, we can leverage the IDE's call stack pane to fulfill the same functionality as the original `recover()` function (https://github.com/posit-dev/positron/issues/3078).

### Fixes

- Minor: We no longer have any size limitation on lines of inputs. They were previously limited to 4096 bytes, the maximum size of the ReadConsole input buffer.

- Restart during debug sessions now works as expected: https://github.com/posit-dev/positron/issues/6553

- We now interrupt ongoing evaluations before restarting. This shouldn't have any impact on Positron, which sends an interrupt before restarting, but should make restarts behave better with other frontends. This is also a nice defensive measure even with Positron.

- We now flush autoprint output when an error is thrown by a print method. I was quite confused by this bug because it also swallowed debugging output!

- The set of syntax errors that throw an R error instead of returning an error code no longer cause a leaked backtrace in the console (#598, #722)

These fixes (and more) are covered by integration tests.

### Approach

- We parse inputs into a vector of complete expressions. These expressions are stored in the read-console state in a new `PendingInputs` struct that implements a stack interface, with a pop method. This stack of pending inputs replaces the stack pending lines we implemented in https://github.com/posit-dev/ark/pull/536.

- When read-console is called by R for a new input, it is evaluated and the result is stored in `base::.ark_last_value`. We store it in the base environment for robust access from (almost) any evaluation environment (global on top level but the debugger can evaluate in any environment). We only require the presence of `::` so that we can reach into base.

- Read-console returns to R with the string input `"base::.ark_last_value"` if evaluation returned visibly, or `"base::invisible(base::.ark_last_value)"` if it returned invisibly. The R REPL then parses this expression and evaluates it.

- Because we are still hooked into the R REPL, it has a chance to do top-level stuff like printing warnings, running task callbacks, updating `.Last.value`, etc.

While this approach works, I hit two main issues that we need to workaround:

- When an error happens, the `R_EvalDepth` variable is not reset before the next evaluation (it's normally reset here: https://github.com/r-devel/r-svn/blob/811080fb/src/main/main.c#L260). This could be bad especially if that error is an exceeded evaluation depth error and we're looping back to evaluate a new expression without having reset it.

- When a nested debug REPL returns to a parent REPL, the `R_ConsoleIob` parse buffer is normally reset after returning from `eval()` (here: https://github.com/r-devel/r-svn/blob/811080fb/src/main/main.c#L279). With this approach, it doesn't have a chance to reset before entering read-console again.

To solve the first issue, we evaluate a dummy input (`"base::.Last.value"` to keep the last value stable, this is tested) after an error occurs. This gives R a chance to fully reset state. To solve the second issue, we send a dummy parse causing a `PARSE_NULL` everytime we detect a nested read-console has returned. This causes R to reset the parse buffer here: https://github.com/r-devel/r-svn/blob/811080fb/src/main/main.c#L238.

These workarounds are unfortunate but have been mostly contained in the `r_read_console()` method, which keeps track of all these details and other state.

### Reorganisation

- Read-console is both an event loop (when we're waiting for input from the frontend) and a state machine (to manage execution of inputs). The event loop part is now extracted in a new method `run_event_loop()`. We no longer wait for both input replies (when we're asking the frontend to reply to a `readline()` or `menu()` question) and execution requests (when we're waiting for the frontend to send us the next bit of code to execute) at the same time, it's either one or the other, which simplifies the code. The handler for input requests now invokes the event loop itself, instead of returning to the top level of read-console to fall through the event loop.

- The state machine part of read-console has been simplified in other ways. We now inspect the states from the top level and pass it down through arguments, which makes it easier to reason about. See in particular the new `take_exception()` and `take_result()` methods called from top-level and passed down to the handler for replying to an active request when there are no inputs left to evaluate. This allowed simplifying things quite a bit.

- We now keep track of the number of nested consoles (due to active debug sessions) on the stack. Not used for anything but it seems good information to keep track of. To support this (and other state bookkeeping), we use a new wrapper around `R_ExecWithCleanup()`.

- The handling of debugger commands has been consolidated into a single location.

- A benefit of parsing expressions ourselves is that it is no longer possible for R to send us an incomplete prompt. This has been completely removed from the possible states read-console can be in, which is another nice simplification.

- The `DummyFrontend` for integration tests has now a few more methods like `execute_request()` that capture common patterns of messaging. This removes quite a lot of boilerplate.

### DIFF
--- a/crates/ark/src/interface.rs
+++ b/crates/ark/src/interface.rs
@@ -2616,28 +2616,6 @@ pub unsafe extern "C-unwind" fn r_polled_events() {
     };
 }
 
-// For integration tests
-use std::sync::Condvar;
-pub static CLEANUP_SIGNAL: (Mutex<bool>, Condvar) = (Mutex::new(false), Condvar::new());
-
-#[no_mangle]
-pub extern "C-unwind" fn r_cleanup_for_tests(_save_act: i32, _status: i32, _run_last: i32) {
-    // Signal that cleanup has started
-    let (lock, cvar) = &CLEANUP_SIGNAL;
-
-    let mut started = lock.lock().unwrap();
-    *started = true;
-
-    cvar.notify_all();
-    drop(started);
-
-    // Sleep to give tests time to complete before we panic
-    std::thread::sleep(std::time::Duration::from_secs(5));
-
-    // Fallthrough to R which will call `exit()`. Note that panicking from here
-    // would be UB, we can't panic over a C stack.
-}
-
 // This hook is called like a user onLoad hook but for every package to be
 // loaded in the session
 #[harp::register]
--- a/crates/ark/src/sys/unix/interface.rs
+++ b/crates/ark/src/sys/unix/interface.rs
@@ -7,6 +7,8 @@
 
 use std::ffi::CStr;
 use std::os::raw::c_char;
+use std::sync::Condvar;
+use std::sync::Mutex;
 
 use libr::ptr_R_Busy;
 use libr::ptr_R_ReadConsole;
@@ -37,6 +39,9 @@ use crate::interface::r_suicide;
 use crate::interface::r_write_console;
 use crate::interface::RMain;
 use crate::signals::initialize_signal_handlers;
+
+// For shutdown signal in integration tests
+pub static CLEANUP_SIGNAL: (Mutex<bool>, Condvar) = (Mutex::new(false), Condvar::new());
 
 pub fn setup_r(args: &Vec<String>) {
     unsafe {
@@ -78,10 +83,7 @@ pub fn setup_r(args: &Vec<String>) {
         // condition variable sends a notification, which occurs in this cleanup method
         // that is called during R's shutdown process.
         if stdext::IS_TESTING {
-            libr::set(
-                libr::ptr_R_CleanUp,
-                Some(crate::interface::r_cleanup_for_tests),
-            );
+            libr::set(libr::ptr_R_CleanUp, Some(r_cleanup_for_tests));
         }
 
         // In tests R may be run from various threads. This confuses R's stack
@@ -132,4 +134,22 @@ pub fn run_activity_handlers() {
             fdset = R_checkActivity(0, 1);
         }
     }
+}
+
+#[no_mangle]
+pub extern "C-unwind" fn r_cleanup_for_tests(_save_act: i32, _status: i32, _run_last: i32) {
+    // Signal that cleanup has started
+    let (lock, cvar) = &CLEANUP_SIGNAL;
+
+    let mut started = lock.lock().unwrap();
+    *started = true;
+
+    cvar.notify_all();
+    drop(started);
+
+    // Sleep to give tests time to complete before we panic
+    std::thread::sleep(std::time::Duration::from_secs(5));
+
+    // Fallthrough to R which will call `exit()`. Note that panicking from here
+    // would be UB, we can't panic over a C stack.
 }
--- a/crates/ark/src/fixtures/dummy_frontend.rs
+++ b/crates/ark/src/fixtures/dummy_frontend.rs
@@ -10,7 +10,6 @@ use amalthea::fixtures::dummy_frontend::DummyConnection;
 use amalthea::fixtures::dummy_frontend::DummyFrontend;
 
 use crate::interface::SessionMode;
-use crate::interface::CLEANUP_SIGNAL;
 use crate::repos::DefaultRepos;
 
 // There can be only one frontend per process. Needs to be in a mutex because
@@ -66,8 +65,11 @@ impl DummyArkFrontend {
 
     /// Wait for R cleanup to start (indicating shutdown has been initiated).
     /// Panics if cleanup doesn't start within the timeout.
+    #[cfg(unix)]
     #[track_caller]
     pub fn wait_for_cleanup() {
+        use crate::sys::interface::CLEANUP_SIGNAL;
+
         let (lock, cvar) = &CLEANUP_SIGNAL;
         let result = cvar
             .wait_timeout_while(lock.lock().unwrap(), Duration::from_secs(3), |started| {


crates/libr/src/r.rs
```diff
@@ -617,6 +624,14 @@ constant_globals::generate! {
     #[default = std::ptr::null_mut()]
     pub static R_TripleColonSymbol: SEXP;
 
+    #[doc = "\"srcfile\""]
+    #[default = std::ptr::null_mut()]
+    pub static R_SrcfileSymbol: SEXP;
```
DavisVaughan, line 631: Seems like you didn't use `R_SrcfileSymbol`

crates/ark/src/sys/unix/interface.rs
```diff
@@ -73,6 +73,14 @@ pub fn setup_r(args: &Vec<String>) {
         libr::set(ptr_R_Busy, Some(r_busy));
         libr::set(ptr_R_Suicide, Some(r_suicide));
 
+        if stdext::IS_TESTING {
+            use libr::ptr_R_CleanUp;
+
+            use crate::interface::r_cleanup_for_tests;
+
+            libr::set(ptr_R_CleanUp, Some(r_cleanup_for_tests));
+        }
```
DavisVaughan, line 87:
```suggestion
        // TODO!: Comment about why we need this
        if stdext::IS_TESTING {
            libr::set(
                libr::ptr_R_CleanUp,
                Some(crate::interface::r_cleanup_for_tests),
            );
        }
```

I think this needs a big comment like the one below it about why we need this for tests.

I also very much dislike this local `use` style. I've never seen other Rust programmers use it, and it doesn't feel very ergonomic to me.

crates/harp/src/parser/srcref.rs
```diff
@@ -153,6 +153,42 @@ impl TryFrom<&harp::CharacterVector> for SrcFile {
...
+    }
+
+    if harp::r_length(result) < 6 {
+        return RObject::null();
+    }
+
+    RObject::new(result)
+}
+
+pub fn get_block_srcrefs(call: libr::SEXP) -> RObject {
```
DavisVaughan: The name `block` confuses me. Is this just for calls? Can it be `get_call_srcrefs()`?

Is this really different from your `pub fn srcrefs(&self)` helper above? I think we should avoid duplication of work if possible for hygeine reasons.

crates/amalthea/tests/client.rs
```diff
@@ -63,6 +64,34 @@ fn test_amalthea_execute_request() {
...
+    frontend.recv_iopub_busy();
+
+    // Receive the shutdown reply
+    let reply = frontend.recv_control_shutdown_reply();
+    assert_eq!(reply.status, Status::Ok);
+    assert_eq!(reply.restart, false);
+
+    frontend.recv_iopub_idle();
+
+    // Test with restart = true
```
DavisVaughan: I don't think these comments add much

crates/ark/tests/kernel.rs
```diff
@@ -1,9 +1,13 @@
 use amalthea::fixtures::dummy_frontend::ExecuteRequestOptions;
 use amalthea::wire::jupyter_message::Message;
+use amalthea::wire::jupyter_message::Status;
 use amalthea::wire::kernel_info_request::KernelInfoRequest;
 use ark::fixtures::DummyArkFrontend;
 use stdext::assert_match;
 
+// Avoids our global calling handler from rlangifying errors.
+// This causes some test instability across configs.
```
DavisVaughan: I don't think I understand this comment

crates/ark/tests/kernel.rs
```diff
@@ -148,16 +165,54 @@ fn test_execute_request_browser() {
...
     assert_eq!(input.code, code);
 
+    frontend.recv_iopub_stream_stderr("Error: error in parent\n");
     frontend.recv_iopub_idle();
 
     assert_eq!(frontend.recv_shell_execute_reply(), input.execution_count);
+
+    frontend.execute_request("NA", |result| assert!(result.contains("[1] NA")));
+    // Quit the outer browser
+    frontend.execute_request_invisibly("Q");
```
DavisVaughan, line 179: Nice test!

crates/ark/tests/kernel.rs
```diff
@@ -350,17 +444,75 @@ fn test_execute_request_browser_stdin() {
...
 
-    frontend.send_execute_request("stop('foobar')", ExecuteRequestOptions::default());
+    frontend.execute_request_error("stop('foobar')", |error_msg| {
+        assert!(error_msg.contains("foobar"));
+    });
+}
+
+#[test]
+fn test_execute_request_error_with_accumulated_output() {
+    // Test that when the very last input output and then throws an error,
```
DavisVaughan:
```suggestion
    // Test that when the very last input runs and then throws an error,
```

crates/ark/tests/kernel.rs
```diff
@@ -418,28 +613,16 @@ fn test_execute_request_multiple_expressions() {
 fn test_execute_request_single_line_buffer_overflow() {
     let frontend = DummyArkFrontend::lock();
 
+    // This used to fail back when we were passing inputs down to the REPL from
```
DavisVaughan, line 547: Nice!

crates/ark/tests/kernel.rs
```diff
@@ -610,3 +885,133 @@ fn test_env_vars() {
...
+        use nix::sys::signal::signal;
+        use nix::sys::signal::SigHandler;
+        use nix::sys::signal::Signal;
+
+        signal(Signal::SIGINT, SigHandler::Handler(sigint_handler)).unwrap();
+    }
+}
+
+// Note that because of these shutdown tests you _have_ to use `cargo nextest`
+// instead of `cargo test`, so that each test has its own process and R thread.
```
DavisVaughan, line 935: I think you can remove this comment. Really all tests need to go through cargo nextest anyways. These aren't more special than the others.

crates/amalthea/tests/client.rs
```diff
@@ -63,6 +64,34 @@ fn test_amalthea_execute_request() {
     frontend.recv_iopub_idle();
 }
 
+#[test]
+fn test_amalthea_shutdown_request() {
```
DavisVaughan, line 68: It seems weird that this test can send a shutdown request twice. But is the argument that it isn't connected to an R backend that actually does something with the shutdown request?

crates/ark/src/sys/unix/interface.rs
DavisVaughan: IIUC, it is something like

```
// Install a CleanUp hook for integration tests that test the shutdown process.
// We confirm that shutdown occurs by waiting in the test until `CLEANUP_SIGNAL`'s
// condition variable sends a notification, which occurs in this cleanup method
// that is called during R's shutdown process.
```

crates/ark/tests/kernel.rs
```diff
@@ -610,3 +885,133 @@ fn test_env_vars() {
...
+    let reply = frontend.recv_control_shutdown_reply();
+    assert_eq!(reply.status, Status::Ok);
+    assert_eq!(reply.restart, true);
+
+    frontend.recv_iopub_idle();
+
+    DummyArkFrontend::wait_for_cleanup();
+}
+
+static SHUTDOWN_TESTS_ENABLED: bool = false;
```
DavisVaughan, line 972: I think it is worth figuring this out before merging. Otherwise it will just confuse us in the future, especially if one of us that didn't write these tests are looking at them for some reason.

DavisVaughan: Or maybe just put

```rust
if true {
    return;
}
```

directly in the test with a comment that this works most of the time when running locally, but it has race condition issues that prevent us from wanting to run it in CI?

crates/amalthea/src/socket/control.rs
```diff
@@ -102,6 +102,11 @@ impl Control {
         H: FnOnce(JupyterMessage<T>) -> Result<(), Error>,
     {
         // Enter the kernel-busy state in preparation for handling the message.
+        // The protocol specification does not mandate status messages for
+        // Control, but we emit them for compatibility with ipykernel:
+        // https://github.com/ipython/ipykernel/pull/585. These status messages
+        // can be discriminated from those on Shell by examining the parent
+        // header.
```
DavisVaughan: I'm pretty sure it _does_ mandate status messages for Control

https://jupyter-client.readthedocs.io/en/latest/messaging.html#status

> Busy and idle messages should be sent before/after handling every request, not just execution.

If you agree, I think we should remove or tweak this comment

crates/harp/src/object.rs
```diff
@@ -238,6 +238,10 @@ pub fn r_list_poke(x: SEXP, i: R_xlen_t, value: SEXP) {
     }
 }
 
+pub fn r_list_get(x: SEXP, i: R_xlen_t) -> RObject {
+    unsafe { RObject::new(VECTOR_ELT(x, i)) }
+}
```
DavisVaughan: We have `list_get()` and `list_poke()` above. We also have `r_list_poke()` above.

I think we should:

- Remove `list_poke()`
- Remove `list_get()`
- Make this `r_list_get()` return a `SEXP` like `list_get()` did. Same as `r_chr_get()`. The caller should wrap in `RObject` if they need to. That's just how it's been done in these files so far, even for things like `r_alloc_logical()` below. I think it is important to have bare bones and "fast" getters that don't wrap in `RObject`.

crates/harp/src/parse.rs
```diff
@@ -109,17 +107,34 @@ pub fn parse_status<'a>(input: &ParseInput<'a>) -> crate::Result<ParseResult> {
             ParseInput::SrcFile(srcfile) => (srcfile.lines()?, srcfile.inner.clone()),
         };
```
DavisVaughan, line 110: Is removing just `ParseResult` from this function rather than `crate::Result<ParseResult>` not worth it?

I see the `// Should not get here` further below

I was wondering if `ParseResult` should have an `UnknownError` variant or something to encapsulate everything into one type, rather than having unknown errors go through the outer `Result` and parse errors go through `ParseResult::SyntaxError`

crates/harp/src/parse.rs
```diff
@@ -207,15 +222,16 @@ mod tests {
             // Error
             assert_match!(
                 parse_status(&ParseInput::Text("42 + _")),
-                Err(_) => {}
+                Ok(ParseResult::SyntaxError { message }) => {
+                    assert!(message.contains("invalid use of pipe placeholder"));
+                }
             );
 
             // "normal" syntax error
```
DavisVaughan:
```suggestion
            // Syntax error
            assert_match!(
                parse_status(&ParseInput::Text("42 + _")),
                Ok(ParseResult::SyntaxError { message }) => {
                    assert!(message.contains("invalid use of pipe placeholder"));
                }
            );

            // Syntax error
```

They used to be different cases!

crates/ark/src/dap/dap_r_main.rs
```diff
@@ -164,7 +164,11 @@ impl RMainDap {
         }
     }
 
-    pub fn finalize_call_text(&mut self) {
+    pub fn handle_read_console(&mut self) {
+        // Upon entering read-console, finalize any debug call text that we were capturing.
+        // At this point, the user can either advance the debugger, causing us to capture
+        // a new expression, or execute arbitrary code, where we will reuse a finalized
+        // debug call text to maintain the debug state.
         match &self.call_text {
```
DavisVaughan: I think we had discussed maybe moving `call_text` into `RMain` or something. Maybe that's a future task?

crates/ark/src/main.rs
```diff
@@ -83,6 +83,9 @@ fn main() -> anyhow::Result<()> {
     let mut capture_streams = true;
     let mut default_repos = DefaultRepos::Auto;
 
+    // We don't support the asking the user whether to save the workspace data
+    // on exit because calling readline during shutdown puts in a precarious
+    // position. So effectively we're implementing "no-save" by default.
```
DavisVaughan:
```suggestion
```

I think you can remove this since it is below.

crates/ark/src/main.rs
```diff
@@ -325,6 +328,20 @@ fn main() -> anyhow::Result<()> {
...
+        // We don't support the asking the user whether to save the workspace
+        // data on exit because calling readline during shutdown puts in a
+        // precarious position. So effectively we're implementing "no-save" by
+        // default. Note that there is no argument to opt into the "ask"
+        // behaviour, so it can't be reenabled by the user.
+        String::from("--no-save"),
+        // Since we don't save by default, we also don't restore by default for
+        // consistency
+        String::from("--no-restore-data"),
+    ]);
```
DavisVaughan, line 339: Are these really required for this PR? I'm not sure this is really the right place for these and it feels like they warrant more discussion.

We are going to want `commandArgs()` to reflect what the user passes in as options, and this would alter that list
https://github.com/posit-dev/ark/issues/670

It's possible these should be added in something like `interface.rs` so the tests accurately reflect what we do here too. Right now I think the tests don't see this at all?

crates/ark/src/interface.rs
```diff
@@ -259,14 +261,129 @@ pub struct RMain {
...
+    read_console_depth: Cell<usize>,
+
+    /// Set to true when `r_read_console()` exits. Reset to false at the start
+    /// of each `r_read_console()` call. Used to detect if `eval()` returned
+    /// from a nested REPL (the flag will be true when the evaluation returns).
+    read_console_nested_return: Cell<bool>,
+
+    /// Set to true `r_read_console()` exits via an error longjump. Used to
+    /// detect if we need to go return from `r_read_console()` with a dummy
+    /// evaluation to reset things like `R_EvalDepth`.
```
DavisVaughan:
```suggestion
    /// Set to true when `r_read_console()` exits via an error longjump. Used to
    /// detect if we need to return from `r_read_console()` with a dummy
    /// evaluation to reset things like `R_EvalDepth`.
```

crates/ark/src/interface.rs
```diff
@@ -259,14 +261,129 @@ pub struct RMain {
...
     debug_session_index: u32,
+
+    /// Tracks how many nested `r_read_console()` calls are on the stack.
+    /// Incremented when entering `r_read_console(),` decremented on exit.
+    read_console_depth: Cell<usize>,
+
+    /// Set to true when `r_read_console()` exits. Reset to false at the start
+    /// of each `r_read_console()` call. Used to detect if `eval()` returned
+    /// from a nested REPL (the flag will be true when the evaluation returns).
+    read_console_nested_return: Cell<bool>,
```
DavisVaughan, line 302:
```suggestion
    /// from a nested REPL (the flag will be true when the evaluation returns).
    /// In these cases, we need to return from `r_read_console()` with a dummy
    /// evaluation to reset things like `R_ConsoleIob`. 
    read_console_nested_return: Cell<bool>,
```

crates/ark/src/interface.rs
```diff
@@ -259,14 +261,129 @@ pub struct RMain {
...
+
+    /// Set to true `r_read_console()` exits via an error longjump. Used to
+    /// detect if we need to go return from `r_read_console()` with a dummy
+    /// evaluation to reset things like `R_EvalDepth`.
+    read_console_threw_error: Cell<bool>,
+
+    /// Used to track an input to evaluate upon returning to `r_read_console()`,
+    /// after having returned a dummy input to reset `R_ConsoleIob` in R's REPL.
+    read_console_next_input: Cell<Option<String>>,
```
DavisVaughan:
```suggestion
    read_console_nested_return_next_input: Cell<Option<String>>,
```

It would make more sense to me if this and `read_console_nested_return` had a defining common prefix of some kind. If you could also group these together in the RMain struct, that would also be helpful I think.

crates/ark/src/interface.rs
```diff
@@ -259,14 +261,129 @@ pub struct RMain {
...
+
+    /// We've received a Shutdown signal and need to return EOF from all nested
+    /// consoles to get R to shut down
+    read_console_shutdown: Cell<bool>,
+
+    /// Current topmost environment on the stack while waiting for input in ReadConsole
+    read_console_frame: RefCell<RObject>,
```
DavisVaughan: Why isn't this one a `Cell`? Maybe a comment about that?

crates/ark/src/interface.rs
```diff
@@ -259,14 +261,129 @@ pub struct RMain {
...
+    index: isize,
+}
+
+enum ParseResult<T> {
+    Success(Option<T>),
+    SyntaxError(String),
+}
+
+impl PendingInputs {
+    pub(crate) fn read(input: &str) -> anyhow::Result<ParseResult<PendingInputs>> {
```
DavisVaughan, line 335: This and `is_empty()` and `pop()` are all private right? And `struct PendingInput` and `enum ConsoleResult`. If so, removing the pub can help us detect dead code better in the future.

crates/ark/src/interface.rs
```diff
@@ -259,14 +261,129 @@ pub struct RMain {
...
+        let status = match harp::parse_status(&harp::ParseInput::Text(input)) {
+            Err(err) => {
+                // Failed to even attempt to parse the input, something is seriously wrong
+                // FIXME: There are some valid syntax errors going through here, e.g. `identity |> _(1)`.
```
DavisVaughan: Can remove this FIXME now!

crates/ark/src/interface.rs
```diff
@@ -259,14 +261,129 @@ pub struct RMain {
...
+                return Ok(ParseResult::SyntaxError(format!("{err}")));
+            },
+            Ok(status) => status,
+        };
+
+        // - Incomplete inputs put R into a state where it expects more input that will never come, so we
+        //   immediately reject them. Positron should never send us these, but Jupyter Notebooks may.
+        // - Complete statements are obviously fine.
+        // - Syntax errors will cause R to throw an error, which is expected.
```
DavisVaughan:
```suggestion
        // - Syntax errors get bubbled up as R errors via an `ConsoleResult::Error`.
```

Maybe? It's a little different now I think?

crates/ark/src/interface.rs
```diff
@@ -259,14 +261,129 @@ pub struct RMain {
...
+            index,
+        })))
+    }
+
+    pub(crate) fn is_empty(&self) -> bool {
+        self.index >= self.len
+    }
+
+    pub(crate) fn pop(&mut self) -> Option<PendingInput> {
+        if self.index >= self.len {
```
DavisVaughan: Use `self.is_empty()`? To keep the weirdness in one place and keep this readable

crates/ark/src/interface.rs
```diff
@@ -616,14 +741,19 @@ impl RMain {
...
             debug_preserve_focus: false,
             debug_last_stack: vec![],
-            debug_env: None,
             debug_session_index: 1,
+            pending_inputs: None,
+            read_console_depth: Cell::new(0),
+            read_console_nested_return: Cell::new(false),
+            read_console_threw_error: Cell::new(false),
+            read_console_next_input: Cell::new(None),
+            read_console_frame: RefCell::new(RObject::new(unsafe { libr::R_GlobalEnv })),
```
DavisVaughan, line 797: `R_ENVS.global`?

crates/ark/src/interface.rs
```diff
@@ -731,66 +861,57 @@ impl RMain {
...
-            return console_result;
-        };
+        if let Some(exception) = self.take_exception() {
+            // Clear any pending inputs, if any
+            self.pending_inputs = None;
+
+            // Reply to active request with error
+            self.handle_active_request(&info, ConsoleValue::Error(exception));
+        } else if matches!(info.kind, PromptKind::InputRequest) {
+            // Request input reply to the frontend and return it to R
```
DavisVaughan:
```suggestion
            // Request input from the frontend and return it to R
```

?

crates/ark/src/interface.rs
```diff
@@ -941,108 +1073,134 @@ impl RMain {
...
-        // Finally, check if we have an active request from a previous `read_console()`
-        // iteration. If so, we `take()` and clear the `active_request` as we're about
-        // to complete it and send a reply to unblock the active Shell
-        // request.
+        Some(exception)
+    }
+
+    /// Returns:
+    /// - `None` if we should fall through to the event loop to wait for more user input
+    /// - `Some(ConsoleResult)` if we should immediately exit `read_console()`
```
DavisVaughan: I don't think these comments are right. In the `Some(ConsoleResult)` case, we clear the active request and reply but still fall through, right? i.e. we don't exit `read_console()`

crates/ark/src/interface.rs
```diff
@@ -731,66 +861,57 @@ impl RMain {
...
+            self.handle_active_request(&info, ConsoleValue::Error(exception));
+        } else if matches!(info.kind, PromptKind::InputRequest) {
+            // Request input reply to the frontend and return it to R
+            return self.handle_input_request(&info, buf, buflen);
+        } else if let Some(input) = self.pop_pending() {
+            // Evaluate pending expression if there is any remaining
+            return self.handle_pending_input(input, buf, buflen);
+        } else {
+            // Otherwise reply to active request with accumulated result
+            let result = self.take_result();
```
DavisVaughan:
```suggestion
            // Reply to active request with error, then fall through to event loop
            self.handle_active_request(&info, ConsoleValue::Error(exception));
        } else if matches!(info.kind, PromptKind::InputRequest) {
            // Request input reply to the frontend and return it to R
            return self.handle_input_request(&info, buf, buflen);
        } else if let Some(input) = self.pop_pending() {
            // Evaluate pending expression if there is any remaining
            return self.handle_pending_input(input, buf, buflen);
        } else {
            // Otherwise reply to active request with accumulated result, then
            // fall through to event loop
            let result = self.take_result();
```

It feels useful to mention the fall throughs explicitly? It's not super clear since this is a bit complicated.

crates/ark/src/interface.rs
```diff
@@ -830,22 +960,28 @@ impl RMain {
...
+            WaitFor::ExecuteRequest => (Some(select.recv(&r_request_rx)), None),
+            WaitFor::InputReply => (None, Some(select.recv(&stdin_reply_rx))),
+        };
+
         let kernel_request_index = select.recv(&kernel_request_rx);
         let tasks_interrupt_index = select.recv(&tasks_interrupt_rx);
         let polled_events_index = select.recv(&polled_events_rx);
 
-        // Don't process idle tasks in browser prompts. We currently don't want
+        // Don't process idle tasks unless at top level. We currently don't want
```
DavisVaughan:
```suggestion
        // Only process idle tasks at top level. We currently don't want
```

crates/ark/src/interface.rs
```diff
@@ -866,25 +1002,21 @@ impl RMain {
             // reset the flag
             set_interrupts_pending(false);
```
DavisVaughan, line 1063: I'm having trouble commenting on the deleted lines, but there was a comment about a FIXME about a race condition between interrupt and new code requests. Is that fixed?

crates/ark/src/interface.rs
```diff
@@ -941,108 +1073,134 @@ impl RMain {
...
-        // this means some user code is requesting input, e.g. via `readline()`
-        let user_request = !browser && n_frame > 0;
-
-        // The request is incomplete if we see the continue prompt, except if
-        // we're in a user request, e.g. `readline("+ ")`. To guard against
-        // this, we check that we are at top-level (call stack is empty or we
-        // have a debug prompt).
-        let top_level = n_frame == 0 || browser;
-        let incomplete = matches_continuation && top_level;
+        let browser = RE_DEBUG_PROMPT.is_match(&prompt);
```
DavisVaughan, line 1145: We spent a lot of time getting these exactly right. How is it that we can just remove most of this now?

DavisVaughan: Oh I guess most of the simplification is that we no longer have an `Incomplete` state? Wow that's nice.

crates/ark/src/interface.rs
```diff
@@ -941,108 +1073,134 @@ impl RMain {
...
-            }
+        // The output generated by autoprint is emitted as an
+        // `execute_result` message.
+        let mut autoprint = std::mem::take(&mut self.autoprint_output);
+
+        if autoprint.ends_with('\n') {
+            // Remove the trailing newlines that R adds to outputs but that
+            // Jupyter frontends are not expecting. Is it worth taking a
+            // mutable self ref across calling methods to avoid the clone?
+            autoprint.pop();
```
DavisVaughan: IIUC this doesn't clone

crates/ark/src/interface.rs
```diff
@@ -941,108 +1073,134 @@ impl RMain {
...
+            // tree which is just as well since the recursive calls would
+            // push a tree too far to the right.
+            let traceback = r_traceback();
+
+            // Reset error buffer so we don't display this message again
+            let _ = RFunction::new("base", "stop").call();
+
+            Exception {
+                ename: String::from(""),
+                evalue: r_peek_error_buffer(),
```
DavisVaughan, line 1217: I believe this timing is incorrect. Previously we called `r_peek_error_buffer()` and stored the result _before_ calling `base::stop()`. Isn't that right?

crates/ark/src/interface.rs
```diff
@@ -941,108 +1073,134 @@ impl RMain {
...
+        // Flush any accumulated output to StdOut. This can happen if
+        // the last input errors out during autoprint.
+        let autoprint = std::mem::take(&mut self.autoprint_output);
+        if !autoprint.is_empty() {
+            let message = IOPubMessage::Stream(StreamOutput {
+                name: Stream::Stdout,
+                text: autoprint,
+            });
+            self.iopub_tx.send(message).unwrap();
         }
```
DavisVaughan, line 1238: This is new right? Is this something we were not handling well before?

crates/ark/src/interface.rs
```diff
@@ -1105,48 +1265,166 @@ impl RMain {
...
-                match Self::on_console_input(buf, buflen, code) {
-                    Ok(()) => Some(ConsoleResult::NewInput),
-                    Err(err) => Some(ConsoleResult::Error(err)),
-                }
+                // Otherwise we got an empty input, e.g. `""` and there's
+                // nothing to do. Close active request.
+                self.handle_active_request(info, ConsoleValue::Success(Default::default()));
+
+                // And return to event loop
+                None
```
DavisVaughan: Maybe change this to a style of

```
if let = {

} else {

}
```

that way you don't need the early return

I think that makes it a little more clear that the happy path is the `handle_pending_input()` path?

crates/ark/src/interface.rs
```diff
@@ -1105,48 +1265,166 @@ impl RMain {
...
-
-                // Store input in R's buffer and return sentinel indicating some
-                // new input is ready
-                match Self::on_console_input(buf, buflen, code) {
-                    Ok(()) => Some(ConsoleResult::NewInput),
-                    Err(err) => Some(ConsoleResult::Error(err)),
-                }
+                // Otherwise we got an empty input, e.g. `""` and there's
+                // nothing to do. Close active request.
+                self.handle_active_request(info, ConsoleValue::Success(Default::default()));
```
DavisVaughan: I'm still a little weirded out by this one because it forces us to call `handle_active_request()` at a non standard time. I still wonder if we could go through the "standard path" by having this return `ConsoleValue::None` which causes `r_read_console_impl` to write `.ark_last_value` into the buffer and evaluate that, or something similar that causes some kind of similar "no-op" style evaluation to occur where we fall through our iteration of read-console completely, so our standard `handle_active_request` path gets a chance to run again on the next "fresh" iteration of read-console.

crates/ark/src/interface.rs
```diff
@@ -1105,48 +1265,166 @@ impl RMain {
...
+            // Send request to frontend. We'll wait for an `input_reply`
+            // from the frontend in the event loop in `read_console()`.
+            // The active request remains active.
+            self.request_input(req.originator.clone(), String::from(&info.input_prompt));
+
+            // Run the event loop, waiting for stdin replies but not execute requests
+            self.run_event_loop(info, buf, buflen, WaitFor::InputReply)
+        } else {
+            // Invalid input request, propagate error to R
+            self.handle_invalid_input_request(buf, buflen)
```
DavisVaughan, line 1377: Nice and simple!

crates/ark/src/interface.rs
```diff
@@ -1105,48 +1265,166 @@ impl RMain {
...
+        if pending_inputs.is_empty() {
+            self.pending_inputs = None;
+        }
+
+        Some(input)
+    }
+
+    // SAFETY: Call this from a POD frame. Inputs must be protected.
+    unsafe fn eval(expr: libr::SEXP, srcref: libr::SEXP, buf: *mut c_uchar, buflen: c_int) {
+        let frame = harp::r_current_frame();
```
DavisVaughan, line 1449: Hmmmmmmmmmmm, big code smell about this being an `RObject` in a POD frame! Destructor probably won't run if this jumps?

DavisVaughan: Oh hmm, you've moved ownership into a `SEXP` and destroyed the `RObject` in the next line.

That really confused me!

crates/ark/src/interface.rs
```diff
@@ -1353,10 +1621,10 @@ impl RMain {
                 let input = convert_line_endings(&input.value, LineEnding::Posix);
                 match Self::on_console_input(buf, buflen, input) {
                     Ok(()) => ConsoleResult::NewInput,
-                    Err(err) => ConsoleResult::Error(err),
+                    Err(err) => ConsoleResult::Error(format!("{err:?}")),
                 }
             },
-            Err(err) => ConsoleResult::Error(err),
+            Err(err) => ConsoleResult::Error(format!("{err:?}")),
```
DavisVaughan, line 1642: Why remove the `amalthea::Error`?

crates/ark/src/interface.rs
```diff
@@ -2287,9 +2389,106 @@ pub extern "C-unwind" fn r_read_console(
...
+    // from all consoles on the stack to let R shut down with an `exit()`.
+    if main.read_console_shutdown.get() {
+        return 0;
+    }
+
+    // We've finished evaluating a dummy value to reset state in R's REPL,
+    // and are now ready to evaluate the actual input
+    if let Some(next_input) = main.read_console_next_input.take() {
+        RMain::on_console_input(buf, buflen, next_input).unwrap();
+        return 1;
```
DavisVaughan:
```suggestion
    // and are now ready to evaluate the actual input, which is typically
    // just `.ark_last_value`.
    if let Some(next_input) = main.read_console_next_input.take() {
        RMain::on_console_input(buf, buflen, next_input).unwrap();
        return 1;
```

Right? At first I was concerned that we could write something here that would overflow the buffer.

crates/ark/src/interface.rs
```diff
@@ -2287,9 +2389,106 @@ pub extern "C-unwind" fn r_read_console(
...
+    // In case of error, we haven't had a chance to evaluate ".ark_last_value".
+    // So we return to the R REPL to give R a chance to run the state
+    // restoration that occurs between `R_ReadConsole()` and `eval()`:
+    // - R_PPStackTop: https://github.com/r-devel/r-svn/blob/74cd0af4/src/main/main.c#L227
+    // - R_EvalDepth:  https://github.com/r-devel/r-svn/blob/74cd0af4/src/main/main.c#L260
+    //
+    // Technically this also resets time limits (see `base::setTimeLimit()`) but
+    // these aren't supported in Ark because they cause errors when we poll R
+    // events.
+    if main.last_error.is_some() && main.read_console_threw_error.get() {
```
DavisVaughan, line 2440:
```suggestion
    if main.read_console_threw_error.get() {
```

Is one of these enough?

If not, it's mildly confusing that you reset `read_console_threw_error` but not `last_error` in the if block

crates/ark/src/interface.rs
```diff
@@ -2354,6 +2599,28 @@ pub unsafe extern "C-unwind" fn r_polled_events() {
     };
 }
 
+// For integration tests
+use std::sync::Condvar;
+pub static CLEANUP_SIGNAL: (Mutex<bool>, Condvar) = (Mutex::new(false), Condvar::new());
+
+#[no_mangle]
+pub extern "C-unwind" fn r_cleanup_for_tests(_save_act: i32, _status: i32, _run_last: i32) {
```
DavisVaughan: I think we are likely to get confused about the OS level support for this right now.

Can we move all of this into a `unix/` folder's file?

It is only supported by `unix/interface.rs`, so it is misleading to have `CLEANUP_SIGNAL` here.

It becomes very confusing when you look at `DummyArkFrontend::wait_for_cleanup()`, which looks like it can be used on any OS because `CLEANUP_SIGNAL` is available on any OS. But that will never get a notification on Windows! 

So it would be nice if `wait_for_cleanup()` would be marked with `cfg(unix)`, which requires making everything here `cfg(unix)` as well.

crates/libr/src/r.rs
lionel-: Ideally `r.rs` would be an exhaustive reference, so I think we should keep the declaration even though it didn't end up used in the PR.

crates/ark/tests/kernel.rs
lionel-: CI was failing because the error included rlang's `!` prefix. I don't understand why this doesn't happen locally.

crates/ark/tests/kernel.rs
lionel-: Lots of tests still work fine with the regular runner so it's still useful to call this out. Especially with AIs around.

crates/ark/tests/kernel.rs
lionel-: Having fully dead code would cause warnings.

crates/amalthea/src/socket/control.rs
lionel-: I based my comment on this quote.

"not just execution" could just refer to other types of Shell requests, including comm messages. Note that this section is incomplete and does not say anything about the parent header of the status messages, so an implementer might create inadvertently non-nested shell status messages.

I think wrapping control requests in Busy/Idle is a strange choice that unnecessarily complicates the protocol.

crates/harp/src/object.rs
lionel-: I think it's more important to have protection by default than fast getters, which already exist in `libr::`

crates/ark/src/interface.rs
lionel-: yep this is one of the bugfixes mentioned in the PR notes, we have a new test for it.

This is the bug I mentioned in our live review when a print method fails during autoprint.

crates/ark/src/interface.rs
lionel-: It felt like an odd choice since we're in Ark not Amalthea

crates/amalthea/src/socket/control.rs
lionel-: The comment is now:

```
        // The protocol specification is vague about status messages for
        // Control, we mostly emit them for compatibility with ipykernel:
```

crates/amalthea/tests/client.rs
lionel-: I've added this comment:

```
    // Although the R thread has shut down, the Amalthea thread keeps running
    // and is able to reply.
```

crates/ark/src/dap/dap_r_main.rs
lionel-: I've now moved all of `dap_r_main.rs` to `RMain`. This functionality and the debug-related functionality of `interface.rs` are now gathered in `repl_debug.rs`

crates/ark/src/interface.rs
lionel-: I don't see any `pub` symbols?

crates/ark/src/interface.rs
lionel-: I added a comment:

```rs
            // Can't use `R_ENVS.global` here as it isn't initialised yet
```

crates/ark/tests/kernel.rs
lionel-: hmm that comment got moved here but I'm not sure how. I don't remember where it should go so I'll just remove it.

crates/harp/src/parser/srcref.rs
lionel-: A block is a call to `{` but that helper would work with expression vectors too. I've renamed it to `get_srcref_list()`. I've also renamed `get_srcref()` to `srcref_list_get()`.

It does something very different from `fn srcrefs()` since it just extracts the list if it exists.

crates/harp/src/parse.rs
lionel-: hmm I think _unexpected_ errors should be a rust result that callers can handle in the regular way (propagate, log, panic, etc), and expected errors should be variants of the returned value that callers handle in a ad hoc way depending on their needs.

crates/ark/src/main.rs
lionel-: This is related to the work that allowed removing the automatic "y" reply on shutdown.

crates/ark/src/interface.rs
lionel-: I understand, it's not ideal. Alternatives feel like they amount to gymnastics and in the end wouldn't make this much better though.

crates/ark/src/interface.rs
lionel-: I no longer agree with/understand that comment. I think I can see another race though, here is the comment I wrote for it:

```rs
            // NOTE: There is a slight race between interrupt and new execute
            // request. We check for interrupts with interval polling and if an
            // execute request comes after an interrupt was emitted but before
            // the interval timed out, we'll start evaluating the request
            // without having had a chance to reset the "interrupt pending"
            // flag, causing the new evaluation to get incorrectly interrupted.
            // To fix this, we could have the interrupt handler wake our event
            // loop.
```

But honestly this is such a benign race I don't think we should even spend a comment on it.

PR #962, opened 2025-11-14
Make Ark's `stdout` behaviour consistent with Python

kv9898:
Part of https://github.com/posit-dev/positron/issues/10505

### DIFF
--- a/crates/ark/src/interface.rs
+++ b/crates/ark/src/interface.rs
@@ -1882,13 +1882,7 @@ impl RMain {
             Stream::Stderr
         };
 
-        // If active execution request is silent don't broadcast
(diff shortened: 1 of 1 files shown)


lionel-:
Unfortunately this is in contradiction with the Jupyter protocol:

```py
# A boolean flag which, if True, signals the kernel to execute
# this code as quietly as possible.
# silent=True forces store_history to be False,
# and will *not*:
#   - broadcast output on the IOPUB channel
#   - have an execute_result
# The default is False.
'silent' : bool,
```

https://jupyter-client.readthedocs.io/en/stable/messaging.html#execute

PR #965, opened 2025-11-19
Add full statement range support to roxygen `@examples` and `@examplesIf`

DavisVaughan:
Addresses https://github.com/posit-dev/positron/issues/8559 (via the fact that we no longer try to run examples one line at a time, there is still a regression in the frontend code logged elsewhere)

I can't find an issue that talks about wanting full Statement Range support in roxygen comments, but since we have the regression where running one line at a time doesn't work either, this became a much higher priority for me.

---

This PR adds full Statement Range support to roxygen `@examples` and `@examplesIf` blocks.

- Previously we could only run _one line at a time_ (and a regression had broken this too)
- Now we can run _entire blocks of code at once_, just like the regular statement range provider

It works by extracting out the `@examples` or `@examplesIf` blocks into a subdocument, which we parse on its own and run through the statement range provider (adjusting row positions accordingly).

This is such a big QOL improvement for me 😬 

I've also reworked the tests for statement range so that you can debug them one at a time. It was pretty frustrating to debug one individual test when others were in the way. Along the way I made it so that `point_from_cursor()` could accept an alternative cursor token than `@`, because that's what is used to denote roxygen tags! This allowed me to make the statement range tests much more readable.

https://github.com/user-attachments/assets/4dfd37e2-c89a-4db0-9147-6ca3846b4609

### DIFF
--- a/crates/ark/src/lsp/statement_range.rs
+++ b/crates/ark/src/lsp/statement_range.rs
@@ -2088,6 +2088,25 @@ x %>%
     }
 
     #[test]
+    fn test_statement_range_roxygen_multiple_spaces_before_the_next_tag() {
+        let text = "
+#' Hi
+#' @param x foo
+#' @examples
+#' 1 + 1^
+#'     @returns
+2 + 2
+";
+        let (text, point) = statement_range_point_from_cursor(text);
+        let document = Document::new(&text, None);
+        let root = document.ast.root_node();
+        let contents = &document.contents;
+        let (range, code) = find_roxygen_statement_range(&root, contents, point).unwrap();
+        assert_eq!(get_text(range, contents), String::from("#' 1 + 1"));
+        assert_eq!(code.unwrap(), String::from("1 + 1"));
+    }
+
+    #[test]
     fn test_statement_range_roxygen_parse_errors_in_examples() {
         // Still sends "just that line" to avoid jumping around
         let text = "


crates/ark/src/lsp/statement_range.rs
```diff
@@ -92,105 +90,246 @@ fn new_statement_range_response(
...
+    // If we are within `@examples` or `@examplesIf`, first find the range that spans the
+    // full examples section
+    if let Some(range) = find_roxygen_examples_section(node, contents) {
+        // Then narrow in on the exact range of code that the user's cursor covers
+        if let Some((range, code)) = find_roxygen_examples_range(root, range, contents, point) {
+            return Some((range, Some(code)));
+        };
+    }
+
+    // If we aren't in an `@examples` or `@examplesIf` section (or we were, but it is
```
lionel-, line 120: Nice touch

crates/ark/src/lsp/statement_range.rs
```diff
@@ -92,105 +90,246 @@ fn new_statement_range_response(
...
+    // somehow invalid), we still check to see if we are in a roxygen2 comment of any
+    // kind. If so, we send just the current comment line to prevent "jumping" past the
+    // entire roxygen block if you misplace your cursor and `Cmd + Enter`.
+    if as_roxygen_comment_text(&node, contents).is_some() {
```
lionel-, line 124: This relies on the fact that `node` is a single line comment right? If so a note might be nice, I wasn't sure at first how `as_roxygen_comment_text()` worked.

crates/ark/src/lsp/statement_range.rs
```diff
@@ -92,105 +90,246 @@ fn new_statement_range_response(
...
+        // Drop `#'` from the comment's text
+        let sibling_text = RE_ROXYGEN2_COMMENT.replace(&sibling_text, "");
 
-        let sibling = contents.node_slice(&sibling).unwrap().to_string();
-        let sibling = sibling.as_str();
+        // Trim leading whitespace after the `#'`
+        let sibling_text = sibling_text.trim_start();
+
+        // Did we discover a new tag?
+        if sibling_text.starts_with("@") {
```
lionel-, line 194: maybe extract in helper? It's a bit of a heavy repeated read with the comments

lionel-: Would be nice to simplify with iterator adaptors if possible.

(but fine to leave as is)

crates/ark/src/lsp/statement_range.rs
```diff
@@ -92,105 +90,246 @@ fn new_statement_range_response(
...
+        // Drop `#'` from the comment's text
+        let sibling_text = RE_ROXYGEN2_COMMENT.replace(&sibling_text, "");
 
-        // Otherwise, did we discover the `node` was in a different tag?
-        if sibling.starts_with("@") {
+        // Trim leading whitespace after the `#'`
+        let sibling_text = sibling_text.trim_start();
+
+        // Did we discover a new tag?
+        if sibling_text.starts_with("@") {
```
lionel-, line 231: Would be nice to note somewhere that this text scanning approach isn't totally robust but should generally work well. E.g. it'd fail with:

```r
#' "
#' @
#' "
```

crates/ark/src/lsp/statement_range.rs
```diff
@@ -92,105 +90,246 @@ fn new_statement_range_response(
...
+    let Some(slice) = contents.get_byte_slice(range.start_byte..range.end_byte) else {
+        return None;
+    };
+
+    // Trim out leading roxygen comments so we are left with a subdocument of actual code
+    let subcontents = slice.to_string();
+    let subcontents: Vec<String> = subcontents
+        .lines()
+        .map(|line| {
+            // Trim `#'` and at most 1 leading whitespace character. Don't trim more
```
lionel-, line 269: The space trimming is also incorrect in strings (again, a reasonable limitation)

crates/ark/src/lsp/statement_range.rs
```diff
@@ -1572,138 +1714,448 @@ list({
...
+NULL
+";
+        let (text, point) = statement_range_point_from_cursor(text);
+        let document = Document::new(&text, None);
+        let root = document.ast.root_node();
+        let contents = &document.contents;
+        let (range, code) = find_roxygen_statement_range(&root, contents, point).unwrap();
+        assert_eq!(get_text(range, contents), String::from("#' 1 +\n#'   1"));
+        assert_eq!(code.unwrap(), String::from("1 +\n  1"));
     }
```
lionel-, line 2182: Maybe a test where `@returns` has more than one space after the roxy prefix?

PR #967, opened 2025-11-24
Check that profile path exists before sourcing

### LINKED ISSUE #966: Verbose error when site file does not exist
Report from https://positpbc.slack.com/archives/C02P4PH94QL/p1763759108177519, copied below:

Users have `R_PROFILE_USER="../input/.Rprofile"` set in their `Renviron.site`, and it results in an error in the Positron session that does not show up in the R sessions in Workbench when that file is not found.

The R kernel in Positron is outputting those errors, as I don't see this in R, either inside or outside of an RStudio session. It seems like it should behave similarly to R outside and in RStudio, but it's using a kernel to run R instead.
In strace R outside the product, (no errors shown in the R console):

```
1201469 09:53:30.595546 openat(AT_FDCWD, "../input/.Rprofile", O_RDONLY) = -1 ENOENT (No such file or directory)
```

In Rstudio strace output (no errors shown in the R console):

```
1205030 10:14:46.074706 openat(AT_FDCWD</home/mark>, "../input/.Rprofile", O_RDONLY) = -1 ENOENT (No such file or directory) <0.002210>
```

In Positron R kernel:

```r
Error in base::sys.source(file = "../input/.Rprofile", envir = <environment>) :
 '../input/.Rprofile' is not an existing file
Error while sourcing R profile file at path '../input/.Rprofile':
Unexpected longjump
Likely caused by:
Error in base::sys.source(file = "../input/.Rprofile", envir = <environment>) :
 '../input/.Rprofile' is not an existing file
> getwd()
[1] "/home/mark"
> file.exists("../input/.Rprofile")
[1] FALSE
> Sys.getenv("R_PROFILE_USER")
[1] "../input/.Rprofile"
```

And the Positron R kernel logs..

```
2025-11-21 12:44:44.161 [info] ** Begin kernel log for session R 4.4.2 (r-1715af64) at 11/21/2025, 12:44:35 PM **
Streaming kernel log file: /tmp/kernel-RdwVQj/kernel.log
[R] 2025-11-21T17:44:57.656757Z ERROR Error while sourcing R profile at '../input/.Rprofile': Unexpected longjump
[R] Likely caused by:
[R] Error in base::sys.source(file = "../input/.Rprofile", envir = ) :
[R] '../input/.Rprofile' is not an existing file
[R]
[R]
[R]
[R] R thread backtrace:
[R] 0: 
[R] 1: 
[R] 2: 
[R] 3: 
[R] 4: 
[R] 5: 
[R] 6: 
[R] 7: 
[R] 8: 
[R] 9: 
[R] 10: __libc_start_call_main
[R] at ./csu/../sysdeps/nptl/libc_start_call_main.h:58:16
[R] 11: __libc_start_main_impl
[R] at ./csu/../csu/libc-start.c:360:3
[R] 12: 
[R]
[R] 0: 
[R] 1: 
[R] 2: 
[R] 3: 
[R] 4: 
[R] 5: 
[R] 6: 
[R] 7: 
[R] 8: 
[R] 9: 
[R] 10: __libc_start_call_main
[R] at ./csu/../sysdeps/nptl/libc_start_call_main.h:58:16
[R] 11: __libc_start_main_impl
[R] at ./csu/../csu/libc-start.c:360:3
[R] 12: 
[R]
[R] at crates/ark/src/startup.rs:86
[R]
```

lionel-:
Closes #966.

R ignores `R_PROFILE` and `R_PROFILE_USER` when they don't exist: https://github.com/wch/r-source/blob/ab34772818fcb3ec74ad1af4a38e1ac1410d7933/src/main/main.c#L721. We now do the same (with a logged warning) to avoid visible errors streamed on IOPub.

I've added a couple of tests and gathered all site file tests in `kernel-nexstep.rs`. These tests need to be run with `cargo nexstep` as they rely on unique process singletons. I'm not prepared to treat all of `kernel.rs` in this way just yet because many IDE affordances for running/debugging tests do not support nexstep.

#### Bug Fixes

- Fixed a bug in the R backend that caused visible errors when the `R_PROFILE` or `R_PROFILE_USER` environment variables were set to paths that don't exist (https://github.com/posit-dev/ark/issues/966).


crates/ark/src/startup.rs
```diff
@@ -62,6 +62,11 @@ fn source_r_profile(path: &PathBuf) {
 
     log::info!("Found R profile at '{path}', sourcing now");
 
+    if !std::path::Path::new(path).exists() {
+        log::warn!("R profile at '{path}' does not exist, skipping source");
+        return;
+    }
```
DavisVaughan, line 69: `find_user_r_profile()` and `find_site_r_profile()` both already check if `path.exists()` in the non-envvar cases. Should they also check if `path.exists()` in the envvar case to avoid this?

That would help ensure an invariant that if one of the `find_*()` functions returns a PathBuf, then it _really really_ thinks that that path exists, since this is the only case where that doesn't hold.

I'd imagine it would work like

```rust
    // Try from env var first
    match std::env::var("R_PROFILE") {
        Ok(path) => {
            let path = PathBuf::from_str(path.as_str()).log_err()?;
            if !path.exists() {
                log::warn!(
                    "`R_PROFILE` detected, but path does not exist: '{path}'",
                    path = path.to_string_lossy()
                );
                return None;
            }
            return Some(path);
        },
        Err(_) => (),
    };
```

I like that this also tells you that `R_PROFILE` found it

We'd probably still keep this error as an extra check, but upgrade it to a `log::error!()`?

crates/ark/tests/kernel-r-profile.rs
DavisVaughan, line 1: `kernel-nexstep.rs` -> `kernel-nextest.rs`?

DavisVaughan, line 1: I really think we should just call this `r-profile.rs`

IIUC we run tests in one of two ways:

- `just test` at the command line to run all tests with nextest
- The `Run Test` button that VS Code provides via rust-analyzer to run a single test

It is impossible to run the full suite of tests without `cargo nextest`, so it does not feel worth it to me to call out that you need it here. Other tests like `test_inability_to_load_ragg_falls_back_to_base_graphics` would also need such a call out. Our `BUILDING.md` mentions the need for nextest pretty clearly.

Do you run tests in a different way than the two mentioned here? This is the 2nd time we've had differing opinions here, so I think that suggests we somehow have different workflows or worldviews about this

DavisVaughan (approved):
No blocking requests

But we should probably discuss our thoughts about nextest together soon!

crates/ark/tests/kernel-r-profile.rs
lionel-: > it is impossible to run the full suite of tests without cargo nextest, so it does not feel worth it to me to call out that you need it here. 

It's possible to run a _subset_ of the suite of tests with `cargo test`, and comes up regularly for me due to various tooling/IDE integrations. For instance "run tests for this module".

lionel-: `cargo test` is so deeply integrated in various Rust workflows that I think we'll be in an intermediate state for a while until test runners become indirected in these integrations due to the success of nextest.

crates/ark/src/startup.rs
lionel-: I don't mind implementing an upstream check but I'll probably keep the downstream check as a defensive measure since the call site is apart, the check is cheap, and this will spare us documenting an invariant.

crates/ark/tests/kernel-r-profile.rs
lionel-: I renamed this to `kernel-r-profile.rs` as I think these integration tests should have a common prefix.

DavisVaughan: > For instance "run tests for this module"

Yea this is the workflow difference between us that i was missing

That's a button for you in Zed?

DavisVaughan: Oh right, its "Run Tests" in VS Code. I remember opening this now https://github.com/nextest-rs/nextest/discussions/2253

DavisVaughan: I've reopened this as a rust-analyzer issue instead https://github.com/rust-lang/rust-analyzer/issues/21137

PR #972, opened 2025-12-02
Add option to use Windows default DLL search strategy (for Conda)

jmcphers:
Ark's default DLL search strategy doesn't work for Conda R since it depends on libraries shipped with Conda that are not placed in either of the locations Ark currently looks (alongside the R binary and in the same path as `R.dll` itself). 

This change adds a `--dll-search-path` option (only for Windows) that searches `PATH` for DLLs, or (more accurately) it enables Windows' default DLL search behavior, which includes checking `PATH`. 

The change shouldn't affect DLL resolution for other R installations; for those, we continue to use the same strategy we do today. 

Part of https://github.com/posit-dev/positron/issues/4398.

### DIFF
--- a/crates/ark/src/main.rs
+++ b/crates/ark/src/main.rs
@@ -53,21 +53,25 @@ Available options:
 --version                    Print the version of Ark
 --log FILE                   Log to the given file (if not specified, stdout/stderr
                              will be used)
---install                    Install the kernel spec for Ark"#);
+--install                    Install the kernel spec for Ark"#
+    );
 
     // Windows-specific options
     #[cfg(target_os = "windows")]
-    print!(r#"
+    print!(
+        r#"
 --standard-dll-search-order  Use the standard Windows DLL search order (including the PATH environment
                              variable) when loading R. Useful for R installations that depend on DLLs
                              other than those in system folders and R install folder, such as Conda.
-                             The R shared library folder must be on the PATH for this to work."#);
+                             The R shared library folder must be on the PATH for this to work."#
+    );
 
-    print!(r#"
+    print!(
+        r#"
 --help                       Print this help message
 
-"#);
-
+"#
+    );
 }
 
 fn main() -> anyhow::Result<()> {


crates/ark/src/main.rs
```diff
@@ -53,10 +53,21 @@ Available options:
...
---install                Install the kernel spec for Ark
+--install                Install the kernel spec for Ark"#);
+
+    // Windows-specific options
+    #[cfg(target_os = "windows")]
+    print!(r#"
+--dll-search-path        Use the standard Windows DLL search path (including the PATH environment
+                         variable) when loading R. Useful for R installations that depend on DLLs
+                         other than those in system folders and R install folder, such as Conda.
+                         The R shared library folder must be on the PATH for this to work."#);
```
DavisVaughan: Just following the naming from https://learn.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-search-order#standard-search-order-for-unpackaged-apps

```suggestion
--standard-dll-search-order        Use the standard Windows DLL search order (including the PATH environment
                         variable) when loading R. Useful for R installations that depend on DLLs
                         other than those in system folders and R install folder, such as Conda.
                         The R shared library folder must be on the PATH for this to work."#);
```

Will need to be reflowed, but I do think this name is a bit clearer

crates/harp/src/library.rs
```diff
@@ -1,17 +1,41 @@
...
 //
 //
 
 use std::env::consts::DLL_PREFIX;
 use std::env::consts::DLL_SUFFIX;
 use std::path::PathBuf;
+#[cfg(windows)]
+use std::sync::atomic::AtomicBool;
+#[cfg(windows)]
+use std::sync::atomic::Ordering;
```
DavisVaughan: We've tried hard to contain OS specific stuff in `sys/windows/` or `sys/unix/` so we can avoid the cfgs everywhere

See https://github.com/posit-dev/ark/pull/973 for an alternate approach that you can merge into this

crates/ark/src/main.rs
```diff
@@ -53,10 +53,21 @@ Available options:
 --version                Print the version of Ark
 --log FILE               Log to the given file (if not specified, stdout/stderr
                          will be used)
---install                Install the kernel spec for Ark
+--install                Install the kernel spec for Ark"#);
+
+    // Windows-specific options
+    #[cfg(target_os = "windows")]
+    print!(r#"
```
DavisVaughan: When I save in this file, rust-analyzer reformats this a little. Can you please check the formatting first before merging?

jmcphers: Looks like rustfmt was failing silently on my Windows machine. 😭 Fixed.

crates/harp/src/library.rs
jmcphers: thanks, that's better!

PR #974, opened 2025-12-02
Create Rowan parse tree, store document contents as text, and use Biome position encoding converters

lionel-:
Progress towards https://github.com/posit-dev/positron/issues/1766. We'll use the new Rowan infrastructure to inject breakpoints at syntax/text level (rather than in runtime R objects as in RStudio).

Requires https://github.com/posit-dev/air/pull/452

#### Switch to Rowan

This PR ia a first step towards using Rowan in Ark. Rowan is a parser-agnostic reprentation of a CST used in Rust-Analyser, Biome, Ruff, Air, Jarl, and now Ark. Benefits:

- The iterators and tools provided by Rowan are more powerful than those provided by Tree-Sitter.

- The generated typed AST wrappers provide convenient accessors to the R constructs.

- The representation is parser-agnostic. Any functionality written on top of Rowan can be used as is with a different parser than the TS parser we currently use. In particular, we'll be able to use a custom parser in the future that has better error recovery and syntax error reporting.

- Convergence of the Air and Ark infrastructure. Less context switching. Transposability of knowledge and patterns across projects.

- Unlike Tree-Sitter where getting a parent node requires a traversal, Rowan nodes store their parent.

- The CST carries full text representation. Getting the text of a node, token, or refactored tree is a simple method call that doesn't require passing around other state such as file contents.

#### Other changes

This PR also :

- Removes the Rope data structure for representing Document contents in favour of a simple `String`. This is also what Ruff, Biome, and Rust-Analyzer do. This considerably simplifies the gymnastics we had to do to deal with the non-contiguous text representation.

- Switches to biome_line_index (https://github.com/biomejs/biome/tree/main/crates/biome_line_index, originally from Rust-Analyzer) for position encoding conversions.

- Switches to conversion tools exported from Air.

Most of the changes in this PR are related to these two infrastructure changes.

#### In detail

- New dependencies on `aether_parser`, `aether_syntax`, and `aether_lsp_utils`. Aether is our tentative name for Posit's collection of tools related to language analysis for R.

- We store a Rowan parse tree in addition to a Tree-Sitter parse tree. We also create and update a Biome LineIndex that stores

- More consistent handling of offsets, positions, and encoding. Conversions are now operated by Biome's LineIndex data structure.

- Conversion methods involving Tree-Sitter are now gathered as Document methods. They need contents, the line index, and the document's position encoding, all contained in Document.

- These conversion methods now fail with a Result when something goes wrong. This is bubbled up through to LSP handlers.

- In general, more consistent error handling and propagation (either logging with our new `log_err()` method or bubbling up as `Result` until the JSON-RPC boundary).

- `offset.rs` and `encoding.rs` have been deleted in favour of the new imported tools.

### DIFF
--- a/crates/ark/src/lsp/document.rs
+++ b/crates/ark/src/lsp/document.rs
@@ -6,6 +6,7 @@
 //
 
 use aether_lsp_utils::proto::from_proto;
+use aether_lsp_utils::proto::to_proto;
 use aether_lsp_utils::proto::PositionEncoding;
 use tower_lsp::lsp_types;
 use tree_sitter::Parser;
@@ -209,10 +210,8 @@ impl Document {
         &self,
         position: lsp_types::Position,
     ) -> anyhow::Result<tree_sitter::Point> {
-        let offset = from_proto::offset(position, &self.line_index, self.position_encoding)?;
-        let line_col = self.line_index.line_col(offset).ok_or_else(|| {
-            anyhow::anyhow!("Failed to convert LSP position {position:?} to LineCol offset")
-        })?;
+        let line_col =
+            from_proto::line_col_from_position(position, &self.line_index, self.position_encoding);
         Ok(tree_sitter::Point::new(
             line_col.line as usize,
             line_col.col as usize,
@@ -227,22 +226,7 @@ impl Document {
             line: point.row as u32,
             col: point.column as u32,
         };
-
-        match self.position_encoding {
-            PositionEncoding::Utf8 => Ok(lsp_types::Position::new(line_col.line, line_col.col)),
-            PositionEncoding::Wide(wide_encoding) => {
-                let wide_line_col = self
-                    .line_index
-                    .to_wide(wide_encoding, line_col)
-                    .ok_or_else(|| {
-                        anyhow::anyhow!("Failed to convert Tree-Sitter point {point:?} to wide line column for document")
-                    })?;
-                Ok(lsp_types::Position::new(
-                    wide_line_col.line as u32,
-                    wide_line_col.col as u32,
-                ))
-            },
-        }
+        to_proto::position_from_line_col(line_col, &self.line_index, self.position_encoding)
     }
 
     pub fn lsp_range_from_tree_sitter_range(
@@ -261,13 +245,11 @@ impl Document {
         let start_point = self.tree_sitter_point_from_lsp_position(range.start)?;
         let end_point = self.tree_sitter_point_from_lsp_position(range.end)?;
 
-        let start_offset =
-            from_proto::offset(range.start, &self.line_index, self.position_encoding)?;
-        let end_offset = from_proto::offset(range.end, &self.line_index, self.position_encoding)?;
+        let text_range = from_proto::text_range(range, &self.line_index, self.position_encoding)?;
 
         Ok(tree_sitter::Range {
-            start_byte: start_offset.into(),
-            end_byte: end_offset.into(),
+            start_byte: text_range.start().into(),
+            end_byte: text_range.end().into(),
             start_point,
             end_point,
         })
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -209,7 +209,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 [[package]]
 name = "aether_lsp_utils"
 version = "0.0.0"
-source = "git+https://github.com/posit-dev/air?rev=f959e32eee91#f959e32eee91654f04a44a32e97321ef5d510e93"
+source = "git+https://github.com/posit-dev/air?rev=4cbd36d552e27d8930cff1602d56bfd9ce4c1ed1#4cbd36d552e27d8930cff1602d56bfd9ce4c1ed1"
 dependencies = [
  "anyhow",
  "biome_line_index",
@@ -246,7 +246,7 @@ dependencies = [
 [[package]]
 name = "air_r_factory"
 version = "0.0.0"
-source = "git+https://github.com/posit-dev/air?rev=f959e32eee91#f959e32eee91654f04a44a32e97321ef5d510e93"
+source = "git+https://github.com/posit-dev/air?rev=4cbd36d552e27d8930cff1602d56bfd9ce4c1ed1#4cbd36d552e27d8930cff1602d56bfd9ce4c1ed1"
 dependencies = [
  "air_r_syntax",
  "biome_rowan",
@@ -255,7 +255,7 @@ dependencies = [
 [[package]]
 name = "air_r_parser"
 version = "0.0.0"
-source = "git+https://github.com/posit-dev/air?rev=f959e32eee91#f959e32eee91654f04a44a32e97321ef5d510e93"
+source = "git+https://github.com/posit-dev/air?rev=4cbd36d552e27d8930cff1602d56bfd9ce4c1ed1#4cbd36d552e27d8930cff1602d56bfd9ce4c1ed1"
 dependencies = [
  "air_r_factory",
  "air_r_syntax",
@@ -271,7 +271,7 @@ dependencies = [
 [[package]]
 name = "air_r_syntax"
 version = "0.0.0"
-source = "git+https://github.com/posit-dev/air?rev=f959e32eee91#f959e32eee91654f04a44a32e97321ef5d510e93"
+source = "git+https://github.com/posit-dev/air?rev=4cbd36d552e27d8930cff1602d56bfd9ce4c1ed1#4cbd36d552e27d8930cff1602d56bfd9ce4c1ed1"
 dependencies = [
  "biome_rowan",
  "serde",
@@ -472,7 +472,7 @@ checksum = "72b3254f16251a8381aa12e40e3c4d2f0199f8c6508fbecb9d91f575e0fbb8c6"
 [[package]]
 name = "biome_console"
 version = "0.5.7"
-source = "git+https://github.com/biomejs/biome?rev=c13fc60726883781e4530a4437724273b560c8e0#c13fc60726883781e4530a4437724273b560c8e0"
+source = "git+https://github.com/lionel-/biome?rev=41d799cfa4cedd25625fc3f6bd7898532873f051#41d799cfa4cedd25625fc3f6bd7898532873f051"
 dependencies = [
  "biome_markup",
  "biome_text_size",
@@ -485,7 +485,7 @@ dependencies = [
 [[package]]
 name = "biome_diagnostics"
 version = "0.5.7"
-source = "git+https://github.com/biomejs/biome?rev=c13fc60726883781e4530a4437724273b560c8e0#c13fc60726883781e4530a4437724273b560c8e0"
+source = "git+https://github.com/lionel-/biome?rev=41d799cfa4cedd25625fc3f6bd7898532873f051#41d799cfa4cedd25625fc3f6bd7898532873f051"
 dependencies = [
  "backtrace",
  "biome_console",
@@ -505,7 +505,7 @@ dependencies = [
 [[package]]
 name = "biome_diagnostics_categories"
 version = "0.5.7"
-source = "git+https://github.com/biomejs/biome?rev=c13fc60726883781e4530a4437724273b560c8e0#c13fc60726883781e4530a4437724273b560c8e0"
+source = "git+https://github.com/lionel-/biome?rev=41d799cfa4cedd25625fc3f6bd7898532873f051#41d799cfa4cedd25625fc3f6bd7898532873f051"
 dependencies = [
  "quote",
  "serde",
@@ -514,7 +514,7 @@ dependencies = [
 [[package]]
 name = "biome_diagnostics_macros"
 version = "0.5.7"
-source = "git+https://github.com/biomejs/biome?rev=c13fc60726883781e4530a4437724273b560c8e0#c13fc60726883781e4530a4437724273b560c8e0"
+source = "git+https://github.com/lionel-/biome?rev=41d799cfa4cedd25625fc3f6bd7898532873f051#41d799cfa4cedd25625fc3f6bd7898532873f051"
 dependencies = [
  "proc-macro-error2",
  "proc-macro2",
@@ -525,7 +525,7 @@ dependencies = [
 [[package]]
 name = "biome_line_index"
 version = "0.1.0"
-source = "git+https://github.com/biomejs/biome?rev=c13fc60726883781e4530a4437724273b560c8e0#c13fc60726883781e4530a4437724273b560c8e0"
+source = "git+https://github.com/lionel-/biome?rev=41d799cfa4cedd25625fc3f6bd7898532873f051#41d799cfa4cedd25625fc3f6bd7898532873f051"
 dependencies = [
  "biome_text_size",
  "rustc-hash",
@@ -534,7 +534,7 @@ dependencies = [
 [[package]]
 name = "biome_markup"
 version = "0.5.7"
-source = "git+https://github.com/biomejs/biome?rev=c13fc60726883781e4530a4437724273b560c8e0#c13fc60726883781e4530a4437724273b560c8e0"
+source = "git+https://github.com/lionel-/biome?rev=41d799cfa4cedd25625fc3f6bd7898532873f051#41d799cfa4cedd25625fc3f6bd7898532873f051"
 dependencies = [
  "proc-macro-error2",
  "proc-macro2",
@@ -544,7 +544,7 @@ dependencies = [
 [[package]]
 name = "biome_parser"
 version = "0.5.7"
-source = "git+https://github.com/biomejs/biome?rev=c13fc60726883781e4530a4437724273b560c8e0#c13fc60726883781e4530a4437724273b560c8e0"
+source = "git+https://github.com/lionel-/biome?rev=41d799cfa4cedd25625fc3f6bd7898532873f051#41d799cfa4cedd25625fc3f6bd7898532873f051"
 dependencies = [
  "biome_console",
  "biome_diagnostics",
@@ -558,7 +558,7 @@ dependencies = [
 [[package]]
 name = "biome_rowan"
 version = "0.5.7"
-source = "git+https://github.com/biomejs/biome?rev=c13fc60726883781e4530a4437724273b560c8e0#c13fc60726883781e4530a4437724273b560c8e0"
+source = "git+https://github.com/lionel-/biome?rev=41d799cfa4cedd25625fc3f6bd7898532873f051#41d799cfa4cedd25625fc3f6bd7898532873f051"
 dependencies = [
  "biome_text_edit",
  "biome_text_size",
@@ -570,7 +570,7 @@ dependencies = [
 [[package]]
 name = "biome_text_edit"
 version = "0.5.7"
-source = "git+https://github.com/biomejs/biome?rev=c13fc60726883781e4530a4437724273b560c8e0#c13fc60726883781e4530a4437724273b560c8e0"
+source = "git+https://github.com/lionel-/biome?rev=41d799cfa4cedd25625fc3f6bd7898532873f051#41d799cfa4cedd25625fc3f6bd7898532873f051"
 dependencies = [
  "biome_text_size",
  "serde",
@@ -580,7 +580,7 @@ dependencies = [
 [[package]]
 name = "biome_text_size"
 version = "0.5.7"
-source = "git+https://github.com/biomejs/biome?rev=c13fc60726883781e4530a4437724273b560c8e0#c13fc60726883781e4530a4437724273b560c8e0"
+source = "git+https://github.com/lionel-/biome?rev=41d799cfa4cedd25625fc3f6bd7898532873f051#41d799cfa4cedd25625fc3f6bd7898532873f051"
 dependencies = [
  "serde",
 ]
@@ -588,7 +588,7 @@ dependencies = [
 [[package]]
 name = "biome_unicode_table"
 version = "0.5.7"
-source = "git+https://github.com/biomejs/biome?rev=c13fc60726883781e4530a4437724273b560c8e0#c13fc60726883781e4530a4437724273b560c8e0"
+source = "git+https://github.com/lionel-/biome?rev=41d799cfa4cedd25625fc3f6bd7898532873f051#41d799cfa4cedd25625fc3f6bd7898532873f051"
 
 [[package]]
 name = "bitflags"
@@ -1141,7 +1141,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.59.0",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -1713,7 +1713,7 @@ checksum = "3640c1c38b8e4e43584d8df18be5fc6b0aa314ce6ebf51b53313d4306cca8e46"
 dependencies = [
  "hermit-abi 0.5.2",
  "libc",
- "windows-sys 0.59.0",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -1829,7 +1829,7 @@ dependencies = [
 [[package]]
 name = "line_ending"
 version = "0.0.0"
-source = "git+https://github.com/posit-dev/air?rev=f959e32eee91#f959e32eee91654f04a44a32e97321ef5d510e93"
+source = "git+https://github.com/posit-dev/air?rev=4cbd36d552e27d8930cff1602d56bfd9ce4c1ed1#4cbd36d552e27d8930cff1602d56bfd9ce4c1ed1"
 dependencies = [
  "memchr",
  "settings",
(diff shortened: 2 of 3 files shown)


crates/ark/src/lsp/traits/node.rs
```diff
@@ -283,6 +287,17 @@ impl<'tree> NodeExt for Node<'tree> {
...
+    fn node_as_str<'a>(&self, source: &'a str) -> anyhow::Result<&'a str> {
+        self.utf8_text(source.as_bytes()).anyhow()
+    }
+
+    fn node_to_string(&self, source: &str) -> anyhow::Result<String> {
+        Ok(self
+            .node_as_str(source)
+            .map(|s| s.to_string())
+            .map_err(|e| anyhow!(e))?)
+    }
```
DavisVaughan, line 299: I'm not really the biggest fan of methods that hide clones, like how the `to_string()` is hidden here

I guess this was mostly for compatibility with the old system?

In theory most usage of this could probably have switched to `node_as_str()` with a few extra tweaks to the call sites

I think I would have even preferred to see the `to_string()` at the call site rather than here to make it very clear a clone was happening at that call site.

But I see that `node_to_string()` it used a lot, so I don't feel like you need to go back and change them all to that now

crates/ark/Cargo.toml
```diff
@@ -38,16 +42,16 @@ regex = "1.10.0"
...
 serde_json = { version = "1.0.94", features = ["preserve_order"] }
 stdext = { path = "../stdext" }
+streaming-iterator = "0.1.9"
 tokio = { version = "1.26.0", features = ["full"] }
-tower-lsp = "0.19.0"
-tree-sitter = "0.23.0"
-tree-sitter-r = { git = "https://github.com/r-lib/tree-sitter-r", rev = "95aff097aa927a66bb357f715b58cde821be8867" }
+tower-lsp.workspace = true
+tree-sitter = "0.24.7"
+tree-sitter-r = { git = "https://github.com/r-lib/tree-sitter-r", rev = "daa26a2ff0d9546e9125c7d8bcec046027f02070" }
```
DavisVaughan: Can we please keep the exact tree-sitter and tree-sitter-r versions as before?

This actually _downgrades_ tree-sitter-r from where we were previously.

The exact tree-sitter and tree-sitter-r versions are extremely sensitive and typically require a lot of mental thought, and I'd prefer to keep any changes to them isolated to their own PRs if possible.

crates/ark/Cargo.toml
```diff
@@ -16,12 +16,16 @@ amalthea = { path = "../amalthea" }
...
 bus = "2.3.0"
 cfg-if = "1.0.0"
 crossbeam = { version = "0.8.2", features = ["crossbeam-channel"] }
 ctor = "0.1.26"
 dap = { git = "https://github.com/sztomi/dap-rs", branch = "main" }
 dashmap = "5.4.0"
+aether_parser.workspace = true
+aether_syntax.workspace = true
+aether_lsp_utils.workspace = true
 ego-tree = "0.6.2"
```
DavisVaughan: Alpahbetical order please?

crates/ark/src/lsp/hover.rs
```diff
@@ -43,8 +43,8 @@ fn hover_context(node: Node, context: &DocumentContext) -> Result<Option<HoverCo
             return Ok(None);
         }
 
-        let package = context.document.contents.node_slice(&lhs)?.to_string();
-        let topic = context.document.contents.node_slice(&rhs)?.to_string();
+        let package = lhs.node_to_string(context.document.contents.as_str())?;
+        let topic = rhs.node_to_string(context.document.contents.as_str())?;
         return Ok(Some(HoverContext::QualifiedTopic { package, topic }));
```
DavisVaughan, line 48: Like here's a case where I definitely preferred seeing `to_string()` to let me know that I was cloning something at this call site

crates/ark/src/lsp/statement_range.rs
```diff
@@ -129,14 +124,14 @@ fn find_roxygen_statement_range(
...
-fn as_roxygen_comment_text(node: &Node, contents: &Rope) -> Option<String> {
+fn as_roxygen_comment_text(node: &Node, contents: &str) -> Option<String> {
     // Tree sitter doesn't know about the special `#'` marker,
     // but does tell us if we are in a `#` comment
     if !node.is_comment() {
         return None;
     }
 
-    let text = contents.node_slice(node).unwrap().to_string();
+    let text = NodeExt::node_to_string(node, contents).ok()?;
```
DavisVaughan: `node.node_to_string()`?

crates/ark/src/lsp/statement_range.rs
```diff
@@ -176,7 +171,7 @@ fn find_roxygen_examples_section(node: Node, contents: &Rope) -> Option<tree_sit
     // right after that is the `start` node.
     //
     // Note: Cleaner to use `cursor.goto_prev_sibling()` but that seems to have
-    // a bug in it (it gets the `kind()` right, but `utf8_text()` returns off by
+    // a bug in it (it gets the `kind()` right, but `node&()` returns off by
```
DavisVaughan: I think this is wrong?

crates/ark/src/lsp/selection_range.rs
```diff
@@ -105,17 +105,20 @@ fn range_default(node: Node) -> Range {
...
-        let selection = convert_selection_range_from_tree_sitter_to_lsp(*selection, document);
-        Some(Box::new(selection))
-    });
+    let parent = match selection.parent {
+        Some(selection) => {
+            let selection = convert_selection_range_from_tree_sitter_to_lsp(*selection, document)?;
+            Some(Box::new(selection))
+        },
+        None => None,
+    };
```
DavisVaughan, line 119: Interesting how having to bail with the `?` can sometimes make the code slightly less ergonomic, since now you can't use `and_then()`

crates/ark/src/lsp/traits/node.rs
```diff
@@ -268,12 +273,11 @@ impl<'tree> NodeExt for Node<'tree> {
...
-        &self,
-        contents: &ropey::Rope,
-    ) -> impl Iterator<Item = Option<String>> {
-        self.arguments_names().map(|maybe_node| {
-            maybe_node.and_then(|node| match contents.node_slice(&node) {
+    fn arguments_names_as_string(&self, contents: &str) -> impl Iterator<Item = Option<String>> {
+        // Note: capture `contents` by reference into the closure so the iterator
+        // can outlive the stack frame where this method is called.
+        self.arguments_names().map(move |maybe_node| {
+            maybe_node.and_then(|node| match node.node_as_str(&contents) {
```
DavisVaughan:
```suggestion
            maybe_node.and_then(|node| match node.node_as_str(contents) {
```

crates/ark/src/lsp/completions/sources/composite/pipe.rs
```diff
@@ -137,7 +137,7 @@ fn find_pipe_root_name(context: &DocumentContext, node: &Node) -> anyhow::Result
     }
 
     // Try to evaluate the left-hand side
-    let root = context.document.contents.node_slice(&lhs)?.to_string();
+    let root = lhs.node_as_str(&context.document.contents)?.to_string();
```
DavisVaughan, line 140: I definitely prefer the method being on `Node`, that's nice

crates/ark/src/lsp/completions/function_context.rs
```diff
@@ -93,7 +92,7 @@ impl FunctionContext {
             cursor.row == node_range.end_point.row && cursor.column == node_range.end_point.column;
 
         let name = match function_name_node {
-            Some(node) => node_text(&node, &document_context.document.contents).unwrap_or_default(),
+            Some(node) => node.node_to_string(&document_context.document.contents)?,
             None => String::new(),
```
DavisVaughan, line 98: This looks like a change in behavior? Subtle, but maybe keep `unwrap_or_default()`?

crates/ark/src/lsp/completions/function_context.rs
```diff
@@ -174,9 +173,12 @@ fn is_inside_special_function(node: &Node, contents: &ropey::Rope) -> bool {
         return false;
     };
 
-    let call_name = node_text(&call_name_node, contents).unwrap_or_default();
+    let call_name = call_name_node
+        .node_as_str(contents)
+        .log_err()
+        .unwrap_or_default();
```
DavisVaughan, line 181: Yea that `log_err()` pattern is cool. I like the `Option` output.

crates/ark/src/lsp/references.rs
```diff
@@ -75,23 +73,29 @@ struct Context {
...
 }
 
-fn found_match(node: &Node, contents: &Rope, context: &Context) -> bool {
+fn found_match(node: &Node, contents: &str, context: &Context) -> bool {
     if !node.is_identifier() {
         return false;
     }
 
-    let symbol = contents.node_slice(node).unwrap().to_string();
+    let symbol = NodeExt::node_to_string(node, contents).unwrap();
```
DavisVaughan: `node.node_to_string()`?

crates/ark/src/lsp/handlers.rs
```diff
@@ -284,46 +282,35 @@ pub(crate) fn handle_goto_definition(
...
 
     // Get tree-sitter points to return selection ranges for
-    let points: Vec<Point> = params
+    let points: anyhow::Result<Vec<Point>> = params
         .positions
         .into_iter()
-        .map(|position| convert_position_to_point(&document.contents, position))
+        .map(|position| document.tree_sitter_point_from_lsp_position(position))
         .collect();
+    let points = points?;
```
```suggestion
    let points = params
        .positions
        .into_iter()
        .map(|position| document.tree_sitter_point_from_lsp_position(position))
        .collect::<anyhow::Result<Vec<_>>>()?;
```

Just because you use this exact pattern below, so we can make it symmetric

crates/ark/src/lsp/declarations.rs
```diff
@@ -41,7 +38,7 @@ pub(crate) fn top_level_declare(
     let Some(enable) = iter.find_map(|n| node_arg_value(&n, "enable", contents)) else {
         return decls;
     };
-    let Some(enable_text) = node_text(&enable, contents) else {
+    let Ok(enable_text) = enable.node_as_str(contents) else {
         return decls;
     };
```
DavisVaughan, line 44: Maybe `log_err()`?

crates/ark/src/lsp/symbols.rs
```diff
@@ -392,7 +389,7 @@ where
 
     for child in node.children(&mut cursor) {
         if let NodeType::Comment = child.node_type() {
-            let comment_text = contents.node_slice(&child)?.to_string();
+            let comment_text = child.node_to_string(&doc.contents)?;
```
DavisVaughan: Make it `node_to_str()`? Seems easy here

crates/ark/src/lsp/symbols.rs
```diff
@@ -461,119 +458,98 @@ where
...
     symbols: &mut Vec<DocumentSymbol>,
 ) -> anyhow::Result<()> {
     let Some(callee) = node.child_by_field_name("function") else {
         return Ok(());
     };
 
     if callee.is_identifier() {
-        let fun_symbol = contents.node_slice(&callee)?.to_string();
+        let fun_symbol = callee.node_to_string(&doc.contents)?;
         match fun_symbol.as_str() {
```
DavisVaughan: Definitely `node_to_str()` here

See I think this is proving my gut feeling about why we should only expose `node_to_str()` and force the caller to do `to_string()`. I think it would encourage better code with less clones.

crates/ark/src/lsp/symbols.rs
```diff
@@ -614,15 +590,15 @@ fn collect_call_test_that(
     let mut cursor = arguments.walk();
     for child in arguments.children_by_field_name("argument", &mut cursor) {
         if let Some(value) = child.child_by_field_name("value") {
-            collect_symbols(ctx, &value, contents, &mut children)?;
+            collect_symbols(ctx, &value, doc, &mut children)?;
         }
     }
 
-    let name = contents.node_slice(&string)?.to_string();
+    let name = string.node_to_string(&doc.contents)?;
```
DavisVaughan: `node_to_str()`

crates/ark/src/treesitter.rs
```diff
@@ -285,7 +286,7 @@ pub trait NodeTypeExt: Sized {
     fn is_identifier(&self) -> bool;
     fn is_string(&self) -> bool;
     fn is_identifier_or_string(&self) -> bool;
-    fn get_identifier_or_string_text(&self, contents: &ropey::Rope) -> anyhow::Result<String>;
+    fn get_identifier_or_string_text(&self, contents: &str) -> anyhow::Result<String>;
```
DavisVaughan: Can we make this return a `Result<&str>`? The one place where we use it looks like it can take a `str`, and we can use `node_as_str()` instead.

crates/ark/src/treesitter.rs
```diff
@@ -625,15 +624,19 @@ pub(crate) fn node_find_containing_call<'tree>(node: Node<'tree>) -> Option<Node
...
-        let line = contents.line(prev_row as usize);
-        let line_len = line.len_chars().saturating_sub(1); // Subtract 1 for newline
-        tree_sitter::Point {
-            row: prev_row,
-            column: line_len,
+        if let Some(line) = contents.lines().nth(prev_row) {
+            let line_len = line.chars().count();
+            tree_sitter::Point {
+                row: prev_row,
+                column: line_len,
```
DavisVaughan, line 633: So was this a careful analysis that determined that Rope was different from str?

crates/ark/src/lsp/indexer.rs
```diff
@@ -427,14 +417,10 @@ fn index_r6_class_methods(
...
-    let contents_str = contents.to_string();
-
-    for method_node in ts_query.captures_for(*node, "method_name", contents_str.as_bytes()) {
-        let name = contents.node_slice(&method_node)?.to_string();
-        let start = convert_point_to_position(contents, method_node.start_position());
-        let end = convert_point_to_position(contents, method_node.end_position());
+    for method_node in ts_query.captures_for(*node, "method_name", doc.contents.as_bytes()) {
+        let name = method_node.node_to_string(&doc.contents)?;
+        let start = doc.lsp_position_from_tree_sitter_point(method_node.start_position())?;
+        let end = doc.lsp_position_from_tree_sitter_point(method_node.end_position())?;
```
DavisVaughan, line 424: Nice

crates/ark/src/lsp/indent.rs
```diff
@@ -21,13 +19,17 @@ use crate::treesitter::NodeTypeExt;
...
     let text = &doc.contents;
     let ast = &doc.ast;
     let config = &doc.config.indent;
 
-    // Rope counts from 1, we count from 0
-    if line >= text.len_lines() {
+    let line_count = if text.is_empty() {
+        1
+    } else {
+        text.chars().filter(|c| *c == '\n').count() + 1
```
DavisVaughan, line 31: `text.lines().count()`?

crates/ark/src/lsp/document.rs
```diff
@@ -0,0 +1,458 @@
...
+            .set_language(&tree_sitter_r::LANGUAGE.into())
+            .unwrap();
+
+        Self::new_with_parser(contents, &mut parser, version)
+    }
+
+    pub fn new_with_parser(contents: &str, parser: &mut Parser, version: Option<i32>) -> Self {
+        let contents = String::from(contents);
+        let ast = parser.parse(contents.as_str(), None).unwrap();
+        let parse = aether_parser::parse(&contents, Default::default());
```
DavisVaughan, line 73: Do you want to add some notes about how we prefer to use `parse` over `ast` now? Or some kind of legacy comments up in the `Document` struct for `ast`? Something would be nice.

crates/ark/src/lsp/document.rs
```diff
@@ -0,0 +1,458 @@
...
+    fn update(
+        &mut self,
+        parser: &mut Parser,
+        change: &lsp_types::TextDocumentContentChangeEvent,
+    ) -> Result<()> {
+        // Extract edit range. Return without doing anything if there wasn't any actual edit.
+        let range = match change.range {
+            Some(r) => r,
+            None => return Ok(()),
+        };
```
DavisVaughan: I think this is very wrong?

If `range` is missing then `change.text` is considered a full document replacement, not a lack of edit.

crates/ark/src/lsp/document.rs
```diff
@@ -0,0 +1,458 @@
...
+use anyhow::Result;
+use tower_lsp::lsp_types;
+use tree_sitter::InputEdit;
+use tree_sitter::Parser;
+use tree_sitter::Point;
+use tree_sitter::Tree;
+
+use crate::lsp::config::DocumentConfig;
+
+fn compute_point(point: Point, text: &str) -> Point {
```
DavisVaughan: I don't know what this is doing, comments?

crates/ark/src/lsp/document.rs
```diff
@@ -0,0 +1,458 @@
...
+        let ast = parser.parse(self.contents.as_str(), Some(&self.ast));
+        self.ast = ast.unwrap();
+
+        // Update the Rowan tree. This currently reparses with TS. We could pass
+        // down the TS tree if that turns out too expensive, but the long term
+        // plan is to remove any TS usage so we prefer not introduce TS trees in
+        // the public APIs.
+        self.parse = aether_parser::parse(&self.contents, Default::default());
+
+        Ok(())
```
DavisVaughan: I think this is actually the toughest part of the whole PR

I'd like to talk with you about dropping all of this incremental edit messiness and instead mirroring Air's `lsp::documents::Document::on_did_change()` extremely closely.

IIUC the only thing extra that we would need is a `self.ast =` line at the end that reparses a fresh tree. And even that would be temporary as we move away from tree-sitter in favor of `self.parse`.

---

I think the Air version is actually even more efficient than this is in some scenarios because it handles the case of getting multiple `changes` better. When this current Ark implementation gets multiple `changes`, every individual `change` gets a freshly created Rowan tree. When the Air impl gets multiple `changes`, a new Rowan tree is only built once at the end after applying all changes.

I just think that this would get dramatically simpler if we lean into Air's `on_did_change()` impl and the `aether_lsp_utils::from_proto::apply_text_changes()` helper that is so nicely optimized for this already.

crates/ark/src/lsp/document.rs
```diff
@@ -0,0 +1,458 @@
...
+                    .ok_or_else(|| {
+                        anyhow::anyhow!("Failed to convert Tree-Sitter point {point:?} to wide line column for document")
+                    })?;
+                Ok(lsp_types::Position::new(
+                    wide_line_col.line as u32,
+                    wide_line_col.col as u32,
+                ))
+            },
+        }
+    }
```
DavisVaughan, line 230:
```suggestion
    pub fn tree_sitter_point_from_lsp_position(
        &self,
        position: lsp_types::Position,
    ) -> anyhow::Result<tree_sitter::Point> {
        let line_col = from_proto::line_col_from_position(position, &self.line_index, self.position_encoding)?;
        Ok(tree_sitter::Point::new(
            line_col.line as usize,
            line_col.col as usize,
        ))
    }

    pub fn lsp_position_from_tree_sitter_point(
        &self,
        point: tree_sitter::Point,
    ) -> anyhow::Result<lsp_types::Position> {
        let line_col = biome_line_index::LineCol {
            line: point.row as u32,
            col: point.column as u32,
        };
        to_proto::position_from_line_col(line_col, &self.line_index, self.position_encoding)
    }
```

You'll need to merge in https://github.com/posit-dev/air/pull/453, but then these helpers should simplify to this Suggestion.

I was irked that we had to dig into `line_index`'s specific methods here, that felt wrong to me.

I realized that in Air we aren't fully exposing the full set of `LineCol/Offset <-> Position` conversions you can do, and we are feeling that pain here.

In particular, we should not need to go through `offset` in ANY way here to make these conversions. It is currently doing way too much work since we aren't exposing all the conversion methods.

It feels very symmetric and pretty with this Air PR.

crates/ark/src/lsp/document.rs
```diff
@@ -0,0 +1,458 @@
...
+            from_proto::offset(range.start, &self.line_index, self.position_encoding)?;
+        let end_offset = from_proto::offset(range.end, &self.line_index, self.position_encoding)?;
+
+        Ok(tree_sitter::Range {
+            start_byte: start_offset.into(),
+            end_byte: end_offset.into(),
+            start_point,
+            end_point,
+        })
+    }
```
DavisVaughan, line 256:
```suggestion
    pub fn tree_sitter_range_from_lsp_range(
        &self,
        range: lsp_types::Range,
    ) -> anyhow::Result<tree_sitter::Range> {
        let start_line_col = from_proto::line_col_from_position(
            range.start,
            &self.line_index,
            self.position_encoding,
        )?;
        let end_line_col = from_proto::line_col_from_position(
            range.end,
            &self.line_index,
            self.position_encoding,
        )?;

        let start_point =
            tree_sitter::Point::new(start_line_col.line as usize, start_line_col.col as usize);
        let end_point =
            tree_sitter::Point::new(end_line_col.line as usize, end_line_col.col as usize);

        let start_offset = self
            .line_index
            .offset(start_line_col)
            .with_context(|| format!("Position {start:?} is out of range", start = range.start))?;
        let end_offset = self
            .line_index
            .offset(end_line_col)
            .with_context(|| format!("Position {end:?} is out of range", end = range.end))?;

        Ok(tree_sitter::Range {
            start_byte: start_offset.into(),
            end_byte: end_offset.into(),
            start_point,
            end_point,
        })
    }
```

If you add the `line_col_from_position()` method discussed above, then this becomes the most efficient way to get here since `from_proto::offset()` would compute the `line_col` again.

But it would also look quite nice (at the cost of some efficiency) as:

```rust
    pub fn tree_sitter_range_from_lsp_range(
        &self,
        range: lsp_types::Range,
    ) -> anyhow::Result<tree_sitter::Range> {
        let start_line_col = from_proto::line_col_from_position(
            range.start,
            &self.line_index,
            self.position_encoding,
        )?;
        let end_line_col = from_proto::line_col_from_position(
            range.end,
            &self.line_index,
            self.position_encoding,
        )?;

        let start_point =
            tree_sitter::Point::new(start_line_col.line as usize, start_line_col.col as usize);
        let end_point =
            tree_sitter::Point::new(end_line_col.line as usize, end_line_col.col as usize);

        let start_offset = from_proto::offset_from_position(
            range.start,
            &self.line_index,
            self.position_encoding,
        )?;
        let end_offset = from_proto::offset_from_position(
            range.end,
            &self.line_index,
            self.position_encoding,
        )?;

        Ok(tree_sitter::Range {
            start_byte: start_offset.into(),
            end_byte: end_offset.into(),
            start_point,
            end_point,
        })
    }
```

crates/ark/src/lsp/hover.rs
lionel-: I don't really follow because the suffix is the same and has the same connotation of cloning.

crates/ark/Cargo.toml
lionel-: Is it alright to keep the same version of tree-sitter-r but upgrade tree-sitter to 0.24.7? Otherwise I get:

```
error: failed to select a version for `tree-sitter`.
    ... required by package `ark v0.1.222 (/Users/lionel/Sync/Projects/Positron/ark/crates/ark)`
versions that meet the requirements `^0.23.0` are: 0.23.2, 0.23.0

package `tree-sitter` links to the native library `tree-sitter`, but it conflicts with a previous package which links to `tree-sitter` as well:
package `tree-sitter v0.24.7`
    ... which satisfies dependency `tree-sitter = "^0.24.7"` of package `air_r_parser v0.0.0 (https://github.com/posit-dev/air?rev=f959e32eee91#f959e32e)`
    ... which satisfies git dependency `aether_parser` (locked to 0.0.0) of package `ark v0.1.222 (/Users/lionel/Sync/Projects/Positron/ark/crates/ark)`
Only one package in the dependency graph may specify the same links value. This helps ensure that only one copy of a native library is linked in the final binary. Try to adjust your dependencies so that only one package uses the `links = "tree-sitter"` value. For more information, see https://doc.rust-lang.org/cargo/reference/resolver.html#links.
```

crates/ark/src/lsp/symbols.rs
lionel-: I see your point now!

crates/ark/src/lsp/indent.rs
lionel-: I've added this comment:

```rs
        // Note that `lines().count()` doesn't count trailing newlines
```

crates/ark/src/treesitter.rs
lionel-: Rope lines include the `\n` character

crates/ark/src/lsp/document.rs
lionel-: oops!!

crates/ark/src/lsp/document.rs
lionel-: That's from this commit: import amalthea source

crates/ark/src/lsp/document.rs
lionel-: The Air version is actually Rust-Analyzer's version. It's nice to use it here too.

crates/ark/src/lsp/document.rs
lionel-: Much nicer thanks!

PR #977, opened 2025-12-02
Set `.Platform$GUI` to "Positron" in console sessions only

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/9933

With this change, I see results like this in the Positron console:

```
> .Platform
$OS.type
[1] "unix"

$file.sep
[1] "/"

$dynlib.ext
[1] ".so"

$GUI
[1] "Positron"

$endian
[1] "little"

$pkgType
[1] "mac.binary.big-sur-arm64"

$path.sep
[1] ":"

$r_arch
[1] ""
```

And results like this in an R session running in an integrated terminal in Positron (the same as RStudio):

```
> .Platform
$OS.type
[1] "unix"

$file.sep
[1] "/"

$dynlib.ext
[1] ".so"

$GUI
[1] "X11"

$endian
[1] "little"

$pkgType
[1] "mac.binary.big-sur-arm64"

$path.sep
[1] ":"

$r_arch
[1] ""
```

Does `SessionMode::Console` really just mean only in Positron? Is there some better way for ark to know when it's running in Positron at initialization?

### DIFF
--- a/crates/ark/src/modules/positron/positron.R
+++ b/crates/ark/src/modules/positron/positron.R
@@ -5,8 +5,8 @@
 #
 #
 
-.Platform <- base::.Platform
-.Platform$GUI <- "Positron"
 if (Sys.getenv("POSITRON") == 1) {
+    .Platform <- base::.Platform
+    .Platform$GUI <- "Positron"
     env_bind_force(baseenv(), ".Platform", .Platform)
 }


crates/ark/src/interface.rs
```diff
@@ -620,6 +620,19 @@ impl RMain {
...
+                    unlockBinding(".Platform", baseenv())
+                    .Platform$GUI <- "Positron"
+                    lockBinding(".Platform", baseenv())
+                "#);
+                match result {
+                    Ok(_) => log::info!("Set .Platform$GUI to 'Positron' for console session"),
+                    Err(err) => log::warn!("Failed to set .Platform$GUI: {err:?}"),
+                }
+            }
+
```
juliasilge: I'm not sure what the best approach is, but this one is at least easy to reason about.

Some other approaches I tried out included:

- Do something like `r_symbol!(".Platform")` in Rust and find the `GUI` element. I like that this doesn't involve executing R code during some initialization step, but you still have to lock/unlock (just at the C level).
- Do the locking/unlocking in `startup.R`. This feels like a better place to run such R code, but I'm not sure we can check session mode there and the other options there are more user-facing.

RStudio does it here, which is not exactly the same but similar in that in happens post-initialization:
https://github.com/rstudio/rstudio/blob/5381a9a4363eae00007fd170302fa8a5c93050b5/src/cpp/r/R/Options.R#L126

lionel-: I think I'd prefer this to be in an R file, maybe in `modules/positron/positron.R`. Then you can use:

```r
.Platform <- base::.Platform
.Platform$GUI <- "Positron"
env_bind_force(baseenv(), ".Platform", .Platform)
```

crates/ark/src/modules/positron/positron.R
```diff
@@ -0,0 +1,10 @@
+#
+# positron.R
+#
+# Copyright (C) 2025 Posit Software, PBC. All rights reserved.
+#
+#
+
+.Platform <- base::.Platform
+.Platform$GUI <- "Positron"
```
lionel-: Can we make it "Ark" if `Sys.getenv("POSITRON")` is not set? Or leave the default.

juliasilge: Just so I understand, files in `modules/positron` get evaluated outside of Positron? All the options in `options.R`, etc? Can you explain a bit more about this?

jennybc: I think #588 is about this.

lionel-: > Just so I understand, files in modules/positron get evaluated outside of Positron? All the options in options.R, etc? Can you explain a bit more about this?

They do because some of the files in there provide essential Ark functionality, for instance regarding error handling.

Maybe the folder really should be called `ark`? Or maybe we should split `ark` from the `positron` folder and the latter would only get sourced in Positron? That would be a nice way to easily differentiate generic and positron-specific functionality.

We could then have a `.ark.` prefix in addition to the `.ps.` prefix, for symbols exported from the `ark` folder.

juliasilge: I went ahead and checked the env var in 1b115682d44d04814eb6c267f3006be31e43ce37, ahead of improvements like that.

crates/ark/src/modules/positron/positron.R
```diff
@@ -0,0 +1,12 @@
...
+if (Sys.getenv("POSITRON") == 1) {
+    env_bind_force(baseenv(), ".Platform", .Platform)
+}
```
lionel-: Otherwise you'll leave a potentially mismatched `.Platform` in the positron namespace that might be confusing when debugging the environment or looking up `.Platform` from a helper defined there.

```suggestion
if (Sys.getenv("POSITRON") == 1) {
    .Platform <- base::.Platform
    .Platform$GUI <- "Positron"
    env_bind_force(baseenv(), ".Platform", .Platform)
}
```

lionel- (approved):

PR #980, opened 2025-12-05
Add a built-in for setting the PPM repo in a running session

atheriel:
This commit adds a `.ps.set_ppm_repo()` built-in that works the same way as the existing CLI option, but allows the repo to be set at runtime.

It is intended to support making the `positron.r.packageManagerRepository` setting apply without needing to restart/recreate existing R sessions.

Part of https://github.com/posit-dev/positron/issues/10965.

### DIFF
--- a/crates/ark/src/modules/positron/repos.R
+++ b/crates/ark/src/modules/positron/repos.R
@@ -19,10 +19,13 @@ apply_repo_defaults <- function(
         repos <- defaults
     } else {
         if ("CRAN" %in% names(repos) && "CRAN" %in% names(defaults)) {
-            # If a CRAN repository is set to @CRAN@, and a default provides a
-            # specific URL, override it. This is the only instance in which we
-            # replace an already-set repository option with a default.
-            if (identical(repos[["CRAN"]], "@CRAN@")) {
+            # If a CRAN repository is set to @CRAN@ or is marked as having been
+            # updated by "the IDE" *and* a default provides a specific URL,
+            # override it.
+            #
+            # This is the only instance in which we replace an already-set
+            # repository option with a default.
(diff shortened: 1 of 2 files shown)


crates/ark/src/modules/positron/repos.R
```diff
@@ -39,3 +39,34 @@ apply_repo_defaults <- function(
...
+#' it or when it uses placeholder `"@CRAN@"`.
+#'
+#' @param url A PPM repository URL. Must be in the form
+#'   `"https://host/repo/snapshot"`, e.g.,
+#'   `"https://packagemanager.posit.co/cran/latest"`.
+#'
+#' @return `NULL`, invisibly.
+#'
+#' @export
+.ps.setPPMRepo <- function(url) {
```
lionel-:
```suggestion
.ps.set_ppm_repo <- function(url) {
```

crates/ark/src/modules/positron/repos.R
```diff
@@ -39,3 +39,34 @@ apply_repo_defaults <- function(
...
+    # Use Ark's built-in PPM binary URL detection.
+    url <- .ps.Call("ps_get_ppm_binary_url", url)
+
+    # If Ark has already set the repos option, it should be safe to do so again.
+    repos <- getOption("repos")
+    if (isTRUE(attr(repos, "IDE"))) {
+        repos[["CRAN"]] <- url
```
lionel-: Suggestion: Maybe add that updating logic to `apply_repo_defaults()`? E.g. we update `CRAN` if set to `"@CRAN@"` or if the IDE attribute is set.

atheriel: Done!

jmcphers:
@atheriel confirmed this is OK to merge.

PR #981, opened 2025-12-08
Annotate execute request source code with code location

lionel-:
Progress towards https://github.com/posit-dev/positron/issues/1766
Ark side of https://github.com/posit-dev/positron/pull/10815
Branched on #974 

This adds a `#line` directive to parsed code to map srcrefs to the file and line origin.
The code is also mapped to the correct starting column by padding the code with leading whitespace.

With these changes, srcrefs point to the editor from which the code was executed:

https://github.com/user-attachments/assets/dfc87f4c-dc09-4755-a8dc-14e8430341b7

The `#line` directive and padding whitespace are added via Rowan (#974). They could also just be prepended to the code textually, but we'll have to use Rowan for breakpoint injection so I thought we'd get our feet wet with it.

Edit: Simplified to a simple string concatenation in the next PR! Still interesting to see the Rowan trivia pattern here (somewhat complex).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -213,7 +213,7 @@ source = "git+https://github.com/posit-dev/air?rev=4cbd36d552e27d8930cff1602d56b
 dependencies = [
  "anyhow",
  "biome_line_index",
- "biome_text_size 0.5.7 (git+https://github.com/lionel-/biome?rev=41d799cfa4cedd25625fc3f6bd7898532873f051)",
+ "biome_text_size",
  "dissimilar",
  "itertools 0.13.0",
  "line_ending",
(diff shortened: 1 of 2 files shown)


crates/amalthea/src/wire/execute_request.rs
```diff
@@ -33,6 +35,166 @@ pub struct ExecuteRequest {
...
+    pub uri: Url,
+    pub start: Position,
+    pub end: Position,
+}
+
+/// `character` in UTF-8 offset
+#[derive(Debug, Clone)]
+pub struct Position {
+    pub line: u32,
+    pub character: usize,
```
lionel-: should be u32

crates/ark/src/console_debug.rs
DavisVaughan, line 1: Change `repl_debug.rs` to `console_debug.rs` in the comment header in the file

PR #985, opened 2025-12-10
Code-sign ark binaries on Windows

jmcphers:
Signs `ark.exe` on Windows. 

Part of https://github.com/posit-dev/positron/issues/9962.

### DIFF
--- a/.github/workflows/release-windows.yml
+++ b/.github/workflows/release-windows.yml
@@ -18,7 +18,6 @@ jobs:
 
         env:
             GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-            DEBUG_FLAG: ${{ matrix.flavor == 'debug' && '-debug' || '' }}
 
         strategy:
             matrix:


.github/workflows/release-windows.yml
```diff
@@ -45,17 +45,60 @@ jobs:
...
-            - name: Create archive
+            - name: Upload unsigned executable for signing
+              uses: actions/upload-artifact@v4
+              with:
+                  name: ark-${{ matrix.flavor }}-windows-${{ matrix.arch }}-unsigned
+                  path: target\${{ matrix.rust_target_prefix }}-pc-windows-msvc\${{ matrix.flavor }}\ark.exe
+
+    sign_windows:
+        name: "Sign Windows Binaries"
+        uses: posit-dev/posit-gh-actions/.github/workflows/sign-windows.yml@main
```
DavisVaughan, line 55: Is `[windows-latest-8x]` a custom machine of ours?
https://github.com/posit-dev/posit-gh-actions/blob/1a1aba8696f8a20b4061d4e4ca7614fcb93b368b/.github/workflows/sign-windows.yml#L19

.github/workflows/release-windows.yml
```diff
@@ -45,17 +45,60 @@ jobs:
...
+        needs: [build_windows]
+        secrets: inherit
+        strategy:
+            matrix:
+                arch: [x64, arm64]
```
DavisVaughan, line 60: I assume that whatever arch `[windows-latest-8x]` is can codesign both `x64` and `arm64` with no issues?

.github/workflows/release-windows.yml
```diff
@@ -45,17 +45,60 @@ jobs:
                   cargo clean
                   cargo build ${{ matrix.flavor == 'release' && '--release' || '' }} --target ${{ matrix.rust_target_prefix }}-pc-windows-msvc
 
-            - name: Create archive
+            - name: Upload unsigned executable for signing
```
DavisVaughan, line 47: I think you can remove `DEBUG_FLAG` from `build_windows:`?

.github/workflows/release-windows.yml
jmcphers: Kind of! It's one of our "Posit Larger Runners". IIRC it is not hosted by Posit but we do pay for it via Github Enterprise.

.github/workflows/release-windows.yml
jmcphers: Good question; yes, it can. It's been code signing the arm64 version of Positron for a few months now.

.github/workflows/release-windows.yml
jmcphers: yes! https://github.com/posit-dev/ark/pull/985/commits/3156b9737961ade19b60c639771176626743f0a6

PR #987, opened 2025-12-11
Add R backend-controlled format options to data explorer

kv9898:
Addresses https://github.com/posit-dev/positron/issues/4818, in tandem with https://github.com/posit-dev/positron/pull/11051.

The combined end result looks like (using the code example in the original [issue](https://github.com/posit-dev/positron/issues/4818)):
Image

This pull request introduces frontend formatting options for the data explorer backend and ensures that these options are provided in backend state replies. The changes also include the logic to calculate formatting options based on the current R session settings.

**Frontend formatting options integration:**

* Added a new `format_options` field of type `Option<FormatOptions>` to the `BackendState` struct in `crates/amalthea/src/comm/data_explorer_comm.rs` to support frontend display customization.
* When replying with backend state in `RDataExplorer::get_state`, the code now includes current formatting options by calling a new helper method.

**Calculation of formatting options:**

* Introduced the `current_format_options` method in `RDataExplorer` to compute formatting options based on the R session's `scipen` option, determining integral digit limits and other display parameters.
* Imported `get_option` from `harp` to retrieve R session options necessary for formatting calculations.

### DIFF
--- a/.github/workflows/cla.yml
+++ b/.github/workflows/cla.yml
@@ -49,7 +49,7 @@ jobs:
             before we accept your contribution. You can sign the CLA by posting
             a comment on this PR saying:
 
-          allowlist: DavisVaughan, dependabot[bot], dfalbel, isabelizimm, jonvanausdeln, lionel-, nstrayer, petetronic, positron-bot[bot], seeM, sharon-wang, softwarenerd, timtmok, wesm, copilot[bot], copilot-swe-agent[bot], Copilot
+          allowlist: DavisVaughan, dependabot[bot], dfalbel, isabelizimm, jonvanausdeln, lionel-, nstrayer, petetronic, positron-bot[bot], seeM, sharon-wang, softwarenerd, timtmok, wesm, Copilot
 
           # the followings are the optional inputs - If the optional inputs are not given, then default values will be taken
           #remote-organization-name: enter the remote organization name where the signatures should be stored (Default is storing the signatures in the same repository)


juliasilge:
Thank you for this PR, @kv9898! This looks like a good approach to me, and I have asked @dfalbel to do a detailed review.

crates/ark/src/data_explorer/r_data_explorer.rs
```diff
@@ -1299,6 +1301,18 @@ impl RDataExplorer {
         // Call the conversion function with resolved sort keys
         convert_to_code::convert_to_code(params, object_name, &resolved_sort_keys)
     }
+
+    fn current_format_options() -> FormatOptions {
+        let scipen: i64 = get_option("scipen").try_into().unwrap_or(0);
+        let max_integral_digits = (7 + scipen).clamp(1, 20);
```
dfalbel: Can you maybe add a comment on why using this heuristic? I understand the R's `scipen` doesn't translate directly to a `max_integral_digits`, and this heuristic seems to work, but I'd like to understand the reasoning behind this.

Also clamping to 20 seems arbitrary, as users can set `options(scipen = 999)`, and get `.Machine$double.xmax` to display without scientific notation.

dfalbel (changes requested):
Thank you @kv9898 !

We should add a test in https://github.com/posit-dev/ark/blob/f0a0e689892e8a45c2b3b2ef9978d8c805e8c804/crates/ark/tests/data_explorer.rs checking if the state is changed after changing that option.

You can use eg: https://github.com/kv9898/ark/blob/9c14a03e3e517e9b3534a2a9a6fcc428a7b707ab/crates/ark/tests/data_explorer.rs#L826

crates/ark/src/data_explorer/r_data_explorer.rs
```diff
@@ -1299,6 +1301,18 @@ impl RDataExplorer {
...
+        FormatOptions {
+            large_num_digits: 2,
+            small_num_digits: 4,
```
dfalbel: Shouldn't scipen also affect the `small_num_digits`? We use it to determine when to use scientific notation for small numbers too: 

https://github.com/kv9898/ark/blob/9c14a03e3e517e9b3534a2a9a6fcc428a7b707ab/crates/ark/src/data_explorer/format.rs#L293-L296

crates/ark/src/data_explorer/r_data_explorer.rs
kv9898: That makes sense. I have removed the upper limits and rewrote the calculation logic. See my comment below

crates/ark/src/data_explorer/r_data_explorer.rs
kv9898: Yes. I have rewritten the logic

kv9898:
> Thank you @kv9898 !
> 
> We should add a test in https://github.com/posit-dev/ark/blob/f0a0e689892e8a45c2b3b2ef9978d8c805e8c804/crates/ark/tests/data_explorer.rs checking if the state is changed after changing that option.
> 
> You can use eg: https://github.com/kv9898/ark/blob/9c14a03e3e517e9b3534a2a9a6fcc428a7b707ab/crates/ark/tests/data_explorer.rs#L826

The test has been added in https://github.com/posit-dev/ark/pull/987/commits/95697792ff3353369826697d260242eadfa80708!

crates/ark/tests/data_explorer.rs
```diff
@@ -3108,3 +3108,63 @@ fn test_single_row_data_frame_column_profiles() {
...
+    
+    // Get the state again and verify format_options changed
+    TestAssertions::assert_state(&socket, |state| {
+        let format_options = state.format_options.as_ref().expect("format_options should be present");
+        
+        // With scipen=10, digits=10:
+        // max_integral_digits = (10 + 5).max(1) = 15
+        // large_num_digits = (10 - 5).max(0) = 5
+        // small_num_digits = (10 + 6).max(1) = 16
+        assert_eq!(format_options.max_integral_digits, 15);
```
dfalbel, line 3160: Looks like tests are failing with: ```    thread 'test_format_options_state_change' (32147) panicked at crates/ark/tests/data_explorer.rs:3160:9:
    assertion `left == right` failed
      left: 5
     right: 15```

Any idea @kv9898 ?

dfalbel (changes requested):

crates/ark/tests/data_explorer.rs
dfalbel: It only fails for R 4.2 🤔

dfalbel: IDK, this might be related: https://github.com/wch/r-source/commit/7f20c19886fd6412453521fb6954febb979ed034

There's something special about scipen. Because digits are being correctly evaluated. Perhaps we can't use get_option for this, and instead we bay need to evaluate `getOption` from the R interpreter.

kv9898: > It only fails for R 4.2 🤔

Thx for checking that out. Because the test worked for me with R 4.5.2:
```bash
E:\ark  [scipen ≡] > cargo test -p ark --test data_explorer
   Compiling ark v0.1.220 (E:\ark\crates\ark)
warning: unused import: `std::time::Duration`
 --> crates\ark\src\fixtures\dummy_frontend.rs:7:5
  |
7 | use std::time::Duration;
  |     ^^^^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(unused_imports)]` (part of `#[warn(unused)]`) on by default

warning: `ark` (lib) generated 1 warning (run `cargo fix --lib -p ark` to apply 1 suggestion)
    Finished `test` profile [unoptimized + debuginfo] target(s) in 14.20s
     Running tests\data_explorer.rs (target\debug\deps\data_explorer-27782ce78c282663.exe)

running 40 tests
test test_export_with_sort_order ... ok
test test_basic_mtcars ... ok
test test_column_labels_edge_cases ... ok
test test_column_labels ... ok
test test_column_labels_haven_compatibility ... ok
test test_column_labels_missing ... ok
test test_data_explorer_special_values ... ok
test test_data_table_support ... ok
test test_data_update_num_rows ... ok
test test_empty_data_frame_column_profiles ... ok
test test_empty_data_frame_data_values ... ok
test test_empty_data_frame_schema ... ok
test test_export_data ... ok
test test_format_options_state_change ... ok
test test_empty_data_frame_state ... ok
test test_frequency_table ... ok
test test_get_data_values_by_indices ... ok
test test_histogram ... ok
test test_histogram_single_bin_same_values ... ok
test test_invalid_filters_preserved ... ok
test test_live_updates ... ok
test test_matrix_support ... ok
test test_null_counts ... ok
test test_row_names_matrix ... ok
test test_schema_identification ... ok
test test_search_filters ... ok
test test_search_schema_combined_filters ... ok
test test_search_schema_data_type_filters ... ok
test test_search_schema_edge_cases ... ok
test test_search_schema_no_matches ... ok
test test_search_schema_sort_orders ... ok
test test_search_schema_text_filters ... ok
test test_search_schema_text_with_sort_orders ... ok
test test_search_schema_type_sort_orders ... ok
test test_set_membership_filter ... ok
test test_single_row_data_frame_column_profiles ... ok
test test_summary_stats ... ok
test test_tibble_support ... ok
test test_update_data_filters_reapplied ... ok
test test_women_dataset ... ok

test result: ok. 40 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.83s
```

kv9898: ~~Currently, I don't have much of a clue.~~ Feel free to edit my code if you find a solution! Wrapping the call in `as.integer()` seems to work.

kv9898: > IDK, this might be related: [wch/r-source@7f20c19](https://github.com/wch/r-source/commit/7f20c19886fd6412453521fb6954febb979ed034)
> 
> There's something special about scipen. Because digits are being correctly evaluated. Perhaps we can't use get_option for this, and instead we bay need to evaluate `getOption` from the R interpreter.

Tried that but sadly it didn't work.

crates/ark/src/data_explorer/r_data_explorer.rs
```diff
@@ -1299,6 +1301,37 @@ impl RDataExplorer {
...
+    /// options. The `max_value_length` is set to 1000, and `thousands_sep`
+    /// is set to `None`.
+    fn current_format_options() -> FormatOptions {
+        // In R 4.2, Rf_GetOption1 (used by get_option) doesn't work correctly for scipen
+        // due to special handling added in later versions. We need to use the R interpreter's
+        // getOption() function instead. See: https://github.com/wch/r-source/commit/7f20c19
+        // Note: 'digits' works fine with Rf_GetOption1, so we don't need to change it.
+        let scipen: i64 = harp::parse_eval_global("as.integer(getOption('scipen', 0))")
+            .ok()
+            .and_then(|obj| obj.try_into().ok())
```
dfalbel:
```suggestion
            .and_then(|obj| obj.try_into())
```

juliasilge:
I'm going to work on this soon, together with the PR to `/positron`. Thanks!

juliasilge:
Thanks for sharing that, @kv9898!

I think there is quite a bit of work for us to do here, as we need to:

- find a better solution for aligning with folks' expectations for backend formatting
- create new infrastructure for how we manage the formatting state updating

Just as a heads up, we don't have folks working on the Data Explorer right now, so we might be a bit slow to make headway given the remaining design work that is needed. We really appreciate your continued contribution and willingness to just jump in!

PR #988, opened 2025-12-11
Send profvis output to editor

jmcphers:
This change causes profiler output from `profvis` to be shown in a new Positron editor tab.

Rough outline:
- In the UI comm, the `is_plot` flag has been replaced with a `destination` enum that specifies one of _three_ destinations: the viewer pane, the plot pane, or an editor tab
- We set the `profvis.print` option and cause profvis files to be output as temporary HTML files, which we send to Positron to be opened in an editor tab (very similar to the workflow for HTML widgets). 

There are a few unrelated changes that are the result of ark being a little out of sync with the UI comm definition in Positron.

Requires https://github.com/posit-dev/positron/pull/11067.

Part of https://github.com/posit-dev/positron/issues/3269.

### DIFF
--- a/crates/ark/src/modules/positron/options.R
+++ b/crates/ark/src/modules/positron/options.R
@@ -39,9 +39,12 @@ options(shiny.launch.browser = function(url) {
 
 # Show Profvis output in the viewer
 options(profvis.print = function(x) {
+    # Render the widget to a tag list to create standalone HTML output.
+    # (htmltools is a Profvis dependency so it's guaranteed to be available)
+    rendered <- htmltools::as.tags(x, standalone = TRUE)
+
     # Render the HTML content to a temporary file
-    # (htmltools is a profvis dependency so it's guaranteed to be available)
-    tmp_file <- htmltools::html_print(x, viewer = NULL)
+    tmp_file <- htmltools::html_print(rendered, viewer = NULL)
 
     # Pass the file to the viewer
     .ps.Call("ps_html_viewer", tmp_file, "R Profile", -1L, "editor")


crates/ark/src/modules/positron/options.R
```diff
@@ -36,3 +36,12 @@ options(plumber.docs.callback = function(url) {
 options(shiny.launch.browser = function(url) {
     .ps.ui.showUrl(url)
 })
+
+# Show Profvis output in the viewer
+options(profvis.print = function(x) {
+    # Render the HTML content to a temporary file
+    tmp_file <- htmltools::html_print(x, viewer = NULL)
```
lionel-: Probably worth a note that `hmtltools` is an (indirect) dependency of profvis

crates/ark/src/interface.rs
```diff
@@ -337,7 +337,7 @@ impl PendingInputs {
 
         let input = if harp::get_option_bool("keep.source") {
             _srcfile = Some(SrcFile::new_virtual_empty_filename(input.into()));
-            harp::ParseInput::SrcFile(&_srcfile.unwrap())
+            harp::ParseInput::SrcFile(_srcfile.as_ref().unwrap())
```
lionel-: I looked into this again and just learned about a neat Rust feature called _lifetime extension_.

I originally thought I needed `_srcfile` to be stored at top-level so I could pass a reference to it. But that's not what happens even in the original version of the code! Since `unwrap()` takes `self` here, the value is moved out of `_srcfile` into a temporary. We then take a reference to it, and Rust extends the lifetime of the temporary to match that of the reference (which is tied to `input`).

More info at: https://doc.rust-lang.org/reference/destructors.html#temporary-lifetime-extension

With your change, we're creating a new `Option` containing a reference, so unwrapping it doesn't consume `_srcfile`.

But armed with the knowledge of lifetime extension we can write this much cleaner version:

```rust
        let input = if harp::get_option_bool("keep.source") {
            harp::ParseInput::SrcFile(&SrcFile::new_virtual_empty_filename(input.into()))
        } else {
            harp::ParseInput::Text(input)
        };
```

Can you please revert your change? Merging it would cause a conflict with PRs/branches of mine. I'll implement the new approach suggested above in these branches.

crates/ark/src/modules/positron/html_widgets.R
```diff
@@ -14,7 +14,7 @@
     tmp_file <- htmltools::html_print(rendered, viewer = NULL)
 
     # Guess whether this is a plot-like widget based on its sizing policy.
-    is_plot <- isTRUE(x$sizingPolicy$knitr$figure)
+    destination <- if (isTRUE(x$sizingPolicy$knitr$figure)) 'plot' else 'viewer'
```
lionel-:
```suggestion
    destination <- if (isTRUE(x$sizingPolicy$knitr$figure)) "plot" else "viewer"
```

crates/ark/src/interface.rs
jmcphers: The reason I made this change was that locally I can't even build ark without it. The original code generates this fatal error on my machine:

```
error[E0716]: temporary value dropped while borrowed
   --> crates/ark/src/interface.rs:340:40
    |
338 |         let input = if harp::get_option_bool("keep.source") {
    |             ----- borrow later stored here
339 |             _srcfile = Some(SrcFile::new_virtual_empty_filename(input.into()));
340 |             harp::ParseInput::SrcFile(&_srcfile.unwrap())
    |                                        ^^^^^^^^^^^^^^^^^ creates a temporary value which is freed while still in use
341 |         } else {
    |         - temporary value is freed at the end of this statement
    |
    = note: consider using a `let` binding to create a longer lived value
```

Incidentally your proposed version doesn't compile for me either:

```
   Compiling ark v0.1.220 (/Users/jmcphers/git2/ark/crates/ark)
error[E0716]: temporary value dropped while borrowed
   --> crates/ark/src/interface.rs:337:40
    |
336 |         let input = if harp::get_option_bool("keep.source") {
    |             ----- borrow later stored here
337 |             harp::ParseInput::SrcFile(&SrcFile::new_virtual_empty_filename(input.into()))
    |                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ creates a temporary value which is freed while still in use
338 |         } else {
    |         - temporary value is freed at the end of this statement
    |
    = note: consider using a `let` binding to create a longer lived value

```

Probably has to do with the version of Rust I'm using? I'm on 1.87. 

I did revert the change, since it doesn't seem to be causing any trouble in CI.

lionel-: oh I'm on 1.89. Good to know!

This is from the 1.89 changelog and explains the difference: https://github.com/rust-lang/rust/pull/140593

lionel-: I've bumped the minimum version to 1.89 on main to avoid such issues in the future.

crates/ark/src/viewer.rs
```diff
@@ -76,12 +77,18 @@ pub unsafe extern "C-unwind" fn ps_html_viewer(
                     }
                 },
                 SessionMode::Console => {
-                    let is_plot = RObject::view(is_plot).to::<bool>();
-                    let is_plot = match is_plot {
-                        Ok(is_plot) => is_plot,
+                    let destination = match RObject::view(destination).to::<String>() {
+                        Ok(s) => s.parse::<ShowHtmlFileDestination>().unwrap_or_else(|_| {
+                            log::warn!(
+                                "`destination` must be one of 'plot', 'editor', or 'viewer'"
```
DavisVaughan:
```suggestion
                                "`destination` must be one of 'plot', 'editor', or 'viewer', using 'viewer' as a fallback."
```

Same note as the other path? Might make logs clearer if we hit this.

crates/ark/src/modules/positron/options.R
DavisVaughan: Agreed

DavisVaughan: I guess passing the `background` color to `htmltools::html_print()` would be asking too much? 😄 

Image

crates/ark/src/modules/positron/options.R
```diff
@@ -36,3 +36,12 @@ options(plumber.docs.callback = function(url) {
...
+
+    # Pass the file to the viewer
```
DavisVaughan, line 49: It is rather annoying that it doesn't resize. It means the window of text that you are reading is quite small to read.

It does resize in RStudio, I wonder what is different there?

https://github.com/user-attachments/assets/7c6e3e6d-66ae-43a1-b8a9-ae5407fe07a0

crates/ark/src/modules/positron/options.R
jmcphers: When I was first implementing htmlwidgets support I actually hooked this up, which was when I discovered that most htmlwidgets are not theme-aware and always render in dark/black colors, so setting the background to black made them unreadable.

jmcphers: Same deal for profvis 😂 
image

crates/ark/src/modules/positron/options.R
jmcphers: Thanks for the nudge, I think I figured it out ... the trick is to force it to render in standalone mode by rendering to tags first. (That's not how RStudio does it, but works here.) https://github.com/posit-dev/ark/pull/988/commits/db02e6fa3269d14e797d958dd1286c00fd95c7d8

PR #990, opened 2025-12-15
Prevent error when expression is invalid for obtaining a help handler

dfalbel:
Addresses https://github.com/posit-dev/positron/issues/7430#issuecomment-3656663284

When an invalid expression is evaluated, we should return `false` so that we show the 'no R help available for topic' popup instead of an error.

### DIFF
--- a/crates/ark/src/help/r_help.rs
+++ b/crates/ark/src/help/r_help.rs
@@ -283,7 +283,7 @@ impl RHelp {
                 Ok(obj) => obj,
                 Err(err) => {
                     // Could not parse/eval the topic; no custom handler.
-                    log::trace!("Could not parse/eval help topic expression '{}': {:?}", topic, err);
+                    log::warn!("Could not parse/eval help topic expression '{}': {:?}", topic, err);
                     return Ok(None);
                 }
             };


crates/ark/src/help/r_help.rs
```diff
@@ -279,7 +279,15 @@ impl RHelp {
                 RObject::from(R_GlobalEnv)
             })();
 
-            let obj = harp::parse_eval0(topic.as_str(), env.sexp)?;
+            let obj = match harp::parse_eval0(topic.as_str(), env.sexp) {
+                Ok(obj) => obj,
+                Err(err) => {
+                    // Could not parse/eval the topic; no custom handler.
+                    log::trace!("Could not parse/eval help topic expression '{}': {:?}", topic, err);
```
lionel-: Probably should be a `warn` so users can figure out why their gesture doesn't work?

```suggestion
                    log::warn!("Could not parse/eval help topic expression '{}': {:?}", topic, err);
```

I'm on the fence regarding whether it should be a visible notification as in the linked issue, but conservative logging seems like a good first approach.

dfalbel: The error notification seems slightly aggressive IMO. I think the info notification is more consistent with the behavior when you do F1 on a name that does not correspond to a help topic in R too.

PR #992, opened 2025-12-16
Take the DAP lock on the R thread to avoid deadlock

lionel-:
Addresses https://github.com/posit-dev/positron/issues/5024

I can't reproduce the deadlock anymore, but the approach taken here should fix all deadlock issues: We only take the lock on the DAP state when we're on the R thread.

### DIFF
--- a/crates/ark/src/dap/dap_server.rs
+++ b/crates/ark/src/dap/dap_server.rs
@@ -446,19 +446,21 @@ impl<R: Read, W: Write> DapServer<R, W> {
     }
 
     fn collect_r_variables(&self, variables_reference: i64) -> Vec<RVariable> {
-        let state = self.state.lock().unwrap();
-        let variables_reference_to_r_object = &state.variables_reference_to_r_object;
+        // Wait until we're in the `r_task()` to lock
+        // See https://github.com/posit-dev/positron/issues/5024
+        let state = self.state.clone();
 
-        let Some(object) = variables_reference_to_r_object.get(&variables_reference) else {
-            log::error!(
-                "Failed to locate R object for `variables_reference` {variables_reference}."
-            );
-            return Vec::new();
-        };
+        let variables = r_task(move || {
+            let state = state.lock().unwrap();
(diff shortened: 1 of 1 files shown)


DavisVaughan:
I managed to deadlock it in dplyr with dev ark, i.e. `cargo build` without `--release`.

```r
# in dplyr
devtools::load_all()
debugonce(mutate_col)
mutate(mtcars, x = mpg+cyl)
```

I stepped in and out of some functions and it deadlocked while coming out of `eval_all_mutate()`

I think I can confirm your fix here based on the frozen stack in ark:

* the main R thread is waiting on the lock in `debug_start()`
* but the `ark-dap` thread is indeed paused in `collect_r_variables()` waiting on the `r_task()` to run _while its holding the lock on `state`_
* so the `r_task()` can't run because the main R thread can't advance, etc, etc, deadlock

You'll also see a request for variables was just received in the logs, but then we got stuck

Screenshot 2025-12-16 at 10 56 42 AM
Screenshot 2025-12-16 at 10 55 53 AM
Screenshot 2025-12-16 at 10 54 30 AM

crates/ark/src/dap/dap_server.rs
```diff
@@ -446,19 +446,21 @@ impl<R: Read, W: Write> DapServer<R, W> {
...
-
-        let Some(object) = variables_reference_to_r_object.get(&variables_reference) else {
-            log::error!(
-                "Failed to locate R object for `variables_reference` {variables_reference}."
-            );
-            return Vec::new();
-        };
+        // Wait until we're in the `r_task()` to lock
+        // See https://github.com/posit-dev/positron/issues/5024
+        let state = self.state.clone();
```
DavisVaughan, line 451:
```suggestion
        let state = Arc::clone(&self.state);
```

Just so it's very clear this is a cheap clone. I had to go look it up myself to make sure.

This is a typical rust pattern i see with Arc

lionel-:
yep same diagnostics as ^^ was provided in https://github.com/posit-dev/positron/issues/5024

PR #998, opened 2025-12-22
Implement plot metadata call

jmcphers:
Adds an implementation of a method to get a plot's metadata, including a derived unique name and the code that was executed to create the plot.

### DIFF
--- a/crates/ark/src/modules/positron/graphics-kind.R
+++ b/crates/ark/src/modules/positron/graphics-kind.R
@@ -151,30 +151,34 @@ detect_ggplot_kind <- function(gg) {
 # Returns plot kind string or NULL
 detect_kind_from_display_list <- function(dl) {
     # Display list entries are lists where first element is the C function name
-    call_names <- vapply(dl, function(x) {
-        if (is.list(x) && length(x) > 0) {
-            name <- x[[1]]
-            if (is.character(name)) name else ""
-        } else {
-            ""
-        }
-    }, character(1))
+    call_names <- vapply(
+        dl,
+        function(x) {
+            if (is.list(x) && length(x) > 0) {
+                name <- x[[1]]
+                if (is.character(name)) name else ""
+            } else {
+                ""
+            }
+        },
+        character(1)
+    )
 
     # Base graphics C functions to plot types
-    if (any(call_names == "C_plotHist")) return("histogram")
-    if (any(call_names == "C_image")) return("image")
-    if (any(call_names == "C_contour")) return("contour")
-    if (any(call_names == "C_persp")) return("3D surface")
-    if (any(call_names == "C_filledcontour")) return("filled contour")
-
-    # Check for grid graphics (ggplot2, lattice)
-    if (any(grepl("^L_", call_names))) {
-        return("grid")
+    if (any(call_names == "C_plotHist")) {
+        return("histogram")
     }
-
-    # Check for base graphics
-    if (any(grepl("^C_", call_names))) {
-        return("base")
+    if (any(call_names == "C_image")) {
+        return("image")
+    }
+    if (any(call_names == "C_contour")) {
+        return("contour")
+    }
+    if (any(call_names == "C_persp")) {
+        return("3D surface")
+    }
+    if (any(call_names == "C_filledcontour")) {
+        return("filled contour")
     }
 
     NULL


crates/ark/src/plots/graphics_device.rs
```diff
@@ -399,6 +462,34 @@ impl DeviceContext {
...
+                let stored_metadata = self.metadata.borrow();
+                let info = stored_metadata.get(id);
+
+                let plot_metadata = match info {
+                    Some(info) => PlotMetadata {
+                        name: info.name.clone(),
+                        kind: info.kind.clone(),
+                        execution_id: info.execution_id.clone(),
+                        code: info.code.clone(),
+                    },
```
DavisVaughan, line 496: Is there any reason you need a local version of `PlotMetadataInfo` over just storing `PlotMetadata`s in the hash map directly?

crates/ark/src/plots/graphics_device.rs
```diff
@@ -483,24 +582,43 @@ impl DeviceContext {
...
+            metadata.insert(
+                id.clone(),
+                PlotMetadataInfo {
+                    name,
+                    kind,
+                    execution_id,
+                    code,
+                },
+            );
+        }
```
DavisVaughan:
```suggestion
        self.metadata
            .borrow_mut()
            .insert(id.clone(), PlotMetadataInfo {
                name,
                kind,
                execution_id,
                code,
            });
```

I think you can get rid of the scope? We do this in a few other places too

crates/ark/src/plots/graphics_device.rs
```diff
@@ -540,9 +658,28 @@ impl DeviceContext {
...
+            plot_metadata.insert(
+                id.clone(),
+                PlotMetadataInfo {
+                    name,
+                    kind,
+                    execution_id,
+                    code,
+                },
+            );
+        }
```
DavisVaughan:
```suggestion
        self.metadata
            .borrow_mut()
            .insert(id.clone(), PlotMetadataInfo {
                name,
                kind,
                execution_id,
                code,
            });
```

same here

crates/ark/src/plots/graphics_device.rs
```diff
@@ -540,9 +658,28 @@ impl DeviceContext {
     }
 
     #[tracing::instrument(level = "trace", skip_all, fields(id = %id))]
-    fn process_new_plot_jupyter_protocol(&self, id: &PlotId) {
+    fn process_new_plot_jupyter_protocol(&self, id: &PlotId, execution_context: Option<(String, String)>) {
```
DavisVaughan: Can you please reformat this file? I get diffs when I save due to reformatting.

crates/ark/src/interface.rs
```diff
@@ -1265,10 +1273,14 @@ impl RMain {
...
 
+            // Extract execution context before req is consumed
+            let execution_id = req.originator.header.msg_id.clone();
+            let code = req.request.code.clone();
+
             // Check for pending graphics updates
             // (Important that this occurs while in the "busy" state of this ExecuteRequest
             // so that the `parent` message is set correctly in any Jupyter messages)
-            graphics_device::on_did_execute_request();
+            graphics_device::on_did_execute_request(execution_id, code);
```
DavisVaughan: How would you feel about a push model where `handle_execute_request()` _pushes_ the active request's `execution_id` and `code` to the graphics device side via a new `graphics_device::on_execute_request()` hook?

i.e. in `handle_execute_request()` after this

```
                // Save `ExecuteCode` request so we can respond to it at next prompt
                self.active_request = Some(ActiveReadConsoleRequest {
                    exec_count,
                    request: exec_req,
                    originator,
                    reply_tx,
                });
```

you'd also call `graphics_device::on_execute_request()` to push the active request's information over.

My thought was that if you did this, then we could remove the awkward code around _sometimes_ having the `execution_id` and `code` being passed through `process_changes()` and _sometimes_ needing to go request it from `RMain`.

If this idea works out, then `DEVICE_CONTEXT` would always have an up to date `execution_id` and `code` from the `graphics_device::on_execute_request()` hook and I think it would simplify the model a good bit.

(The way this works right now is what took the most time for me to review and understand)

crates/ark/src/modules/positron/graphics.R
```diff
@@ -529,3 +529,168 @@ render_path <- function(id, format) {
     file <- paste0("render-", id, ".", format)
     file.path(directory, file)
 }
+
```
DavisVaughan, line 532: This is kind of a personal request, but would you mind if we isolated this code in `graphics-kind.R` or something like that?

I like that `graphics.R` is currently just about the main flow of handling the graphics devices

I have a feeling this bit of code might require a few rounds of iteration / might be a common place we have to update in the future, so it would be nice to have it tucked away in its own file (for some reason that makes it easier for me to reason about).

crates/ark/src/modules/positron/graphics.R
```diff
@@ -529,3 +529,168 @@ render_path <- function(id, format) {
...
+#' Detect the kind of plot from a recording
+#'
+#' Uses multiple strategies to determine plot type:
+#' 1. Check .Last.value for high-level plot objects (ggplot2, lattice)
+#' 2. Check recording's display list for base graphics patterns
+#' 3. Fall back to generic "plot"
```
DavisVaughan: So the idea is that for ggplot2, the display list isn't really "rich" enough to give you all the details you need to come up with a good name? And instead you ideally need the plot object?

crates/ark/src/modules/positron/graphics.R
```diff
@@ -529,3 +529,168 @@ render_path <- function(id, format) {
...
+
+    # Check for grid graphics (ggplot2, lattice)
+    if (any(grepl("^L_", call_names))) {
+        return("grid")
+    }
+
+    # Check for base graphics
+    if (any(grepl("^C_", call_names))) {
+        return("base")
+    }
```
DavisVaughan: I feel like just `plot` is more useful than these?

crates/ark/src/plots/graphics_device.rs
jmcphers: You're right, I think this type is superfluous. I've removed it.

crates/ark/src/plots/graphics_device.rs
jmcphers: done!

crates/ark/src/interface.rs
jmcphers: yeah! done in https://github.com/posit-dev/ark/pull/998/commits/36941caf2605e1c65877c12be7596ac24683416e.

crates/ark/src/modules/positron/graphics.R
jmcphers: moved it!

crates/ark/src/modules/positron/graphics.R
jmcphers: Yeah, we can just get slightly better names this way

crates/ark/src/modules/positron/graphics.R
jmcphers: updated!

PR #1003, opened 2026-01-13
Add breakpoint support to DAP

lionel-:
Branched from https://github.com/posit-dev/ark/pull/981

Addresses https://github.com/posit-dev/positron/issues/1766
Addresses https://github.com/posit-dev/positron/issues/11403

Frontend-side PR to be opened soon (feature/breakpoints)
Requires https://github.com/lionel-/biome/pull/1
Requires https://github.com/r-lib/pkgload/pull/323 (run `pak::pak("r-lib/pkgload#323`)

This PR implements breakpoint support in the DAP server adds breakpoint support by injecting `browser()` calls at parse time rather than modifying live objects. Key benefits over RStudio's approach:
- In packages, breakpoints work everywhere: functions, R6 methods, `.onLoad()` hooks, top-level code
- In scripts, no need for `debugSource()` - just Cmd+Enter from the editor

This is the culmination of a series of prepatory PRs that implement a new approach for integrating with R that allows injection of breakpoints in a much more general and robust way:

1. Annotate execution request code with source references
   https://github.com/posit-dev/ark/pull/981
   Allows code evaluated from a script to be aware of location in frontend

   Positron side of source ref annotation for execution requests
   https://github.com/posit-dev/positron/pull/10815

2. Integrate Rowan syntax trees in Ark
   https://github.com/posit-dev/ark/pull/974
   Used to inject breakpoints in source code rather than dynamic objects

   Prepare Air for Ark dependency on Rowan
   https://github.com/posit-dev/air/pull/452
  
   Add post-order visit hook
   https://github.com/lionel-/biome/pull/1
   Work around limitation of Biome's Rowan API for tree manipulation

The first series of PRs makes it possible for code evaluated from an editor to carry source references. It also allows this PR to build on this and inject breakpoints before evaluation. With these, users can step through scripts using familiar gestures like Cmd+Enter.

For instance if a breakpoint is set withing an `lapply()` call:

```r
lapply(1:3, function(x) {
  1  # BP
  x
})
```

https://github.com/user-attachments/assets/d0410d2f-a8e8-4902-adff-77f6645873f8

Being able to invoke the debugger with Cmd+Enter without having to source a whole file as in RStudio is much more practical:

- No need to evaluate potentially long-running expressions prior to the breakpoint.
- No need to run expressions at the start of the script that reset state that's potentially important for the debugging session.

To achieve this, Ark is now integrated with the R REPL in a novel way that allows Ark to be in charge of _parsing_, so it can control source references. It's also by necessity in charge of _evaluation_, which will be useful later on to provide `recover()` functionality in Positron (evaluation will happen in the selected call frame).

The second series of PRs sets the stage for a novel approach for injecting breakpoints implemented in this PR. Traditionally, breakpoint injection has followed base R's `setBreakpoint()` approach which uses `trace()` to inject `browser()` calls in runtime functions. While this approach works well, it has inherent limitations and difficulties:

- We're dynamically modifying functions after they've been copied in various places (e.g. imported by packages, exported to the search path, inserted in the S3 method table, or completely ad hoc things like manually copied in a private environment from an onload hook). Modifying one copy is not sufficient for consistent behaviour of the debugger, finding all copies is not possible in principle, and finding the most important ones is not trivial.

- After injection, the existing srcref objects need to be adjusted to the modified AST, which is tricky.

- Sometimes the breakpoint might be set in a place that is hard to reach, like R6 methods. RStudio still doesn't support them https://r6.r-lib.org/articles/Debugging.html

The new approach implemented here is to inject `browser()` calls for breakpoints at parse time. By injecting breakpoints before R ever sees the code, we sidestep all of these limitations and difficulties. R only evaluates code that already contains breakpoint calls, so we never miss copies and source references are correct from the start.

With this approach, we support breakpoints anywhere in a package, including R6 methods, `.onLoad()` hooks, or even at top-level, without needing ongoing maintenance work, e.g. to add support to new object systems like S7.

https://github.com/user-attachments/assets/86422664-525a-428d-babc-d377d061242f

#### Breakpoint injection

An injected breakpoint looks like this:

```r
.ark_auto_step(.ark_breakpoint(browser(), "*url*", "*id*"))
#line *line* "*url*"
expression_to_break_on
```

- `.ark_auto_step()` is an identity function that serves as sentinel when R steps through code. If the user steps on an injected breakpoint, we detect the auto-step call in the `debug at` message emitted by R and automatically step over it (i.e. call `n`).

- `.ark_breakpoint()` takes a `browser()` call promised in the current environment, a URL, and the breakpoint's unique ID. It only forces the browser argument if the breakpoint is active. Since the argument is promised in the call-site environment, this causes R to mark that environment as being debugged with the `RDEBUG()` flag.

  It does not stop quite at the right place though, inside the `.ark_breakpoint()` wrapper, with `.ark_auto_step()` on the stack as well. To solve this, there is a second condition triggering auto-stepping in ReadConsole: if the function of the top stack frame is `.ark_breakpoint()` (which we detect through a class assigned to the function), then we auto-step. This causes R to resume evaluation and, since the call-site environment is being debugged (`RDEBUG()` is set on the environment), it stops at the next expression automatically, in this case `expression_to_break_on`.

- The `#line` directive right above `expression_to_break_on` maps the source references to the original location in the source document. When R stops on the expression, it emits the original location, allowing the DAP to communicate the appropriate stopping place to the frontend.

#### Source instrumentation

`base::source()` and `devtools::load_all()` need breakpoint injection as described above, as well as top-level adjustments so it's possible to step through a script or top-level package file.

If the sourced file looks like:

```r
1
2
```

The instrumented version ends up as:

```r
{
#line 1 "file:///file.R"
1
base::.ark_auto_step(base::.ark_verify_breakpoints_range("file:///test.R", 1L, 2L))
#line 2 "file:///file.R"
2
}
```

- The whole source is wrapped in `{}` to allow R to step through the code.
- Line directives map each expression to original source.
- An auto-stepped `.ark_verify_breakpoints_range()` call after each expression lets the DAP know that any breakpoints spanned by the last expression are now "verified", i.e. the breakpoints have been injected and the code containing them has been evaluated.

  A subtlety: verify calls are _not_ injected after trailing expressions in a `{}` list. The issue is that every injected call (breakpoint or verify) needs a `#line` directive afterwards to restore correct source reference mapping. These directives are attached as leading trivia to the next sibling node. But a trailing expression has no next sibling, so there's nowhere to attach the directive.

  Instead, trailing expressions defer their verification to the parent `{}` list. For deeply nested trailing breakpoints, the verification bubbles up to the outermost list. In `source()`'s case, this unconditionally adds a final verify call covering all lines.

To hook into `source()`, we rebind `base::source` to a wrapper that optionally performs breakpoint injection. The hook has several safeguards to fall back to the original `base::source()` implementation:

- If the global option `ark.source_hook` is set to `FALSE`.
- If there are no breakpoints for the file URI.
- If `file` is not supplied or can't be converted to a URI.
- If any argument other than `echo` (used by Positron) or `local` (useful user argument) is supplied. This is conservative but avoids edge cases with argument combinations we haven't tested.

When the hook activates, it reads the file, calls into Rust to annotate the source with breakpoints, parses the annotated code (which is now a single `{...}` expression) as R AST, and evaluates it in the appropriate environment.

`pkgload::load_all()` hooks similarly but is simpler, see https://github.com/r-lib/pkgload/pull/323.

#### Auto-stepping over/outside injected expressions

To prevent injected expressions from interfering with debug-stepping, we automatically step over to the next statement when the debugger stops at an injected expression. Two cases:

- Not in debug mode, we'll stop only _in_ injected breakpoint calls. Auto-step to next expression (the user expression we should actually stop at).
- In debug mode, we'll step _at_ injected calls (breakpoint or verification calls). In this case, auto-step to the next expression.

This happens in `ReadConsole`.

#### Limitations

- While the general user expectation would be for breakpoints to just work right after activating them in the UI, the reality is that the breakpoint is "Unverified" (inactive) at first. To activate the breakpoint, the user _has_ to send the code with new breakpoints to the R session, either by evaluating or sourcing / calling `load_all()`.

  Alternative approaches that don't have these limitations:

  - Having R manage a list of breakpoints that can be compared against srcrefs of executed code. The interpreter would break when entering an expression that matches a breakpoint. This is the ideal setup.

  - Try and update live objects when user sets a breakpoint but then we're back to RStudio's limitations regarding finding the live objects and all their copies.

  A corollary is that when user has multiple sessions, breakpoint state is tied to session state. Breakpoints that are verified in one session will not be verified in another session, they need to be sourced one way or another in all sessions of interest.

- Verbose console output:

  ```r
  debug at file:///Users/lionel/Desktop/breakpoints.R#13: base::.ark_auto_step(base::.ark_verify_breakpoints_range("file:///Users/lionel/Desktop/breakpoints.R",
    7L, 13L))
  Browse[1]> n
  debug at file:///Users/lionel/Desktop/breakpoints.R#15: [1] 1
  Browse[1]> n
  debug at file:///Users/lionel/Desktop/breakpoints.R#16: [1] 2
  Browse[1]> n
  debug at file:///Users/lionel/Desktop/breakpoints.R#18: base::.ark_auto_step(base::.ark_breakpoint(browser(), "file:///Users/lionel/Desktop/breakpoints.R",  "2"))
  Called from: base::.ark_breakpoint(browser(), "file:///Users/lionel/Desktop/breakpoints.R",  "5")
  debug at file:///Users/lionel/Desktop/breakpoints.R#20: [1]
  ```

  Verbose printed functions after injection:

  ```r
  # Browse[1]> sys.function()
  function() {
    base::.ark_auto_step(base::.ark_breakpoint(browser(), "file:///Users/lionel/Desktop/breakpoints.R", "5"))
    #line 20 "file:///Users/lionel/Desktop/breakpoints.R"
    1 # BP1
    base::.ark_auto_step(base::.ark_verify_breakpoints_range("file:///Users/lionel/Desktop/breakpoints.R", 20L, 21L))
    #line 21 "file:///Users/lionel/Desktop/breakpoints.R"
    2
  }
  ```

  Injection with `trace()` is not as verbose because it changes the function's class and print method.

  It's unclear how to do better. Filtering from `WriteConsole` is not appealing because of its streaming nature, hard to match with regex when you don't have the full context, and hard to have a filter that's 100% correct (although not necessarily a goal, seems like a useful property).

  We could potentially wrap functions with injected breakpoints in a constructor at parse-time, but this could cause unexpected behaviour and side effects in edge cases.

#### Breakpoint invalidation and adjustment

Some breakpoints can't hit precisely on the line requested by the user, but we're able to move them to valid lines:

- Breakpoints inside multi-line expressions like

  ```r
  list(
    1,  # BP here
    2
  )
  ```

  These breakpoints are moved _up_ to the start of the statement.

- Breakpoints on whitespace or comments

  ```r
  # BP Here
  1
  ```

  These breakpoints are moved _down_ to the start of the next statement.

Some breakpoints are invalid and can never be hit:

- Breakpoints on a closing `}` line.
- Breakpoints inside empty multiline braces `{ }` (rare edge case)

#### Permanent DAP session

Previously, the DAP session was only connected when a browser REPL was active. We used to send a notification `start_debug` to request the frontend to connect with a DAP client. This is no longer the case, the DAP is now expected to always be connected so that we can receive notification about the state of breakpoints in the frontend. Without this, we would not know that we need to inject breakpoints in executed code.

The `start_debug` and `stop_debug` messages sent via the Jupyter comm are still used but have become hints for the frontend to show or hide the debug toolbar, not session lifecycle events.

When R enters the debugger, we send `start_debug` followed by a `Stopped` event. When R exits the debugger, we send `stop_debug` followed by a `Continued` event (not `Terminated`). The DAP connection remains active throughout, allowing the backend to:

- Receive new breakpoints as the user sets them in the UI
- Send verification events as code is evaluated
- Preserve breakpoint state across debug sessions

#### DAP event notifications

Breakpoint state changes are communicated to the frontend via DAP `Breakpoint` events sent through a backend channel:

1. **At parse time**: After `annotate_input()` or `annotate_source()`, we call `notify_invalid_breakpoints()` to inform the frontend about any breakpoints marked invalid during tree rewriting. This notification contains a message indicating reason of invalidaty and is shown on hover in the frontend UI.

2. **At breakpoint hit**: When `.ark_breakpoint()` is about to force its `browser()` argument, it calls `ps_verify_breakpoint()` which marks the breakpoint as verified and sends a `BreakpointState` event. This ensures the breakpoint turns red immediately when first hit, even before the expression containing it finishes evaluating.

3. **After expression evaluation**: The `.ark_verify_breakpoints_range()` calls injected after expressions call `ps_verify_breakpoints_range()`, which loops over breakpoints in the line range and sends `BreakpointState` events for any that transition from unverified to verified.

   Note that the frontend only reacts to line-adjustments for _verified_ breakpoints, so it would not be helpful to notify these earlier at parse time.

These events allow the frontend to update breakpoint appearance (gray dot to red dot, or show error message on hover) in real-time as code executes.

#### Breakpoint state management

Each breakpoint has a state that determines its behavior and appearance:

- **Unverified**: Initial state. The breakpoint hasn't been injected into evaluated code yet. Shown as a gray dot.
- **Verified**: The breakpoint has been injected and the containing code has been evaluated. Shown as a red dot.
- **Disabled**: A previously verified breakpoint that the user unchecked. When the user unchecks a breakpoint in the frontend UI, it appears as a deletion to the backend (omitted from the `SetBreakpoints` request). We preserve it internally as Disabled so that when the user re-enables it, we can restore its Verified state directly without requiring them to source again.
- **Invalid**: The breakpoint is at an invalid location (closing brace, empty braces). Shown as gray with a hover message explaining why.

Breakpoints also track:

- **`injected`**: Whether the breakpoint was actually injected into code during annotation. This is crucial: `verify_breakpoints()` only verifies breakpoints where `injected == true`. This prevents a bug where a breakpoint added _after_ parsing gets incorrectly verified when stopping at another breakpoint in the same function.

- **`line` vs `original_line`**: `original_line` is what the frontend sent. `line` is the adjusted/anchored line where R will actually stop. We need both because the frontend doesn't know about our line adjustments, so subsequent `SetBreakpoints` requests always use the original line numbers. We match against `original_line` to find existing breakpoints and preserve their state. The frontend displays the adjusted `line` after verification.

When multiple breakpoints anchor to the same expression (e.g., one on a blank line above and one inside a multi-line call), only one `.ark_breakpoint()` call is injected (using the first breakpoint's ID), but all matching breakpoints get their `line` adjusted and `injected` flag set. They all become verified together and display at the same location.

#### Document hashing and reconnection

Breakpoints are associated with a blake3 hash of the document content. This allows breakpoint state to survive DAP server disconnections, which happen when:

- The user uses the disconnect command (the frontend automatically reconnects).
- The console session goes to the background (the LSP is also disabled in this case, so we don't receive document change notifications).

When the server comes back online and receives a `SetBreakpoints` request, it compares the current document content against the stored hash. If the hash matches, existing breakpoint states (verified, adjusted lines, etc.) are preserved. If the document changed, all breakpoints for that URI are reset to unverified.

When a document changes while connected, `did_change_document()` removes all breakpoints for that URI and sends unverified events to the frontend, since the old line numbers may no longer be valid.

#### Tree rewriting implementation

Depends on https://github.com/lionel-/biome/pull/1

Breakpoint injection uses Biome's `SyntaxRewriter`, a tree visitor with preorder and postorder hooks that allows replacing nodes on the way out.

- Preorder (`visit_node`): Cache line information for braced expressions. We record where each expression starts and its line range. This must be done before any modifications because the postorder hook sees a partially rebuilt tree with shifted token offsets.

- Postorder (`visit_node_post`): Process expression lists bottom-up. For lists inside braces, we inject breakpoint calls, add `#line` directives, and mark remaining breakpoints (e.g. on closing braces) as invalid. The cached line info represents original source positions, which is exactly what we need for anchoring breakpoints to document lines.

We use `SyntaxRewriter` instead of `BatchMutation` because the latter doesn't handle insertions in lists (only replacements), and doesn't handle nested changes in a node that is later replaced. For example with a breakpoint on both `{` and `1`:

```r
{     # BP 1
   1  # BP 2
}
```

BP 2 causes changes inside the braces. Then BP 1 causes the whole brace expression to be replaced with a variant that has a `#line` directive. `BatchMutation` can't express both changes because it takes modifications upfront. `SyntaxRewriter` lets us replace nodes bottom-up as we go.

#### Activating breakpoints

- There are multiple code paths that should all behave the same:
  - Evaluation from an editor (Cmd/Ctrl+Enter)
  - Sourcing file with and without echo
  - `devtools::load_all()`

- Script breakpoints can be activated simply by evaluating from the editor with Cmd/Ctrl+Enter, either with the statement range or a selection. You can even immediately drop into the debugger, e.g. with:

  ```r
  lapply(1:3, function(x) {
    1 # BP
    2
    3
  })
  ```

  This is a significant improvement over RStudio which requires `debugSource()` instead.

- Package breakpoints require the pkgload PR (`pak::pak("r-lib/pkgload#323")`). With it installed, `load_all()` should allow breakpoints to work anywhere in the package: at top-level, in `.onLoad` hooks, in functions, in R6 methods, etc. This is a significant improvement over RStudio which doesn't support R6 methods.

- We're hooking into `base::source()` and it should inject and verify breakpoints but only if `file` is supplied as sole argument. If any other argument is supplied, we use the regular base `source()` function. The exceptions are `echo` (which Positron may pass in) and `local` (useful user argument). If the global option `ark.source_hook` is set to `FALSE`, we always use regular `source()`.

#### Verification behavior

- Normally breakpoints are verified _after_ an evaluation has finished. But if the evaluation causes R to stop on a breakpoint, it should become verified right away. For instance with:

  ```r
  lapply(1:3, function(x) {
    x  # BP
  })
  ```

  The breakpoint on `x` should become a full red dot on the first stop there.

- Inner breakpoints should be verified after stepping over them, both at top-level and inside brace lists.

  ```r
  local({
    fn <- function() { # BP that causes to break in `local({ ... })`
      1 # Internal BP
    }
    fn # fn's breakpoints should be verified when stepping here
  })
  ```

- If a breakpoint is added after a function was parsed/evaluated, stopping at another breakpoint in that function should not incorrectly verify the new breakpoint. For example:

  1. Parse and evaluate:
     ```r
     fn <- function() {
       1 # BP1
       2
     }
     ```
     BP1 is injected and verified.

  2. Add an unverified breakpoint on line 2 (BP2). It's not injected yet since we haven't reparsed.

  3. Call `fn()` and stop at BP1. BP2 should remain Unverified since it was never actually injected.

- When you source a file that causes an error mid-script, the breakpoints preceding the error site should be verified. The breakpoints following the error site remain unverified.

#### Adjustment and invalidation

- After verification, all BPs inside multiline expressions should be mapped to the start line of the expression.

  ```r
  local({
    list(  # Start of expression in { list where BPs get mapped to
      1, # BP inside multiline expression
         # BP on empty line
      2  # BP inside multiline expression
    )
  })
  ```

- Breakpoints are invalid and should remain Unverified at all times when set:
  - In empty braces:
    ```r
    {
      # BP
    }
    ```
  - On a closing brace `}`

  You can inspect the reason by hovering over the breakpoint.

#### Enable/disable

- Breakpoints can be disabled then reenabled. If they were already verified prior to disabling, they should be restored as verified when reenabled.

- When a breakpoint was verified then disabled by the user, stopping at the breakpoint with the debugger (via `debug()` or via another earlier breakpoint) should not bring back the disabled breakpoint online. Although we hold onto it internally in case it's re-enabled by user, it should remain inert until they do so.

- Set a breakpoint, source, remove the breakpoint, source again, then re-add the breakpoint. The breakpoint shoulde be unverified, and re-verified on the next source. This should be the case both with evaluation from an editor and with `source()` or `load_all()`.

#### Debug session behavior

- The debug toolbar should appear when stopped at a breakpoint (or more generally in the R debugger) and disappear when continuing/exiting the debugger.

- The `Q` command should cleanly exit the debugger and hide the toolbar. Same for `n` and `f` is they exit a debugger scope.

- When you've stopped at a breakpoint, stepping onto a line where there is another breakpoint should just stop there normally (no double-stop on the same line because we automatically step over injected breakpoints).

- If you set a breakpoint at top-level in a `{}` block, and step through it, evaluating another `{}` block without breakpoints shouldn't drop you in the debugger (I had to work around base R marking the global env as being debugged). This is covered by an Ark-side test.

#### Session and connection management

- There should only be a single DAP session at any time. If there are multiple, you'll see the running sessions in the call stack viewpane of the frontend. When there is only a single session running, the session is normally (there are edge cases not discussed here) not shown at all in that viewpane.

- Stopping, starting, and restarting sessions should all work as before.

- Switch console sessions while debugging session is ongoing. The debugging session of the now background console should have quit with `Q`.

- After switching sessions back and forth, breakpoint state should be restored.

- Disconnecting a debug session from the frontend via the debug toolbar or Shift-F5 will not only exit the debugger on the R side but also disconnect from the DAP server. The frontend starts up a new session automatically, allowing the backend to receive breakpoint updates. In other words, a hard disconnect shouldn't interfere with breakpoint functionality.

- Editing a file while stopped at a breakpoint should instantly invalidate breakpoints for that file. All injected breakpoints are conditional, invalidated breakpoints never cause R to break in the debugger. The goal is to prevent stepping through an editor using a stale source view. After invalidation, the user must re-source the breakpoints.

- LSP session switching code is implicated. Should still work as before.

##### Backend tests (Ark)

The backend has extensive unit tests for code annotation in `console_annotate.rs` (43 tests):

- **Input annotation**: Basic `#line` directive injection, line/character offsets, handling of existing whitespace and comments.
- **Breakpoint injection**: Single and multiple breakpoints, nested brace lists, multiline expressions anchoring to start, blank lines anchoring to next expression, closing brace invalidation, empty braces invalidation.
- **Source annotation**: Wrapping in `{}`, verification call injection, top-level breakpoints.
- **Edge cases**: Doubly/triply nested braces, if-else branches, multiple breakpoints collapsing to same line.

Integration tests in `kernel-debugger.rs` cover browser REPL behavior (entering/exiting debugger, nested debugging, stepping, etc.) but don't yet cover breakpoint-specific scenarios.

These tests currently do not cover the Breakpoint events emitted by the backend (verification status, adjusted lines, reason of invalidity, etc).

##### Planned: DAP protocol tests

We plan to add protocol-level tests that communicate with the DAP server directly to verify breakpoint events:

- `SetBreakpoints` response contains correct initial state (unverified, with correct IDs)
- `Breakpoint` events are sent when breakpoints become verified
- `Breakpoint` events are sent when breakpoints become invalid (with correct reason message)
- Line adjustment is reflected in events after verification
- Document hash changes cause breakpoints to reset to unverified
- Disabled breakpoints are preserved and restored correctly

##### Suggested: Frontend UI tests

Important UI scenarios to cover with automated tests:

- **Basic flow**: Set breakpoint -> source file -> breakpoint turns red -> hit breakpoint -> toolbar appears -> continue -> toolbar disappears
- **Breakpoint adjustment**: Set breakpoint inside multiline expression -> after verification, dot moves to expression start
- **Invalid breakpoints**: Set breakpoint on closing brace -> remains gray with hover message
- **Disable/enable**: Verify breakpoint -> uncheck -> re-check -> still verified (no re-sourcing needed)
- **Document edit**: Stop at breakpoint -> edit file -> breakpoints turn gray -> never activates again
- **Session switching**: Breakpoint state is attached to active console session. Editing the file should disable breakpoints of background sessions too.

See also all the tricky usage patterns in QA notes above.

### DIFF
--- a/crates/ark/src/console_annotate.rs
+++ b/crates/ark/src/console_annotate.rs
@@ -183,9 +183,8 @@ pub(crate) fn annotate_input(
     let annotated_code = if let Some(breakpoints) = breakpoints {
         let parse = aether_parser::parse(code, Default::default());
         if let Some(err) = parse.error() {
-            return Err(anyhow!(
-                "Unexpected parse error in `annotate_input()`: {err}"
-            ));
+            // Shouldn't happen since we're only operating on code that was parsed by R
+            return Err(anyhow!("Parse error in `annotate_input()`: {err}"));
         }
 
         let root = parse.tree();
@@ -246,7 +245,13 @@ pub(crate) fn annotate_source(
     let wrapped = format!("{{\n{code}\n}}");
     let line_index = LineIndex::new(&wrapped);
 
-    let root = aether_parser::parse(&wrapped, Default::default()).tree();
+    let parse = aether_parser::parse(&wrapped, Default::default());
+    if let Some(err) = parse.error() {
+        // Shouldn't happen since we're only operating on code that was parsed by R
+        return Err(anyhow!("Parse error in `annotate_input()`: {err}"));
+    }
+
+    let root = parse.tree();
 
     // `line_offset` = -1 because:
     // - Wrapped line 0 is `{`
--- a/crates/ark/src/interface.rs
+++ b/crates/ark/src/interface.rs
@@ -354,9 +354,18 @@ impl PendingInputs {
         breakpoints: Option<&mut [Breakpoint]>,
     ) -> anyhow::Result<ParseResult<PendingInputs>> {
         let input = if let Some(location) = location {
-            let annotated_code = annotate_input(code, location, breakpoints)?;
-            log::trace!("Annotated code: \n```\n{annotated_code}\n```");
-            harp::ParseInput::SrcFile(&SrcFile::new_virtual_empty_filename(annotated_code.into()))
+            match annotate_input(code, location, breakpoints) {
+                Ok(annotated_code) => {
+                    log::trace!("Annotated code: \n```\n{annotated_code}\n```");
+                    harp::ParseInput::SrcFile(&SrcFile::new_virtual_empty_filename(
+                        annotated_code.into(),
+                    ))
+                },
+                Err(err) => {
+                    log::warn!("{err:?}");
+                    harp::ParseInput::Text(code)
+                },
+            }
         } else if harp::get_option_bool("keep.source") {
             harp::ParseInput::SrcFile(&SrcFile::new_virtual_empty_filename(code.into()))
         } else {


crates/harp/src/vector/integer_vector.rs
```diff
@@ -19,7 +19,7 @@ use crate::vector::Vector;
 
 #[harp_macros::vector]
 pub struct IntegerVector {
-    object: RObject,
+    pub object: RObject,
```
DavisVaughan, line 22: Is this really needed?

crates/ark/src/dap/dap.rs
```diff
@@ -1,7 +1,7 @@
 //
 // dap.rs
```
DavisVaughan, line 2: The debugging experience in packages does not feel right.

When I have `devtools::load_all()`-ed a package, I expect that I can add a breakpoint and have it immediately show up as verified.

Right now it:
- Does not show as verified
- Is not hit when I run `mutate(mtcars, across(mpg, (identity)))`

https://github.com/user-attachments/assets/aa54942d-c57e-4a87-b16d-348c54b1b162

crates/ark/src/dap/dap.rs
```diff
@@ -1,7 +1,7 @@
...
 //
-// Copyright (C) 2023 Posit Software, PBC. All rights reserved.
+// Copyright (C) 2023-2026 Posit Software, PBC. All rights reserved.
 //
```
DavisVaughan, line 5: Somehow pkgload + ark has completely borked my loaded `mutate()`

I don't have a reproducible example but all I did was add a breakpoint in `mutate()` and `across()` and `load_all()` a few times while moving between the two files.

Image

DavisVaughan: Actually I can fully reproduce this.

In a fresh R session in the dplyr project:

- Put a breakpoint on https://github.com/tidyverse/dplyr/blob/2be1ab6ad9e5377ce23e513b0fbc44e205c4cfbe/R/mutate.R#L181
- Run `devtools::load_all()`
- Look at `mutate`

DavisVaughan (changes requested):

crates/ark/src/dap/dap.rs
DavisVaughan: Notably it really is that specific line.

Placing the breakpoint one line above or below that line works?

crates/harp/src/vector/integer_vector.rs
lionel-: yes, used in SrcFile

crates/ark/src/dap/dap.rs
lionel-: Looks like the mutate apparent borkedness is due to source ref manipulations, the source is actually fine.

Regarding breakpoint, I may need to special case {{

crates/ark/src/dap/dap.rs
lionel-: As discussed RStudio is currently a little more dynamic. We could improve on this in the future with:

- A step machine approach that allows hitting breakpoints without any injection
- `debug()` dynamic injection for functions with known places

This approach would get the dynamism closer to RStudio's and would enable more workflows not currently supported in RStudio, such as being able to break into loops from withing the debugger:

```r
{
  1 # Stopped here

  for (i in x) {
    i # Can add a breakpoint here, hit `c`, and immediately break
  }
  
  2 # Then from the loop can add a breakpoint here, hit `f`, and immediately break
}
```

crates/ark/src/dap/dap.rs
lionel-: Instead of special-casing {{, I made it so that breakpoints never anchor on statements that start on the same line as `{`. In that case they anchor to a parent statement if applicable. That rule prevents the embrace operator from ever being matched to a breakpoint.

Regarding srcref borkedness, I think there is an off by one in the srcref's line:

```
> lobstr::src(mutate)
<closure>
├─attr("srcref"): <srcref>
│ ├─location: 146:11 - 148:1
│ └─attr("srcfile"): <srcfilecopy> @001
│   ├─Enc: "unknown"
│   ├─filename: "/Users/lionel/Sync/Projects/R/ti..."
│   ├─fixedNewlines: TRUE
│   ├─isFile: TRUE
│   ├─lines<chr [542]>: "#' Create, m...", "#'", "#' `mutate()...", ...
│   ├─parseData<int [22,160]>: 1, 1, 1, ...
│   ├─timestamp: "2026-01-16 21:08:32"
│   └─wd: "/Users/lionel/Sync/Projects/R/ti..."
└─body(): <{>
  ├─attr("srcref"): <list>
  │ ├─[[1]]: <srcref>
  │ │ ├─location: 146:32 - 146:32
  │ │ └─attr("srcfile"): @001
  │ └─[[2]]: <srcref>
  │   ├─location: 147:3 - 147:21
  │   └─attr("srcfile"): @001
  ├─attr("srcfile"): @001
  └─attr("wholeSrcref"): <srcref>
    ├─location: 1:0 - 148:1
    └─attr("srcfile"): @001
> attr(mutate, "srcref") |> attr("srcfile") |> _$lines |> _[146:148]
[1] "  UseMethod(\"mutate\")" "}"                       ""                       
> mutate
d("mutate")
}
```

lionel-: If you update to latest pkgload PR this should be fixed.

lionel-: @DavisVaughan I also fixed the weirdness we observed for:

> Set a breakpoint, source, remove the breakpoint, source again, then re-add the breakpoint. The breakpoint shoulde be unverified, and re-verified on the next source. This should be the case both with evaluation from an editor and with `source()` or `load_all()`.

DavisVaughan: Works nicely now

https://github.com/user-attachments/assets/299197cb-ee62-4f3d-8269-a3c54c627cbf

crates/ark/src/interface.rs
```diff
@@ -652,6 +666,11 @@ impl RMain {
                 log::error!("Error setting default repositories: {err:?}");
             }
 
+            // Finish initilization of modules
+            RFunction::from("initialize")
+                .call_in(ARK_ENVS.positron_ns)
+                .log_err();
+
```
DavisVaughan: I find it fairly confusing that we have `modules::initialize()` above which is about "initialization of modules" and then again here.

Can we just simplify this to `debug_initialize()` with a comment of `// Set up debugging symbols`?

In the future if it needs to be more generic, we can collapse things back together. But for now I think that actually makes it more confusing (and harder to find if you are just looking for all the debug specific hooks)

crates/ark/src/interface.rs
```diff
@@ -885,6 +916,23 @@ impl RMain {
         &self.iopub_tx
     }
 
+    // Async messages for the Console. Processed at interrupt time.
```
DavisVaughan, line 924: "Processed at interrupt time" is that right? It feels like it processes the notifications immediately since it is in its own thread

crates/ark/src/interface.rs
```diff
@@ -954,9 +1002,10 @@ impl RMain {
             // from the last expression we evaluated
             self.debug_is_debugging = true;
             self.debug_start(self.debug_preserve_focus);
-        } else if self.debug_is_debugging {
-            self.debug_is_debugging = false;
-            self.debug_stop();
+
+            // Note that for simplicity this state is reset on exit via the
+            // cleanups registered in `r_read_console()`. Ideally we'd clean
+            // from here for symmetry.
```
DavisVaughan, line 1013: I guess an alternative is for `r_read_console()` to initialize a `mut bool is_debugging = false` that it passes through to here, where it gets flipped to `true`.

And then `r_read_console()` only resets the state if it sees that `is_debugging` is `true`.

That way you at least have a variable threaded through that connects the state setting to the state resetting

crates/ark/src/interface.rs
```diff
@@ -1345,20 +1436,47 @@ impl RMain {
...
+                // Remove disabled breakpoints. Their verification state is now stale since
+                // they weren't injected during this annotation. If the user re-enables
+                // them, they'll be treated as new unverified breakpoints.
+                if let Some(uri) = &uri {
+                    if let Some((_, bps)) = dap_guard.breakpoints.get_mut(uri) {
+                        bps.retain(|bp| !matches!(bp.state, BreakpointState::Disabled));
+                    }
+                }
+
+                drop(dap_guard);
```
DavisVaughan: I have a few comments about this section.

Two minor ones:

- When defining `breakpoints` above, I think you can start from `uri` instead of going through `loc` again, since all you need from it is the `uri`

- I feel like we are calling `dap_guard.breakpoints.get()` on the same `uri` a lot. Can we just reuse the `breakpoints` we already stored above?

One major one:

I feel quite uncertain about the ownership of `breakpoints` here. It seems like a lot of people can mutate the same object? Like the `breakpoints` ultimately live in one `HashMap` in the DAP state, but then we pass out a mutate handle to `PendingInputs::read()`, and we modify it here with `retain()`. Is that right? Do they all really need to share the same copy? Can `AnnotationRewriter` have a clone of the breakpoints instead?

One thing I'm very uncertain of is the fact that `PendingInputs::read()` gets a fixed mutable slice of breakpoints (with some fixed length), but then later on we _shorten_ the underlying vector with `bps.retain()`. That feels very sketchy to me! Is the mutable slice now incorrect??

crates/ark/src/interface.rs
```diff
@@ -1345,20 +1436,47 @@ impl RMain {
...
+                        dap_guard.notify_invalid_breakpoints(bps);
+                    }
+                }
+
+                // Remove disabled breakpoints. Their verification state is now stale since
+                // they weren't injected during this annotation. If the user re-enables
+                // them, they'll be treated as new unverified breakpoints.
+                if let Some(uri) = &uri {
+                    if let Some((_, bps)) = dap_guard.breakpoints.get_mut(uri) {
+                        bps.retain(|bp| !matches!(bp.state, BreakpointState::Disabled));
```
DavisVaughan: Instead of `get_mut()`, can you use the same `dap_guard.notify_*()` strategy used above to notify about disabled breakpoints as well? That seems like a nice pattern in general (allowing the DAP to mostly be the one in charge of breakpoint mutation seems good)

crates/ark/src/interface.rs
```diff
@@ -2562,6 +2689,10 @@ fn r_read_console_impl(
                     main.read_console_nested_return.set(false);
                 }
 
+                // We verify breakpoints _after_ evaluation is complete. An
+                // error will prevent verification.
+                main.verify_breakpoints(RObject::from(srcref));
```
DavisVaughan, line 2696: Hmm doesn't an `RObject` have a destructor?

Referring to the comment from above:

```
    // NOTE: Keep this function a "Plain Old Frame" without any
    // destructors. We're longjumping from here in case of interrupt.
```

crates/ark/src/console_debug.rs
```diff
@@ -358,3 +402,70 @@ fn as_frame_info(info: libr::SEXP, id: i64) -> Result<FrameInfo> {
...
+/// Verify a single breakpoint by ID.
+/// Called when a breakpoint expression is about to be evaluated.
+#[harp::register]
+pub unsafe extern "C-unwind" fn ps_verify_breakpoint(uri: SEXP, id: SEXP) -> anyhow::Result<SEXP> {
+    let uri: String = RObject::view(uri).try_into()?;
+    let id: String = RObject::view(id).try_into()?;
+
+    let Ok(uri) = Url::parse(&uri) else {
+        return Ok(libr::R_NilValue);
+    };
```
DavisVaughan: Should this be a `?` that would log?

crates/ark/src/console_debug.rs
```diff
@@ -358,3 +402,70 @@ fn as_frame_info(info: libr::SEXP, id: i64) -> Result<FrameInfo> {
...
+
+    Ok(libr::R_NilValue)
+}
+
+/// Verify breakpoints in the line range covered by a srcref.
+/// Called after each expression is successfully evaluated in source().
+#[harp::register]
+pub unsafe extern "C-unwind" fn ps_verify_breakpoints(srcref: SEXP) -> anyhow::Result<SEXP> {
+    let srcref = RObject::view(srcref);
+    RMain::get().verify_breakpoints(srcref.clone());
```
DavisVaughan: No `clone()` needed?

crates/ark/src/console_debug.rs
```diff
@@ -358,3 +402,70 @@ fn as_frame_info(info: libr::SEXP, id: i64) -> Result<FrameInfo> {
...
+    start_line: SEXP,
+    end_line: SEXP,
+) -> anyhow::Result<SEXP> {
+    let uri: String = RObject::view(uri).try_into()?;
+    let start_line: i32 = RObject::view(start_line).try_into()?;
+    let end_line: i32 = RObject::view(end_line).try_into()?;
+
+    let Ok(uri) = Url::parse(&uri) else {
+        return Ok(libr::R_NilValue);
+    };
```
DavisVaughan: Same question about wanting to use `?` instead for logging

crates/ark/src/modules/positron/utils.R
```diff
@@ -182,6 +182,27 @@ system_path <- function(pkg) {
...
+    }
+
+    # On Windows, paths like "C:/foo" need to become "file:///C:/foo"
+    # On Unix, paths like "/foo" need to become "file:///foo"
+    if (startsWith(path, "/")) {
+        paste0("file://", path)
+    } else {
+        paste0("file:///", path)
+    }
+}
```
DavisVaughan, line 205: I'm not sure what the best way to detect this is

I think I've seen some VS Code code that looks for "1 character drive letter plus a colon", i.e. `C` + `:` and used that as the way to detect windows, and anything else is unix-ey.

Do you think there is a chance we could have Unix paths come through that don't start with `/`?

crates/ark/src/modules/positron/debug.R
```diff
@@ -27,8 +22,15 @@ debugger_stack_info <- function(
...
-    # Top level call never has source references.
-    # It's what comes through the console input.
+    if (n == 0L) {
+        return(list(frame_info_from_srcref(
+            source_name = "<global>.R",
+            frame_name = "<global>",
+            srcref = context_srcref,
+            environment = NULL
+        )))
+    }
```
DavisVaughan, line 45: Is that really correct though? We have 0 frames, so what is the pseudo frame needed for?

Do you have an example?

crates/ark/src/modules/positron/debug.R
```diff
@@ -252,6 +251,10 @@ frame_info_from_srcref <- function(
         return(NULL)
     }
 
+    if (is_string(info$file)) {
+        source_name <- basename(info$file)
+    }
+
     new_frame_info(
         source_name = source_name,
         frame_name = frame_name,
```
DavisVaughan, line 275: I feel like the argument to this function should be `fallback_source_name`?

So it's like

```r
    if (is_string(info$file)) {
        source_name <- basename(info$file)
    } else {
        source_name <- fallback_source_name
    }
```

crates/ark/src/modules/positron/debug.R
```diff
@@ -700,3 +695,72 @@ non_parseable_pattern_info <- function(pattern, replacement) {
...
+is_breakpoint_enabled <- function(uri, id) {
+    .ps.Call("ps_is_breakpoint_enabled", uri, id)
+}
+
+verify_breapoint <- function(uri, id) {
+    .ps.Call("ps_verify_breakpoint", uri, id)
+}
+
+# Injected breakpoint. This receives a `browser()` call in the `expr` argument.
+# The argument if forced if the breakpoint is enabled. Since `expr` is promised
```
```suggestion
# The argument is forced if the breakpoint is enabled. Since `expr` is promised
```

crates/ark/src/modules/positron/debug.R
```diff
@@ -700,3 +695,72 @@ non_parseable_pattern_info <- function(pattern, replacement) {
...
+# in the calling frame environment, that environment is marked by R as being
+# debugged (with `SET_RDEBUG`), allowing to step through it. We're stopped in
+# the wrong frame (`.ark_breakpoint()`'s) but the console automatically steps to
+# the next expression whenever it detects that the current function (retrieved
+# with `sys.function()`) inherits from `ark_breakpoint`.
+#' @export
+.ark_breakpoint <- structure(
+    function(expr, uri, id) {
+        # Verify breakpoint right away, if not already the case We normally
```
```suggestion
        # Verify breakpoint right away, if not already the case. We normally
```

crates/ark/src/modules/positron/debug.R
```diff
@@ -700,3 +695,72 @@ non_parseable_pattern_info <- function(pattern, replacement) {
...
+# `source()`.
+.ark_verify_breakpoints_range <- function(uri, start_line, end_line) {
+    .ps.Call("ps_verify_breakpoints_range", uri, start_line, end_line)
+}
+
+debug_initialize <- function() {
+    # Store `.ark_breakpoint` and friends in base namespace so they're maximally
+    # reachable. We might want to do that for all symbols exported from the
+    # Ark/Positron namespace.
+    node_poke_cdr(as.symbol(".ark_annotate_source"), .ark_annotate_source)
```
DavisVaughan: `.ark_annotate_source` doesn't live here, should it? Or should there be another `source_initialize()`?

crates/ark/src/modules/positron/debug.R
```diff
@@ -700,3 +695,72 @@ non_parseable_pattern_info <- function(pattern, replacement) {
...
+#' @export
+.ark_auto_step <- function(expr) {
+    expr
+}
+
+# Verify breakpoints in a line range. Called after each top-level expression in
+# `source()`.
+.ark_verify_breakpoints_range <- function(uri, start_line, end_line) {
+    .ps.Call("ps_verify_breakpoints_range", uri, start_line, end_line)
+}
```
DavisVaughan, line 769: FWIW this one is missing the `@export` tag, and all the other `.ark_` prefixed ones have it

crates/ark/src/modules/positron/hooks_source.R
```diff
@@ -0,0 +1,133 @@
...
+        } else if (isFALSE(local)) {
+            .GlobalEnv
+        } else if (is.environment(local)) {
+            local
+        } else {
+            return(fall_back)
+        }
+
+        text <- paste(
+            readLines(uri, encoding = encoding, warn = FALSE),
```
DavisVaughan, line 100: It looks like you are implying that `encoding` is also an allowed option?

crates/ark/src/modules/positron/hooks_source.R
```diff
@@ -0,0 +1,133 @@
...
+
+        # Fall back if hook is disabled
+        if (!isTRUE(getOption("ark.source_hook", default = TRUE))) {
+            return(fall_back)
+        }
+
+        call <- match.call()
+
+        # Ignore `echo` and `local` arguments
+        call$echo <- NULL
```
DavisVaughan, line 75: You never actually branch on `echo` in our hook. What is this supposed to do in our code?

crates/ark/src/lsp/backend.rs
```diff
@@ -513,7 +516,7 @@ pub fn start_lsp(
         let (shutdown_tx, mut shutdown_rx) = tokio::sync::mpsc::channel::<()>(1);
 
         let init = |client: Client| {
-            let state = GlobalState::new(client);
+            let state = GlobalState::new(client, console_notification_tx.clone());
```
DavisVaughan:
```suggestion
            let state = GlobalState::new(client, console_notification_tx);
```

crates/ark/src/dap/dap_server.rs
```diff
@@ -297,19 +316,195 @@ impl<R: Read, W: Write> DapServer<R, W> {
...
-            thread_id: Some(THREAD_ID),
-            preserve_focus_hint: Some(false),
-            text: None,
-            all_threads_stopped: None,
-            hit_breakpoint_ids: None,
-        }))
+        self.send_event(Event::Thread(ThreadEventBody {
+            reason: ThreadEventReason::Started,
+            thread_id: THREAD_ID,
+        }));
```
DavisVaughan, line 510: Were we sending the wrong thing before?

crates/ark/src/dap/dap_server.rs
```diff
@@ -297,19 +316,195 @@ impl<R: Read, W: Write> DapServer<R, W> {
...
+    // content against our stored hash to detect if breakpoints are now stale.
+    //
+    // Key implementation details:
+    // - We use `original_line` for lookup since the frontend doesn't know about
+    //   our line adjustments and always sends back the original line numbers.
+    // - When a user unchecks a breakpoint, it appears as a deletion (omitted
+    //   from the request). We preserve verified breakpoints as Disabled so we
+    //   can restore their state when re-enabled without requiring re-sourcing.
+    fn handle_set_breakpoints(&mut self, req: Request, args: SetBreakpointsArguments) {
+        let path = args.source.path.clone().unwrap_or_default();
```
DavisVaughan: I think you can avoid the `clone()` and the confusing `unwrap_or_default()` by doing something like this


```rust
        let Some(path) = args.source.path.as_ref() else {
            log::warn!("Missing a path to set breakpoints for.");
            let rsp = req.error(&format!("Missing a path to set breakpoints for"));
            self.respond(rsp);
            return;
        };
```

you don't seem to need to own or mutate the `path`

you also seem to rely on a `path` being here (the other option is that `source.sourceReference` would be set? this never occurs?), so it feels better to be explicit about the missing case rather than letting the error below of "Failed to convert path to URI" trigger, which wouldn't make much sense when it is missing and you just filled in the "default" of an empty string.

crates/ark/src/dap/dap_server.rs
```diff
@@ -297,19 +316,195 @@ impl<R: Read, W: Write> DapServer<R, W> {
...
+        let doc_content = match std::fs::read_to_string(&path) {
+            Ok(content) => content,
+            Err(err) => {
+                // TODO: What do we do with breakpoints in virtual documents?
+                log::error!("Failed to read file '{path}': {err}");
+                let rsp = req.error(&format!("Failed to read file: {path}"));
+                self.respond(rsp);
+                return;
+            },
+        };
```
DavisVaughan, line 364: Yea this might be related `source.sourceReference` instead of `source.path`?

We might get a `sourceReference` id that I think maps into our map of virtual documents? And we've got the contents of them stored somewhere already? I don't fully remember.

crates/ark/src/dap/dap_server.rs
```diff
@@ -297,19 +316,195 @@ impl<R: Read, W: Write> DapServer<R, W> {
...
+                        original_line: line,
+                        state: BreakpointState::Unverified,
+                        injected: false,
+                    }
+                })
+                .collect()
+        } else {
+            log::trace!("DAP: Document unchanged for {uri}, preserving breakpoint states");
+
+            // Unwrap Safety: `doc_changed` is false, so `existing_breakpoints` is Some
```
DavisVaughan:
```suggestion
            // Unwrap Safety: `doc_changed` is false, so `old_breakpoints` is Some
```

crates/ark/src/dap/dap_server.rs
```diff
@@ -297,19 +316,195 @@ impl<R: Read, W: Write> DapServer<R, W> {
...
+            new_breakpoints
+        );
+
+        state
+            .breakpoints
+            .insert(uri, (doc_hash, new_breakpoints.clone()));
+
+        drop(state);
+
+        let response_breakpoints: Vec<dap::types::Breakpoint> = new_breakpoints
```
DavisVaughan: Can probably avoid the `clone()` if you build `response_breakpoints()` first.

crates/ark/src/dap/dap.rs
```diff
@@ -158,12 +250,15 @@ impl Dap {
         self.is_debugging = false;
 
         if self.is_connected {
```
DavisVaughan, line 252: Do you want something like `reset_variables_reference_count()` but for the `current_breakpoint_id`?

crates/ark/src/console_annotate.rs
```diff
@@ -1,63 +1,914 @@
...
+//
+// ```r
+// {
+// #line 1 "file:///file.R"
+// 1
+// base::.ark_auto_step(base::.ark_verify_breakpoints_range("file:///test.R", 1L, 2L))
+// #line 2 "file:///file.R"
+// 2
+// base::.ark_auto_step(base::.ark_verify_breakpoints_range("file:///test.R", 2L, 3L))
+// }
```
DavisVaughan, line 122:
```suggestion
// #line 1 "file:///file.R"
// 1
// base::.ark_auto_step(base::.ark_verify_breakpoints_range("file:///file.R", 1L, 2L))
// #line 2 "file:///file.R"
// 2
// base::.ark_auto_step(base::.ark_verify_breakpoints_range("file:///file.R", 2L, 3L))
// }
```

Same or different files?

crates/ark/src/console_annotate.rs
```diff
@@ -1,63 +1,914 @@
...
+// 2
+// ```
+//
+// The instrumented version ends up as:
+//
+// ```r
+// {
+// #line 1 "file:///file.R"
+// 1
+// base::.ark_auto_step(base::.ark_verify_breakpoints_range("file:///test.R", 1L, 2L))
```
DavisVaughan: The documentation from above suggests that `.ark_auto_step()` should come _before_ the `#line` directive?

crates/ark/src/console_annotate.rs
```diff
@@ -1,63 +1,914 @@
...
+pub(crate) fn annotate_input(
+    code: &str,
+    location: CodeLocation,
+    breakpoints: Option<&mut [Breakpoint]>,
+) -> anyhow::Result<String> {
+    // First, inject breakpoints into the original code. This must happen before
+    // we add the outer line directive, otherwise the coordinates of inner line
+    // directives are shifted by 1 line.
+    let annotated_code = if let Some(breakpoints) = breakpoints {
+        let root = aether_parser::parse(code, Default::default()).tree();
```
DavisVaughan: Oh one thing that is missing here is that you aren't checking the `error` status of the `Parse` result

If we see a parse error, we should not try and annotate anything (i think).

Otherwise you're just asking for chaos!

crates/ark/src/console_annotate.rs
```diff
@@ -1,63 +1,914 @@
...
+    breakpoints: &mut [Breakpoint],
+) -> anyhow::Result<String> {
+    // Wrap code in braces first. This:
+    // 1. Allows R to step through top-level expressions
+    // 2. Makes all breakpoints valid (they're now inside braces, at top-level they'd be invalid)
+    // This enables uniform treatment by `AnnotationRewriter` for input and source cases.
+    let wrapped = format!("{{\n{code}\n}}");
+    let line_index = LineIndex::new(&wrapped);
+
+    let root = aether_parser::parse(&wrapped, Default::default()).tree();
```
DavisVaughan: Same here about bailing on parse error somehow

crates/ark/src/console_annotate.rs
```diff
@@ -1,63 +1,914 @@
...
+}
+
+/// Rewriter that handles all code annotation inside braced expression lists:
+/// - Breakpoint calls on statements
+/// - Verification calls after statements containing breakpoints
+/// - `#line` directives after injected calls to restore sourceref bookkeeping
+struct AnnotationRewriter<'a> {
+    uri: &'a Url,
+    /// Breakpoints in document coordinates, will be mutated to mark invalid ones
+    breakpoints: &'a mut [Breakpoint],
```
DavisVaughan:
```suggestion
    /// Breakpoints in document coordinates, will be mutated to mark invalid ones,
    /// which is propagated back through the DAP to the frontend
    breakpoints: &'a mut [Breakpoint],
```

?

crates/ark/src/console_annotate.rs
```diff
@@ -1,63 +1,914 @@
...
+        // (partially) rebuilt tree with invalid line information.
+        //
+        // Note: we intentionally cache line info from the original parse tree.
+        // Downstream injections (breakpoint calls, `#line` trivia, verify calls)
+        // can change token offsets and make line lookups on the rebuilt nodes
+        // unreliable, but they do not change where the original expressions
+        // lived in the source. We only rely on these cached source positions
+        // when anchoring and invalidating breakpoints, tasks for which we need
+        // the _original_ coordinates, not the new ones.
+        if let Some(braced) = RBracedExpressions::cast(node.clone()) {
```
DavisVaughan:
```suggestion
        if let Some(braced) = RBracedExpressions::cast_ref(&node) {
```

This has the cool property of only cloning the node if the condition is hit, should be much cheaper in the hot path of the code walker

crates/ark/src/console_annotate.rs
```diff
@@ -1,63 +1,914 @@
...
+        if self.err.is_some() {
+            // Something is wrong but we can't short-circuit the visit. Just
+            // visit nodes until exhaustion.
+            return node;
+        }
+
+        // Only process expression lists
+        if node.kind() != RSyntaxKind::R_EXPRESSION_LIST {
+            return node;
+        }
```
DavisVaughan, line 530: Oh should this be up in `visit_node()` as well? Then you can use `RBracedExpressions::unwrap_cast(syntax)` and flatten a layer of nesting

crates/ark/src/console_annotate.rs
```diff
@@ -1,63 +1,914 @@
...
+    if let Some((_, breakpoints)) = dap_guard.breakpoints.get(&uri) {
+        dap_guard.notify_invalid_breakpoints(breakpoints);
+    }
+
+    // Remove disabled breakpoints. Their verification state is now stale since
+    // they weren't injected during this annotation. If the user re-enables
+    // them, they'll be treated as new unverified breakpoints.
+    if let Some((_, breakpoints)) = dap_guard.breakpoints.get_mut(&uri) {
+        breakpoints.retain(|bp| !matches!(bp.state, BreakpointState::Disabled));
+    }
```
DavisVaughan, line 920: This looks like another place where I naively expected to see another `dap_guard.notify_*()` instead

crates/ark/src/interface.rs
lionel-: It's an async loop processed by Console at interrupt/idle time.

crates/ark/src/interface.rs
lionel-: It's past the jumpy area of the function.

crates/ark/src/interface.rs
lionel-: I kept the current setup, which I think is more conventional (e.g. similar to rlang and vctrs's C files init) but consolidated it to `modules::initialize()`. The errors module init now also follows same pattern, although it required a different less safe invocation path to avoid condition handlers on the stack. This should nicely address the issues you brought up.

crates/ark/src/interface.rs
lionel-: I just tried this but didn't find it much clearer or cleaner. I found that stopping unconditionally is simpler overall.

crates/ark/src/interface.rs
lionel-: oh we did have a preexisting instance of this pattern. `register_hooks()` is now `initialize_hooks()`, and called from `initialize()`.

crates/ark/src/interface.rs
lionel-: > I feel like we are calling dap_guard.breakpoints.get() on the same uri a lot. Can we just reuse the breakpoints we already stored above?

Not without getting into borrow checker issues because of the mut ref held while trying to get another ref.

> I feel quite uncertain about the ownership of breakpoints here. It seems like a lot of people can mutate the same object? Like the breakpoints ultimately live in one HashMap in the DAP state, but then we pass out a mutate handle to PendingInputs::read(), and we modify it here with retain(). Is that right? Do they all really need to share the same copy? Can AnnotationRewriter have a clone of the breakpoints instead?

We modify the breakpoints state from multiple places (e.g. read might invalidate breakpoints, eval might verify them) but I think Rust's strong ownership makes this safe and easy to reason about. I don't particularly feel the need to go for a functional approach here.

> One thing I'm very uncertain of is the fact that PendingInputs::read() gets a fixed mutable slice of breakpoints (with some fixed length), but then later on we shorten the underlying vector with bps.retain(). That feels very sketchy to me! Is the mutable slice now incorrect??

The context with the slice has ended by then, and we're able to safely modify the vector.

crates/ark/src/interface.rs
lionel-: Good idea. I've also streamlined the signature so the DAP method retrieves its own breakpoints, this simplifies the call sites.

crates/ark/src/console_debug.rs
lionel-: hmm I think that might show the error in the console. I'll use the new `log_err()` method instead to get:

```rs
    let Some(uri) = Url::parse(&uri).log_err() else {
        return Ok(libr::R_NilValue);
    };
```

crates/ark/src/modules/positron/utils.R
lionel-: I've now tweaked Windows drive detection along the lines suggested.

> Do you think there is a chance we could have Unix paths come through that don't start with /?

I don't think so because of the `normalizePath()` call just above.

crates/ark/src/modules/positron/debug.R
lionel-: I think calling `browser()` at top-level will trigger this.

crates/ark/src/modules/positron/hooks_source.R
lionel-: We could pass the default `getOption("encoding")` directly but then there'd be a repetition.

crates/ark/src/modules/positron/hooks_source.R
lionel-: It would be hard to support it so currently it's just ignored in the injection code path. I've added a note about this.

crates/ark/src/dap/dap_server.rs
lionel-: We're using our own mechanism for virtual docs, so there are no DAP source references (i.e. vdoc handles) supported by Ark.

https://github.com/posit-dev/ark/pull/852

crates/ark/src/dap/dap_server.rs
lionel-: With the permanent DAP session we're now starting in Running state rather Stopped state.

crates/ark/src/dap/dap.rs
lionel-: Probably not since breakpoints persist across DAP sessions.

crates/ark/src/console_annotate.rs
lionel-: Breakpoints verification comes after each top-level expression.

crates/ark/src/console_annotate.rs
lionel-: ~It's not really possible to get here with parse errors (or it means that our parser failed when R's didn't) but I'll add a defensive check.~

hmm it's definitively possible, I was confused!

lionel-: I'll also add a fallback handling in `read()` when annotation fails.

crates/ark/src/console_annotate.rs
lionel-: Good catches, this is now:

Screenshot 2026-01-22 at 16 51 09

PR #1004, opened 2026-01-20
Add viewer support for database connections in Variables Pane

dfalbel:
The PR adds the `ark_positron_variable_view` method, which is called by the variables pane upon clicking on the view button. Packages can add support for a view action for any object that appears in the variables pane by implementing `ark_positron_has_viewer` and `ark_positron_variable_view`, they can use `ark_positron_variable_kind` to customize the viewer button icon.

Additionally, it makes the connections pane focus when a connection is opened. 

Adresses:

- https://github.com/posit-dev/positron/issues/8991
- https://github.com/posit-dev/positron/issues/11357

The PR implements support for odbc and bigrquery. Ideally those methods would live in their packages. But it will require new package releases to really fix  https://github.com/posit-dev/positron/issues/11357 - which is somewhat urgent. Happy to move the implementation there though, if we think it's bad to keep it here.

I found other packages that use the same addTaskCallback pattern:

- [dplyr.teradata](https://github.com/hoxo-m/dplyr.teradata/blob/c19c589f1bee6b90459505a855cb896947efb803/R/Driver.R#L134-L153)
- [looker-R](https://github.com/AsobaCloud/looker-R/blob/95c1f0027c6d1d193d0d354a16e13c811fae7883/R/LookerSDK.R#L143-L147) (not on CRAN)

### DIFF
--- a/crates/ark/src/modules/positron/connection.R
+++ b/crates/ark/src/modules/positron/connection.R
@@ -21,8 +21,7 @@
     .ps.Call("ps_connection_updated", id)
 }
 
-#' @export
-.ps.connection_focus <- function(id) {
+connection_focus <- function(id) {
     .ps.Call("ps_connection_focus", id)
 }
 
@@ -50,7 +49,7 @@
         for (id in ls(envir = connections)) {
             con <- get(id, envir = connections)
             if (identical(con$host, host) && identical(con$type, type)) {
-                .ps.connection_focus(id)
+                connection_focus(id)
                 return(invisible(id))
             }
         }
@@ -73,7 +72,7 @@
             # until the end of the connection.
             objectTypes = connection_flatten_object_types(listObjectTypes())
         )
-        .ps.connection_focus(id)
+        connection_focus(id)
         invisible(id)
     }
 
@@ -178,7 +177,7 @@ connection_flatten_object_types <- function(object_tree) {
 }
 
 # Helper to reconstruct ODBC connection code from connection info
-.ps.odbc_connection_code <- function(info) {
+odbc_connection_code <- function(info) {
     # Try to reconstruct a reasonable connection string
     # Priority: DSN > connection string parameters
     if (!is.null(info$sourcename) && nzchar(info$sourcename)) {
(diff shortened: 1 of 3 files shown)


crates/ark/src/modules/positron/connection.R
```diff
@@ -170,6 +177,36 @@ connection_flatten_object_types <- function(object_tree) {
     return(TRUE)
 }
 
+# Helper to reconstruct ODBC connection code from connection info
+.ps.odbc_connection_code <- function(info) {
```
lionel-: This is not exported, so no need for `.ps.` prefix

crates/ark/src/modules/positron/connection.R
```diff
@@ -21,6 +21,11 @@
     .ps.Call("ps_connection_updated", id)
 }
 
+#' @export
+.ps.connection_focus <- function(id) {
```
lionel-: I see this is used in internal handlers below. Does it also need to be exported?

crates/ark/src/modules/positron/connection.R
```diff
@@ -283,6 +320,107 @@ connection_flatten_object_types <- function(object_tree) {
...
+                # Use bigrquery's built-in connection observer integration
+                bigrquery:::on_connection_opened(x, code = code)
+                invisible(TRUE)
+            }
+        )
+    }
+)
+
+# Helper to reconstruct BigQuery connection code
+.ps.bigrquery_connection_code <- function(con) {
```
lionel-: No need for export prefix

crates/ark/src/variables/variable.rs
```diff
@@ -608,10 +604,30 @@ fn has_viewer(value: SEXP) -> bool {
...
+        // No method found, fall back to default logic for data frames/matrices
+        Ok(None) => r_is_data_frame(value) || r_is_matrix(value),
+    }
+}
+
+/// Try to view an object using a custom view method.
+/// This dispatches to the `ark_positron_variable_view` method.
+/// Returns Ok(true) if a custom view was handled, Ok(false) if no method found,
+/// or Err if the method failed.
+pub fn try_custom_view(value: SEXP) -> anyhow::Result<bool> {
```
lionel-: `try_dispatch_view()`? For consistency with `try_dispatch()`.

crates/ark/src/modules/positron/connection.R
dfalbel: Correct, no need to export it

Issue #1006, opened 2026-01-20
Respect `browserNLdisabled` in our custom parse/eval loop so `Enter` advances the debugger


DavisVaughan:
https://github.com/posit-dev/positron/issues/11440

- `debugonce(dplyr:::mutate.data.frame)`
- `dplyr::mutate(mtcars, x = 1)`
- Press `Enter` while in a blank console, notice how the debugger does not advance

@juliasilge has confirmed that this used to advance the debugger. It is very likely that https://github.com/posit-dev/ark/pull/960 is what changed this behavior.

In RStudio, and following the intention of `browserNLdisabled`, if you follow the instructions above then `Enter` in a blank console is supposed to count as `n` when in `debug()` mode.
https://github.com/wch/r-source/blob/ca917f5f467e9dd6ee4b867deaabff1f78a8b3e2/src/main/main.c#L232-L237

It seems like we are missing out on this branch, or maybe we have to implement it manually. It seems like a nice feature to ensure we support.

After manually adding support for the `Enter` feature, we'd also want to make sure that `options(browserNLdisabled = TRUE)` turns this back _off_

PR #1010, opened 2026-01-26
Interpret empty lines as `n` debug command

lionel-:
Addresses https://github.com/posit-dev/ark/issues/1006

### DIFF
--- a/crates/ark/src/interface.rs
+++ b/crates/ark/src/interface.rs
@@ -140,6 +140,12 @@ use crate::ui::UiCommSender;
 pub static CAPTURE_CONSOLE_OUTPUT: AtomicBool = AtomicBool::new(false);
 static RE_DEBUG_PROMPT: Lazy<Regex> = Lazy::new(|| Regex::new(r"Browse\[\d+\]").unwrap());
 
+/// All debug commands as documented in `?browser`
+const DEBUG_COMMANDS: &[&str] = &["c", "cont", "f", "help", "n", "s", "where", "r", "Q"];
+
+// The subset of debug commands that continue execution
+const DEBUG_COMMANDS_CONTINUE: &[&str] = &["n", "f", "c", "cont"];
+
 /// An enum representing the different modes in which the R session can run.
 #[derive(PartialEq, Clone, Copy)]
 pub enum SessionMode {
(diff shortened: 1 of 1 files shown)


crates/ark/src/interface.rs
DavisVaughan: From an interpretation perspective, this is a rare case where I think more nesting might be clearer.

The first layer of nesting handles "do we have _any_ input?"

The second layer of nesting in the `else` handles "ok, what do we do with no input?"

```rust
if there_is_an_input {
  handle_pending_input(input)
} else {
  // Otherwise we got an empty input, e.g. `""`
  if self.debug_is_debugging && !harp::options::get_option_bool("browserNLdisabled") {
    // Treat empty input as `n` while in the debugger
    Some(self.debug_forward_continue_command(buf, buflen, String::from("n")))
  } else {
    // Close active request
    self.handle_active_request(info, ConsoleValue::Success(Default::default()));
    // And return to event loop
    None
  }
}
```

Currently it feels like that middle `self.debug_is_debugging` branch is different from the "we have no input case", which mildly confused me. Is it just me?

DavisVaughan: Hmmmmmmmmmm, I was a bit confused by the number of places we are calling `on_console_input()` from now

I feel like this might be clearer if you go ahead and also move `if DEBUG_COMMANDS_CONTINUE.contains(&&sym[..])` into `debug_forward_continue_command()` and call it `debug_forward_command` instead

```rust
fn debug_forward_command(
        &mut self,
        buf: *mut c_uchar,
        buflen: c_int,
        cmd: String,
        continue: bool
    ) -> ConsoleResult {
      if DEBUG_COMMANDS_CONTINUE.contains(&cmd) {
          // For continue-like commands, we do not preserve focus,
          // i.e. we let the cursor jump to the stopped
          // position. Set the preserve focus hint for the
          // next iteration of ReadConsole.
          self.debug_preserve_focus = false;

          // Let the DAP client know that execution is now continuing
          self.debug_send_dap(DapBackendEvent::Continued);
      }

        // Forward the command to R's base REPL.
        // Unwrap safety: A debug command fits in the buffer.
        Self::on_console_input(buf, buflen, cmd).unwrap();
        ConsoleResult::NewInput
}
```

I feel like the less places we call `Self::on_console_input()`, the better, in terms of just understanding it

lionel-: That makes sense!

PR #1016, opened 2026-01-27
Fix breakpoint issues on Windows

lionel-:
Addresses https://github.com/posit-dev/positron/issues/11558

This fixes two issues:

- We were passing file URIs to `normalizePath()`. On Unix this is fine, it doesn't find the path on disk and so returns the input since `mustWork` is FALSE:

  ```r
  normalizePath("file:///home/user/test.R", mustWork = FALSE)
  #> [1] "file:///home/user/test.R"
  ```

  But on Windows this produces paths like `C:\Users\lione\Projects\dplyr\file:\C:\Users\lione\OneDrive\Desktop\test.R`

  To fix this, we detect all URIs rather than just Ark ones.

- On Windows, Positron sends encoded file URIs via to the LSP server and via execute-request code locations which look like `c%03` instead of `C:`. These URIs didn't match the ones we constructed from file paths in the DAP. To fix this, we now normalise the URIs coming from the LSP (did change events for breakpoint invalidation) and from execute requests (so cmd+enter may validate breakpoints).

  Ideally we'd sanitize all URIs at the DAP/LSP/Jupyter boundaries to a common encoding, but such a large change felt too risky at this point. We will be able to do this when we introduce the notion of a `FileId`.

https://github.com/user-attachments/assets/bda40e71-75f8-4d9c-b62d-afc91143542a

### DIFF
--- a/crates/ark/src/url.rs
+++ b/crates/ark/src/url.rs
@@ -49,11 +49,14 @@ impl ExtUrl {
 
         // Round-trip through filesystem path to get canonical form.
         // This decodes URL-encoded characters like %3A -> :
-        let Ok(path) = uri.to_file_path() else {
+        let Some(uri) = uri
+            .to_file_path()
+            .ok()
+            .and_then(|path| Url::from_file_path(&path).ok())
+        else {
+            log::warn!("Failed to normalize file URI: {uri}");
             return uri;
         };
-
-        let uri = Url::from_file_path(&path).unwrap_or(uri);
         uppercase_windows_drive_in_uri(uri)
     }
 


crates/ark/src/sys/windows/traps.rs
```diff
@@ -9,7 +9,13 @@ use crate::traps::backtrace_handler;
...
-        libc::signal(libc::SIGSEGV, backtrace_handler as libc::sighandler_t);
-        libc::signal(libc::SIGILL, backtrace_handler as libc::sighandler_t);
+        libc::signal(
+            libc::SIGSEGV,
+            backtrace_handler as *const () as libc::sighandler_t,
+        );
+        libc::signal(
+            libc::SIGILL,
+            backtrace_handler as *const () as libc::sighandler_t,
+        );
```
DavisVaughan, line 19: Needed? New windows rust check?

crates/ark/src/url.rs
```diff
@@ -0,0 +1,197 @@
...
+            return uri;
+        }
+
+        // Round-trip through filesystem path to get canonical form.
+        // This decodes URL-encoded characters like %3A -> :
+        let Ok(path) = uri.to_file_path() else {
+            return uri;
+        };
+
+        let mut uri = Url::from_file_path(&path).unwrap_or(uri);
```
DavisVaughan: Can we do a consistent thing on failure?

It doesn't feel like this should fail so maybe if either of these fail we consistently do a `log::warn!()` and return the `uri` as is in both cases?

crates/ark/src/url.rs
```diff
@@ -0,0 +1,197 @@
...
+    /// No-op on non-Windows platforms.
+    #[cfg(not(windows))]
+    pub fn normalize(uri: Url) -> Url {
+        uri
+    }
+}
+
+/// Uppercase the drive letter in a Windows file URI for consistent hashing.
+#[cfg(windows)]
+fn uppercase_windows_drive_in_uri(uri: &mut Url) {
```
DavisVaughan: It feels way more Rusty to me to just take `uri: Url` and return `Url`. i.e. do the full ownership transfer. Same end result, but feels more pure to me?

crates/ark/src/url.rs
```diff
@@ -0,0 +1,197 @@
...
+    let is_windows_path = matches!(
+        (chars.next(), chars.next(), chars.next()),
+        (Some('/'), Some(drive), Some(':')) if drive.is_ascii_alphabetic()
+    );
+
+    if !is_windows_path {
+        return;
+    }
+
+    let drive = path.chars().nth(1).unwrap();
```
DavisVaughan:
```suggestion
    // Unwrap: Checked existence of `drive` before this
    let drive = path.chars().nth(1).unwrap();
```

DavisVaughan: (You could likely avoid the `unwrap()` altogether by turning the `matches!()` call above into a true `match`, and then putting the code below this in the branch that matches on `(Some('/'), Some(drive), Some(':')) if drive.is_ascii_alphabetic()`. that would probably be my preference and would be more clear)

crates/ark/src/url.rs
```diff
@@ -0,0 +1,197 @@
...
+    let upper = drive.to_ascii_uppercase();
+
+    if drive != upper {
+        let new_path = format!("/{upper}{}", &path[2..]);
```
DavisVaughan:
```suggestion
        let new_path = format!("/{upper}:{}", &path[3..]);
```

am i crazy in thinking its clearer to just go ahead and write in everything you "know" to be there?

crates/ark/src/sys/windows/traps.rs
lionel-: Just a drive by fix for Windows-only compilation warnings

crates/ark/src/url.rs
lionel-: That's one way to do Rust yes :)
Note that this has more risks of unnecessarily copying the value, not that it matters here. Happy to change it.

PR #1018, opened 2026-01-28
Shut down DAP server gracefully on disconnection

lionel-:
Branched on #1016.

Follow up to https://github.com/posit-dev/positron/pull/11407 and https://github.com/posit-dev/positron/pull/11497 where we disabled the exthost restart test on Windows (Slack discussion at https://positpbc.slack.com/archives/C06DNFJSHPD/p1769113239000079).

Not sure what this only surfaced on Windows, but the cause was an Ark crash due to sloppy Unwraps on connection events. This PR makes sure connection errors are propagated, logged, and cause the DAP server to shut down gracefully.

It was a bit tricky to debug because the logs were not showing the crash. The only clue that surfaced was a message in the Console:

Screenshot 2026-01-28 at 12 07 27

and these messages in the Supervisor logs:

```log
10:00:05 [DEBUG] (2) kcserver::kernel_session::process: End of output stream (kind: Stderr)
10:00:05 [DEBUG] (18) kcserver::kernel_session::process: End of output stream (kind: Stdout)
10:00:05 [INFO] Child process for session r-68064bdf exited with status: exit code: 0xc0000409
10:00:05 [DEBUG] (17) kcserver::kernel_state: [session r-68064bdf] status 'idle' => 'exited' (child process exited)
10:00:05 [DEBUG] (17) kcserver::heartbeat: [session r-68064bdf] Stopping heartbeat monitor (exit event signaled).
```

`0xc0000409` is the code for `STATUS_STACK_BUFFER_OVERRUN`.

Attaching Ark with a debugger then restarting the exthost gave enough time to the process to emit log messages before exit, revealing the DAP panic.

### DIFF
--- a/crates/ark/src/dap/dap_server.rs
+++ b/crates/ark/src/dap/dap_server.rs
@@ -202,7 +202,7 @@ fn listen_dap_events<W: Write>(
 
                 let mut output = output.lock().unwrap();
                 if let Err(err) = output.send_event(event) {
-                    log::warn!("DAP: Failed to send event, closing: {err}");
+                    log::warn!("DAP: Failed to send event, closing: {err:?}");
                     return;
                 }
             },
@@ -246,7 +246,7 @@ impl<R: Read, W: Write> DapServer<R, W> {
             Ok(Some(req)) => req,
             Ok(None) => return false,
             Err(err) => {
-                log::warn!("DAP: Connection closed: {err}");
+                log::warn!("DAP: Connection closed: {err:?}");
                 return false;
             },
         };
@@ -291,7 +291,7 @@ impl<R: Read, W: Write> DapServer<R, W> {
         };
 
         if let Err(err) = result {
-            log::warn!("DAP: Handler failed, closing connection: {err}");
+            log::warn!("DAP: Handler failed, closing connection: {err:?}");
             return false;
         }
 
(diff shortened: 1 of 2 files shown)


crates/ark/src/dap/dap_server.rs
```diff
@@ -150,9 +151,18 @@ fn listen_dap_events<W: Write>(
     loop {
         select!(
             recv(backend_events_rx) -> event => {
+                let event = match event {
+                    Ok(event) => event,
+                    Err(err) => {
+                        // Channel closed, sender dropped
+                        log::info!("DAP: Event channel closed: {err:?}");
+                        return;
```
DavisVaughan, line 159: So is it getting dropped on the sender side when the ext host restarts suddenly?

crates/ark/src/dap/dap_server.rs
```diff
@@ -191,7 +201,10 @@ fn listen_dap_events<W: Write>(
                 };
 
                 let mut output = output.lock().unwrap();
-                output.send_event(event).unwrap();
+                if let Err(err) = output.send_event(event) {
+                    log::warn!("DAP: Failed to send event, closing: {err}");
```
DavisVaughan: You use `{err:}?` in some places, but not consistently. Maybe `{err:?}` everywhere? (More than just this one place)

crates/ark/src/dap/dap_server.rs
lionel-: I don't think we've hit that code path, it's a defensive fix. It would happen if the DAP comm closed before the DAP connection I think.

PR #1019, opened 2026-01-28
Fix `source()` hook fallback with `local = TRUE`

lionel-:
@DavisVaughan This fixes the `devtools::test()` issue in dplyr. The fallback path wasn't properly forwarding the local environment when `local = TRUE` was passed.

### DIFF
--- a/crates/ark/src/modules/positron/hooks_source.R
+++ b/crates/ark/src/modules/positron/hooks_source.R
@@ -38,7 +38,8 @@ make_ark_source <- function(original_source) {
         # fallback calls.
         use_file <- missing(exprs)
 
-        # Capture environment early if local evaluation is requested
+        # Capture environment early if local evaluation is requested.
+        # This is necessary if we have to fallback when `local = TRUE`.
         if (isTRUE(local)) {
             local <- parent.frame()
         }
@@ -102,10 +103,9 @@ make_ark_source <- function(original_source) {
         }
 
         env <- if (isTRUE(local)) {
-            # That would be very unexpected since we captured the local
-            # environment in `local` when set to `TRUE`. But just for peace of
-            # mind we handle it here.
-            parent.frame()
+            stop(
+                "Internal error: `local = TRUE` should have been converted to an environment above."
+            )
         } else if (isFALSE(local)) {
             .GlobalEnv
         } else if (is.environment(local)) {


crates/ark/src/modules/positron/hooks_source.R
```diff
@@ -38,6 +38,11 @@ make_ark_source <- function(original_source) {
         # fallback calls.
         use_file <- missing(exprs)
 
+        # Capture environment early if local evaluation is requested
+        if (isTRUE(local)) {
+            local <- parent.frame()
+        }
```
DavisVaughan, line 45: Looking at `base::source()`, I see

```r
envir <- if (isTRUE(local)) 
        parent.frame()
    else if (isFALSE(local)) 
        .GlobalEnv
    else if (is.environment(local)) 
        local
    else stop("'local' must be TRUE, FALSE or an environment")
```

So, I would be in favor of moving your branch from below up as

```r
        # Handle `local` early so we capture the correct parent frame
        env <- if (isTRUE(local)) {
            parent.frame()
        } else if (isFALSE(local)) {
            .GlobalEnv
        } else if (is.environment(local)) {
            local
        } else {
            # Same error as `base::source()`
            stop("'local' must be TRUE, FALSE or an environment")
        }
```

Then you only use `env` from there on out. And you pass `env` to the fallback path.

lionel-: This has the inconvenient that we are reproducing more of the base behaviour (throwing an error) ourselves.

crates/ark/src/modules/positron/hooks_source.R
```diff
@@ -38,31 +38,46 @@ make_ark_source <- function(original_source) {
         # fallback calls.
         use_file <- missing(exprs)
 
+        # Capture environment early if local evaluation is requested
+        if (isTRUE(local)) {
```
DavisVaughan:
```suggestion
        # Capture environment early if local evaluation is requested.
        # This is necessary if we have to fallback when `local = TRUE`.
        if (isTRUE(local)) {
```

crates/ark/src/modules/positron/hooks_source.R
```diff
@@ -87,6 +102,9 @@ make_ark_source <- function(original_source) {
         }
 
         env <- if (isTRUE(local)) {
+            # That would be very unexpected since we captured the local
+            # environment in `local` when set to `TRUE`. But just for peace of
+            # mind we handle it here.
             parent.frame()
```
DavisVaughan:
```suggestion
            stop("Internal error: `local = TRUE` should have been converted to an environment above.")
```

This is a case where it feels way less confusing to me to just assert our invariants.

6 months from now me will be able to more easily match up this error to the `parent.frame()` handling above

DavisVaughan: It's like an `// Unwrap:` safety comment

PR #1021, opened 2026-01-29
Fix virtual document names in release builds too

lionel-:
Addresses https://github.com/posit-dev/positron/issues/11591

I had actually fixed this in https://github.com/posit-dev/ark/pull/1003 by adding (what I thought was) a new `as_label()` function extracted from rlang:

https://github.com/posit-dev/ark/blob/8c819ef1431596553dc9162e9cbef760c79c42af/crates/ark/src/modules/positron/calls_deparse.R#L12

Turns out, there was already a much more bare bones `as_label()` in the repo: https://github.com/posit-dev/ark/blob/8c819ef1431596553dc9162e9cbef760c79c42af/crates/ark/src/modules/positron/utils.R#L66

Here's the thing:

- When loading the R files from disk, as happens in a _debug_ build, the new `as_label()` overwrote the old one. So my new implementation was working as expected.

- When loading the R files from the binary blob, as happens in a _release_ build, rust-embed loads the files alphabetically (https://docs.rs/crate/rust-embed-utils/8.11.0/source/src/lib.rs#19), and the new impl gets overwritten by the old one!

This had me very confused for a while!

---

This PR removes the dangling `as_label()`, and adjusts the modules loading mechanism to match rust-embed so this doesn't happen again in the future.

### DIFF
--- a/crates/ark/src/modules.rs
+++ b/crates/ark/src/modules.rs
@@ -151,9 +151,7 @@ pub fn initialize() -> anyhow::Result<RObject> {
     }
 
     // Finish initialization of modules
-    RFunction::from("initialize")
-        .call_in(namespace.sexp)
-        .log_err();
+    RFunction::from("initialize").call_in(namespace.sexp)?;
 
     // Do this separately with a bare eval because `errors_initialize()` should
     // be called without any condition handlers on the stack
--- a/crates/ark/src/console.rs
+++ b/crates/ark/src/console.rs
@@ -648,10 +648,12 @@ impl Console {
                     .log_err();
             }
 
-            // Initialize support functions (after routine registration, after r_task initialization)
+            // Initialize support functions (after routine registration, after
+            // r_task initialization). Intentionally panic if module loading
+            // fails. Modules are critical for ark to function.
             match modules::initialize() {
                 Err(err) => {
-                    log::error!("Can't load R modules: {err:?}");
+                    panic!("Failed to load R modules: {err:?}");
                 },
                 Ok(namespace) => {
                     console.positron_ns = Some(namespace);


crates/ark/src/modules.rs
```diff
@@ -268,13 +268,22 @@ mod debug {
...
+            })
+            .collect();
+        entries.sort_by_key(|entry| entry.path());
 
         for entry in entries {
-            match entry {
-                Ok(entry) => import_file(&entry.path(), src, env)?,
-                Err(err) => log::error!("Can't load modules from file: {err:?}"),
-            };
+            import_file(&entry.path(), src, env)?;
```
DavisVaughan, line 284: We've got a mix of `?` and log + continue style behavior here

The modules feel very important to me, I wonder if `log::error!("Can't read directory entry: {err:?}")` should really be turned into a `?` that propagates upwards rather than logging + continue?

lionel-: hmm I think I agree with your sentiment that we're currently too lenient in our init routine.

I've implemented your suggestion and turned the upstream logging into a loud panic. Note that this only concerns the debug path for local module files. We're still lenient in release builds, we should probably fail early there too. WDYT?

Also note that the hot-reloading path still only logs, otherwise we'd get panic for things like syntax errors.

crates/ark/src/modules.rs
```diff
@@ -125,12 +125,14 @@ pub fn initialize() -> anyhow::Result<RObject> {
                 &root.join("positron"),
                 RModuleSource::Positron,
                 namespace.sexp,
-            )?;
+            )
+            .unwrap();
```
DavisVaughan: I would be in favor of a more explicit `panic!()` with a comment to show that _we really mean this one_ and this isn't just an accidental `unwrap()` that we need to remove

crates/ark/src/modules.rs
DavisVaughan: > We're still lenient in release builds, we should probably fail early there too.

I do agree that even in release builds we should probably fail loudly as long as its fairly informative logging.

Modules just feel so critical to ark working, that a loud panic feels right?

lionel-: We now panic consistently if modules fail to load or initialise.

PR #1024, opened 2026-01-30
Rename `interface.rs` to `console.rs` and `RMain` to `Console`

lionel-:
The core of Ark has historically been implemented in `interface.rs` and consolidated in a singleton struct `RMain` in https://github.com/posit-dev/ark/pull/52

Really this struct is mostly about the ReadConsole and WriteConsole handlers, and we've been thinking about renaming `interface.rs` to `console.rs`, and `RMain` to `Console` for a while. Some files already use the `console_` prefix, e.g. `console_debug.rs` and `console_annotate.rs`. This PR gets us through the other side of that ongoing change.

### DIFF
--- a/crates/ark/src/lsp/backend.rs
+++ b/crates/ark/src/lsp/backend.rs
@@ -530,7 +530,7 @@ pub fn start_lsp(
             r_task({
                 let events_tx = events_tx.clone();
                 move || {
-                    Console::with_mut(|con| con.set_lsp_channel(events_tx));
+                    Console::get_mut().set_lsp_channel(events_tx);
                 }
             });
 
@@ -577,7 +577,7 @@ pub fn start_lsp(
         // from `Console`, at least until someone starts the LSP up again.
         r_task({
             move || {
-                Console::with_mut(|con| con.remove_lsp_channel());
+                Console::get_mut().remove_lsp_channel();
             }
         });
     })
--- a/crates/ark/src/browser.rs
+++ b/crates/ark/src/browser.rs
@@ -26,17 +26,15 @@ pub unsafe extern "C-unwind" fn ps_browse_url(url: SEXP) -> anyhow::Result<SEXP>
 }
 
 fn is_help_url(url: &str) -> bool {
-    Console::with(|con| con.is_help_url(url))
+    Console::get().is_help_url(url)
 }
 
 fn handle_help_url(url: String) -> anyhow::Result<()> {
-    Console::with(|con| {
-        let event = HelpEvent::ShowHelpUrl(ShowHelpUrlParams {
-            url,
-            kind: ShowHelpUrlKind::HelpProxy,
-        });
-        con.send_help_event(event)
-    })
+    let event = HelpEvent::ShowHelpUrl(ShowHelpUrlParams {
+        url,
+        kind: ShowHelpUrlKind::HelpProxy,
+    });
+    Console::get().send_help_event(event)
 }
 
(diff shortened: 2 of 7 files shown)


crates/ark/src/connections/r_connection.rs
```diff
@@ -361,11 +361,11 @@ pub unsafe extern "C-unwind" fn ps_connection_updated(id: SEXP) -> Result<SEXP,
 
 #[harp::register]
 pub unsafe extern "C-unwind" fn ps_connection_focus(id: SEXP) -> Result<SEXP, anyhow::Error> {
-    if !RMain::is_initialized() {
+    if !Console::is_initialized() {
         return Ok(R_NilValue);
     }
 
-    let main = RMain::get();
+    let main = Console::get();
```
DavisVaughan: You're going to have a lot of these, which are going to be quite confusing to read 😢 

If you can get these to update to `let console =` then I'd be in favor of that

crates/ark/src/lsp/completions/sources/composite/search_path.rs
```diff
@@ -51,9 +51,9 @@ fn completions_from_search_path(
 
     unsafe {
         // Iterate through environments starting from the current frame environment.
-        #[cfg(not(test))] // Unit tests do not have an `RMain`
-        // Mem-Safety: Object protected by `RMain` for the duration of the `r_task()`
-        let mut env = crate::interface::RMain::get().read_console_frame().sexp;
+        #[cfg(not(test))] // Unit tests do not have an `Console`
```
DavisVaughan, line 54:
```suggestion
        #[cfg(not(test))] // Unit tests do not have a `Console`
```

crates/ark/src/lsp/backend.rs
```diff
@@ -522,16 +522,16 @@ pub fn start_lsp(
...
-            // `RMain` should be initialized by now, since the caller of this
+            // `Console` should be initialized by now, since the caller of this
             // function waits to receive the init notification sent on
             // `kernel_init_rx`. Even if it isn't, this should be okay because
             // `r_task()` defensively blocks until its sender is initialized.
             r_task({
                 let events_tx = events_tx.clone();
                 move || {
-                    RMain::with_mut(|main| main.set_lsp_channel(events_tx));
+                    Console::with_mut(|con| con.set_lsp_channel(events_tx));
```
DavisVaughan: I greatly prefer `console` over `con`. I immediately think "connection" when I see `con`

crates/ark/src/sys/unix/console.rs
```diff
@@ -1,13 +1,158 @@
...
+use crate::console::r_show_message;
+use crate::console::r_suicide;
+use crate::console::r_write_console;
+use crate::console::Console;
+use crate::signals::initialize_signal_handlers;
+
+// For shutdown signal in integration tests
+pub static CLEANUP_SIGNAL: (Mutex<bool>, Condvar) = (Mutex::new(false), Condvar::new());
+
+pub fn setup_r(args: &Vec<String>) {
```
DavisVaughan, line 46: I'm going to trust that nothing changed in here beyond moving it from `interface.rs` to `console.rs` 😄

crates/ark/src/lsp/backend.rs
lionel-: I see where you're coming from. That said the con abbreviation for console is a very common one: https://github.com/search?q=%22con+%3D+console%22&type=code

Another abbrev I've been using for connection is "conn".

For readability I think it's important to have an abbreviation for small scopes especially tiny ones like this lambda.

PR #1028, opened 2026-02-03
Return an `LspResult` from LSP Request handlers

DavisVaughan:
Update: I did not actually use a custom jsonrpc error in #1040, instead we modeled a "rejection" as part of the `Response` type. However, I still feel like it is useful for us to be able to do this, so I think we should consider this still.

---

I'm starting to look at https://github.com/posit-dev/positron/issues/8350, and I believe that `statement_range()` is going to need to be able to return a structured `jsonrpc::Error` with a well defined `code` and `message` that we can pick up on on the frontend.

In particular, I'm likely going to have it return `RequestFailed`
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#responseMessage

```ts
/**
	 * A request failed but it was syntactically correct, e.g the
	 * method name was known and the parameters were valid. The error
	 * message should contain human readable information about why
	 * the request failed.
	 *
	 * @since 3.17.0
	 */
	export const RequestFailed: [integer](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#integer) = -32803;
```

Currently all of our LSP Request handlers return an `anyhow::Result`, and at _the very last second_ we convert any `anyhow::Error`s into a `jsonrpc::Error` with an arbitrary `-1` code. This is no longer going to be good enough for cases where we need to return structured `jsonrpc::Error`s with a specific `code`.

Luckily I think we can get the best of both worlds, here's what I've come up with for the return type of all Request handlers:

```rs
pub(crate) type LspResult<T> = std::result::Result<T, LspError>;

#[derive(Debug)]
pub(crate) enum LspError {
    JsonRpc(jsonrpc::Error),
    Anyhow(anyhow::Error),
}
```

This comes with nice `From` methods

```rs
// For the ability to `?` a `jsonrpc::Error` into an `LspError`
impl From<jsonrpc::Error> for LspError {
    fn from(error: jsonrpc::Error) -> Self {
        Self::JsonRpc(error)
    }
}

// For the ability to `?` an `anyhow::Error` into an `LspError`
impl From<anyhow::Error> for LspError {
    fn from(error: anyhow::Error) -> Self {
        Self::Anyhow(error)
    }
}
```

which means there is very little code to change. We can still use `?` on any helper method that returns an `anyhow::Result`, and the `anyhow::Error` will get wrapped into `LspError::Anyhow` automatically.

I haven't actually done anything in `statement_range()` yet, but I'm planning on returning something like this using this new `LspResult` feature:

```rs
// Unfortunately `tower_lsp::jsonrpc::ErrorCode::RequestFailed` isn't provided yet.
const REQUEST_FAILED: i64 = -32803;
return LspResult::Err(LspError::JsonRpc(jsonrpc::Error {
    code: jsonrpc::ErrorCode::ServerError(REQUEST_FAILED),
    message: Cow::Owned(String::from("Parse failure")),
    data: None,
}));
```

### DIFF
--- a/crates/ark/src/lsp/backend.rs
+++ b/crates/ark/src/lsp/backend.rs
@@ -75,7 +75,9 @@ macro_rules! cast_response {
         match $target {
             RequestResponse::Result(Ok($pat(resp))) => Ok(resp),
             RequestResponse::Result(Ok(_)) => {
-                panic!("Unexpected variant while casting to {}", stringify!($pat))
+                let message = format!("Unexpected variant while casting to {}", stringify!($pat));
+                log::error!("{message}");
+                Err(new_jsonrpc_error(message))
             },
             RequestResponse::Result(Err(err)) => match err {
                 LspError::JsonRpc(err) => Err(err),


crates/ark/src/lsp/backend.rs
```diff
@@ -65,16 +65,22 @@ use crate::r_task;
 // would flood the LSP logs with irrelevant backtraces.
 pub(crate) enum RequestResponse {
     Disabled,
-    Crashed(anyhow::Result<LspResponse>),
-    Result(anyhow::Result<LspResponse>),
+    Crashed(anyhow::Error),
```
DavisVaughan, line 68: I recognized along the way that `Crashed` is only ever intended to contain an `anyhow::Error`, i.e. it will never hold the `Ok()` variant of an `anyhow::Result<>`, so we can simplify this a little.

crates/ark/src/lsp/backend.rs
```diff
@@ -65,16 +65,22 @@ use crate::r_task;
...
         match $target {
             RequestResponse::Result(Ok($pat(resp))) => Ok(resp),
-            RequestResponse::Result(Err(err)) => Err(new_jsonrpc_error(format!("{err:?}"))),
+            RequestResponse::Result(Ok(_)) => {
+                panic!("Unexpected variant while casting to {}", stringify!($pat))
+            },
+            RequestResponse::Result(Err(err)) => match err {
+                LspError::JsonRpc(err) => Err(err),
+                LspError::Anyhow(err) => Err(new_jsonrpc_error(format!("{err:?}"))),
+            },
```
DavisVaughan, line 85: This is most of the magic. We can now build a context aware `jsonrpc::Error` at the failure site and pipe it all the way through to be sent to the frontend. I think this will be useful for more than just `statement_range()` going forward.

crates/ark/src/lsp/main_loop.rs
```diff
@@ -478,22 +480,19 @@ fn respond<T>(
...
+        RequestResponse::Result(Err(ref error)) => {
+            Err(anyhow!("Error while handling request:\n{error:?}"))
+        },
+        RequestResponse::Crashed(ref error) => {
+            Err(anyhow!("Crashed while handling request:\n{error:?}"))
+        },
+        RequestResponse::Disabled => Err(anyhow!("Received impossible `Disabled` response state")),
     };
 
     // Ignore errors from a closed channel. This indicates the request has
```
DavisVaughan, line 498: A little bit of fiddling to get this right, but I think it reads a little clearer in the end

crates/ark/src/lsp/backend.rs
```diff
@@ -65,16 +65,22 @@ use crate::r_task;
...
 }
 
 // Based on https://stackoverflow.com/a/69324393/1725177
 macro_rules! cast_response {
     ($self:expr, $target:expr, $pat:path) => {{
         match $target {
             RequestResponse::Result(Ok($pat(resp))) => Ok(resp),
-            RequestResponse::Result(Err(err)) => Err(new_jsonrpc_error(format!("{err:?}"))),
+            RequestResponse::Result(Ok(_)) => {
+                panic!("Unexpected variant while casting to {}", stringify!($pat))
```
lionel-: hmm I'd still log and return an error here. Maybe use `debug_panic` instead?

lionel- (approved):
Seems like a nice improvement!

PR #1030, opened 2026-02-04
Refine statement range parse error behavior

DavisVaughan:
Part of https://github.com/posit-dev/positron/issues/8350
Part of https://github.com/posit-dev/positron/issues/5023

Part 1 of the statement range refinement plan

Only refuse to provide a statement range after detecting the first top-level expression that contains a parse error somewhere within it. We assert that anything before this has parsed correctly, so if the user's cursor is up there, we should be able to compute a statement range (we are mostly confident in this assertion, I cannot think of any examples where it would not hold).

---

This matches Mine's expectations over in https://github.com/posit-dev/positron/issues/8350 of

> Regardless of broken/buggy code _further down in the script_, a "good" pipeline with line breaks should be sent to the Console entirely.

### DIFF
--- a/crates/ark/src/lsp/statement_range.rs
+++ b/crates/ark/src/lsp/statement_range.rs
@@ -374,14 +374,6 @@ fn expand_range_across_semicolons(mut node: Node) -> tree_sitter::Range {
 }
 
 fn find_statement_range(root: &Node, row: usize) -> Option<tree_sitter::Range> {
-    // Refuse to provide a statement range in the face of parse errors, we are
-    // unlikely to be able to provide anything useful, and are more likely to provide
-    // something confusing. Instead, return `None` so that the frontend sends code to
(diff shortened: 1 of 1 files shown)


crates/ark/src/lsp/statement_range.rs
```diff
@@ -374,21 +374,27 @@ fn expand_range_across_semicolons(mut node: Node) -> tree_sitter::Range {
...
 
     let children = root.children(&mut cursor);
 
     let mut node = None;
 
     for child in children {
+        // Refuse to provide a statement range after detecting a top-level `child` with a
+        // parse error. We expect that everything before this top-level `child` has parsed
+        // correctly, so if the cursor was above this first parse error, we can execute
+        // that code fearlessly. After the first parse error, all bets are off. Even if
```
jennybc, line 387: Fearlessly!!!

fearless

PR #1032, opened 2026-02-09
Merge CommManager into Shell and IOPub

lionel-:
Branched from #1027 (DAP integration tests)
Progress towards #1033 (Make DAP protocol tests deterministic)
Progress towards #1026 

This PR merges the comm manager into:

- The IOPub thread, for outgoing `comm_msg` emission
- The Shell thread, for incoming `comm_msg` dispatching and outgoing `comm_open`/`comm_close` emission

The main goal is for outgoing `comm_msg` and other IOPub messages to go through the same crossbeam channel. Previously, these took separate paths to the frontend:

- IOPub messages: sender > `iopub_tx` > IOPub thread > ZMQ socket
- Comm messages: sender > `outgoing_tx` on CommSocket > CommManager thread > `iopub_tx` > IOPub thread > ZMQ socket

The comm manager intermediary caused non-deterministic message ordering between comm and IOPub messages. When R code on the same thread emits a stream output followed by a comm message, or a comm message followed by an Idle boundary, the frontend could receive them in either order. After this PR, all comm outgoing messages go directly through `iopub_tx`, the same channel used by all other IOPub messages. Ordering is now deterministic: messages sent sequentially from the same thread arrive at the frontend in that order.

## Motivation

While working on #1027 I observed non-deterministic ordering between `busy`/`idle` status messages (sent via IOPub) and `stop_debug`/`start_debug` comm messages (sent via the DAP comm through CommManager). The frontend handled this fine since these messages are consumed by different components, but it caused significant problems writing robust and deterministic protocol tests: tests could not make sequential assertions about message ordering without resorting to fuzzy accumulator-based matching that was fragile and hid bugs.

More generally, any code that depends on the relative ordering of comm messages and other IOPub messages has a potential race. With the unified path, ordering matches programmer expectations by construction, eliminating this class of issues.

The change also reduces overall complexity. The CommManager was a dedicated thread that existed solely to bridge comm channels to IOPub, using `crossbeam::Select` over a dynamic set of receivers and maintaining a lookup table for RPC headers. Removing it eliminates a thread and ~260 lines. The responsibilities it held are absorbed by Shell and IOPub, where they naturally belong: Shell already handles `comm_open`, `comm_msg`, and `comm_close` from the frontend, so it's the right place to manage the set of open comms. IOPub already serializes all messages to the frontend, so it's the right place to handle comm outgoing messages.

The direct payoff is the follow-up PR (#1033), which leverages the ordering guarantee to replace fuzzy message matching with strict sequential assertions.

#### Delete CommManager

The `CommManager` thread is removed entirely (~260 lines). It previously:
- Held the list of open comms
- Used `crossbeam::Select` over a dynamic set of comm `outgoing_rx` channels
- Maintained a `pending_rpcs` HashMap to match RPC replies to their original Jupyter headers
- Relayed comm messages to IOPub

All of these responsibilities are absorbed by Shell and IOPub.

#### `CommOutgoingTx` wrapper

A new `CommOutgoingTx` type wraps `Sender<IOPubMessage>` with a `comm_id`. When a comm handler calls `outgoing_tx.send(msg)`, it sends `IOPubMessage::CommOutgoing(comm_id, msg)` through the IOPub channel. This is the key mechanism allowing message determinism.

The `outgoing_rx` side of `CommSocket` is removed. Comm handlers no longer have a dedicated receive channel for outgoing messages since there's no CommManager polling them.

#### `CommMsg::Rpc` carries the parent header

Previously, when the frontend sent an RPC to a comm, the Shell thread would store the Jupyter header in CommManager's `pending_rpcs` map. When the comm replied, CommManager would look up the header to create a properly parented IOPub message.

Now `CommMsg::Rpc` is a named struct that carries the `parent_header` directly:

```rust
CommMsg::Rpc {
    id: String,
    parent_header: JupyterHeader,
    data: Value,
}
```

The header travels with the message through the comm handler and back, so a lookup table is no longer needed.

#### Shell manages open comms

Shell now holds `open_comms: Vec<CommSocket>` and handles comm lifecycle directly:

- Frontend-initiated comms (`comm_open`, `comm_msg`, `comm_close` on the Shell socket) are handled in-place
- Backend-initiated comms (from `comm_manager_rx`) arrive when R code opens a comm (data explorer, variables pane, connections, etc.)

The challenge is that Shell is already in a `zmq_poll()` loop waiting on its ZMQ socket, but backend comm events arrive on a crossbeam channel. You can't mix ZMQ `poll()` and crossbeam `Select` in a single wait. To solve this problem, I generalised the solution implemented for StdIn in https://github.com/posit-dev/ark/pull/58

The bridge mechanism was cleaned up and generalised to allow management of several sockets/entities. As part of this work, the "notifier" thread is now called the "channel bridge" thread, and the "forwarding" thread is called the "socket bridge" thread.

#### IOPub processes `CommOutgoing`

A new `IOPubMessage::CommOutgoing(String, CommMsg)` variant is handled in the IOPub thread. It flushes any buffered stream output before forwarding the comm message, preserving the ordering from the sender's perspective. This unconditional flush on every comm message is the key to the ordering guarantee. For the expected message rates it has negligible overhead (stream flushing is a no-op when the buffer is empty).

## What's mechanical

About 70% of the diff is mechanical plumbing:
- Threading `iopub_tx` through `CommSocket::new()` and every `start()` function that creates comms (data explorer, variables, connections, reticulate, plots, help)
- Adapting tests to receive comm responses from an IOPub channel instead of `outgoing_rx`
- Updating `CommMsg::Rpc(id, data)` to `CommMsg::Rpc { id, parent_header, data }` at all call sites

The core logic changes are in:
- `crates/amalthea/src/kernel.rs` (channel bridge thread, Forwarder)
- `crates/amalthea/src/socket/shell.rs` (comm management, RefCell removal)
- `crates/amalthea/src/socket/iopub.rs` (CommOutgoing handling)
- `crates/amalthea/src/socket/comm.rs` (CommOutgoingTx)

### DIFF
--- a/crates/amalthea/src/kernel.rs
+++ b/crates/amalthea/src/kernel.rs
@@ -487,7 +487,7 @@ fn socket_bridge_thread(
     };
 
     // Drain all pending outbound messages and forward to ZMQ.
-    let drain_outbound = || {
+    let forward_outbound = || {
         while let Ok(outbound_msg) = zmq_outbound_rx.try_recv() {
             let result = match outbound_msg {
                 OutboundMessage::StdIn(msg) => msg.send(&stdin_socket),
@@ -536,7 +536,7 @@ fn socket_bridge_thread(
 
         for _ in 0..n {
             if consume_outbound_notification() {
-                drain_outbound();
+                forward_outbound();
                 continue;
             }
 


crates/ark/src/variables/r_variables.rs
```diff
@@ -501,19 +512,12 @@ impl RVariables {
...
             Ok(data) => {
-                // If we were given a request ID, send the response as an RPC;
-                // otherwise, send it as an event
-                let comm_msg = match request_id {
-                    Some(id) => CommMsg::Rpc(id, data),
-                    None => CommMsg::Data(data),
-                };
-
-                self.comm.outgoing_tx.send(comm_msg).unwrap()
+                self.comm.outgoing_tx.send(CommMsg::Data(data)).log_err();
```
DavisVaughan, line 520: What's happened with `request_id` here?

crates/ark/src/sys/unix/console.rs
```diff
@@ -136,7 +136,7 @@ pub fn run_activity_handlers() {
     }
 }
 
-#[no_mangle]
+#[cfg_attr(not(test), no_mangle)]
```
DavisVaughan, line 139: What's going on here?

When...not running tests we don't mangle? So in `release` we don't mangle? But in tests we mangle?

What am I missing?

crates/ark/src/connections/r_connection.rs
```diff
@@ -59,12 +60,14 @@ impl RConnection {
     pub fn start(
         metadata: Metadata,
         comm_manager_tx: Sender<CommManagerEvent>,
+        iopub_tx: Sender<IOPubMessage>,
```
DavisVaughan: I'm seeing both `comm_manager_tx: Sender<CommManagerEvent>` and `iopub_tx: Sender<IOPubMessage>` side by side in a lot of the function signatures now.

That feels very good!

Something about that seems a lot more correct, but it's hard to describe exactly what I mean.

crates/ark/src/connections/r_connection.rs
```diff
@@ -325,7 +328,12 @@ pub unsafe extern "C-unwind" fn ps_connection_opened(
         code: RObject::view(code).to::<Option<String>>().unwrap_or(None),
     };
 
-    if let Err(err) = RConnection::start(metadata, console.get_comm_manager_tx().clone(), id) {
+    if let Err(err) = RConnection::start(
+        metadata,
+        console.get_comm_manager_tx().clone(),
+        console.get_iopub_tx().clone(),
```
DavisVaughan: Like this feels more correct to me now, but it's hard to explain why!

crates/ark/src/dap/dap_server.rs
```diff
@@ -760,7 +761,7 @@ impl<R: Read, W: Write> DapServer<R, W> {
             // were sent by the user. This ensures prompts are updated.
             let msg = amalthea::comm_rpc_message!("execute", command = debug_request_command(cmd));
 
-            tx.send(msg).unwrap();
+            tx.send(msg).log_err();
```
DavisVaughan, line 764: 2 places in this file it switched away from `unwrap` during a `send()`

but we still have things like `self.r_request_tx.send(RRequest::DebugCommand(cmd)).unwrap()`

I thought we had said that for a `send()`, we typically just assume that is going to work, and thats one of the few things we unconditionally `unwrap`?

crates/amalthea/src/kernel.rs
```diff
@@ -460,21 +486,17 @@ fn zmq_forwarding_thread(
...
-            OutboundMessage::StdIn(msg) => msg.send(&stdin_socket)?,
-            OutboundMessage::IOPub(msg) => msg.send(&iopub_socket)?,
-        };
-
-        // Notify back
-        outbound_notif_socket.send(zmq::Message::new())?;
-
-        Ok(())
+    // Drain all pending outbound messages and forward to ZMQ.
+    let drain_outbound = || {
```
DavisVaughan: Could this still keep "forward" in the name somewhere? It does more than just drain. I actually like the old name more

crates/amalthea/src/socket/iopub.rs
```diff
@@ -88,12 +90,12 @@ pub enum IOPubMessage {
     ExecuteInput(ExecuteInput),
     Stream(StreamOutput),
     CommOpen(CommOpen),
-    CommMsgReply(JupyterHeader, CommWireMsg),
-    CommMsgEvent(CommWireMsg),
     CommClose(CommClose),
```
DavisVaughan: What is the difference between `IOPubMessage::CommClose` and `IOPubMessage::CommOutgoing` with `CommMsg::Close`?

crates/amalthea/src/socket/iopub.rs
```diff
@@ -88,12 +90,12 @@ pub enum IOPubMessage {
...
     DisplayData(DisplayData),
     UpdateDisplayData(UpdateDisplayData),
     Wait(Wait),
+    /// Outgoing comm message from a backend. The String is the comm_id.
+    CommOutgoing(String, CommMsg),
```
DavisVaughan, line 96: Does it make sense to put this up with `CommOpen` and `CommClose`?

crates/amalthea/src/socket/iopub.rs
```diff
@@ -132,50 +141,69 @@ impl IOPub {
...
+    /// - Subscription messages from the ZMQ socket
+    /// - Periodic stream buffer flushing
     pub fn listen(&mut self) {
-        // Flush the active stream (either stdout or stderr) at regular
-        // intervals
-        let flush_interval = *StreamBuffer::interval();
-        let flush_interval = tick(flush_interval);
+        let flush_tick = tick(*StreamBuffer::interval());
 
         loop {
```
DavisVaughan, line 143: I think I actually found the `select!` more intuitive here. It feels like it flowed better. Can we not use that now?

crates/amalthea/src/socket/iopub.rs
```diff
@@ -132,50 +141,69 @@ impl IOPub {
...
         }
     }
 
-    /// Process an IOPub message from another thread.
-    fn process_outbound_message(&mut self, message: IOPubMessage) -> crate::Result<()> {
+    /// Process an outbound IOPub message received from another thread.
+    ///
+    /// These messages originate from Shell, Control, or comm handlers and are
+    /// forwarded to the frontend via the IOPub ZMQ socket.
+    fn process_iopub_message(&mut self, message: IOPubMessage) -> crate::Result<()> {
```
DavisVaughan, line 182: I like the directional `outbound` name that was here, and the `inbound` one for the subscription message

Would you be ok with a very long name of `process_outbound_iopub_message()` and `process_inbound_subscription_message()`? That helps me understand it since iopub does both directions here

crates/amalthea/src/comm/event.rs
```diff
@@ -23,29 +21,6 @@ pub enum CommManagerEvent {
     /// second value is the message.
```
DavisVaughan, line 19: I would be fine with `CommManagerEvent` -> `CommEvent` if you think that makes sense at this point

crates/amalthea/src/socket/comm.rs
```diff
@@ -159,9 +197,13 @@ impl CommSocket {
...
 
-        let response = CommMsg::Rpc(id, json);
+        let response = CommMsg::Rpc {
+            id,
+            parent_header,
+            data: json,
+        };
 
-        self.outgoing_tx.send(response).unwrap();
+        self.outgoing_tx.send(response).log_err();
```
DavisVaughan, line 206: I'm noticing more of the `unwrap` -> `log_err` transition

I feel like these are probably pretty critical messages?

DavisVaughan: I notice you put something in AGENTS about this, and that is probably contributing to this. Maybe send() should be an exception that unwraps() still.

crates/amalthea/src/socket/shell.rs
```diff
@@ -232,39 +375,40 @@ impl Shell {
...
-            .unwrap();
-
-        // Wait on the reply
-        let CommManagerInfoReply { comms } = rx.recv().unwrap();
-
         // Convert to a JSON object
         let mut info = serde_json::Map::new();
 
-        for comm in comms.into_iter() {
+        for comm in open_comms.iter() {
```
DavisVaughan, line 397: Nice simplication to not have to ask for these from the comm manager

crates/amalthea/src/socket/shell.rs
```diff
@@ -276,22 +420,24 @@ impl Shell {
 
     /// Handle a request to open a comm
     fn handle_comm_open(
-        &self,
+        iopub_tx: &Sender<IOPubMessage>,
         shell_handler: &mut Box<dyn ShellHandler>,
+        server_handlers: &HashMap<String, Arc<Mutex<dyn ServerHandler>>>,
+        open_comms: &mut Vec<CommSocket>,
         msg: &CommOpen,
```
DavisVaughan, line 429: It feels a bit weird to me that this isn't `&mut self`

I see you supply `open_comms`, which is from `self`, and you modify that directly.

Is this just a weird case where you'd "over borrow" too much if you did `&mut self` and Rust wouldn't be happy about that?

DavisVaughan: Like right now it makes the method look "pure" but it's really not, as it modifies `self` state. And that's somewhat confusing IMO

crates/amalthea/src/socket/shell.rs
```diff
@@ -499,12 +663,24 @@ impl Shell {
     }
 
     /// Handle a request to close a comm
-    fn handle_comm_close(&self, msg: &CommClose) -> crate::Result<()> {
-        // Send a notification to the comm message listener thread notifying it that
-        // the comm has been closed
-        self.comm_manager_tx
-            .send(CommManagerEvent::Closed(msg.comm_id.clone()))
-            .unwrap();
+    fn handle_comm_close(open_comms: &mut Vec<CommSocket>, msg: &CommClose) -> crate::Result<()> {
```
DavisVaughan, line 670: Same here of course about feeling like this should be `&self mut`

crates/ark/src/sys/unix/console.rs
lionel-: That is fixing a gnarly linker error in tests. I didn't do a deep dive so can't tell more.

crates/ark/src/dap/dap_server.rs
lionel-: Now I typically log_err since it's so easy. Especially with these non-core components, no need to bring down entire threads (and potentially the main thread along with it) if the R session might still work.

I'd unwrap/panic if IOPub is down for instance, but not the DAP or one of the UI comms.

crates/amalthea/src/socket/comm.rs
lionel-: See comment above

crates/ark/src/sys/unix/console.rs
DavisVaughan: It's pretty strange to me that we'd need this because I thought that we'd want `no_mangle` no matter what

crates/amalthea/src/comm/event.rs
lionel-: Done!

crates/amalthea/src/socket/iopub.rs
lionel-: None!

I've now unified these types and integrated `CommOpen` into `CommMsg`.

The enum is used both for incoming and outgoing messages, and a CommOpen is never expected as incoming message on a comm, but that's a small inconsistency.

crates/amalthea/src/socket/shell.rs
lionel-: The change to a free method was motivated by the borrow checker. Ideally it'd have remained a `self` method.

crates/ark/src/sys/unix/console.rs
lionel-: Explanation from Claude:

When running tests, `ark` gets compiled twice:
1. As a library dependency (via `ark_test`)
2. As the test binary itself

Functions marked `#[no_mangle]` export with a fixed C symbol name like `r_read_console`. Both compilations produced this same symbol, so the linker saw two `r_read_console` symbols and failed with a duplicate symbol error.

crates/ark/src/variables/r_variables.rs
lionel-: It was dead code, requests are handled by `handle_request()`

PR #1033, opened 2026-02-09
Make DAP protocol tests deterministic

### LINKED ISSUE #1026: DAP protocol tests

lionel-:
Branched from https://github.com/posit-dev/ark/pull/1032 (Merge CommManager into Shell and IOPub)
Closes #1026

This PR is the payoff for the IOPub/Comm unification of #1032. It leverages deterministic message ordering to replace fuzzy message matching with strict sequential assertions in the test infrastructure. It also fixes several bugs made obvious by the new assertions, and removes emission of debug events during auto-stepping so that code injected for breakpoints is invisible to the frontend.

- Replaces fuzzy matching with strict sequential assertions, making tests both more readable and more accurate. Before:

  ```rust
  frontend.recv_iopub_until(|acc| {
      acc.has_comm_method_count("start_debug", 2) &&
          acc.has_comm_method_count("stop_debug", 2) &&
          acc.saw_idle()
  });
  ```

  After (the different counts are due to other fixes, see below):

  ```rust
  frontend.recv_iopub_start_debug();
  frontend.assert_stream_stdout_contains("Called from:");
  frontend.assert_stream_debug_at(&file);
  frontend.recv_iopub_idle();
  ```

- Eliminates workarounds like `recv_iopub_busy_skip_streams`, `recv_iopub_execute_input_skip_streams`, `recv_iopub_async`, and the entire `MessageAccumulator` / `recv_iopub_until` pattern

As part of a complete test review I also fixed:

- Off-by-one line number issue in sourced code (missing initial `#line` directive)
- Missing source file assertions on stack frames
- Superfluous DAP events from auto-stepping. Code injected for breakpoints no longer emits start_debug/stop_debug cycles

## Stream buffering

The one remaining source of non-determinism is R's stream output. R may batch or split `stdout`/`stderr` messages arbitrarily, independently of the ordering of non-stream messages. The infrastructure around stream assertions was reworked to robustly prevent test flakiness caused by stream buffering.

All `recv_iopub_*` methods now automatically buffer Stream messages and _ignore them_. The accumulated streams are now tested/asserted separately via `assert_stream_stdout_contains()` / `assert_stream_stderr_contains()`. Stream assertions are still order-sensitive: each match drains the buffer up to that point. But a stray stream can no longer cause other assertions to fail.

`recv_iopub_idle()` acts as a synchronization point: at that point, we know for sure R has finished streaming all output emitted during the request handling. The method panics if streams were received but not asserted, then clears the buffers.

There must be at least one stream assertion per request handling if output is emitted. Unfortunately this setup still allows unexpected stream output to go unnoticed as long as there is one stream assertion in the test. Due to arbitrary splitting though, I don't think we can do better. Unless we go for full snapshots, but this approach has issues of its own (scrubbing of stateful output, and stability across platforms and versions of R).

## Auto-stepping before `debug_start()`

Previously, auto-stepping through injected breakpoint wrappers happened _after_ `debug_start()` had already been sent to the frontend. Every breakpoint hit produced visible intermediate events (start_debug, Stopped, Continued, Continued, start_debug, Stopped). Tests had to account for all of these with `recv_auto_step_through()` and multiple start/stop cycles.

Now `maybe_auto_step()` is called _before_ `debug_start()`. If we detect we're at injected code, we immediately return `n` to R without emitting any debug notifications. The frontend only sees a single start_debug + Stopped at the user expression.

Similarly, when `pending_inputs` is non-empty (multiple expressions queued), we skip `debug_start()` for intermediate browser prompts.

The main insight is that we never start a new debug session unless we're ready to close the active execution request, and we never close the active execution request until R has stopped at the final user-visible location.

### DIFF
--- a/crates/ark/tests/dap.rs
+++ b/crates/ark/tests/dap.rs
@@ -143,13 +143,13 @@ fn test_dap_error_during_debug() {
     let stack = dap.stack_trace();
     assert!(stack.len() >= 1, "Should have at least 1 frame");
 
-    // Step to the error line - this moves us TO stop() but doesn't execute it yet
+    // Step to execute the error
     frontend.debug_send_step_command("n", &file);
     dap.recv_continued();
+    dap.recv_stopped();
 
-    // After error in sourced code, R exits the debug session.
-    // We received Continued but no subsequent Stopped event - the debug session ended.
-    // The DapClient::drop() will verify no unexpected messages remain.
+    frontend.debug_send_quit();
+    dap.recv_continued();
 }
 
 #[test]
--- a/crates/ark_test/src/dap_client.rs
+++ b/crates/ark_test/src/dap_client.rs
@@ -674,8 +674,29 @@ impl Drop for DapClient {
     fn drop(&mut self) {
         // Don't try to disconnect if we're already panicking, as this could
         // obscure the original error
-        if !std::thread::panicking() {
-            self.disconnect();
+        if std::thread::panicking() {
+            return;
         }
+
+        // Check for unhandled messages using non-blocking mode.
+        // Must happen before disconnect() which drains events while waiting for response.
+        let _ = self.reader.get_ref().set_nonblocking(true);
+
+        let mut unexpected_messages: Vec<Sendable> = Vec::new();
+        while let Ok(msg) = self.recv() {
+            unexpected_messages.push(msg);
+        }
+
+        let _ = self.reader.get_ref().set_nonblocking(false);
+
+        if !unexpected_messages.is_empty() {
+            panic!(
+                "DAP socket has {} unexpected message(s) on exit:\n{:#?}",
+                unexpected_messages.len(),
+                unexpected_messages
+            );
+        }
+
+        self.disconnect();
     }
 }


crates/ark/src/snapshots/ark__console_annotate__tests__annotate_source_breakpoint_on_function_definition_line.snap
```diff
@@ -3,6 +3,7 @@ source: crates/ark/src/console_annotate.rs
 expression: result
 ---
 {
+#line 1 "file:///test.R"
 base::.ark_auto_step(base::.ark_breakpoint(browser(), "file:///test.R", "1"))
 #line 1 "file:///test.R"
```
DavisVaughan, line 8: It's supposed to be on line 6 and 8?

crates/ark_test/Cargo.toml
```diff
@@ -22,5 +22,6 @@ serde = { version = "1.0.183", features = ["derive"] }
 serde_json = "1.0.94"
 stdext = { path = "../stdext" }
 tempfile = "3"
+regex = "1"
```
DavisVaughan: alphabetical order?

crates/ark_test/src/dummy_frontend.rs
```diff
@@ -83,6 +114,467 @@ impl DummyArkFrontend {
...
+        } else {
+            None
+        }
+    }
+
+    /// Receive from IOPub with a timeout.
+    /// Returns `None` if the timeout expires before a message arrives.
+    ///
+    /// On Windows ARM, ZMQ poll with timeout blocks forever instead of
+    /// respecting the timeout. Use non-blocking poll with manual timing.
```
DavisVaughan, line 148: WAT

crates/ark_test/src/dummy_frontend.rs
```diff
@@ -43,9 +43,40 @@ use crate::MessageAccumulator;
...
+/// This handles the non-deterministic interleaving of streams (due to arbitrary
+/// batching and splitting of messages by R) with other messages.
 pub struct DummyArkFrontend {
     guard: MutexGuard<'static, DummyFrontend>,
+    /// Accumulated stdout stream content
+    stream_stdout: RefCell<String>,
+    /// Accumulated stderr stream content
+    stream_stderr: RefCell<String>,
+    /// Put-back queue for non-stream messages encountered during stream assertions
+    pending_messages: RefCell<VecDeque<Message>>,
```
DavisVaughan: Should probably say in the name that they are iopub messages

crates/ark_test/src/dummy_frontend.rs
```diff
@@ -83,6 +114,467 @@ impl DummyArkFrontend {
...
+
+            // Read more (with short timeout to allow checking deadline)
+            let remaining = deadline.saturating_duration_since(Instant::now());
+            let poll_timeout = remaining.min(Duration::from_millis(100));
+
+            if let Some(msg) = self.recv_iopub_with_timeout(poll_timeout) {
+                trace_iopub_msg(&msg);
+                match msg {
+                    Message::Stream(ref data) => self.buffer_stream(&data.content),
+                    other => self.pending_messages.borrow_mut().push_back(other),
```
DavisVaughan: I'm not super familiar with how this `trace_iopub_msg()` stuff works or why its useful, but isn't it weird that you can trace twice for some messages?

For a message you put in `pending_messages`, it is traced here and when you pop it off.

DavisVaughan: same in assert_stream_matches_re

crates/ark_test/src/dummy_frontend.rs
```diff
@@ -83,6 +114,467 @@ impl DummyArkFrontend {
...
+                        break;
+                    },
+                },
+                None => break,
+            }
+        }
+
+        DrainedStreams {
+            stdout: std::mem::take(&mut *self.stream_stdout.borrow_mut()),
+            stderr: std::mem::take(&mut *self.stream_stderr.borrow_mut()),
```
DavisVaughan:
```suggestion
            stdout: std::mem::take(&mut self.stream_stdout.borrow_mut()),
            stderr: std::mem::take(&mut self.stream_stderr.borrow_mut()),
```

i think

crates/ark_test/src/dummy_frontend.rs
```diff
@@ -83,6 +114,467 @@ impl DummyArkFrontend {
...
+        }
+    }
+
+    // Shadow DummyFrontend's stream methods to prevent bypassing the buffering layer.
+    // These methods read directly from the IOPub socket, which breaks the stream
+    // accumulation invariant. Use `assert_stream_*_contains()` instead.
+
+    #[deprecated = "Use assert_stream_stdout_contains() instead"]
+    #[allow(unused)]
+    pub fn recv_iopub_stream_stdout(&self, _expect: &str) {
```
DavisVaughan, line 395: Well, should we just remove them then?

crates/ark_test/src/dummy_frontend.rs
```diff
@@ -83,6 +114,467 @@ impl DummyArkFrontend {
...
+        panic!(
+            "Use assert_stream_stderr_contains() or assert_stream_stderr_matches() \
+             instead of recv_iopub_stream_stderr_with()"
+        );
+    }
+
+    // Overriding methods for base DummyFrontend's `recv_iopub_` methods. These
+    // use `recv_iopub_next()` to auto-skip streams. Question: Maybe this
+    // behaviour should live in the base dummy frontend? The stream issues are
+    // likely not R specific.
```
DavisVaughan, line 429: Yea having to override things when we also own the upstream DummyFrontend is feeling very complicated.

Thinking about touching any of this code a few months from now is already giving me anxiety haha

lionel-: yep the only thing that gave me pause is the possibility of an amalthea kernel with more predictable streams? But that seems unlikely.

crates/ark_test/src/dummy_frontend.rs
lionel-: Windows ARM is very worrying. Couldn't find any upstream issues though!

Here is another issue (this time flaky) I just observed:

https://github.com/posit-dev/ark/actions/runs/21896771039/job/63214590132?pr=1037#step:9:864

```
    failures:
        test_dap_evaluate_in_different_frames

    test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 7 filtered out; finished in 6.48s

  stderr ───

    thread 'test_dap_evaluate_in_different_frames' (3508) panicked at crates\ark\tests\dap_evaluate.rs:107:19:
    Failed to receive DAP message: A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond. (os error 10060)

        FAIL [   6.611s] (538/765) ark::dap_evaluate test_dap_evaluate_error
```

crates/ark/src/snapshots/ark__console_annotate__tests__annotate_source_breakpoint_on_function_definition_line.snap
lionel-: I'm not sure what you mean

crates/ark_test/src/dummy_frontend.rs
lionel-: Good catch!

crates/ark_test/src/dummy_frontend.rs
lionel-: As the comment indicates these are shadowing methods to prevent calling into the Amalthea methods. I've removed the deprecated tags to not be as misleading.

PR #1036, opened 2026-02-10
Add support for breaking on exceptions and pausing R at any time

lionel-:
Branched from [#1033](https://github.com/posit-dev/ark/pull/1043) (for show hidden frames option)
Addresses https://github.com/posit-dev/positron/issues/11797
Addresses https://github.com/posit-dev/positron/issues/11799
Addresses https://github.com/posit-dev/positron/issues/12085

Adds support for the DAP feature of exception breakpoints, and for pausing the debugger at any time. Both features rely on:

- Calling `browser()` from a global handler for the relevant condition (error, warning, or interrupt)

- Skipping the top-level frame which points to our global handlers.

  When these handlers trigger, the R stack includes the handler frame at the top. Since this frame is not useful to users, we track _why_ we entered the debugger via `DebugStoppedReason` (`Condition` for errors/warnings, `Pause` for interrupts) and discard the top frame in `start_debug()` for these cases.

  Note that the console will still evaluate in this frame until we fix https://github.com/posit-dev/positron/issues/9156#issuecomment-3816658521.

  Also note that for errors and warnings, we'd ideally automatically skip the part of the call stack involved in emitting/handling the condition (partly the base emitting context with `withRestart()` etc, and partly the package-level emitting context, e.g. based on rlang's `error_call = parent.frame()` pattern).

The condition breakpoints are surfaced like this in the UI:

Screenshot 2026-02-10 at 15 19 59

We could add messages as well but I thought that'd crowd the UI, and it doesn't feel as useful to easily break on those. Could be added as a user setting in the future.

Pausing can be done by calling the existing VS Code command "Pause":

Screenshot 2026-02-10 at 15 03 39

### QA Notes

**Exception breakpoints (break on error/warning):**

- In the Debug Console's breakpoint picker, enable "Uncaught Errors" or "Warnings".

- Run code that triggers an error (e.g. `stop("oops")`) or warning (e.g. `warning("hmm")`). The debugger should pause at the point of the condition.

- The call stack should show user code (for errors or warnings thrown from C) or `stop()`/`warning()`, not R's internal condition handling frames (`.handleSimpleError`, `doWithOneRestart`, etc.).

- The exception type and message should appear in the editor and the debug UI (e.g. "simpleError: oops").

- Continue (`c`) should resume execution and propagate the error/warning normally.

**Pause:**

- While R is executing a long-running operation (e.g. `Sys.sleep(100)`, an infinite loop, or I/O), run the `Debug: Pause` command from the command palette. R should interrupt and drop into the debugger at the current execution point.

- The call stack should reflect where R was interrupted.

- `Q` should exit the debugger. `c` / Continue should _resume_ running code. Note that with curl downloads or I/O there are known issues.

### DIFF
--- a/crates/ark/src/dap/dap.rs
+++ b/crates/ark/src/dap/dap.rs
@@ -237,8 +237,6 @@ impl Dap {
         self.is_debugging = true;
         self.fallback_sources.extend(fallback_sources);
 
-        remove_condition_handling_frames(&mut stack, &stopped_reason);
-
         self.load_variables_references(&mut stack);
         self.stack = Some(stack);
 
@@ -551,46 +549,6 @@ impl ServerHandler for Dap {
         });
 
         return Ok(());
-    }
-}
-
-/// Discard top frames that are part of the debug infrastructure rather than
-/// user code.
-fn remove_condition_handling_frames(
-    stack: &mut Vec<FrameInfo>,
-    stopped_reason: &DebugStoppedReason,
-) {
-    // Discard top frame when stopped due to exception breakpoint or pause,
-    // it points to our global handler that calls `browser()`
-    if matches!(
-        stopped_reason,
-        DebugStoppedReason::Condition { .. } | DebugStoppedReason::Pause
-    ) && !stack.is_empty()
-    {
-        stack.remove(0);
-    }
-
-    // Then discard base R's own condition handling/emitting frames, if any
-    remove_frame_prefix(stack, &[".handleSimpleError()"]);
-    remove_frame_prefix(stack, &[
-        "doWithOneRestart()",
-        "withOneRestart()",
-        "withRestarts()",
-        ".signalSimpleWarning",
-    ]);
-}
-
-/// Remove frames from the top of the stack that match the given prefixes in order.
-fn remove_frame_prefix(stack: &mut Vec<FrameInfo>, prefixes: &[&str]) {
-    for prefix in prefixes {
-        if stack
-            .first()
-            .is_some_and(|frame| frame.frame_name.starts_with(prefix))
-        {
-            stack.remove(0);
-        } else {
-            break;
-        }
     }
 }
 
--- a/crates/ark/src/console_debug.rs
+++ b/crates/ark/src/console_debug.rs
@@ -79,7 +79,7 @@ impl Console {
         debug_stopped_reason: DebugStoppedReason,
     ) {
         match self.debug_stack_info() {
-            Ok(stack) => {
+            Ok(mut stack) => {
                 // Figure out whether we changed location since last time,
                 // e.g. because the user evaluated an expression that hit
                 // another breakpoint. In that case we do want to move
@@ -99,6 +99,14 @@ impl Console {
                 self.debug_last_stack = stack_id;
 
                 let preserve_focus = same_stack && debug_preserve_focus;
+
+                let show = get_show_hidden_frames();
+                if !show.internal {
+                    remove_condition_handling_frames(&mut stack, &debug_stopped_reason);
+                }
+                if !show.fenced {
+                    remove_fenced_frames(&mut stack);
+                }
 
                 let mut dap = self.debug_dap.lock().unwrap();
                 dap.start_debug(
@@ -277,10 +285,6 @@ impl Console {
 
             log::trace!("DAP: Current call stack:\n{out:#?}");
 
-            if !harp::get_option_bool("ark.debugger.show_hidden_frames") {
-                filter_hidden_frames(&mut out);
-            }
-
             Ok(out)
         }
     }
@@ -352,6 +356,77 @@ impl Console {
     }
 }
 
+/// Controls which categories of hidden frames to show (i.e., not filter out).
+/// Parsed from the `ark.debugger.show_hidden_frames` R option.
+struct ShowHiddenFrames {
+    /// Show frames fenced by `..stacktraceon..`/`..stacktraceoff..` sentinels
+    fenced: bool,
+    /// Show internal condition-handling frames (e.g. `.handleSimpleError()`)
+    internal: bool,
+}
+
+/// Read the `ark.debugger.show_hidden_frames` R option via the R-side
+/// `debugger_show_hidden_frames()` helper, which validates and normalises
+/// the option into a character vector of categories to show.
+fn get_show_hidden_frames() -> ShowHiddenFrames {
+    let values =
+        match RFunction::new("", "debugger_show_hidden_frames").call_in(ARK_ENVS.positron_ns) {
+            Ok(obj) => Vec::<String>::try_from(obj).unwrap_or_default(),
+            Err(err) => {
+                log::warn!("Failed to read `ark.debugger.show_hidden_frames`: {err:?}");
+                return ShowHiddenFrames {
+                    fenced: false,
+                    internal: false,
+                };
+            },
+        };
+
+    ShowHiddenFrames {
+        fenced: values.iter().any(|v| v == "fenced"),
+        internal: values.iter().any(|v| v == "internal"),
+    }
+}
+
+/// Discard top frames that are part of the debug infrastructure rather than
+/// user code.
+fn remove_condition_handling_frames(
+    stack: &mut Vec<FrameInfo>,
+    stopped_reason: &DebugStoppedReason,
+) {
+    // Discard top frame when stopped due to exception breakpoint or pause,
+    // it points to our global handler that calls `browser()`
+    if matches!(
+        stopped_reason,
+        DebugStoppedReason::Condition { .. } | DebugStoppedReason::Pause
+    ) && !stack.is_empty()
+    {
+        stack.remove(0);
+    }
+
+    // Then discard base R's own condition handling/emitting frames, if any
+    remove_frame_prefix(stack, &[".handleSimpleError()"]);
+    remove_frame_prefix(stack, &[
+        "doWithOneRestart()",
+        "withOneRestart()",
+        "withRestarts()",
+        ".signalSimpleWarning",
+    ]);
+}
+
+/// Remove frames from the top of the stack that match the given prefixes in order.
+fn remove_frame_prefix(stack: &mut Vec<FrameInfo>, prefixes: &[&str]) {
+    for prefix in prefixes {
+        if stack
+            .first()
+            .is_some_and(|frame| frame.frame_name.starts_with(prefix))
+        {
+            stack.remove(0);
+        } else {
+            break;
+        }
+    }
+}
+
 /// Removes frames fenced between `..stacktraceon..` and `..stacktraceoff..`
 /// markers (used by Shiny to hide internal frames from error stack traces).
 ///
@@ -373,7 +448,7 @@ impl Console {
 ///
 /// The topmost frame (index 0) is never filtered out so the user always
 /// sees where they are stopped.
-fn filter_hidden_frames(frames: &mut Vec<FrameInfo>) {
+fn remove_fenced_frames(frames: &mut Vec<FrameInfo>) {
     let mut hidden_depth: u32 = 0;
     let mut first = true;
 
@@ -615,14 +690,14 @@ mod tests {
     #[test]
     fn test_filter_hidden_frames_empty() {
         let mut frames = vec![];
-        filter_hidden_frames(&mut frames);
+        remove_fenced_frames(&mut frames);
         assert!(frames.is_empty());
     }
 
     #[test]
     fn test_filter_hidden_frames_no_sentinels() {
         let mut frames = vec![frame("a()"), frame("b()"), frame("c()")];
-        filter_hidden_frames(&mut frames);
+        remove_fenced_frames(&mut frames);
         assert_eq!(names(&frames), vec!["a()", "b()", "c()"]);
     }
 
@@ -635,7 +710,7 @@ mod tests {
             frame("..stacktraceoff..()"),
             frame("outer()"),
         ];
-        filter_hidden_frames(&mut frames);
+        remove_fenced_frames(&mut frames);
         assert_eq!(names(&frames), vec!["user_code()", "outer()"]);
     }
 
@@ -656,7 +731,7 @@ mod tests {
             frame("outer_off()"),
             frame("top()"),
         ];
-        filter_hidden_frames(&mut frames);
+        remove_fenced_frames(&mut frames);
         assert_eq!(names(&frames), vec![
             "user_code()",
             "inner_off()",
@@ -685,7 +760,7 @@ mod tests {
             frame("..stacktraceoff..(captureStackTraces)"),
             frame("shiny::runApp()"),
         ];
-        filter_hidden_frames(&mut frames);
+        remove_fenced_frames(&mut frames);
         assert_eq!(names(&frames), vec![
             "renderPlot()",
             "output$distPlot()",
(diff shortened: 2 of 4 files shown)


crates/ark/src/dap/dap.rs
```diff
@@ -133,6 +144,9 @@ pub struct Dap {
     /// Known breakpoints keyed by URI, with document hash
     pub breakpoints: HashMap<Url, (blake3::Hash, Vec<Breakpoint>)>,
 
+    /// Filters for enabled condition breakpoints
+    pub breakpoints_conditions: Vec<String>,
```
DavisVaughan: Based on my reading of the spec and how this is actually used, can we name this 

```rust
pub exception_breakpoint_filters
```

That maps to the DAP type of ExceptionBreakpointsFilter and already goes with your helper name of `is_exception_breakpoint_filter_enabled`

---

The current name confuses me due to the lack of "filter" in the name in particular

crates/ark/src/dap/dap.rs
```diff
@@ -511,6 +549,46 @@ impl ServerHandler for Dap {
...
+) {
+    // Discard top frame when stopped due to exception breakpoint or pause,
+    // it points to our global handler that calls `browser()`
+    if matches!(
+        stopped_reason,
+        DebugStoppedReason::Condition { .. } | DebugStoppedReason::Pause
+    ) && !stack.is_empty()
+    {
+        stack.remove(0);
+    }
```
DavisVaughan: Can you match on a prefix here too? Just in case we are wrong? It also feels more self documenting if a prefix is shown in the code here about what we are trying to remove

crates/ark/src/dap/dap_server.rs
```diff
@@ -578,12 +612,13 @@ impl<R: Read, W: Write> DapServer<R, W> {
     fn handle_set_exception_breakpoints(
         &mut self,
         req: Request,
-        _args: SetExceptionBreakpointsArguments,
+        args: SetExceptionBreakpointsArguments,
     ) -> Result<(), ServerError> {
+        let mut state = self.state.lock().unwrap();
+        state.breakpoints_conditions = args.filters;
+        drop(state);
```
DavisVaughan: I do feel like a `with_lock(|state| { })` style helper here would be less of a footgun

crates/ark/src/dap/dap_server.rs
```diff
@@ -578,12 +612,13 @@ impl<R: Read, W: Write> DapServer<R, W> {
...
         let rsp = req.success(ResponseBody::SetExceptionBreakpoints(
-            SetExceptionBreakpointsResponse {
-                breakpoints: None, // TODO
-            },
+            SetExceptionBreakpointsResponse { breakpoints: None },
```
DavisVaughan: Interesting, so we don't have to return anything?

I read https://microsoft.github.io/debug-adapter-protocol//specification.html#Requests_SetExceptionBreakpoints

I saw

> For backward compatibility both the breakpoints array and the enclosing body are optional. If these elements are missing a client is not able to show problems for individual exception breakpoints or filters.

That struck me as odd, implying that we should be trying to send something back here

crates/ark/src/dap/dap_server.rs
```diff
@@ -753,6 +788,16 @@ impl<R: Read, W: Write> DapServer<R, W> {
         self.respond(rsp)
     }
 
+    fn handle_pause(&mut self, req: Request, _args: PauseArguments) -> Result<(), ServerError> {
+        self.state.lock().unwrap().is_interrupting_for_debugger = true;
```
DavisVaughan, line 816: I see this is only reset to `false` in the global calling handler

Is that safe? Should it be reset to `false` more often in case we never hit the global calling handler somehow or something?

crates/ark/src/modules/positron/errors.R
```diff
@@ -50,6 +59,20 @@
...
 #' @export
+.ps.errors.globalWarningHandler <- function(cnd) {
+    # If the debugger requested that we break on warnings, drop into `browser()`
+    if (debug_should_break_on_condition("warning")) {
+        debug_set_stopped_reason(cnd)
+
+        # Drop in the debugger
+        browser()
+
+        # Fall through to our error handler when we're done debugging
```
DavisVaughan, line 70: Remove

crates/ark/src/modules/positron/errors.R
```diff
@@ -78,6 +101,18 @@
     invokeRestart("muffleMessage")
 }
 
+globalInterruptHandler <- function(cnd) {
```
DavisVaughan, line 104: you might as well just call this `.ps.errors.globalInterruptHandler` for consistency?

crates/ark/src/modules/positron/errors.R
```diff
@@ -78,6 +101,18 @@
...
+    if (is_interrupting_for_debugger()) {
+        debug_set_stopped_reason_pause()
+        browser()
+        base::.tryResumeInterrupt()
+    }
+}
+
+debug_set_stopped_reason_pause <- function() {
```
DavisVaughan: maybe move this down with `debug_set_stopped_reason` and the other debug_ helpers?

crates/ark/src/console.rs
```diff
@@ -2679,6 +2694,7 @@ pub extern "C-unwind" fn r_read_console(
             // For a more practical example see Shiny app example in
             // https://github.com/rstudio/rstudio/pull/14848
             console.debug_is_debugging = false;
+            console.debug_stopped_reason = None;
             console.debug_stop();
```
DavisVaughan: should `debug_stop()` just set both of these?

crates/ark/tests/dap_exception_breakpoints.rs
```diff
@@ -0,0 +1,448 @@
...
+    dap.set_exception_breakpoints(&["error"]);
+
+    frontend.send_source("stop('test error')");
+
+    let (text, description) = dap.recv_stopped_exception();
+    assert!(text.contains("simpleError"));
+    assert!(description.contains("test error"));
+
+    let stack = dap.stack_trace();
+    assert!(!stack.is_empty());
```
DavisVaughan, line 31: Do we care about what it looks like? Snapshot it maybe?

crates/ark/src/dap/dap.rs
```diff
@@ -20,6 +20,7 @@ use stdext::result::ResultExt;
 use stdext::spawn;
 use url::Url;
```
DavisVaughan, line 22: Should rlang `abort()` frames be removed?

```r
# Check "stop on Errors" in the debug menu

# in dplyr
devtools::load_all() 
mutate(mtcars, x = 1 + "x")
```

Image

crates/ark/src/dap/dap.rs
```diff
@@ -20,6 +20,7 @@ use stdext::result::ResultExt;
...
 
+use crate::console::DebugStoppedReason;
 use crate::console_debug::FrameInfo;
```
DavisVaughan, line 24: Additional question related to the image in 

https://github.com/posit-dev/ark/pull/1036/changes#r2834408419

Should the `Called from:` debug output in the Console be showing the internal global calling handler for errors?

crates/ark/src/dap/dap.rs
```diff
@@ -20,6 +20,7 @@ use stdext::result::ResultExt;
...
 use crate::dap::dap_server;
 use crate::request::RRequest;
```
DavisVaughan, line 26: Additionally, hitting `next` after hitting this rlang error puts me in ark internals

Image

crates/ark/src/dap/dap.rs
```diff
@@ -109,13 +110,23 @@ pub enum DapBackendEvent {
         verified: bool,
```
DavisVaughan, line 110: Super useful in data.table, where you know it errors somewhere in their `[` method, but its >1k lines long and you want to quickly isolate where in there that it errors

Image

DavisVaughan: Note that `eval()` shows twice here because it drops into a primitive

reproduce yourself with

```r
df <- data.table::data.table(x = 1:5)
df[, y := x + "1"]
```

DavisVaughan: Also, should it be showing `Called from: h(simpleError())` in the Console?

Hmm, probably right? You're filtering them out of the Call Stack menu, but there's not much you can do about what R itself is showing the user

crates/ark/src/dap/dap.rs
DavisVaughan: And FWIW I also get here with the base R error thrown by data.table, so this one isn't rlang specific.

crates/ark/src/dap/dap.rs
```diff
@@ -109,13 +110,23 @@ pub enum DapBackendEvent {
...
         message: Option<String>,
     },
+
+    /// Event sent when an exception/error occurs
+    Exception(DapExceptionEvent),
 }
```
DavisVaughan, line 116: What is the expected behavior when you throw an error _while paused in the browser_?

Right now it prints the error out in red and pauses you at the internal `browser()` call

```r
df <- data.table::data.table(x = 1:5)
df[, y := x + "1"] # get in the debugger with this
stop("oh no") # run this while in the debugger
```

Image

crates/ark/src/dap/dap.rs
DavisVaughan: Same for `rlang::warn()` of course

DavisVaughan: With tests!

crates/ark/src/dap/dap_server.rs
```diff
@@ -192,6 +193,19 @@ fn listen_dap_events<W: Write>(
                         })
                     },
 
+                    DapBackendEvent::Exception(DapExceptionEvent { class, message, preserve_focus }) => {
+                        let text = format!("<{class}> {message}");
```
DavisVaughan: Can you conditionally add a newline after `<{class}>` if `message` contains a newline?

That way this displays better

Image

DavisVaughan: Or unconditionally add a newline there, since `<rlang_warning>` is like a "header"

crates/ark/src/dap/dap_server.rs
```diff
@@ -325,6 +340,25 @@ impl<R: Read, W: Write> DapServer<R, W> {
...
+                    condition_description: None,
+                },
+                types::ExceptionBreakpointsFilter {
+                    filter: String::from("warning"),
+                    label: String::from("Warnings"),
+                    description: Some(String::from("Break on R warnings")),
+                    default: Some(false),
+                    supports_condition: Some(false),
+                    condition_description: None,
+                },
```
DavisVaughan, line 368: I would not be opposed to a filter for `interrupt` as well

_You and I_ know that you can do `Debug: Pause` to accomplish this, but when I was testing this locally _even I_ forgot how to do this, and was staring at the `Errors` and `Warnings` check boxes and wondering why there wasn't one for `Interrupts`

I think that will be very useful for people that are stuck in some infinite loop somewhere, and forcing them to call `Debug: Pause` feels like we are just hiding away some awesome feature

crates/ark/src/dap/dap.rs
lionel-: This is out of scope for this PR. For rlang we need some kind of protocol that allows the IDE to track the relevant caller. And ideally we'd support the error-caller pattern too. Maybe rlang could set a flag in the user-facing calling environment that Ark could detect. That's kind of an intrusive approach though so I don't know.

crates/ark/src/dap/dap.rs
lionel-: We'll filter all debug output in an ulterior PR.

crates/ark/src/dap/dap.rs
lionel-: I noticed that too but wasn't sure how to address this. Now I have an idea (adding return and stepping into to trick R into stepping out). But that means our default handler won't run on resume.

Edit: oh we can just run `f` instead of `n` when we're in that frame.

crates/ark/src/modules/positron/errors.R
lionel-: Or the other way around? I was wondering if we really needed to export these. Nowadays we can access with `.ps.internal()`.

crates/ark/src/dap/dap.rs
lionel-: hmm... for some reason I'm not a fan of this idea, perhaps because it introduces a stringy dependency on fragile internal details.

crates/ark/src/dap/dap_server.rs
lionel-: hmm. Not sure about that, a block would be a more conventional alternative I think.

crates/ark/src/dap/dap_server.rs
lionel-: I read it as `breakpoints` being an `Option<>` because it is a protocol extension. And the response in only useful for errors / invalid breakpoints, which doesn't apply to us since we always succeed.

See the `?` in the typescript type:

```ts
breakpoints?: Breakpoint[];
```

crates/ark/src/modules/positron/errors.R
lionel-: Why do you want to remove?

crates/ark/src/dap/dap.rs
lionel-: We have no control over this output emitted by R but I plan to strip all debug output in an ulterior PR.

crates/ark/src/dap/dap_server.rs
lionel-: Good idea.

crates/ark/src/dap/dap_server.rs
lionel-: That's a great idea

crates/ark/src/dap/dap_server.rs
lionel-: yep good catch! Fixed with a test.

crates/ark/tests/dap_exception_breakpoints.rs
lionel-: hmm I prefer not to maintain potentially per-version snapshots and this doesn't seem critical to test

crates/ark/src/dap/dap.rs
lionel-: I'm not sure how you get there, because our global error handler doesn't run for transient evals that error out. R's own `error` calling handler intercepts the error to restart to the browser context. That's why we're not able to provide structured backtraces for these errors.

However, now that we are in control of evaluation, we can attach our handlers for every `eval()` calls. We no longer need to call into `globalCallingHandlers()` at all! And that should solve all weirdness related to calling handlers in debugging contexts.

crates/ark/src/modules/positron/errors.R
DavisVaughan: It's not about the error handler, its the warning handler. So I think we should at least change that.

But it also doesn't do anything right now and the comment below this handles it so i figured we could just remove it

crates/ark/src/modules/positron/errors.R
DavisVaughan: I'm fine with it if you want to unexport them! It sounds better to have them as internal

crates/ark/src/dap/dap.rs
lionel-: Now done, and that solves https://github.com/posit-dev/positron/issues/12085

PR #1037, opened 2026-02-10
Add Watch Pane support

lionel-:
Branched from #1033 (for DAP tests)
Addresses https://github.com/posit-dev/positron/issues/1765

Adds support for evaluating expressions from the debug watch pane via DAP `evaluate` requests. One such request is sent for every watch expression.

Since we're asynchronously getting these requests, and R might no longer be idle at that time, the `evaluate` requests run as _idle tasks_ (see https://github.com/posit-dev/ark/pull/371) that only run once R is safely idle (waiting in `ReadConsole`). The regular idle tasks spawned with `spawn_idle()` do not run when the debugger is active. This is the safe default for general purpose tasks, but for `evaluate` requests we obviously need them to run when R is paused. I've introduced `spawn_idle_any()` for tasks that also need to run at debug prompts.

Also adds infrastructure for capturing console output during idle tasks processing `evaluate` requests. This way the output is logged at info level instead of being emitted as orphaned IOPub streams.

I've also added support for _printing_ in the watch pane by prefixing the expression with `/print `. In that case we return the captured output instead of the object. That doesn't work that well though because you need to hover to see the captured outpub. This feature is almost free though so I kept it.

https://github.com/user-attachments/assets/e79d424f-2870-492b-b982-8f830f34100d

### QA Notes

- Expressions in the watch pane should be sensitive to selected frame, and should update at each step.

- Expressions evaluate to the same kind of object as in the variable pane. Complex values are expandable.

- By prefixing with the slash command `/print`, you can get the printed output, e.g. `/print mtcars`. Truncated but full output available on hover.

- Watch expressions at top-level browser (e.g. `browser()` called directly) should evaluate in the global environment.

- Invalid expressions should show error messages.

### DIFF
--- a/crates/ark/src/dap/dap.rs
+++ b/crates/ark/src/dap/dap.rs
@@ -438,7 +438,7 @@ impl Dap {
                     Ok(object_variable(String::new(), value.sexp))
                 }
             },
-            Err(err) => Err(format!("{err}")),
+            Err(err) => Err(evaluate_error_message(err)),
         }
     }
 
@@ -616,6 +616,13 @@ impl Dap {
                     BreakpointState::Verified | BreakpointState::Unverified
                 )
         })
+    }
+}
+
+fn evaluate_error_message(err: harp::Error) -> String {
+    match err {
+        harp::Error::TryCatchError { message, .. } => message,
+        err => format!("{err}"),
     }
 }
 
--- a/crates/ark/tests/dap_evaluate.rs
+++ b/crates/ark/tests/dap_evaluate.rs
@@ -189,6 +189,10 @@ local({
         err.contains("not found"),
         "Expected 'not found' error, got: {err}"
     );
+    assert!(
+        !err.contains("backtrace"),
+        "Error should not contain backtrace, got: {err}"
+    );
 
     // Evaluate incomplete code
     let err = dap.evaluate_error("1 +", Some(frame_id));
@@ -202,6 +206,10 @@ local({
     assert!(
         err.contains("intentional error"),
         "Expected error message, got: {err}"
+    );
+    assert!(
+        !err.contains("backtrace"),
+        "Error should not contain backtrace, got: {err}"
     );
 
     // Debug session should still be alive and stopped after all these errors


lionel-:
TODO: Error responses should not include backtrace

Screenshot 2026-02-19 at 10 09 57

crates/ark/src/dap/dap.rs
```diff
@@ -280,11 +280,11 @@ impl Dap {
...
-            let environment = frame.environment.take();
-
-            let Some(environment) = environment else {
-                continue;
-            };
+            // If the frame has no environment (e.g. top-level browser), use global env.
+            let environment = frame
+                .environment
+                .take()
+                .unwrap_or_else(|| RThreadSafe::new(RObject::new(unsafe { libr::R_GlobalEnv })));
```
DavisVaughan: `R_ENVS.global`?

crates/ark/src/dap/dap_server.rs
```diff
@@ -118,22 +123,21 @@ pub fn start_dap(
         );
 
         let (backend_events_tx, backend_events_rx) = unbounded::<DapBackendEvent>();
+        let (responses_tx, responses_rx) = unbounded::<Response>();
         let (done_tx, done_rx) = bounded::<bool>(0);
+        server.responses_tx = Some(responses_tx);
```
DavisVaughan: It looks to me like `responses_tx` could be an input to `DapServer::new()` and doesn't necessarily need to be `Option<>`

crates/ark/src/dap/dap_server.rs
```diff
@@ -681,6 +708,66 @@ impl<R: Read, W: Write> DapServer<R, W> {
...
+        let state = self.state.clone();
+
+        let Some(responses_tx) = self.responses_tx.clone() else {
+            let rsp = req.error("DAP: No responses channel for `evaluate`");
+            self.respond(rsp)?;
+            return Ok(());
+        };
+
+        log::trace!("DAP: Spawning idle task for evaluate");
+        spawn_idle_any_prompt(move |mut capture| async move {
```
DavisVaughan, line 780: `move |mut capture| async move` that is a mouthful

crates/ark/src/dap/dap_server.rs
```diff
@@ -681,6 +708,66 @@ impl<R: Read, W: Write> DapServer<R, W> {
...
+                            Some(obj) => {
+                                let mut state = state.lock().unwrap();
+                                state.insert_variables_reference_object(obj)
+                            },
+                            None => 0,
+                        };
+
+                        req.success(ResponseBody::Evaluate(EvaluateResponse {
+                            result: variable.value,
+                            type_field: variable.type_field,
```
DavisVaughan: We have

```rust
fn into_variables(&self, variables: Vec<RVariable>) -> Vec<Variable> {
```

which I quite like. In particular I like that it captures the mutation of `state.insert_variables_reference_object(obj)` in a nicely named helper, so you know exactly why we are doing this.

Would you be open to adding

```rust
fn into_evaluate_response(&self, variable: RVariable) -> EvaluateResponse {
```

which would serve a similar purpose and would live next to `into_variables()`?

 I think it would be nice to have all references to `state.insert_variables_reference_object(obj)` close together.

crates/ark/src/dap/dap_server.rs
```diff
@@ -769,6 +856,58 @@ impl<R: Read, W: Write> DapServer<R, W> {
...
+    let Some(obj) = state
+        .variables_reference_to_r_object
+        .get(&variables_reference)
+    else {
+        return Err(format!(
+            "Unknown `variables_reference`: {variables_reference}"
+        ));
+    };
+
+    Ok(obj.get().sexp)
```
DavisVaughan: nit: i dont really have a concrete comment here, but something feels weird about having a function that must be called on the main R thread live here

crates/ark/src/dap/dap_variables.rs
```diff
@@ -156,6 +156,12 @@ fn object_variable(name: String, x: SEXP) -> RVariable {
     }
 }
 
+/// Create an `RVariable` from an evaluated expression result.
+/// Used for watch expressions in the DAP evaluate request.
+pub(super) fn object_variable_from_value(x: SEXP) -> RVariable {
+    object_variable(String::new(), x)
+}
```
DavisVaughan: Maybe move it up with the other entry point of this file, `pub(super) fn object_variables(x: SEXP)`?

Maybe it should be `dap_evaluate_variable()`?

---

Oh, or what if you actually just expose `object_variable()` itself instead of this and move `object_variable()` up with `object_variables()`? And then at the call site on the DAP side you just supply `String::new()` and mention that the name won't matter in this case?

That feels nice to me, where this is the only public API for this file

```rust
pub(super) fn object_variable(name: String, x: SEXP)
pub(super) fn object_variables(x: SEXP)
```

crates/ark/src/console.rs
```diff
@@ -273,9 +271,8 @@ pub struct Console {
     /// Stored in `Console` to avoid memory leakage when `Rf_error()` jumps.
     r_error_buffer: Option<CString>,
 
-    /// `WriteConsole` output diverted from IOPub is stored here. This is only used
-    /// to return R output to the debugger.
-    pub(crate) captured_output: String,
+    /// When `Some`, console output is captured here instead of being sent to IOPub.
+    pub(crate) captured_output: Option<String>,
```
DavisVaughan, line 284: Should you mention that you should only interact with this via `ConsoleOutputCapture`?

crates/ark/src/console.rs
```diff
@@ -902,6 +962,18 @@ impl Console {
         &self.iopub_tx
     }
 
+    /// Start capturing console output.
+    /// Returns a guard that restores the previous state on drop.
+    pub(crate) fn start_capture(&mut self) -> ConsoleOutputCapture {
+        let was_capturing = self.captured_output.is_some();
+        self.captured_output = Some(String::new());
```
DavisVaughan: "restores the previous state on drop"

but...not really? If there is preexisting output in `self.captured_output` when you start capturing, then it gets lost completely right? Since you're just tracking a boolean.

crates/ark/src/console.rs
```diff
@@ -1117,16 +1190,20 @@ impl Console {
...
-        // channels.
         let tasks_idle_index = if matches!(info.kind, PromptKind::TopLevel) {
             Some(select.recv(&tasks_idle_rx))
         } else {
             None
         };
 
+        // "Idle any" tasks run at both top-level and browser prompts
+        let tasks_idle_any_index =
+            if matches!(info.kind, PromptKind::TopLevel | PromptKind::Browser) {
```
DavisVaughan, line 1221: I would not be upset if these were called "debug" tasks or "idle debug" tasks instead of "idle any" tasks

It seems likely to me that these will only be used for debugger related things, and "debug" in the name would discourage us from accidentally using them for other things.

Mostly I just don't feel like `idle_any` invokes much meaning about what it does, especially since the main difference is that they also run at `Browser` time

crates/ark/src/debug.rs
```diff
@@ -162,14 +160,19 @@ pub fn tidy_kind(kind: libr::SEXPTYPE) -> &'static str {
 /// This should only be used in a debugging context where leaking is not an
 /// issue.
 pub fn capture_console_output(cb: impl FnOnce()) -> *const ffi::c_char {
-    let old = CAPTURE_CONSOLE_OUTPUT.swap(true, Ordering::SeqCst);
```
DavisVaughan, line 165: It smells like you should be using the new `ConsoleOutputCapture` guard here?

crates/ark/src/r_task.rs
```diff
@@ -290,6 +303,47 @@ where
     spawn_ext(fun, false)
 }
 
+/// Spawn an async task that runs when R is at any idle prompt (top-level or browser).
+/// Unlike `spawn_idle` which only runs at top-level, this also runs during debug sessions.
+///
+/// The closure receives a `ConsoleOutputCapture` that can be used to capture
+/// console output during the task. Call `take()` on it to retrieve output, which
+/// can be done multiple times. Any remaining output is logged when the capture is dropped.
+pub(crate) fn spawn_idle_any_prompt<F, Fut>(fun: F)
```
DavisVaughan, line 312: Yea I think I would be happy if this was `spawn_idle_debug()` which means "top level or debug" prompt

lionel-: to me that name makes it sounds like it's debug-specific

crates/ark/src/dap/dap_server.rs
lionel-: Good idea. I've also moved them to DAP so the methods don't take a lock internally.

crates/ark/src/dap/dap_server.rs
lionel-: I see what you mean!

lionel-: I've moved this as methods on `Dap` in `dap.rs`, a much better fit.

PR #1038, opened 2026-02-12
Reset global env debug flag on error too

lionel-:
@DavisVaughan It's because of this bug I had to restart R right in the middle of the debugger demo!

### DIFF
--- a/crates/ark/src/console.rs
+++ b/crates/ark/src/console.rs
@@ -1043,8 +1043,6 @@ impl Console {
 
             // Reply to active request with error, then fall through to event loop
             self.handle_active_request(&info, ConsoleValue::Error(exception));
-
-            self.reset_global_env_rdebug();
         } else if matches!(info.kind, PromptKind::InputRequest) {
             // Request input from the frontend and return it to R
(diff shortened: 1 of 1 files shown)


crates/ark/src/console.rs
```diff
@@ -1055,15 +1057,7 @@ impl Console {
...
-            // Reset debug flag on the global environment. This is a workaround
-            // for when a breakpoint was entered at top-level, in a `{}` block.
-            // In that case `browser()` marks the global environment as being
-            // debugged here: https://github.com/r-devel/r-svn/blob/476ffd4c/src/main/main.c#L1492-L1494.
-            // Only do it when the call stack is empty, as removing the flag
-            // prevents normal stepping with `source()`.
-            if harp::r_n_frame().unwrap_or(0) == 0 {
-                unsafe { libr::SET_RDEBUG(libr::R_GlobalEnv, 0) };
-            }
+            self.reset_global_env_rdebug();
```
DavisVaughan: Should this be inside `handle_active_request()`?

lionel-: Good idea!

PR #1040, opened 2026-02-13
Implement `StatementRangeSuccess` and `StatementRangeRejection`

DavisVaughan:
Branched from #1028 
Goes with https://github.com/posit-dev/positron/pull/11907, and all documentation is there

### DIFF
--- a/crates/ark/src/lsp/statement_range.rs
+++ b/crates/ark/src/lsp/statement_range.rs
@@ -2023,6 +2023,7 @@ fn <- function() {
     }
 
     // Useful when testing the success case
+    #[track_caller]
     fn find_roxygen_statement_range_success(
         root: &Node,
         contents: &str,
@@ -2051,6 +2052,7 @@ fn <- function() {
     }
 
     // Useful when testing the rejection case
+    #[track_caller]
     fn find_roxygen_statement_range_rejection(
         root: &Node,
         contents: &str,


crates/ark/src/lsp/statement_range.rs
```diff
@@ -37,16 +40,89 @@ pub struct StatementRangeParams {
...
+
+#[derive(Debug, Eq, PartialEq)]
+enum ArkStatementRangeResponse {
+    Success(ArkStatementRangeSuccess),
+    Rejection(ArkStatementRangeRejection),
+}
+
+#[derive(Debug, Eq, PartialEq)]
+struct ArkStatementRangeSuccess {
+    range: tree_sitter::Range,
```
DavisVaughan, line 83: It got very annoying to try and have a `StatementRangeSuccess` with a LSP `Range` deep within the roxygen2 code path that I had to convert back to a tree-sitter `Range` to perform the "subdocument adjustment" on.

I think it is better to have our own `Ark*` types here that use our internal data structure, and only convert to and from the LSP types once at the boundaries.

This feels similar to what rust-analyzer does?

crates/ark/src/lsp/statement_range.rs
```diff
@@ -328,7 +427,25 @@ fn find_roxygen_examples_range(
...
+        ArkStatementRangeRejection::Syntax(subdocument_rejection) => {
+            // Adjust line number of the syntax rejection to reflect original document
+            Some(ArkStatementRangeRejection::Syntax(
+                ArkStatementRangeSyntaxRejection {
+                    line: subdocument_rejection.line + row_adjustment as u32,
+                },
+            ))
+        },
+    }
 }
```
DavisVaughan, line 447: Small bit of magic here, but it's not really that fancy!

crates/ark/src/lsp/statement_range.rs
```diff
@@ -389,11 +506,14 @@ fn find_statement_range(root: &Node, row: usize) -> Option<tree_sitter::Range> {
...
+        // anywhere past the first parse error, returning a syntax rejection that points
+        // to the first line in this child node, which the frontend will show the user
+        // (posit-dev/positron#5023, posit-dev/positron#8350).
         if node_has_error_or_missing(&child) {
-            return None;
+            let line = child.start_position().row as u32;
+            return Ok(Some(ArkStatementRangeResponse::Rejection(
+                ArkStatementRangeRejection::Syntax(ArkStatementRangeSyntaxRejection { line }),
+            )));
         }
```
DavisVaughan, line 515: Here's where we newly switch from `None` to a `StatementRangeSyntaxRejection`, which propagates up through both the normal R code path and the roxygen2 code path

crates/ark/src/lsp/statement_range.rs
```diff
@@ -1816,25 +1984,31 @@ fn3 <- function() {@ # Can't execute this!
...
     1 + 1@
     sum(
 }
 ",
         );
-        assert!(range.is_none());
+        assert_matches::assert_matches!(
+            rejection,
+            ArkStatementRangeRejection::Syntax(ArkStatementRangeSyntaxRejection { line: 1 })
+        );
```
DavisVaughan, line 2009: There's a lot of churn in the tests, but changes like this one in our existing tests are the meaningful bits

crates/ark/src/lsp/statement_range.rs
```diff
@@ -37,16 +40,89 @@ pub struct StatementRangeParams {
...
+
+#[derive(Debug, Eq, PartialEq)]
+struct ArkStatementRangeSyntaxRejection {
+    line: u32,
+}
+
+// Sole conversion method between `ArkStatementRangeResponse` and `StatementRangeResponse`,
+// which handles all Tree-sitter to LSP conversion at the method boundary
+fn statement_range_response_from_ark_statement_range_response(
+    response: ArkStatementRangeResponse,
```
lionel-: That name is rather long, maybe make it an `into_lsp_response()` on `ArkStatementRangeResponse`?

crates/ark/src/lsp/statement_range.rs
lionel-: yep I think that's generally our ideal pattern in Ark/Air

(Ideally that'd be something based on TextPos, now that we've started using Rowan, but it does seem fine to have a TS type for now since that's what being consumed downstream.)

crates/ark/src/lsp/statement_range.rs
```diff
@@ -1850,6 +2024,63 @@ fn <- function() {
         (text, point)
     }
 
+    // We typically want the `.unwrap().unwrap()` behavior during tests because we are
+    // testing the actual range and code returned
+    fn find_roxygen_statement_range_unsafe(
```
lionel-: nit (feel free to ignore): `assert_` prefix would be a better than `_unsafe` suffix here? We're not doing anything unsafe in the Rust sense.

lionel-: `#[track_caller]`?

Issue #1035, opened 2026-02-10
`source()` hook should return last value


lionel-:
Split from https://github.com/posit-dev/positron/issues/11784

Source a file with a breakpoint. `source()` currently returns `NULL` instead of the `withVisible()` last value:

https://github.com/user-attachments/assets/37d5f258-76fa-4d2b-81d7-e2563279fe96

PR #1041, opened 2026-02-16
Return last value from annotated source

lionel-:
Addresses https://github.com/posit-dev/ark/issues/1035
Progress towards https://github.com/posit-dev/positron/issues/11784

`annotate_source()` created a trailing breakpoint verification call that overrode the last value, breaking e.g. `shiny::runApp()` (https://github.com/rstudio/shiny/pull/4352).

Now we wrap the annotated source in `base::list(*source*, *verify_call*)[[1]]` to:
1. Execute the code block
2. Execute the trailing verify call
3. Return only the code block's result via `[[1]]`

Also fixes the `source()` hook to return invisibly the last value wrapped in `withVisible()` to match `base::source()` behaviour.

### QA Notes

- `source()` with breakpoints should return the same structure as `base::source()`: `list(value = <last_expr>, visible = <TRUE/FALSE>)`
- `source()` should return invisibly

Manual testing: Shiny apps with breakpoints should launch via `runApp()` correctly.

### DIFF
--- a/crates/ark/tests/kernel-hooks-source.rs
+++ b/crates/ark/tests/kernel-hooks-source.rs
@@ -1,5 +1,6 @@
 use std::io::Write;
 
+use amalthea::fixtures::dummy_frontend::ExecuteRequestOptions;
 use ark_test::DummyArkFrontend;
 use ark_test::SourceFile;
 
@@ -153,6 +154,44 @@ invisible(42)
             result
         );
     });
+}
+
+#[test]
+fn test_source_error_propagates_correctly() {
+    let frontend = DummyArkFrontend::lock();
+    let mut dap = frontend.start_dap();
+
+    // Create a file that will error when sourced.
+    let file = SourceFile::new(
+        "
+foo <- function() {
+    1
+}
+1 + 'x'
+",
+    );
+
+    // Set a breakpoint inside the function (never called) to trigger annotation.
+    let breakpoints = dap.set_breakpoints(&file.path, &[3]);
+    assert_eq!(breakpoints.len(), 1);
+
+    // Source the file - should error.
+    let code = format!(r#"source("{}")"#, file.path);
+    frontend.send_execute_request(&code, ExecuteRequestOptions::default());
+    frontend.recv_iopub_busy();
+    frontend.recv_iopub_execute_input();
+    dap.recv_breakpoint_verified();
+
+    // The error should propagate correctly through the list() wrapper.
+    let error = frontend.recv_iopub_execute_error();
+    assert!(
+        error.contains("non-numeric argument"),
+        "Expected type error, got: {}",
+        error
+    );
+
+    frontend.recv_iopub_idle();
+    frontend.recv_shell_execute_reply_exception();
 }
 
 #[test]


crates/ark/src/modules/positron/hooks_source.R
```diff
@@ -133,12 +133,13 @@ make_ark_source <- function(original_source) {
...
-            log_trace("`source()`: Expected a single `{}` expression")
+            log_trace("`source()`: Expected a single `list()[[1]]` expression")
         }
 
         # `eval()` loops over the expression vector, handling gracefully
-        # unexpected lengths (0 or >1)
-        eval(parsed, env)
+        # unexpected lengths (0 or >1). Return withVisible() result to match
+        # base::source() structure.
+        invisible(withVisible(eval(parsed, env)))
```
DavisVaughan:
```suggestion
        # unexpected lengths (0 or >1). Return withVisible() result invisibly
        # to match base::source() structure.
        invisible(withVisible(eval(parsed, env)))
```

Slight tweak since this is slightly confusing

crates/ark/src/snapshots/ark__console_annotate__tests__annotate_source_breakpoint_on_function_definition_line.snap
```diff
@@ -2,7 +2,7 @@
 source: crates/ark/src/console_annotate.rs
 expression: result
 ---
-{
+base::list({
```
DavisVaughan, line 5: Do you happen to have any tests that test what happens when you put something like `1 + "x"` in a file and then `source()` it? i.e. something that errors?

It works nicely, i just wanted to make sure that this `list()` didn't affect anything

crates/ark/src/console_annotate.rs
```diff
@@ -277,29 +277,32 @@ pub(crate) fn annotate_source(
...
+    // 3. Return only the code block's result via `[[1]]`
     //
-    // This is unconditional for simplicity.
+    // This ensures breakpoints in the last expression get verified while
+    // preserving the return value for callers that depend on it.
+    // Note: visibility is not preserved (invisible results become visible), but
+    // this is acceptable since `source()` always returns invisibly, and
+    // `load_all()` doesn't look at return values at all, and `runApp()` doesn't
+    // return the result to the user. If visibility is important, wrap your code
+    // in a `withVisible()` before annotating with Ark.
```
DavisVaughan: I'm not totally sure I understand this

How could visibility be preserved even with `base::source()`? What does that even mean? It seems to always return invisibly and use `withVisible()` no matter what.

So it feels like we are doing the same thing here

Mostly im just confused by this comment

lionel-: yep this is confusing. And now that I take a second look at this, I do think we're going to get the wrong `withVisible()` value from our `source()` hook (i.e. always visible). So even though user-visible behaviour is not affected, programmatic behaviour might be.

I'll double check. That may be a limitation that we're going to have to document though.

Edit: oh maybe I can wrap in `withVisible({...})` before annotating (i.e. do what my comment advises).

lionel-: I've added a `withVisible = FALSE` argument to `.ark_annotate_source()`, which our source hook sets to `TRUE`. The comment is now:

```
    // When `with_visible` is true, wrap the code block in `withVisible()` so
    // callers can access the visibility of the last expression. This is needed
    // e.g. for `source()` which returns `list(value = ..., visible = ...)`.
```

crates/ark/src/snapshots/ark__console_annotate__tests__annotate_source_breakpoint_on_function_definition_line.snap
lionel-: We now have one

PR #1042, opened 2026-02-17
Fix truncated stack trace in DAP

lionel-:
Branched from #1041 
Addresses https://github.com/posit-dev/positron/issues/11891

The `total_frames` field in the `StackTrace` response was reporting the size of the current page rather than the full stack size. Frontends that paginate (e.g. requesting `start_frame + levels`) would see `total_frames`: 20 for a 20-frame page and never request more, resulting in truncated deep stacks.

Also fixes some unwraps that could take the DAP down in case of protocol-level errors.

### QA Notes

Place a breakpoint somewhere with a very large stack, such as in a Shiny app. The call stack in the frontend should match what `sys.calls()` returns in the R console.

### DIFF
--- a/crates/ark/src/dap/dap_server.rs
+++ b/crates/ark/src/dap/dap_server.rs
@@ -593,39 +593,49 @@ impl<R: Read, W: Write> DapServer<R, W> {
         req: Request,
         args: StackTraceArguments,
     ) -> Result<(), ServerError> {
-        let state = self.state.lock().unwrap();
-        let stack = &state.stack;
-        let fallback_sources = &state.fallback_sources;
-
-        let stack = match stack {
-            Some(stack) => stack
-                .into_iter()
-                .map(|frame| into_dap_frame(frame, fallback_sources))
-                .collect(),
-            _ => vec![],
+        let stack = {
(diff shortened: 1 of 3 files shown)


crates/ark/src/dap/dap_server.rs
```diff
@@ -593,39 +593,49 @@ impl<R: Read, W: Write> DapServer<R, W> {
...
         } else {
             n_usize
         };
 
+        let Ok(total_frames) = i64::try_from(n_usize) else {
+            let rsp = req.error(&format!("Stack frame count overflows i64: {n_usize}"));
+            return self.respond(rsp);
+        };
         let stack = stack[start..end].to_vec();
-        let n = stack.len().try_into().unwrap();
```
DavisVaughan, line 621: It took me a minute, but the main problem is the ordering here right? i.e. this would have been correct right?

```
        let n = stack.len().try_into().unwrap();
        let stack = stack[start..end].to_vec();
```

I appreciate all the other safety improvements, just checking my understanding

lionel-: yep

PR #1043, opened 2026-02-17
Filter out internal Shiny stack

lionel-:
Branched from #1042
Addresses https://github.com/posit-dev/positron/issues/11892

Filters out the call stack between Shiny trace fences: https://github.com/rstudio/shiny/blob/08383ad8b978ed4add7dedcb3a16929c2610d2d8/R/conditions.R#L617-L634. This can be disabled by setting `ark.debugger.show_hidden_frames` to `TRUE`.

Here is how it looks like currently. This could be improved with better fencing, I've opened an issue on the Shiny side (https://github.com/rstudio/shiny/issues/4357).

shiny-stack-positron

Note that the resulting stack is different from RStudio's when internal frames are filtered out:

shiny-stack-rstudio

That's because in RStudio, any frame that doesn't have srcrefs is an internal frame. We could add an option for this in Ark but I think that sort of filtering is a bit too aggressive given our long term goal of having srcrefs for all installed packages.

Edit: Here is the call stack with latest Shiny installed:

Screenshot 2026-02-25 at 15 03 27

### DIFF
--- a/crates/ark/src/console_debug.rs
+++ b/crates/ark/src/console_debug.rs
@@ -368,6 +368,7 @@ fn filter_hidden_frames(frames: &mut Vec<FrameInfo>) {
 
     // `Vec::retain` iterates front-to-back (guaranteed by std)
     frames.retain(|frame| {
+        // Always preserve the topmost frame so the user sees where they're stopped
         if first {
             first = false;
             return true;
@@ -378,6 +379,11 @@ fn filter_hidden_frames(frames: &mut Vec<FrameInfo>) {
             return false;
         }
         if frame.frame_name.starts_with("..stacktraceoff..") {
+            if hidden_depth == 0 {
+                log::trace!(
+                    "Unbalanced `..stacktraceoff..` without prior `..stacktraceon..` in call stack"
+                );
+            }
             hidden_depth = hidden_depth.saturating_sub(1);
             return false;
         }


crates/ark/src/console_debug.rs
```diff
@@ -321,6 +325,62 @@ impl Console {
...
+/// Since our stack is innermost-first (opposite to Shiny's traversal), the
+/// semantics invert: `..stacktraceon..` *enters* a hidden region (going
+/// outward) and `..stacktraceoff..` *exits* it. We use `hidden_depth` as an
+/// equivalent to Shiny's score, where `depth == 0` means visible.
+///
+/// Note: Shiny also uses `..stacktracefloor..` to truncate stacks entirely
+/// below that point. We don't handle this since showing the full context
+/// (e.g. `shiny::runApp()`) is useful in a debugger. Shiny's
+/// `shiny.fullstacktrace` option disables filtering; our equivalent is
+/// `ark.debugger.show_hidden_frames`.
```
DavisVaughan, line 361: Should you just use the `shiny.fullstacktrace` option as well? That feels right to me

crates/ark/src/console_debug.rs
```diff
@@ -321,6 +325,62 @@ impl Console {
...
+///
+/// The topmost frame (index 0) is never filtered out so the user always
+/// sees where they are stopped.
+fn filter_hidden_frames(frames: &mut Vec<FrameInfo>) {
```
DavisVaughan, line 365: Since this is literally just shiny, I'd love if this was called `shiny_filter_hidden_frames()` or something with shiny in the name

crates/ark/src/console_debug.rs
```diff
@@ -250,6 +250,10 @@ impl Console {
 
             log::trace!("DAP: Current call stack:\n{out:#?}");
 
+            if !harp::get_option_bool("ark.debugger.show_hidden_frames") {
+                filter_hidden_frames(&mut out);
+            }
```
DavisVaughan, line 271: If you don't want to use `shiny.fullstacktrace`, then I think I'd like to see `shiny` in the option name somewhere. This feels like too valuable of an option name to give out to a shiny specific feature to me

DavisVaughan: For example, the tryCatch frames could be considered "hidden frames" and we filter those out even if you set this option. So putting "shiny" in the name somewhere would clarify that confusion

crates/ark/src/console_debug.rs
DavisVaughan, line 365: This is so ridiculously shiny specific that I would not be against moving this and all its tests to `console_debug_shiny.rs` to not clutter this file up

lionel-: well it's a protocol/convention made by Shiny, but anyone can use it to filter out frames.

crates/ark/src/console_debug.rs
lionel-: I had thought about extension, and I think this could be:

- TRUE/FALSE to enable/disable all filtering
- an enum character vector to opt into specific filtering

The fact that it's a single option makes it easier to reason about from a user perspective.

(As mentioned above I would not use the "shiny" name in the enum though, since I think the protocol has a larger scope.)

crates/ark/src/console_debug.rs
lionel-: I made it intentionally generic...

crates/ark/src/console_debug.rs
lionel-: I was going to document this in the debugger doc as a generic way of filtering the stack.

See also below.

PR #1044, opened 2026-02-17
Disable JIT compiler before stepping into

lionel-:
Closes https://github.com/posit-dev/positron/issues/11890

We now disable the JIT compiler right before stepping-into (`s`). The previous setting is restored when entering read-console again.

### DIFF
--- a/crates/ark/src/console.rs
+++ b/crates/ark/src/console.rs
@@ -307,6 +307,10 @@ pub struct Console {
     /// Reset after `debug_stop()`, not between debug steps.
     pub(crate) debug_current_frame_id: i64,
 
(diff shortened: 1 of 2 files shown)


crates/ark/src/console.rs
```diff
@@ -836,6 +840,7 @@ impl Console {
             debug_last_stack: vec![],
             debug_session_index: 1,
             debug_current_frame_id: 0,
+            debug_jit_level: None,
```
DavisVaughan, line 843: It's starting to feel like we need `DebugConsole` to hold all of these 😛

DavisVaughan (approved):

PR #1045, opened 2026-02-17
Don't index or diagnose virtual documents

lionel-:
Addresses https://github.com/posit-dev/positron/issues/11887

Debugger virtual documents (`ark://` URIs) were getting distracting diagnostics squiggles on foreign code the user can't edit. We now skip non-file URIs in `diagnostics_refresh_all()`.

Also added a symmetric guard in `indexer::update()` to match what `indexer::create()` already had, and unified both to use a new `ExtUrl::is_local()` helper. We're now checking at both main loop and handler level, for defensiveness.

Unfortunately it's not possible to test diagnostics because we don't have integration tests for the LSP yet.

### DIFF
--- a/crates/ark/src/lsp/indexer.rs
+++ b/crates/ark/src/lsp/indexer.rs
@@ -260,6 +260,8 @@ pub fn filter_entry(entry: &DirEntry) -> bool {
     true
 }
 
+// Only called for actual files during workspace walking. Documents managed by
+// the LSP go through `update()` instead.
 pub(crate) fn create(uri: &Url) -> anyhow::Result<()> {
     if uri.scheme() != "file" {
         return Ok(());
--- a/crates/ark/src/lsp/main_loop.rs
+++ b/crates/ark/src/lsp/main_loop.rs
@@ -1008,7 +1008,7 @@ pub(crate) fn diagnostics_refresh_all(state: WorldState) {
     );
 
     for (uri, _document) in state.documents.iter() {
-        if !ExtUrl::is_indexable(uri) {
+        if !ExtUrl::should_diagnose(uri) {
             continue;
         }
 
--- a/crates/ark/src/url.rs
+++ b/crates/ark/src/url.rs
@@ -52,10 +52,17 @@ impl ExtUrl {
         Ok(Self::normalize(url))
     }
 
-    /// Whether this URI should be indexed and diagnosed. Currently uses an
-    /// exclude list: only `ark://` virtual documents are excluded since they
-    /// show foreign code the user can't edit.
+    /// Whether this URI should be indexed. Currently uses an exclude list:
+    /// only `ark://` virtual documents are excluded since they show foreign
+    /// code the user can't edit.
     pub fn is_indexable(uri: &Url) -> bool {
+        !Self::is_ark_virtual_doc(uri)
+    }
+
+    /// Whether this URI should get diagnostics. Currently uses the same
+    /// exclude list as [`Self::is_indexable`] but kept separate so the
+    /// criteria can diverge independently.
+    pub fn should_diagnose(uri: &Url) -> bool {
         !Self::is_ark_virtual_doc(uri)
     }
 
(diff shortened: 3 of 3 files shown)


crates/ark/src/lsp/main_loop.rs
```diff
@@ -1012,3 +1021,9 @@ pub(crate) fn diagnostics_refresh_all(state: WorldState) {
             .unwrap_or_else(|err| lsp::log_error!("Failed to queue diagnostics refresh: {err}"));
     }
 }
+
+/// Virtual documents (e.g. `ark://` debugger vdocs) should not be indexed
+/// or diagnosed since they show foreign code the user can't edit.
+fn is_indexable(uri: &Url) -> bool {
+    ExtUrl::is_local(uri)
+}
```
DavisVaughan: Can you please allow the `untitled` scheme as well? And add a test for them?

It is important that we show diagnostics in untitled R files, and currently this PR regresses on that.

We specially handle this in Air

https://github.com/posit-dev/air/blob/4cbd36d552e27d8930cff1602d56bfd9ce4c1ed1/crates/lsp/src/workspaces.rs#L136

https://github.com/posit-dev/air/blob/4cbd36d552e27d8930cff1602d56bfd9ce4c1ed1/crates/lsp/src/handlers_format.rs#L637-L642

And both Air and positron-r have a special document selector allowing `untitled` documents through

https://github.com/posit-dev/positron/blob/1f1ff9501fdd1b1019592f114b339c5bf759c535/extensions/positron-r/src/provider.ts#L27

https://github.com/posit-dev/air/blob/4cbd36d552e27d8930cff1602d56bfd9ce4c1ed1/editors/code/src/lsp.ts#L179

crates/ark/src/url.rs
```diff
@@ -123,6 +128,24 @@ mod tests {
...
+        assert!(!ExtUrl::is_local(&uri));
+    }
+
+    #[test]
+    fn test_is_local_inmemory_uri() {
+        let uri = Url::parse("inmemory://console/input.R").unwrap();
+        assert!(!ExtUrl::is_local(&uri));
+    }
+
     #[test]
```
DavisVaughan, line 179: If you want, I believe you can walk back this pull request from @seeM now

https://github.com/posit-dev/positron/pull/8176

I'm not actually sure if _any_ LSP features are useful in the `assistant-code-confirmation-widget`, but maybe I'm missing something? 

If we determine that the `assistant-code-confirmation-widget` is a useful place to have some LSP features, then we probably need to actually special case that instead.

Probably _some_ kind of action should be made here one way or the other

crates/ark/src/lsp/indexer.rs
```diff
@@ -256,10 +261,7 @@ pub fn filter_entry(entry: &DirEntry) -> bool {
 }
 
 pub(crate) fn create(uri: &Url) -> anyhow::Result<()> {
-    // Only index R files for file URIs. This discards `inmemory` (Console) and
-    // `ark` schemes in particular.
-
-    if uri.scheme() != "file" {
+    if !ExtUrl::is_local(uri) {
```
DavisVaughan:
```suggestion
    // Defensive, callers are expected to filter non-file URIs before queuing
    if !ExtUrl::is_local(uri) {
```

Mirroring comment

crates/ark/src/lsp/main_loop.rs
lionel-: oh you're right. Needs some thought about how to organise the criteria because we don't currently index `untitled` URIs, we only do it for `file` URIs.

crates/ark/src/url.rs
lionel-: It's probably fine to filter from both sides?

crates/ark/src/lsp/main_loop.rs
lionel-: I guess we could start indexing `untitled` URIs though. I can't see a reason why not?

We're currently not strict enough regarding the scope of indexed symbols (made available in all other files) but that's a separate issue.

PR #1046, opened 2026-02-18
Synchronise console evaluations with selected call stack frames

lionel-:
Branched from #1037 (watch pane support)
Backend-side of https://github.com/posit-dev/positron/pull/11941

Addresses https://github.com/posit-dev/positron/issues/3078
Addresses https://github.com/posit-dev/positron/issues/7667
Addresses https://github.com/posit-dev/positron/issues/11604

This PR adds support for evaluating expressions in a user-selected stack frame during debugging. When the user clicks a different frame in the call stack pane, subsequent console evaluations run in that frame's environment.

This required reworking how console evaluations interact with the debug session. Previously, every evaluation (typing `x` or `1 + 1` at the `Browse>` prompt) triggered a full `debug_stop` / `debug_start` cycle, which sent `Continued` + `Stopped` DAP events. Each `Stopped` event causes the frontend to reset its call stack UI, losing the user's frame selection. Console evaluations are now treated as "transient" within the debug session, sending an `Invalidated` event instead so the frontend refreshes variables without resetting its frame selection.

This new approach is much cleaner and results in better experience when evaluating in the console during debugging, for instance it fixes https://github.com/posit-dev/positron/issues/7667.

### Transient evaluations

Console evaluations are now classified as "transient" -- they don't represent debugger navigation, just expression evaluation within the current session. The `debug_preserve_focus` flag is replaced by `debug_transient_eval`, which controls a broader set of behaviours:

- `debug_stop()` returns early, preserving all Console-level state (frame IDs, session index, selected frame, last stack)
- `debug_start()` sends an `Invalidated` DAP event (variables area only) instead of the `Continued` + `Stopped` cycle
- The frontend refreshes its variables pane without resetting its frame selection or call stack UI

Debug commands (`n`, `f`, `c`, `cont`, `Q`) set `debug_transient_eval = false` and go through the normal stop/start path.

If a transient eval changes the stack (e.g. the expression hits a `browser()`), we detect this via the existing stack comparison heuristic, clean up Console-level state, send a compensating `Continued`, and start a new debug session normally.

### Frame selection

The frontend notifies the kernel of frame selection changes via a sentinel DAP `evaluate` expression (`.positron_selected_frame`). When set, `Console::eval_frame()` resolves the selected frame ID to its R environment via the DAP state, and `Rf_eval` runs in that environment instead of the current frame.

The selected frame is cleared whenever a new debug session starts (either by exiting debug normally, or when a transient eval itself pushes a new debugger onto the stack).

### QA Notes

See frontend PR.

### DIFF
--- a/crates/ark/src/console_debug.rs
+++ b/crates/ark/src/console_debug.rs
@@ -92,8 +92,6 @@ impl Console {
                 let stack_id: Vec<FrameInfoId> = stack.iter().map(|f| f.into()).collect();
                 let stack_changed = stack_id != self.debug_last_stack;
 
-                self.debug_last_stack = stack_id;
-
                 // Transient eval with unchanged stack: just refresh variables
                 if transient_eval && !stack_changed {
                     let dap = self.debug_dap.lock().unwrap();
@@ -102,12 +100,13 @@ impl Console {
                 }
 
                 // If we skipped `debug_stop` during a transient eval but the
-                // stack changed, clean up Console-level state now.
+                // stack changed, clean up and notify frontend before starting
+                // the new session.
                 if transient_eval {
-                    self.clear_fallback_sources();
-                    self.debug_session_index += 1;
-                    self.set_debug_selected_frame_id(None);
+                    self.debug_stop_session();
                 }
+
+                self.debug_last_stack = stack_id;
 
                 // Initialize fallback sources for this stack
                 let fallback_sources = self.load_fallback_sources(&stack);
@@ -121,12 +120,6 @@ impl Console {
                 }
 
                 let mut dap = self.debug_dap.lock().unwrap();
-
-                // Send Continued so the frontend knows the previous session ended.
-                if transient_eval {
-                    dap.stop_debug();
-                }
-
                 dap.start_debug(stack, fallback_sources, debug_stopped_reason)
             },
             Err(err) => log::error!("ReadConsole: Can't get stack info: {err:?}"),
@@ -144,6 +137,10 @@ impl Console {
         self.debug_is_debugging = false;
         self.debug_stopped_reason = None;
         self.debug_last_stack = vec![];
+        self.debug_stop_session();
+    }
+
+    fn debug_stop_session(&mut self) {
         self.clear_fallback_sources();
         self.debug_session_index += 1;
         self.set_debug_selected_frame_id(None);


crates/ark/src/dap/dap_server.rs
```diff
@@ -185,18 +189,26 @@ fn listen_dap_events<W: Write>(
...
                             all_threads_stopped: Some(true),
                             hit_breakpoint_ids: None,
                         })
                     },
 
+                    DapBackendEvent::Invalidated => {
+                        Event::Invalidated(InvalidatedEventBody {
+                            areas: Some(vec![types::InvalidatedAreas::Variables]),
+                            thread_id: Some(THREAD_ID),
+                            stack_frame_id: None,
```
DavisVaughan, line 221: Mildly surprised you don't need a stack frame id

crates/ark/src/console.rs
```diff
@@ -1605,12 +1615,11 @@ impl Console {
         // Default: preserve current focus for evaluated expressions.
         // This only has an effect if we're debugging.
         // https://github.com/posit-dev/positron/issues/3151
-        self.debug_preserve_focus = true;
+        self.debug_transient_eval = true;
```
DavisVaughan: Not sure if you want to update this comment

crates/ark/src/console_debug.rs
DavisVaughan: Possibly update comment about focus vs transience?

crates/ark/src/console_debug.rs
```diff
@@ -87,27 +87,53 @@ impl Console {
...
+                    return;
+                }
+
+                // If we skipped `debug_stop` during a transient eval but the
+                // stack changed, clean up Console-level state now.
+                if transient_eval {
+                    self.clear_fallback_sources();
+                    self.debug_reset_frame_id();
+                    self.debug_session_index += 1;
+                    self.set_debug_selected_frame_id(None);
```
DavisVaughan: Mehhhh I feel a bit gross about having this repeated? Feels very easy for this to get out of sync.

Maybe abstract into a cleanup/reset helper?

crates/ark/src/console_debug.rs
```diff
@@ -87,27 +87,53 @@ impl Console {
...
                 // Initialize fallback sources for this stack
                 let fallback_sources = self.load_fallback_sources(&stack);
 
-                self.debug_last_stack = stack_id;
+                let mut dap = self.debug_dap.lock().unwrap();
 
-                let preserve_focus = same_stack && debug_preserve_focus;
+                // Send Continued so the frontend knows the previous session ended.
+                if transient_eval {
+                    dap.stop_debug();
```
DavisVaughan: I think I see a way to nicely tighten this up

- only 1 `if transient_eval {` branch rather than 2
- the `if transient_eval {` branch becomes literally just `debug_stop()`, so hopefully we can remove the duplication
- `self.debug_last_stack = stack_id;` moves to _after_ the `if transient_eval {` branch, so you reset it to `vec[]!` in the `if transient_eval {` branch (ideally this happens via `debug_stop()`) then set it after that branch

What do you think? It's possible we need `debug_stop_internal()` or something, or the `if self.debug_transient_eval {` needs to be removed from `debug_stop()` to make it work. But it feels nice?

```rust
                // Figure out whether we changed location since last time,
                // e.g. because the user evaluated an expression that hit
                // another breakpoint. In that case we do want to move
                // focus, even though the user didn't explicitly used a step
                // gesture. Our indication that we changed location is
                // whether the call stack looks the same as last time. This
                // is not 100% reliable as this heuristic might have false
                // negatives, e.g. if the control flow exited the current
                // context via condition catching and jumped back in the
                // debugged function.
                let stack_id: Vec<FrameInfoId> = stack.iter().map(|f| f.into()).collect();
                let stack_changed = stack_id != self.debug_last_stack;

                // Transient eval with unchanged stack: just refresh variables
                if transient_eval && !stack_changed {
                    let dap = self.debug_dap.lock().unwrap();
                    dap.send_invalidated();
                    return;
                }

                // If we skipped `debug_stop()` during a transient eval but the
                // stack changed, invoke it now
                if transient_eval {
                    // (DAVIS): This is literally just `debug_stop()` now!
                    self.debug_last_stack = vec![];
                    self.clear_fallback_sources();
                    self.debug_reset_frame_id();
                    self.debug_session_index += 1;
                    self.set_debug_selected_frame_id(None);
                    let mut dap = self.debug_dap.lock().unwrap();
                    dap.stop_debug();
                }

                // Record new stack id
                self.debug_last_stack = stack_id;

                // Initialize fallback sources for this stack
                let fallback_sources = self.load_fallback_sources(&stack);

                let mut dap = self.debug_dap.lock().unwrap();
                dap.start_debug(stack, fallback_sources)
```

crates/ark_test/src/tracing.rs
```diff
@@ -111,6 +111,14 @@ fn format_dap_event(event: &Event) -> String {
                 id, line, verified
```
DavisVaughan, line 111: Random comment:

Do you know why stepping gets "flickery" when I enter `DataMask$new()` here?

```r
debugonce(mutate.data.frame)
mutate(mtcars, x = 1 + "x")
```

https://github.com/user-attachments/assets/38eb9104-87ad-4d4e-9c9b-6c1776555d9e

crates/ark/src/dap/dap_server.rs
lionel-: We invalidate everything in bulk. Might be over cautious, but an eval may affect other envs too

crates/ark_test/src/tracing.rs
lionel-: hmm that's jarring, could it be related to stepping through a fallback vdoc?

Related anti-flickering measure: https://github.com/posit-dev/positron/blob/fa16bbdc20f74e3b3afdf2bb61cf50ffc073f27e/extensions/positron-supervisor/src/DapComm.ts#L175-L176

PR #1048, opened 2026-02-18
Include function names in virtual documents

lionel-:
Addresses https://github.com/posit-dev/positron/issues/11889

When stepping into a function without source references, the debugger creates a virtual document by deparsing the function. This PR ensures the content is prefixed with a named assignment when appropriate:

- `foo(1)` → `foo <- function(...) { ... }`
- `base::identity(42)` → `identity <- function(...) { ... }`
- `` `my fun`(1) `` → `` `my fun` <- function(...) { ... }``
- `obj$method(1)` → `function(...) { ... }` (no prefix)

The prefix makes virtual documents self-documenting and helps users understand which function they're stepping through.

The PR also adds integration tests for virtual document (vdoc) content in the debugger.

### QA Notes

To test this, you need to trigger the vdoc fallback case, which is no longer easy since all evals now parse with source references.

One way to do it is to debug within functions from a package that was not installed with srcrefs and for which `debug()` or `debugonce()` was not called (that would trigger srcref generation):

```r
dplyr::mutate(mtcars, browser())
```

For functions with normal names, you should see the vdoc starting with `name <- function()`

Screenshot 2026-02-25 at 11 43 19

If the function has a namespace prefix, you should see the assignment without the namespace prefix:

Screenshot 2026-02-25 at 11 50 39

For functions with complex calls, you should see `function()` without the assignment, as before.

Screenshot 2026-02-25 at 11 43 29

For functions with complex names, you should see `` `name` <- function()``, but there is no examples of that in the dplyr call stack. Since this is tested on the backend side I wouldn't worry about it.

### DIFF
--- a/crates/ark/tests/dap_vdoc.rs
+++ b/crates/ark/tests/dap_vdoc.rs
@@ -32,16 +32,12 @@ fn test_dap_vdoc_fn_name_stepping() {
     let frontend = DummyArkFrontend::lock();
     let mut dap = frontend.start_dap();
 
-    frontend.debug_send_browser();
-    dap.recv_stopped();
-
     frontend.debug_enter_debugonce(
         "foo <- eval(parse(text = 'function(x) {\n  x + 1\n  x + 2\n}', keep.source = FALSE))\n\
          debugonce(foo)\n\
          foo(1)",
     );
 
-    dap.recv_continued();
     dap.recv_stopped();
 
     // Initial stop is at the `{ body }` which can't be matched, so position is 0
@@ -73,15 +69,8 @@ fn test_dap_vdoc_fn_name_stepping() {
     assert_eq!(stack[0].end_line, Some(4));
     assert_eq!(stack[0].end_column, Some(10));
 
-    // Step out: function returns, back to outer browser
-    frontend.debug_send_vdoc_step_out("n");
-    dap.recv_continued();
-    dap.recv_stopped();
-
-    let stack = dap.stack_trace();
-    assert_eq!(stack.len(), 1);
-
-    frontend.debug_send_quit();
+    // Step out: function returns
+    frontend.debug_finish("n");
     dap.recv_continued();
 }
 
--- a/crates/ark_test/src/dummy_frontend.rs
+++ b/crates/ark_test/src/dummy_frontend.rs
@@ -61,6 +61,8 @@ pub struct DummyArkFrontend {
     /// If stream is emitted during a test, there must be at least one stream
     /// assertion.
     streams_handled: Cell<bool>,
+    /// Whether we're currently in a debug context (between start_debug and stop_debug)
+    in_debug: Cell<bool>,
 }
 
 /// Result of draining accumulated streams
@@ -118,6 +120,7 @@ impl DummyArkFrontend {
             stream_stderr: RefCell::new(String::new()),
             pending_iopub_messages: RefCell::new(VecDeque::new()),
             streams_handled: Cell::new(false),
+            in_debug: Cell::new(false),
         }
     }
 
@@ -674,6 +677,7 @@ impl DummyArkFrontend {
             },
             other => panic!("Expected CommMsg with start_debug, got {:?}", other),
         }
+        self.in_debug.set(true);
     }
 
     /// Receive from IOPub and assert a `stop_debug` comm message.
@@ -692,6 +696,12 @@ impl DummyArkFrontend {
             },
             other => panic!("Expected CommMsg with stop_debug, got {:?}", other),
         }
+        self.in_debug.set(false);
+    }
+
+    /// Whether the frontend is currently in a debug context.
+    pub fn in_debug(&self) -> bool {
+        self.in_debug.get()
     }
 
     /// Send an execute request with tracing
@@ -1005,7 +1015,9 @@ impl DummyArkFrontend {
         self.send_execute_request(code, ExecuteRequestOptions::default());
         self.recv_iopub_busy();
         self.recv_iopub_execute_input();
-        self.recv_iopub_stop_debug();
+        if self.in_debug() {
+            self.recv_iopub_stop_debug();
+        }
         self.assert_stream_stdout_contains("debugging in:");
         self.recv_iopub_start_debug();
         self.recv_iopub_idle();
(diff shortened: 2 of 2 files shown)


crates/ark/src/modules/positron/calls_deparse.R
```diff
@@ -28,7 +28,7 @@ as_label <- function(x) {
 }
 
 is_simple_call <- function(x) {
-    call_print_type(x) == "call"
+    call_print_type(x) == "prefix"
```
lionel-, line 31: Preexisting bug for a function that was still unused.

crates/ark/tests/dap_vdoc.rs
```diff
@@ -0,0 +1,801 @@
...
+
+// Tests for virtual document function name prefix in the debugger.
+// When stepping into a function without source references, the virtual document
+// content should be prefixed with `name <- ` when the frame call has a simple
+// symbol in function position (e.g., `foo <- function(x) { ... }`).
+
+use amalthea::fixtures::dummy_frontend::ExecuteRequestOptions;
+use ark_test::DummyArkFrontend;
+
+/// Basic test: step into a function without srcrefs, verify the frame name
```
lionel-, line 16: The agent created a whole suite of tests for vdoc, unprompted.

I haven't checked too closely. The tests that interested me are below, for vdoc contents.

crates/ark/tests/dap_vdoc.rs
```diff
@@ -0,0 +1,801 @@
...
+    dap.recv_continued();
+    dap.recv_stopped();
+
+    frontend.debug_send_quit();
+    dap.recv_continued();
+}
+
+/// Test that virtual document content contains the prefix assignment.
+#[test]
+fn test_dap_vdoc_content_has_prefix() {
```
lionel-, line 541: The integration tests of interest start here, i.e. all those that call `frontend.get_virtual_document()`.

crates/ark/src/modules/positron/calls_deparse.R
```diff
@@ -305,3 +305,65 @@ call_parse_type <- function(call) {
 
     ""
 }
+
+backtick <- function(x) {
```
DavisVaughan:
```suggestion
backtick_if_needed <- function(x) {
```

It really does sound like an unconditional backtick otherwise

crates/ark/src/modules/positron/calls_deparse.R
```diff
@@ -305,3 +305,65 @@ call_parse_type <- function(call) {
...
+    if (needs_backticks(x)) {
+        paste0("`", x, "`")
+    } else {
+        x
+    }
+}
+
+needs_backticks <- function(str) {
```
DavisVaughan: A much simpler approach would be to copy tibble, which we've used for many years

```
> tibble:::is_syntactic_impl
function (x) 
{
    unchanged_after_repair <- (x == make.names(x))
    dot_dot_dot_or_numbers <- grepl("^(?:(?:[.][.][.])|(?:[.][.][0-9]+))$", 
        x)
    unchanged_after_repair & !dot_dot_dot_or_numbers
}
```

```
> tibble:::tick_if_needed
function (x) 
{
    needs_ticks <- !is_syntactic(x)
    x[needs_ticks] <- tick(x[needs_ticks])
    x
}
```

DavisVaughan: Or even better, expose and use `sym_quote_invalid()` from `harp/src/utils.rs`!

Would be very nice to have 1 consistent helper for this

crates/ark/src/modules/positron/debug.R
```diff
@@ -196,7 +201,8 @@ frame_info <- function(
     fn,
     environment,
     call_text,
-    last_start_line
+    last_start_line,
+    fn_name = NULL
```
DavisVaughan: I intentionally made all arguments required, as I wanted to be intentional about passing them all through at every call site. Can we remove `NULL`?

crates/ark/src/modules/positron/debug.R
```diff
@@ -149,7 +150,8 @@ context_frame_info <- function(
         fn,
         environment,
         call_text,
-        last_start_line
+        last_start_line,
+        fn_name = fn_name
```
```suggestion
        fn_name
```

I think we should use an all or nothing approach on naming call arguments, and right now its nothing, so I'd prefer to keep it that way, since this just looks weird

crates/ark/src/modules/positron/debug.R
```diff
@@ -215,6 +221,13 @@ frame_info <- function(
     # Only deparse if `srcref` failed!
     fn_text <- deparse_string(fn)
 
+    # Prefix with a named assignment if we have a known function name.
+    # This makes the virtual document self-documenting.
+    # https://github.com/posit-dev/positron/issues/11889
+    if (!is.null(fn_name)) {
+        fn_text <- paste0(fn_name, " <- ", fn_text)
+    }
```
DavisVaughan: Do you want to mention that this should not affect debug stepping because we aren't adding any line breaks? Am I correct about that?

crates/ark_test/src/dummy_frontend.rs
```diff
@@ -973,6 +973,96 @@ impl DummyArkFrontend {
...
+    /// Returns `None` if the document is not found (or has empty content).
+    ///
+    /// Note: This method handles debug mode by consuming stop_debug/start_debug messages.
+    #[track_caller]
+    pub fn get_virtual_document(&self, uri: &str) -> Option<String> {
+        // Use cat() which handles NULL gracefully (outputs nothing)
+        let code = format!(
+            "cat(.ps.internal(.ps.Call(\"ps_get_virtual_document\", \"{}\")), sep = \"\\n\")",
+            uri
+        );
```
DavisVaughan, line 1076: I can't tell if I like this or hate this 😆

crates/ark/tests/dap_vdoc.rs
```diff
@@ -0,0 +1,801 @@
...
+///
+/// After stepping to `x + 1`, the position should be line 3.
+/// After stepping to `x + 2`, the position should be line 4.
+#[test]
+fn test_dap_vdoc_fn_name_stepping() {
+    let frontend = DummyArkFrontend::lock();
+    let mut dap = frontend.start_dap();
+
+    frontend.debug_send_browser();
+    dap.recv_stopped();
```
DavisVaughan: why does it send a browser up front?

DavisVaughan (approved):
Just a question about `needs_backticks` and some preference requests in `debug.R`

crates/ark/src/modules/positron/calls_deparse.R
lionel-: If we're going to use a Rust impl, let's reimplement the rlang implementation (not tibble which calls into base R) in Rust then. Our current Rust-side regex is less correct than the rlang one.

crates/ark/src/modules/positron/calls_deparse.R
lionel-: Now `sym_quote_invalid()` to mirror Rust side

crates/ark/src/modules/positron/debug.R
lionel-: yep

crates/ark/tests/dap_vdoc.rs
lionel-: Likely because the test helper was assuming being called from debug context. I've now made them agnostic to context.

PR #1049, opened 2026-02-19
Capture debugger evaluation environment reliably inside C-level evals

lionel-:
Addresses https://github.com/posit-dev/positron/issues/11780

When `browser()` is called inside `tryCatch()` (or `withCallingHandlers()`), expressions typed in the debug console were evaluated in the wrong environment.

The root cause is how we captured the evaluation environment. Previously, `r_read_console()` called `harp::r_current_frame()` to retrieve the environment from the R-level call stack. This works for normal browser sessions but fails inside `tryCatch()` because R evaluates the `tryCatch` expression body via a C-level `Rf_eval()`, and there is no corresponding R-level frame for that evaluation.

#### Two-pass environment capture

There is no way to reliably get the browser's evaluation environment from ReadConsole:

- Evaluating R code requires supplying an environment, which interferes with how `parent.frame()` resolves the environment.

- `sys.frames()` and friends require the environment to appear on the call stack.

The trick is to get R itself to evaluate `parent.frame()`. This requires two passes through `r_read_console()`:

1. On the first browser prompt, instead of doing entry bookkeeping, we return `base::.ark_capture_top_level_environment()` to R. This function calls `parent.frame()` and stores the result in a base symbol.

2. On the next `r_read_console()` call, we read the captured environment back from the symbol, unbind it to make it available for GC, and proceed with normal entry bookkeeping using the correct environment.

The state machine is encoded in `ReadConsolePendingAction`, which replaces the old `Option<String>` field and makes the multi-step logic explicit:

```rust
enum ReadConsolePendingAction {
    None,
    CaptureEnv,
    ExecuteInput(String),
}
```

If capture somehow fails, we fall back to `r_current_frame()` as before.

#### Synthetic stack frame

When the captured top-level environment differs from the R-level top frame (indicating a C-level eval context), `debugger_stack_info` appends a synthetic `<top>` frame so the variables pane shows the correct bindings.

#### Refactoring

- Part of the ReadConsole state machine is now encoded in `ReadConsolePendingAction`, as mentioned above.

- Setup and cleanup are now performed by `read_console_enter()` and `read_console_exit()`.

- **Environment stack**: `read_console_frame: RefCell<RObject>` is replaced by `read_console_env_stack: RefCell<Vec<RObject>>`. The old code saved and restored a single frame via `replace()` in the entry/cleanup pair. The stack mirrors the actual nesting of `r_read_console()` calls and is managed by `read_console_enter()` / `read_console_exit()`.

- **`base_bind()` alias**: All `node_poke_cdr(as.symbol(...), value)` calls are replaced with `base_bind(sym, value)` to make the intent clearer. The underlying mechanism is the same (poking SYMVALUE/CDR), but the name now communicates that we're binding a value in the base namespace.

- `console::eval_env()` is a new free function for external callers, that falls back to global env if `Console` isn't initialized. `Console::eval_env()`is a stateful method for use inside other `Console` methods.

  TODO: Reconciliate with https://github.com/posit-dev/ark/pull/1046 once merged.

- Made search path loop more obviously safe. Added `RObject::replace()` in the process.

### QA Notes

When stepping through blocks in contexts like `identity()` or `tryCatch()`, Ark should use the correct environment in the Variables pane, in local completions, and in console evaluations.

```r
local({
  foo <- 1
  identity({
    browser()
    2
  })
})
```

### DIFF
--- a/crates/ark/tests/dap.rs
+++ b/crates/ark/tests/dap.rs
@@ -758,3 +758,45 @@ f(99)
     frontend.debug_send_quit();
     dap.recv_continued();
 }
+
+/// Same as above but with `identity()`, a non-condition related example.
+#[test]
+fn test_dap_browser_in_identity() {
+    let frontend = DummyArkFrontend::lock();
+    let mut dap = frontend.start_dap();
+
+    let _file = frontend.send_source(
+        "
+f <- function(my_var) {
+  identity({
+    browser()
+    my_var
+  })
+}
+f(42)
+",
+    );
+    dap.recv_stopped();
+
+    let stack = dap.stack_trace();
+    let frame_id = stack[0].id;
+    let scopes = dap.scopes(frame_id);
+    let variables = dap.variables(scopes[0].variables_reference);
+
+    let var = variables.iter().find(|v| v.name == "my_var").unwrap();
+    assert_eq!(var.value, "42");
+
+    // Evaluate `my_var` from the console: must resolve to the argument.
+    // Transient evals send Invalidated instead of Continued+Stopped.
+    frontend.send_execute_request("my_var", ExecuteRequestOptions::default());
+    frontend.recv_iopub_busy();
+    frontend.recv_iopub_execute_input();
+    frontend.assert_stream_stdout_contains("[1] 42");
+    frontend.recv_iopub_idle();
+    frontend.recv_shell_execute_reply();
+
+    dap.recv_invalidated();
+
+    frontend.debug_send_quit();
+    dap.recv_continued();
+}


crates/ark/src/modules/positron/debug.R
```diff
@@ -125,8 +130,20 @@ debugger_stack_info <- function(
...
+    # When `top_env` differs from `context_environment`, it means that (a) the
+    # current expression is evaluated from C, e.g. we might be forcing a promise
+    # via `list(...)`, and (b) we're adding a synthetic top environment based on
+    # the context srcref. Unfortunately, we don't have any srcref information
+    # for the R-level top in that case, as neither `R_Srcref` nor the last call
+    # in `sys.calls()` point to the location of the call. So we let it fall back
+    # to the virtual source from the function.
+    has_synthetic_top <- !is.null(top_env) &&
+        !identical(top_env, context_environment)
+    last_context_srcref <- if (has_synthetic_top) NULL else context_srcref
```
```suggestion
    if (has_synthetic_top) {
      top_srcref <- context_srcref
      context_srcref <- NULL
    } else {
      top_srcref <- NULL
    }
```

Would you be open to doing this so that all arguments to `context_frame_info()` are still named `context_*`? It seems a bit nicer that way, i.e. this is us just "adjusting" the `context_srcref` if we need to.

crates/ark/src/modules/positron/debug.R
```diff
@@ -140,6 +157,21 @@ debugger_stack_info <- function(
...
 
+    # If the captured top-level environment differs from the R-level top frame,
+    # append a synthetic frame. This happens when evaluating from C (e.g.
+    # promise forcing).
+    if (has_synthetic_top) {
+        top_frame_info <- frame_info_from_srcref(
+            source_name = "<top>.R",
+            frame_name = "<top>",
+            srcref = context_srcref,
+            environment = top_env
```
DavisVaughan: then this becomes

```suggestion
            srcref = top_srcref,
            environment = top_env
```

which also feels a bit better

crates/ark/src/console.rs
```diff
@@ -2735,25 +2854,9 @@ pub extern "C-unwind" fn r_read_console(
...
-
-    // Keep track of state that we use for workarounds while interacting
-    // with the R REPL and force it to reset state
-
-    // - Reset flag that helps us figure out when a nested REPL returns
-    console.read_console_nested_return.set(false);
-
-    // - Reset flag that helps us figure out when an error occurred and needs a
-    //   reset of `R_EvalDepth` and friends
-    console.read_console_threw_error.set(true);
```
DavisVaughan, line 2860: I think we lost a lot of important documentation here

crates/ark/src/console.rs
```diff
@@ -2766,31 +2869,7 @@ pub extern "C-unwind" fn r_read_console(
...
-
-            // We're exiting, decrease depth of nested consoles
-            console
-                .read_console_depth
-                .set(console.read_console_depth.get() - 1);
-
-            // Set flag so that parent read console, if any, can detect that a
-            // nested console returned (if it indeed returns instead of looping
-            // for another iteration)
-            console.read_console_nested_return.set(true);
```
DavisVaughan, line 2883: and lost docs here

crates/ark/tests/dap.rs
```diff
@@ -669,3 +669,98 @@ outer_user()
     assert_eq!(stack[1].name, "shiny_wrapper()");
     assert_eq!(stack[2].name, "outer_user()");
 }
+
+/// https://github.com/posit-dev/positron/issues/11780
+/// `browser()` inside `tryCatch()` must evaluate in the function's environment,
+/// not a parent one. Here `by` shadows `base::by` and must resolve to `1`.
```
DavisVaughan: wrong comment. not about `by`

crates/ark/tests/dap.rs
```diff
@@ -669,3 +669,98 @@ outer_user()
...
+    frontend.recv_iopub_busy();
+    frontend.recv_iopub_execute_input();
+    frontend.recv_iopub_stop_debug();
+    frontend.recv_iopub_start_debug();
+    frontend.assert_stream_stdout_contains("[1] 99");
+    frontend.recv_iopub_idle();
+    frontend.recv_shell_execute_reply();
+    dap.recv_continued();
+    dap.recv_stopped();
+
```
DavisVaughan, line 757: could add the `identity` example too if you wanted? as a non-condition related example

crates/harp/src/object.rs
```diff
@@ -321,6 +327,11 @@ impl RObject {
         }
     }
 
+    pub fn replace(&mut self, data: SEXP) -> SEXP {
+        self.sexp = data;
+        unsafe { protect_replace(self.cell, data) }
+    }
```
DavisVaughan: I am deeply uncomfortable with the idea of being able to swap out the SEXP inside an `RObject`.

I don't really want to think through all the implications of this, or the bad ways we might accidentally use it, and it looks like you only use it in one place in search_path.rs where it seems like you can easily work around this by just updating a SEXP on every iteration instead, which is what we were doing before, and the SEXP is protected by the environment provided to ENCLOS so it should be safe.

Im imagining you'd do

```rs
let env_obj = console::eval_env(); // this protects the initial env SEXP
let mut env = env.sexp;

// then in the loop
env = ENCLOS(env)
```

Would it be ok if we remove this?

lionel-: That's basically REPROTECT, but yes I can work around

PR #1052, opened 2026-02-21
Synchronise Positron's Variables pane with debug environment

lionel-:
Branched from #1050

Addresses https://github.com/posit-dev/positron/issues/12131

The Variables pane was hardcoded to show the global environment. When entering debug mode (`browser()`), it kept showing global bindings instead of switching to the debug frame's local environment.

This PR wires up `console::eval_env()` so the Variables pane tracks the current evaluation environment, switching to the debug environment on entry and back to the global environment on exit.

The Variables comm is now exposed to our test infrastucture, so we can test Jupyter + DAP + comm patterns.

#### Restructured `update()`

The refresh (full rescan) and incremental (diff) paths in `update()` were interleaved: the refresh branch set a flag, then fell through into the diff logic which was a no-op. The two paths are now clearly separated in a single `r_task`:

- **Refresh path**: environment changed or initial startup. Collects all bindings, sends `Refresh` event, returns early.
- **Incremental path**: diffs old vs new bindings, sends `Update` event if anything changed.

#### Deferred initial scan

The initial environment scan was previously synchronous via `r_task()`, which is unsafe at startup. It's now deferred to an idle task that sends a prompt signal to the variables thread. The first `update()` sees `needs_initial_refresh = true` and sends the `Refresh` event that the frontend requires for initialization.

Two scenarios:
- R already idle: idle task fires promptly, frontend gets its initial `Refresh`.
- R busy: idle task queues up, fires at next idle. The subsequent prompt-driven `update()` may also arrive, which is a harmless no-op since bindings haven't changed.

#### `list_variables()` is now a pure read

The `List` RPC handler no longer calls `update_env()` or `update_bindings()`. Since RPCs only arrive when R is idle (after a prompt-driven `update()` has already synced state), `current_bindings` is always up to date by the time a `List` request comes in.

### QA Notes

This is tested on the backend side.

Positron's Variables pane should now track the selected call stack frame, just like the debugger's Variables pane.

### DIFF
--- a/crates/ark/tests/variables_debug.rs
+++ b/crates/ark/tests/variables_debug.rs
@@ -70,6 +70,9 @@ fn test_variables_pane_shows_debug_env() {
     frontend.recv_iopub_execute_input();
     frontend.recv_iopub_idle();
     frontend.recv_shell_execute_reply();
+
+    // Consume the Update triggered by the rm()
+    let _update = frontend.recv_variables_update();
 }
 
 /// When entering debug via a function call, the variables pane should show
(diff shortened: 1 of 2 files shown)


crates/ark/src/variables/r_variables.rs
```diff
@@ -249,13 +256,7 @@ impl RVariables {
     fn list_variables(&mut self) -> Vec<Variable> {
         let mut variables: Vec<Variable> = vec![];
         r_task(|| {
```
DavisVaughan, line 258: So the `List` request used to request a bindings update, but we don't want to do that anymore? We are trusting that it should be update to date because it would have updated after every top level eval?

crates/ark/tests/variables.rs
```diff
@@ -112,6 +112,22 @@ fn test_variables_list() {
         Rf_defineVar(sym, Rf_ScalarInteger(42), *test_env);
```
DavisVaughan, line 112: I was confused for a minute about why `.` prefixed names werent showing up

I wonder if the Variables Pane should instead list all `.` prefixed names _except_ for `.Random.seed`, which it should always filter out, rather than not listing any `.` prefixed names.

Image

crates/ark/src/variables/r_variables.rs
lionel-: yep

PR #1060, opened 2026-02-26
Notify Variables pane when frame is selected

lionel-:
Follow-up to #1046 and #1052
Addresses https://github.com/posit-dev/positron/issues/12140

When the user selects a different stack frame in the call stack pane during debugging, Positron's Variables pane does not automatically update to show that frame's environment. It only refreshes after a console evaluation.

`set_debug_selected_frame_id()` now emits a `console_prompt` signal when a frame is selected (`Some`), which causes the Variables pane (and other listeners) to re-inspect context and update. The signal is not emitted when the frame is cleared (`None`) during `debug_stop_session()`, since R is no longer idle at that point.

The other listener for `console_prompt` is the Data Explorer. Its `update()` checks whether the underlying data object has changed by comparing the SEXP pointer of a specific variable in the environment where `View()` was called. This is independent of the console's evaluation environment, so the extra signal is a harmless no-op for open data viewers.

### QA Notes

Tested on the backend side.

```r
f <- function(foo = "foo") {
  browser()
}
g <- function(bar = "bar") {
  f()
}

g()
```

1. Enter debug mode
2. The Variables pane should show the current frame's variables.
3. Click a different frame in the call stack pane.
4. The Variables pane should immediately refresh to show that frame's variables, without needing to evaluate anything in the console.

https://github.com/user-attachments/assets/b5202d1f-bc6d-45e7-8984-a4406bc628f2

### DIFF
--- a/crates/ark/src/console.rs
+++ b/crates/ark/src/console.rs
@@ -1187,8 +1187,7 @@ impl Console {
             self.refresh_lsp();
         }
 
-        // Signal prompt
-        EVENTS.console_prompt.emit(());
+        EVENTS.environment_changed.emit(());
 
         self.run_event_loop(&info, buf, buflen, WaitFor::ExecuteRequest)
     }
@@ -2778,7 +2777,7 @@ impl Console {
 
         // Signal listeners (e.g. the Variables pane) that they can update state
         if frame_id.is_some() {
-            EVENTS.console_prompt.emit(());
+            EVENTS.environment_changed.emit(());
         }
     }
 
--- a/crates/ark/src/data_explorer/r_data_explorer.rs
+++ b/crates/ark/src/data_explorer/r_data_explorer.rs
@@ -249,11 +249,10 @@ impl RDataExplorer {
     }
 
     pub fn execution_thread(mut self) {
-        // Register a handler for console prompt events. We do this before sending
-        // `CommManagerEvent::Opened` to ensure we can't miss any console prompt events
-        // (#781).
+        // Register a handler for environment change events. We do this before sending
+        // `CommManagerEvent::Opened` to ensure we can't miss any events (#781).
         let (prompt_signal_tx, prompt_signal_rx) = unbounded::<()>();
-        let listen_id = EVENTS.console_prompt.listen({
+        let listen_id = EVENTS.environment_changed.listen({
             move |_| {
                 prompt_signal_tx.send(()).unwrap();
             }
@@ -281,8 +280,8 @@ impl RDataExplorer {
         // Set up event loop to listen for incoming messages from the frontend
         loop {
             select! {
-                // When a console prompt event is received, check for updates to
-                // the underlying data
+                // When the environment changes, check for updates to the
+                // underlying data
                 recv(&prompt_signal_rx) -> msg => {
                     if let Ok(()) = msg {
                         match self.update() {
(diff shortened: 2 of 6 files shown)


crates/ark/src/console.rs
```diff
@@ -2775,6 +2775,11 @@ impl Console {
 
     pub(crate) fn set_debug_selected_frame_id(&self, frame_id: Option<i64>) {
         self.debug_selected_frame_id.set(frame_id);
+
+        // Signal listeners (e.g. the Variables pane) that they can update state
+        if frame_id.is_some() {
+            EVENTS.console_prompt.emit(());
+        }
```
lionel-, line 2781: That's the fix, the rest is all tests.

crates/ark/src/console.rs
```diff
@@ -2775,6 +2775,11 @@ impl Console {
 
     pub(crate) fn set_debug_selected_frame_id(&self, frame_id: Option<i64>) {
         self.debug_selected_frame_id.set(frame_id);
+
+        // Signal listeners (e.g. the Variables pane) that they can update state
+        if frame_id.is_some() {
+            EVENTS.console_prompt.emit(());
```
DavisVaughan: I disagree with using the `console_prompt` event in this way. This is not a new `console_prompt`. If we have future listeners that truly only want to update on a new `console_prompt`, then this is incorrect.

Can we add a new `EVENTS.debug_frame_changed` that we would emit from here instead?

Then in `r_variables.rs`, this:

```rust
let (prompt_signal_tx, prompt_signal_rx) = unbounded::<()>();
```

could become

```rust
let (update_signal_tx, update_signal_rx) = unbounded::<()>();
```

and we'd have 3 signalers

```rust
        // Schedule the initial environment scan as an idle task. If R is
        // already idle this runs promptly, otherwise it runs at next idle.
        // This and the prompt-driven `update()` may arrive in either order,
        // which is safe: the first sends a `Refresh`, the second is a no-op.
        r_task::spawn_idle({
            let update_signal_tx = update_signal_tx.clone();
            async move |_| {
                update_signal_tx.send(()).log_err();
            }
        });

        // Register a handler for console prompt events
        let console_prompt_id = EVENTS.console_prompt.listen({
            let update_signal_tx = update_signal_tx.clone();
            move |_| {
                log::info!("Got console prompt signal.");
                update_signal_tx.send(()).log_err();
            }
        });

        // Register a handler for debug frame change events
        let debug_frame_change_id = EVENTS.debug_frame_change.listen({
            let update_signal_tx = update_signal_tx.clone();
            move |_| {
                log::info!("Got debug frame change signal.");
                update_signal_tx.send(()).log_err();
            }
        });
```

lionel-: I was thinking of renaming `console_prompt` to something more agnostic. For the purpose of external listeners, whether R is waiting for a new prompt doesn't seem relevant. The signal here is really "R is idle and the environment might have changed".

lionel-: If we're going to refactor, I'd like to go beyond the cross-thread event/listener paradigm and do something more robust for environment inspection while R is idle (out of scope for this PR).

DavisVaughan: How about `EVENTS.active_environment_changed`?

lionel-: I've made it `EVENTS.environment_changed`

PR #1062, opened 2026-02-26
Simplify `.Last.value` handling

DavisVaughan:
While reviewing some of @lionel-'s tweaks to `variables.rs`, it occurred to me that the `.Last.value` handling is quite complicated and could be heavily simplified

I think the existing tests were also actually wrong. If the `.Last.value` is being shown and was set to `NULL` and after executing another R statement it is _still_ `NULL`, then we should not get an update for that. The pointer value didn't change, so nothing should update.

The only way to actually test ^ is to really send R code through `r_read_console()` so I've switched to using frontend powered tests for `.Last.value` testing.

### DIFF
--- a/crates/amalthea/src/fixtures/dummy_frontend.rs
+++ b/crates/amalthea/src/fixtures/dummy_frontend.rs
@@ -606,31 +606,6 @@ impl DummyFrontend {
             (data.content.comm_id, data.content.target_name, data.content.data)
         })
     }
-
-    pub fn is_installed(&self, package: &str) -> bool {
-        let code = format!(".ps.is_installed('{package}')");
-        self.send_execute_request(&code, ExecuteRequestOptions::default());
-        self.recv_iopub_busy();
-
-        let input = self.recv_iopub_execute_input();
-        assert_eq!(input.code, code);
-
-        let result = self.recv_iopub_execute_result();
-
-        let out = if result == "[1] TRUE" {
-            true
-        } else if result == "[1] FALSE" {
-            false
-        } else {
-            panic!("Expected `TRUE` or `FALSE`, got '{result}'.");
-        };
-
-        self.recv_iopub_idle();
-
-        assert_eq!(self.recv_shell_execute_reply(), input.execution_count);
-
-        out
-    }
 }
 
 impl Default for ExecuteRequestOptions {
--- a/crates/ark_test/src/dummy_frontend.rs
+++ b/crates/ark_test/src/dummy_frontend.rs
@@ -1690,6 +1690,31 @@ impl DummyArkFrontend {
             Some(content.to_string())
         }
     }
+
+    pub fn is_installed(&self, package: &str) -> bool {
+        let code = format!(".ps.is_installed('{package}')");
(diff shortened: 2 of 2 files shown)


crates/ark/src/variables/r_variables.rs
```diff
@@ -53,17 +55,6 @@ use crate::variables::variable::try_dispatch_view;
 use crate::variables::variable::PositronVariable;
 use crate::view::view;
 
-/// Enumeration of treatments for the .Last.value variable
-pub enum LastValue {
-    /// Always show the .Last.value variable in the Variables pane. This is used
-    /// by tests to show the value without changing the global option.
-    Always,
```
DavisVaughan, line 60: Only used by tests before, but we don't use this now. We literally set the option like an R user would.

crates/ark/src/variables/r_variables.rs
```diff
@@ -256,10 +219,6 @@ impl RVariables {
     fn list_variables(&mut self) -> Vec<Variable> {
         let mut variables: Vec<Variable> = vec![];
         r_task(|| {
-            if let Some(last_value) = self.last_value() {
-                variables.push(last_value.var());
-            }
-
             for binding in self.current_bindings.get() {
```
DavisVaughan, line 222: I thought it was very strange that we had all this special handling around `.Last.value`.

I realized it could be greatly simplified by shoving `.Last.value` handling into `bindings()`.

Either:
- The option is on, and `.Last.value` is returned from `bindings()`
- The option is off, and `.Last.value` is not returned from `bindings()`

Everything else (including `assigned` and `removed` handling below) just correctly falls out from this.

crates/ark/src/variables/r_variables.rs
```diff
@@ -721,6 +632,12 @@ impl RVariables {
 
         let mut bindings: Vec<Binding> = env.iter().filter_map(|b| b.ok()).collect();
 
+        if self.show_last_value() {
+            if let Some(last_value) = Self::last_value().log_err() {
+                bindings.push(last_value);
+            }
+        }
```
DavisVaughan, line 639: This is the one special place we handle it now

crates/ark/tests/kernel-variables.rs
```diff
@@ -0,0 +1,68 @@
...
+        .map(|v| v.display_name.as_str())
+        .collect();
+    assert!(names.contains(&"test_var"));
+    assert!(!names.contains(&".Last.value"));
+
+    // Request that Ark start showing `.Last.value`
+    frontend.execute_request_invisibly("options(positron.show_last_value = TRUE)");
+
+    // The variables pane should have sent an Update, and now `.Last.value` should be in there
+    let update = frontend.recv_variables_update();
```
DavisVaughan, line 24: @lionel- I think the CI errors are due to the Variables `Update` message racing with the IOPub `Status` message that we are expecting in `execute_request_invisibly`. I think the `Update` comes through too quickly?

It looks like you have some infra for this in `recv_iopub_next()` that maybe `execute_request_invisibly()` should be using, but its all a bit confusing to me. Can you help me take a look?

PR #1064, opened 2026-02-27
Implement conditional breakpoints

lionel-:

Adds support for conditional breakpoints in the DAP server. When a breakpoint has a condition expression, it is evaluated at hit time in the breakpoint's local environment. The breakpoint only stops execution if the condition is truthy.

The condition is stored as a string in the Rust `Breakpoint` struct and evaluated on the R thread via a new `ps_should_break` registered function that combines the enabled check with condition evaluation. The DAP mutex is dropped before calling back into R to avoid deadlock.

Condition results are coerced via `as.logical()` so R's standard truthy/falsy rules apply: `0` skips, `1` stops, `nrow(df)` works naturally. If evaluation errors or the result can't be coerced to scalar logical (e.g. an environment), the breakpoint fires to avoid silently swallowing typos in conditions. From a quick search this seems standard among DAPs (e.g. js/ts, python). If we feel strongly that we should be stricter, we can revert the last commit.

Conditions that fail with an error always trigger the breakpoint to avoid silently skipping them. The error is logged in the console.The message is fenced with a breakpoint header that includes a link to the breakpoint location:

Screenshot 2026-03-04 at 15 45 49

Screenshot 2026-03-04 at 18 33 47

 If there is any output, warning, or messages, we log in the console too. 

Screenshot 2026-03-04 at 15 39 57

### QA Notes

Comprehensively tested on the backend side (10 integration tests covering true/false conditions, loop iteration, local variables, error conditions, numeric coercion, mixed breakpoints, and live condition updates).

To test manually:

```r
foo <- function(x) {
  y <- x * 2
  y + 1
}
for (i in 1:5) {
  foo(i)
}
```

1. Set a conditional breakpoint on `y <- x * 2` with condition `x == 3`. Run the loop, should stop only when `x` is 3.
2. Edit the condition to `x > 4` without re-sourcing. Run the loop again, should now stop at `x == 5`.
3. Try a numeric condition like `x - 2` (should skip when `x == 2` since `0` is falsy).
4. Try a broken condition like `nonexistent` — should stop on every hit rather than silently skipping. You should see the error in the Console with a link to jump to the breakpoint.
5. Output, warnings, and messages emitted by the breakpoint conditions are logged in the Console.

### DIFF
--- a/crates/ark/src/console_debug.rs
+++ b/crates/ark/src/console_debug.rs
@@ -560,13 +560,14 @@ pub unsafe extern "C-unwind" fn ps_should_break(
             let ((should_break, error), captured_output) =
                 Console::with_capture(|| eval_condition(condition, env));
 
-            let output = match error {
-                Some(err) if captured_output.is_empty() => err,
-                Some(err) => format!("{captured_output}{err}\n"),
-                None => captured_output,
-            };
-
-            emit_condition_output(&uri, bp_line, condition, &output);
+            if !captured_output.trim().is_empty() || error.is_some() {
+                let mut output = format!("Code: `{condition}`\n");
+                output.push_str(&captured_output);
+                if let Some(err) = error {
+                    output.push_str(&err);
+                }
+                emit_breakpoint_block(&uri, bp_line, &output);
+            }
             should_break
         },
     };
@@ -574,9 +575,9 @@ pub unsafe extern "C-unwind" fn ps_should_break(
     Ok(RObject::from(should_break).sexp)
 }
 
-/// Emit any output from condition evaluation to stderr.
-fn emit_condition_output(uri: &UrlId, line: u32, condition: &str, captured: &str) {
-    let Some(text) = format_condition_output(uri, line, condition, captured) else {
+/// Emit a fenced breakpoint block to stderr.
+fn emit_breakpoint_block(uri: &UrlId, line: u32, content: &str) {
+    let Some(text) = format_breakpoint_block(uri, line, content) else {
         return;
     };
 
@@ -589,13 +590,8 @@ fn emit_condition_output(uri: &UrlId, line: u32, condition: &str, captured: &str
         .unwrap();
 }
 
-fn format_condition_output(
-    uri: &UrlId,
-    line: u32,
-    condition: &str,
-    captured: &str,
-) -> Option<String> {
-    if captured.trim().is_empty() {
+fn format_breakpoint_block(uri: &UrlId, line: u32, content: &str) -> Option<String> {
+    if content.trim().is_empty() {
         return None;
     }
 
@@ -607,10 +603,10 @@ fn format_condition_output(
     let display_line = line + 1;
     let label = ansi_file_link(uri, line, &format!("{filename}#{display_line}"));
 
-    let mut text = format!("```breakpoint {label}\n#> {condition}\n");
+    let mut text = format!("```breakpoint {label}\n");
 
-    text.push_str(captured);
-    if !captured.ends_with('\n') {
+    text.push_str(content);
+    if !content.ends_with('\n') {
         text.push('\n');
     }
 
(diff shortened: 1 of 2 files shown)


crates/ark/src/console.rs
```diff
@@ -1009,6 +1009,16 @@ impl Console {
...
 
+    /// Run a closure while capturing console output.
+    /// Returns the closure's result paired with any captured output.
+    pub(crate) fn with_capture<T>(f: impl FnOnce() -> T) -> (T, String) {
+        let mut capture = Console::get_mut().start_capture();
+        let result = f();
+        let output = capture.take();
+        drop(capture);
+        (result, output)
+    }
```
DavisVaughan, line 1034:
```suggestion
    pub(crate) fn with_capture<T>(f: impl FnOnce() -> T) -> (T, String) {
        let mut capture = Console::get_mut().start_capture();
        let result = f();
        let output = capture.take();
        (result, output)
    }
```

That seems a bit superfluous?

crates/ark/src/console_debug.rs
```diff
@@ -515,21 +520,292 @@ fn as_frame_info(info: libr::SEXP, id: i64) -> Result<FrameInfo> {
...
+
+/// Format an OSC 8 hyperlink pointing to a file location.
+/// Terminals that support OSC 8 render `display` as a clickable link.
+fn ansi_file_link(uri: &Url, line: u32, display: &str) -> String {
+    let display_line = line + 1;
+    format!("\x1b]8;line={display_line};{uri}\x07{display}\x1b]8;;\x07")
+}
+
+#[cfg(test)]
+mod tests_condition_output {
```
DavisVaughan: can we move tests to the very bottom? i totally skipped over `eval_condition()` because i thought this was the end of the file

crates/ark/src/console_debug.rs
```diff
@@ -515,21 +520,292 @@ fn as_frame_info(info: libr::SEXP, id: i64) -> Result<FrameInfo> {
...
+///
+/// Returns `(should_break, warnings, diagnostic)`. Warnings are captured via
+/// `withCallingHandlers` in R since R defers warnings until the prompt by default.
+/// When evaluation fails or produces a non-logical result, `should_break` is `true`
+/// so that typos in conditions cause a visible stop rather than a silently ignored
+/// breakpoint.
+fn eval_condition(condition: &str, envir: RObject) -> (bool, String, Option<String>) {
+    // `if` coerces via `asLogicalNoNA` (not the generic `as.logical`)
+    // and errors on NA, length != 1, and non-coercible types.
+    let code = format!("base::.ark_eval_capture(if ({{ {condition} }}) TRUE else FALSE)");
```
DavisVaughan: feels like the right semantics to me

crates/ark/src/console_debug.rs
```diff
@@ -515,21 +520,292 @@ fn as_frame_info(info: libr::SEXP, id: i64) -> Result<FrameInfo> {
...
+    let list_result = harp::list_get(result.sexp, 0);
+    let list_conditions = harp::list_get(result.sexp, 1);
+
+    let conditions = format_captured_conditions(list_conditions);
+
+    // `if` guarantees the result is TRUE or FALSE (never NA, never non-scalar),
+    // so a failed conversion here is unexpected.
+    match bool::try_from(RObject::view(list_result)) {
+        Ok(val) => (val, conditions, None),
+        Err(err) => (true, conditions, Some(format!("Error: {err}"))),
```
DavisVaughan: I had wondered what this mysterious 3rd `diagnostic` field was when reading the rest of the code.

...it's just what we get when a conversion from an R bool to Rust bool fails?

That feels like maybe a bit of overkill? Can we just tie that in with `conditions` or something instead?

I think it would have cleared things up for me quite a bit while reading the rest of the code.

crates/ark/src/modules/positron/eval.R
```diff
@@ -0,0 +1,34 @@
...
+#
+
+initialize_eval <- function() {
+    base_bind(as.symbol(".ark_eval_capture"), .ark_eval_capture)
+}
+
+# Evaluate an expression while capturing warnings and messages that R
+# would otherwise defer or send to the console. Returns
+# `list(result, conditions)` where `result` is the raw evaluation result
+# and `conditions` is a list of condition objects (inheriting from
```
lionel-: We can actually do the following to print warnings immediately:

```r
options(warn = 1)
```

I think we should do it from the OutputCapture guard. What do you think?

DavisVaughan: I'm not confident in all of the implications of that but it seems kind of reasonable?

crates/ark/src/console_debug.rs
lionel-: I may misunderstand but it's also the path for R-level errors

PR #1065, opened 2026-02-27
Restore output capture of R tasks correctly

lionel-:
Previously we were only restoring the output capture on Drop, when the idle task was completely finished. If the task yielded with `.await` and went pending, giving control back to R, the output capture would remain in place! This PR fixes that by restoring the output capture at yield points.

Fixes some test failures I observed in the quiet debugger branch.

And we can now test the R task infra!

### DIFF
--- a/crates/ark/tests/kernel-captured-output.rs
+++ b/crates/ark/tests/kernel-captured-output.rs
@@ -5,27 +5,7 @@
 //
 // Integration tests verifying that pending async idle tasks don't swallow
 // console output. See the `captured_output` save/restore mechanism in
-// `poll_task`.
-//
-// The flow exercised here:
-//
-// 1. `.Call("ps_test_spawn_pending_task")` spawns an async idle task that
-//    holds a `ConsoleOutputCapture` and awaits a oneshot channel.
-//
-// 2. On the next event-loop iteration the task is polled, `captured_output`
-//    is set, and the future stays `Pending`. `poll_task` saves the capture
-//    into `pending_futures` and clears `captured_output`. The task sets
-//    the R option `ark.test.task_polled` to `TRUE` so the test can detect
-//    when the poll has happened.
-//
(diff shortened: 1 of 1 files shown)


juliasilge:
We chatted over the weekend and I agree it's OK for the release to go out as is, without this fix.

Thanks for working on this, @lionel-! 🙌

crates/ark/src/r_task.rs
```diff
@@ -372,5 +373,72 @@ where
...
+//    into `pending_futures` and clear `captured_output`.
+//
+// 3. The test busy-loops with `getOption("ark.test.task_polled")` until it
+//    sees `TRUE`, confirming the task has been polled.
+//
+// 4. The test sends another execute request (e.g. `cat("hello\n")`).
+//    Because `captured_output` has been cleared, the output reaches IOPub.
+//
+// 5. Test calls `.Call("ps_test_complete_pending_task")` to unblock the
+//    oneshot, letting the idle task finish and drop its capture cleanly.
```
DavisVaughan, line 396: we duplicate this on the test file side. i think we should only put this in one place, its easy for this to get out of date

crates/stdext/src/result.rs
```diff
@@ -17,6 +17,20 @@ macro_rules! debug_panic {
     };
 }
 
+#[macro_export]
+macro_rules! soft_assert {
```
DavisVaughan, line 21: why not `debug_assert!`, like `debug_panic!`?

they both have the same implication:
- panic, but only die during debug
- assert, but only die during debug

lionel-: That's the name I initially picked, but it conflicts with a macro in the standard library. I thought it's fine to have `stdext::debug_assert!` as a disambiguator but then ran into complications due to scoping of macro exports at top-level that I don't fully understand.

PR #1067, opened 2026-02-27
Add plot file/origin attribution

jmcphers:
Companion to https://github.com/posit-dev/positron/pull/12168; adds additional metadata to plots annotating where they originated, to make it easy to navigate to that location. See that PR for more info.

For linewise execution, this borrows from the source location infrastructure @lionel- set up already for debugging; for execution via `source()`, some slightly more involved work is needed to store the source context so we can get at it later.

### DIFF
--- a/crates/ark/src/plots/graphics_device.rs
+++ b/crates/ark/src/plots/graphics_device.rs
@@ -384,13 +384,15 @@ impl DeviceContext {
     /// and edge cases). The fallback path does not include `code_location`.
     fn capture_execution_context(&self) -> ExecutionContext {
         // Check if we have a stored execution context from on_execute_request()
-        if let Some(ctx) = self.execution_context.borrow().as_ref() {
+        let stored = self.execution_context.borrow();
+        if let Some(ctx) = stored.as_ref() {
             return ExecutionContext {
                 execution_id: ctx.execution_id.clone(),
                 code: ctx.code.clone(),
                 code_location: ctx.code_location.clone(),
             };
         }
+        drop(stored);
 
         // Fall back to getting context from Console (for edge cases).
         // This path does not provide code_location.


crates/ark/src/modules/positron/hooks_source.R
```diff
@@ -70,6 +70,24 @@ make_ark_source <- function(original_source) {
...
+        # source file. This is best-effort; if it fails we proceed without attribution.
+        source_uri <- tryCatch(path_to_file_uri(file), error = function(e) NULL)
+
+        # Push source context for plot attribution (if we have a URI).
+        # The on.exit ensures we always pop, even if source() errors.
+        # Use tryCatch so that source() still works if the native function
+        # is not yet available (e.g. during development with mismatched builds).
+        if (!is.null(source_uri)) {
+            pushed <- tryCatch(
+                { .ps.Call("ps_graphics_push_source_context", source_uri); TRUE },
```
DavisVaughan: Can you please reformat with Air? I should probably turn that on as a CI check...

crates/ark/src/modules/positron/hooks_source.R
```diff
@@ -70,6 +70,24 @@ make_ark_source <- function(original_source) {
...
+                error = function(e) FALSE
+            )
+            if (pushed) {
+                on.exit(tryCatch(.ps.Call("ps_graphics_pop_source_context"), error = function(e) NULL), add = TRUE)
```
DavisVaughan: We are trying to use `defer()` instead of `on.exit()`, which comes with the right defaults

crates/ark/src/modules/positron/hooks_source.R
```diff
@@ -70,6 +70,24 @@ make_ark_source <- function(original_source) {
             args$catch.aborts <- NULL
         }
 
+        # Try to resolve the file URI early so we can attribute plots to this
+        # source file. This is best-effort; if it fails we proceed without attribution.
+        source_uri <- tryCatch(path_to_file_uri(file), error = function(e) NULL)
```
DavisVaughan: I see no reason to not just consistently call it `uri` everywhere rather than renaming it later on

crates/ark/src/console.rs
```diff
@@ -1535,10 +1427,16 @@ impl Console {
                     reply_tx,
                 });
 
-                // Push execution context to graphics device for plot attribution
+                // Push execution context to graphics device for plot attribution.
+                // Extract code_location from the execute request for source file origin.
+                let code_location = exec_req.code_location().log_err().flatten().map(|mut loc| {
```
DavisVaughan, line 1540: @lionel- should `code_location()` just `normalize()` the URI internally? It seems like the one other place we call `code_location()` also normalizes after the fact.

crates/ark/src/plots/graphics_device.rs
```diff
@@ -1189,3 +1321,20 @@ unsafe extern "C-unwind" fn ps_graphics_get_metadata(id: SEXP) -> anyhow::Result
...
+    Ok(harp::r_null())
+}
+
+/// Pop a source file URI from the source context stack.
+/// Called from the `source()` hook when leaving a sourced file.
+#[harp::register]
+unsafe extern "C-unwind" fn ps_graphics_pop_source_context() -> anyhow::Result<SEXP> {
+    DEVICE_CONTEXT.with_borrow(|cell| cell.pop_source_context());
+    Ok(harp::r_null())
+}
```
DavisVaughan, line 1344: IIUC the desire is to be able to query the name of the file that is currently being `source()`-ed at any time while inside the execution of that file.

That doesn't feel super specific to plots, so I would not be against that being a generic feature that lives in `Console` instead.

@lionel- what do you think?

crates/ark/src/plots/graphics_device.rs
```diff
@@ -213,29 +235,51 @@ impl DeviceContext {
...
+    fn current_source_uri(&self) -> Option<String> {
+        self.source_context_stack.borrow().last().cloned()
     }
 
     /// Create a new id for this new plot page (from Positron's perspective)
     /// and note that this is a new page
     fn new_positron_page(&self) {
         self.is_new_page.replace(true);
         self.id.replace(Self::new_id());
+        *self.pending_origin.borrow_mut() = None;
```
DavisVaughan: Use `self.pending_origin.replace(None)` like we've done for the other fields?

Here and elsewhere.

crates/ark/src/plots/graphics_device.rs
```diff
@@ -316,18 +369,65 @@ impl DeviceContext {
...
+        // Check if we have a stored execution context from on_execute_request()
+        let stored = self.execution_context.borrow();
+        if let Some(ctx) = stored.as_ref() {
+            return ExecutionContext {
+                execution_id: ctx.execution_id.clone(),
+                code: ctx.code.clone(),
+                code_location: ctx.code_location.clone(),
+            };
+        }
+        drop(stored);
```
DavisVaughan, line 395:
```suggestion
        if let Some(ctx) = self.execution_context.borrow().as_ref() {
            return ExecutionContext {
                execution_id: ctx.execution_id.clone(),
                code: ctx.code.clone(),
                code_location: ctx.code_location.clone(),
            };
        }
```

crates/ark/src/plots/graphics_device.rs
```diff
@@ -966,6 +1084,12 @@ pub(crate) fn on_did_execute_request() {
     DEVICE_CONTEXT.with_borrow(|cell| {
         cell.process_changes();
         cell.clear_execution_context();
+        // Clear any unconsumed pending origin so it doesn't leak into the
+        // next execute request. `process_changes()` above already consumed it
+        // if this execution produced a new plot; this handles the case where
+        // drawing occurred (e.g. `lines()` inside `source()`) but only as an
+        // update to an existing plot, leaving the pending origin unclaimed.
+        *cell.pending_origin.borrow_mut() = None;
```
DavisVaughan: Maybe we want `set_pending_origin()` and `clear_pending_origin()` to go along with `set_execution_context()` and `clear_execution_context()`? That probably cleans this up a little bit.

crates/ark/src/modules/positron/hooks_source.R
```diff
@@ -70,6 +70,24 @@ make_ark_source <- function(original_source) {
             args$catch.aborts <- NULL
         }
 
+        # Try to resolve the file URI early so we can attribute plots to this
```
DavisVaughan, line 73: Since @lionel- is the author of this bit I think he should sign off on the changes in this file

crates/ark/src/modules/positron/hooks_source.R
jmcphers: Yeah, it looks like a lot of the `.R` files are not air-compliant. I went ahead and formatted the whole workspace (LMK if you'd like to keep the changes scoped to just this new file)

crates/ark/src/modules/positron/hooks_source.R
jmcphers: done!

crates/ark/src/modules/positron/hooks_source.R
jmcphers: done!

crates/ark/src/plots/graphics_device.rs
jmcphers: done!

crates/ark/src/plots/graphics_device.rs
jmcphers: done!

crates/ark/src/modules/positron/hooks_source.R
```diff
@@ -70,6 +70,30 @@ make_ark_source <- function(original_source) {
...
+                    TRUE
+                },
+                error = function(e) FALSE
+            )
+            if (pushed) {
+                defer(tryCatch(
+                    .ps.Call("ps_graphics_pop_source_context"),
+                    error = function(e) NULL
+                ))
+            }
```
lionel-: nit: This could equivalently be run unconditionally in the try branch, after the .Call returns.

crates/ark/src/plots/graphics_device.rs
```diff
@@ -316,18 +381,64 @@ impl DeviceContext {
...
+            code_location: None,
+        }
+    }
+
+    /// Determine the plot origin for a new plot.
+    ///
+    /// Checks three sources in priority order:
+    /// 1. Source context stack (inside `source("file.R")`) -- file-level, no range
+    /// 2. Execute request's `code_location` (code run from a file via the IDE) -- with range
+    /// 3. None (code typed at the console)
```
lionel-, line 416: 👍

crates/ark/src/plots/graphics_device.rs
lionel-: That makes sense to me, although I can't think of a reason to do it proactively.

I was initially thinking it may help with transitions of comms to Console thread, as the stack could be made available via the comm context. But the graphics device is not a comm so I don't think that will help there.

Feel free to add the source context to the Console though Jonathan.

In that case it might make sense to include the execute request's location in the stack so that priority resolution is consistently handled.

crates/amalthea/src/comm/plot_comm.rs
```diff
@@ -66,6 +69,34 @@ pub struct PlotSize {
 	pub width: i64
 }
 
+/// The origin (source) of a plot
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+pub struct PlotOrigin {
```
lionel-, line 74: Since the structure is very similar to CodeLocation, I think I would have expected `PlotLocation`. I guess `Origin` has a slightly different nuance though, and indicates the original location might have changed.

crates/ark/src/console.rs
lionel-: That's a good idea!

`code_location()` lives in Amalthea and `ExtUrl::normalize()` is in Ark, so a nice place to implement it would be on the `ExtUrl` extension trait:

```rust
let code_location = ExtUrl::from_execute_request(&exec_req);
```

crates/ark/tests/plots.rs
```diff
@@ -355,3 +360,158 @@ fn test_plot_get_metadata() {
         "Metadata should contain kind 'plot', got:\n{result}"
     );
 }
+
```
lionel-, line 363: Nice to see the new test infra put to good use!

I think it'd be worth testing the use case of sourcing a file A that sources a file B, to exercise the stacked aspect of locations.

crates/ark/src/plots/graphics_device.rs
jmcphers: Thanks, agree that could be helpful but won't do it proactively here!

crates/amalthea/src/comm/plot_comm.rs
jmcphers: Yes, it's hard to name this correctly! I think `PlotLocation` might be misread as "where the plot _itself_ is located" vs "where the plot came from"

crates/ark/tests/plots.rs
jmcphers: good idea! done in https://github.com/posit-dev/ark/pull/1067/commits/08d50b847faa8ef2baa84cb9567e754b09cc03f3

crates/ark/src/modules/positron/hooks_source.R
jmcphers: yep https://github.com/posit-dev/ark/pull/1067/commits/c1f9b563dc658b1a3627b8aced108d2f1a3ab201

crates/ark/src/console.rs
lionel-: Actually I realised while working on https://github.com/posit-dev/ark/pull/1082 that we should not normalise/canonicalise URIs that flow back to the frontend, because it might think it's a different file. I noticed Positron would open a canonicalised path in a new editor instead of reusing an existing one.

Fixed in the linked PR.

PR #1068, opened 2026-02-27
Filter debug messages from console output

lionel-:
Branched from #1065.
Addresses https://github.com/posit-dev/positron/issues/12230

R's debug repl emits messages like `debug at file.R#10: `, `Called from: `, `debugging in: `, and `exiting from: ` to the console. In a terminal these help the user stay oriented through debug sessions, and IDEs can parse them to extract stepping information. In Positron, however, the GUI already surfaces this information through the call stack pane and step highlighting, making the console messages redundant and distracting.

With the injected breakpoints we use for auto-stepping, the noise unfortunately increased quite a bit. This makes the console quite a hostile place for users during debugging sessions, when they really should feel at home and be able to easily find output of interest.

Screenshot 2026-02-27 at 18 18 05

This PR introduces a `ConsoleFilter` state machine that intercepts stdout at `WriteConsole` time, detects debug message prefixes, and suppresses the stepping noise at `ReadConsole` time (`debug at`, `debug:`, `Called from:`) while leaving function entry/exit messages (`debugging in:`, `exiting from:`) visible.

### What R emits

R's debug handler emits four families of messages, all from C code in `eval.c`.

**`debug at file.R#N: ` or `debug: `** (the latter when no srcref is available). Emitted on each step through braces, control flow, and functions. Followed by `do_browser()`:

https://github.com/r-devel/r-svn/blob/35ef64e6e7f7ac5423747d9e7be70bbde47f0b96/src/main/eval.c#L2988-L2990

```c
SrcrefPrompt("debug", R_Srcref);
PrintValue(CAR(args));        // the expression about to be evaluated
do_browser(call, op, R_NilValue, rho);
```

**`Called from: `**. Emitted by `browser()` itself before entering the REPL. Followed by `do_browser()`.

**`debugging in: `**. Emitted when entering a closure via `debug()` or `s`. Followed by `do_browser()`:

https://github.com/r-devel/r-svn/blob/35ef64e6e7f7ac5423747d9e7be70bbde47f0b96/src/main/eval.c#L2369C1-L2371C34

```c
Rprintf("debugging in: ");
PrintCall(call, rho);         // the call being entered
SrcrefPrompt("debug", R_Srcref);
PrintValue(body);             // the function body
do_browser(call, op, R_NilValue, newrho);
```

**`exiting from: `**. Emitted when leaving a debugged closure (in `R_execClosure`). Unlike the other three, this is NOT followed by `do_browser()` — user code resumes immediately after:

```c
Rprintf("exiting from: ");
PrintCall(call, rho);         // the call being left
// no do_browser() — execution continues in the calling function
```

Each prefix is followed by a `PrintValue()` or `PrintCall()` that can produce arbitrary multi-line output (long calls wrapping across lines, literal vectors, even large objects injected by metaprogramming). There's no syntactic delimiter marking the end of a debug message, so we can't use parse-based detection.

The key insight that makes reliable filtering possible is that three of the four prefixes are immediately followed by `do_browser()`, which enters `ReadConsole`. No user code runs between the prefix and that `ReadConsole`, so everything in between is debug noise that can be (mostly) safely accumulated and suppressed.

`exiting from:` is the exception. It is emitted after the browser's `ReadConsole` has already returned (the user typed `c`, `n`, etc.) but before R reaches the _next_ `ReadConsole`. User code runs after the prefix: the function returns a value and that value may get printed. There is no delimiter between the `PrintCall` output and subsequent user code output, making it impossible to safely strip `exiting from:` without risking swallowed user output.

### Filter mechanism

We filter only the stepping noise, i.e. the prefixes that fire on every step and are always immediately followed by `do_browser()`:

- **`debug at` / `debug:`**
- **`Called from:`**

We do not filter function entry/exit transitions:

- **`exiting from:`** can't be filtered because user code runs after it
- **`debugging in:`** not filtered for symmetry with `exiting from:`

This is a good trade-off. The stepping noise is by far the most frequent and distracting — especially with injected breakpoints producing a flurry of `Called from:` + `debug at` pairs on every auto-step. The entry/exit messages are comparatively rare (once per function boundary) and useful for orientation.

The filter is a state machine with two states:

- **Passthrough**: Default. Content is emitted to IOPub immediately. At line boundaries, checks if new content matches a full filtered debug prefix. On match, transitions directly to Filtering. We don't need to worry about partial matching since R writes each prefix with a single `Rprintf` call so it always arrives as one complete chunk to WriteConsole.

- **Filtering**: Full prefix match confirmed. Accumulates all subsequent content until `ReadConsole` resolves whether this is real debug output or adversarial user output.

Resolution happens at `ReadConsole` time:

- **Debug context** (`was_debugging || is_browser`, we either just left a browser session or are entering one): content is suppressed as real debug output, and the captured expression is extracted for auto-stepping.
- **Top-level prompt**: content was user output that happened to match a prefix, so it's emitted back to the user.

A timeout mechanism (50ms, 500ms in tests) ensures content is never held indefinitely if `ReadConsole` doesn't arrive promptly.

Also noteworthy:

- Debug messages also reach autoprint output (the `execute_result` path) when stepping through top-level braced expressions. `strip_step_lines()` cleans autoprint at browser prompts by truncating at the first line matching `Called from:`, `debug at`, or `debug:`. This is only applied at browser prompts — at top-level prompts these strings could be legitimate user output from print methods.

- The old `debug_handle_write_console` / `debug_handle_read_console` pair that captured `debug: ` and `debug at ` for auto-stepping is replaced by the filter's `finalize_capture`. The `DebugCallText::Capturing` variant is removed since the filter now handles accumulation. The filter produces `DebugCallTextUpdate` values that the console applies at `ReadConsole` time.

### Safety guarantees

The filter only suppresses output when it can confirm a debug context: either a browser prompt is reached, or the prefix was matched while already in a debug session. At top-level prompts outside of debug sessions, any prefix-matching content is emitted back to the user. Combined with the timeout fallback, this guarantees the filter has no impact on regular (non-debug) usage.

User `cat()` output that matches a filtered debug prefix (`Called from:`, `debug at`, `debug:`) inside a browser session IS suppressed. We can't distinguish it from real debug output because both are followed by a browser `ReadConsole` prompt. This is documented in the adversarial tests.

Filtering can be disabled with `options(ark.debugger.filter_debug_output = FALSE)`. When `NULL` (the default), filtering is enabled. When disabled, the filter still runs (extracting debug info for auto-stepping) but emits all content instead of suppressing it.

### QA Notes

Comprehensively tested with both unit tests (state machine) and integration tests (full kernel), including adversarial scenarios where user code produces output matching debug prefixes, and nested debug scenarios (debugged function calling another debugged function, debugged function returning to non-debugged caller).

Debug sessions should no longer show `Called from:`, `debug at`, or `debug:` messages in the console. `debugging in:` and `exiting from:` messages remain visible. Normal output during debugging (e.g. `cat()`, `print()`) should be unaffected.

### DIFF
--- a/crates/ark/src/console.rs
+++ b/crates/ark/src/console.rs
@@ -297,6 +297,10 @@ pub struct Console {
     /// Whether or not we are currently in a debugging state.
     pub(crate) debug_is_debugging: bool,
 
+    /// Filter for debug console output. Removes R's internal debug messages
+    /// from user-visible console output.
+    pub(crate) debug_filter: ConsoleFilter,
+
     /// The current call emitted by R as `debug: <call-text>`.
     pub(crate) debug_call_text: Option<DebugCallText>,
 
@@ -355,10 +359,6 @@ pub struct Console {
     /// Pushed on entry to `r_read_console()`, popped on exit.
     /// This is a RefCell since we require `get()` for this field and `RObject` isn't `Copy`.
     pub(crate) read_console_env_stack: RefCell<Vec<RObject>>,
-
-    /// Filter for debug console output. Removes R's internal debug messages
-    /// from user-visible console output.
-    pub(crate) filter: ConsoleFilter,
 }
 
 /// Stack of pending inputs
@@ -950,7 +950,7 @@ impl Console {
             read_console_pending_action: Cell::new(ReadConsolePendingAction::None),
             read_console_env_stack: RefCell::new(Vec::new()),
             read_console_shutdown: Cell::new(false),
-            filter: ConsoleFilter::new(),
+            debug_filter: ConsoleFilter::new(),
         }
     }
 
@@ -1114,7 +1114,7 @@ impl Console {
         let is_browser = matches!(info.kind, PromptKind::Browser);
 
         let suppress = filter_debug_output();
-        self.filter.set_suppress(suppress);
+        self.debug_filter.set_suppress(suppress);
 
         // Debug prefixes (`Called from:`, `debug at`, `debug:`) reach
         // autoprint when stepping in top-level braced expressions. We only
@@ -1130,7 +1130,7 @@ impl Console {
         // A browser prompt means filtered content was real debug output (which
         // we suppress). Top-level prompt means it was user output matching a
         // prefix (which we emit).
-        let (emit, debug_update) = self.filter.on_read_console(is_browser);
+        let (emit, debug_update) = self.debug_filter.on_read_console(is_browser);
         if let Some(text) = emit {
             self.emit_stdout(text);
         }
@@ -2486,7 +2486,7 @@ impl Console {
 
         if stream == Stream::Stderr {
             // Flush any buffered stdout so it appears before this stderr
-            if let Some(text) = console.filter.flush() {
+            if let Some(text) = console.debug_filter.flush() {
                 console.emit_stdout(text);
             }
 
@@ -2499,7 +2499,7 @@ impl Console {
             return;
         }
 
-        let emits = console.filter.feed(&content);
+        let emits = console.debug_filter.feed(&content);
         for text in emits {
             console.emit_stdout(text);
         }
@@ -2572,7 +2572,7 @@ impl Console {
         // accumulated content is emitted. This allows user code to produce
         // output that looks like debug lines emitted by R without them getting
         // filtered out or held up too long.
-        if let Some(text) = self.filter.check_timeout() {
+        if let Some(text) = self.debug_filter.check_timeout() {
             self.emit_stdout(text);
         }
 


crates/ark/src/console_debug.rs
```diff
@@ -149,61 +155,13 @@ impl Console {
...
-            DebugCallText::Finalized(_, _) => (),
-            // If capturing, transition to finalized.
-            DebugCallText::Capturing(call_text, kind) => {
-                self.debug_call_text = DebugCallText::Finalized(call_text.clone(), *kind)
+    /// Apply a debug call text update from the stream filter
+    pub(crate) fn debug_update_call_text(&mut self, update: DebugCallTextUpdate) {
+        match update {
+            DebugCallTextUpdate::Finalized(text, kind) => {
+                self.debug_call_text = DebugCallText::Finalized(text, kind);
             },
```
DavisVaughan: Something seems outdated here.

`DebugCallText` seems like it should really just be `None` and `Some` now, or something like that.

There is no need for a `Finalized` state if there is no longer a `Capturing` state?

DavisVaughan: And why are `DebugCallTextUpdate` and `DebugCallText` different? It seems like we only need 1 type here.

crates/ark/src/console_filter.rs
```diff
@@ -0,0 +1,768 @@
...
+
+    /// Confirmed prefix match. Accumulating all subsequent content until
+    /// `ReadConsole` resolves whether this is real debug output or adversarial
+    /// user output.
+    Filtering {
+        pattern: MatchedPattern,
+        buffer: String,
+        timestamp: Instant,
+        /// Whether the console was in a debug session when this match started.
+        was_debugging: bool,
```
DavisVaughan: How could it not be?

crates/ark/src/console_filter.rs
```diff
@@ -0,0 +1,768 @@
...
+    is_debugging: bool,
+    /// Whether to suppress confirmed debug output. When `false`, the
+    /// filter still extracts debug info for auto-stepping but emits
+    /// the content instead of dropping it.
+    suppress: bool,
+}
+
+fn get_timeout() -> Duration {
+    if stdext::IS_TESTING {
+        Duration::from_millis(500)
```
DavisVaughan: why do you need this and `new_with_timeout`?

crates/ark/src/console_filter.rs
```diff
@@ -0,0 +1,768 @@
...
+            ConsoleFilterState::Filtering {
+                pattern,
+                buffer,
+                was_debugging,
+                ..
+            } => {
+                if was_debugging || is_browser {
+                    debug_update = finalize_capture(pattern, &buffer);
+                }
+                if !(was_debugging || is_browser) || !self.suppress {
```
DavisVaughan:
```suggestion
                let needs_capture = was_debugging || is_browser;
                if needs_capture {
                    debug_update = finalize_capture(pattern, &buffer);
                }
                if !needs_capture || !self.suppress {
```

Easier to read I think?

crates/ark/src/console_filter.rs
```diff
@@ -0,0 +1,768 @@
...
+    }
+
+    /// Check for timeout and handle state transitions.
+    /// Timeout means we didn't reach ReadConsole to confirm debug output,
+    /// so we emit the accumulated content back to the user.
+    pub fn check_timeout(&mut self) -> Option<String> {
+        self.drain_on_timeout()
+    }
+
+    fn drain_on_timeout(&mut self) -> Option<String> {
```
DavisVaughan, line 267: I'm not sure I understand the timeout. Is it really needed?

crates/ark/src/console_filter.rs
```diff
@@ -0,0 +1,768 @@
...
+                pattern, buffer, ..
+            } => {
+                format!("{}{}", pattern.prefix(), buffer)
+            },
+        };
+        self.state = ConsoleFilterState::Passthrough {
+            at_line_start: text.ends_with('\n'),
+        };
+        Some(text)
+    }
```
DavisVaughan, line 299:
```suggestion
    fn drain(&mut self) -> Option<String> {
        match &self.state {
            ConsoleFilterState::Passthrough { .. } => {
                self.state = ConsoleFilterState::Passthrough {
                    at_line_start: true,
                };
                None
            },
            ConsoleFilterState::Filtering {
                pattern, buffer, ..
            } => {
                let text = format!("{}{}", pattern.prefix(), buffer);
                self.state = ConsoleFilterState::Passthrough {
                    at_line_start: text.ends_with('\n'),
                };
                Some(text)
            },
        }
    }
```

This feels a little complicated vs just this

crates/ark/src/console_filter.rs
```diff
@@ -0,0 +1,768 @@
...
+/// Strip `Called from:` / `debug at` / `debug:` lines from autoprint.
+///
+/// These prefixes appear after user output (e.g. `{ browser(); 1; 2 }`
+/// at top level produces user output followed by `debug at` lines).
+/// Truncating at the first match also removes multi-line `PrintValue`
+/// continuations.
+///
+/// Must only be called at browser prompts: at top level, user print
+/// methods could legitimately produce output matching these prefixes.
+pub fn strip_step_lines(text: &mut String) {
```
DavisVaughan, line 367: Is the prefix not always in the first line?

crates/ark/src/console.rs
```diff
@@ -357,6 +358,10 @@ pub struct Console {
     /// Pushed on entry to `r_read_console()`, popped on exit.
     /// This is a RefCell since we require `get()` for this field and `RObject` isn't `Copy`.
     pub(crate) read_console_env_stack: RefCell<Vec<RObject>>,
+
+    /// Filter for debug console output. Removes R's internal debug messages
+    /// from user-visible console output.
+    pub(crate) filter: ConsoleFilter,
```
DavisVaughan:
```suggestion
    pub(crate) debug_filter: ConsoleDebugFilter,
```

This feels like it could use a very debug specific name?

Otherwise I think it could be invoked at any time to filter something

crates/ark/src/console_debug.rs
lionel-: Good point, this is now

```rust
pub enum DebugCallText {
    /// `debug: <expr>` - emitted when stepping without srcrefs
    Debug(String),
    /// `debug at <path>#<line>: <expr>` - emitted when stepping with srcrefs
    DebugAt(String),
}
```

crates/ark/src/console_filter.rs
lionel-: `new_with_timeout()` is used for unit tests so they don't wait as much.

crates/ark/src/console_filter.rs
lionel-: This is part of our guarantee that output is never swallowed if the user doesn't actually drop in the debugger. We only hold onto filtered output until a small timeout of 50ms. This way if the user has a `cat("debug: ")` call in their code, we won't hold all their subsequent output until the next ReadConsole (which might never come in a server-like situation), they'll just see a small delay.

It's still possible that we swallow user output in this way, but only if we eventually hit the debugger within the timeout delay, which makes the risk acceptable.

crates/ark/src/console.rs
lionel-: hmm the filter _is_ always active, including in top-level repls. The initial debug output comes _before_ dropping in a nested debug repl.

crates/ark/src/console_filter.rs
lionel-: This is called for top-level filtering of execute results, so there could be user output before.

```
[1] 1
debug at #3: 2
```

crates/ark/src/console.rs
lionel-: I let Claude settle it :)

> `debug_filter`. Davis is right that `filter` is too generic on a struct this large. And the fact that it's always active doesn't make it less debug-specific in purpose.

Issue #1069, opened 2026-02-27
Kernel Test Bed


rgbkrk:
Hey all! I put together a [kernel test bed](https://runtimed.github.io/kernel-testbed/) to evaluate how much jupyter kernels are properly obeying the jupyter protocol while also making sure that the runtimed crates (jupyter related rust crates) are working well with all the available kernels.

Along its colleagues deno and julia, [ark is at a 92% pass rate](https://runtimed.github.io/kernel-testbed/kernel/ark/), tying for second place!

<a href="https://runtimed.github.io/kernel-testbed/kernel/ark/"><img width="1200" height="630" alt="Image" src="https://github.com/user-attachments/assets/1066201f-9ec1-468e-bb16-11a01afb0786" /></a>

The only two the test suite were unable to exercise for ark were `history_request` and `comm_info_request`.

PR #1072, opened 2026-03-02
Handle absent target name in comm info request

lionel-:
Progress towards #1069.

The `target_name` field of `comm_info_request` is optional in the Jupyter protocol. When absent, the server should provide info for all comms.

Currently we're treating `""` as the all comm sentinel, but the protocol really specifies the field as optional (can be absent). To fix this, the field is now an `Option<String>` instead of a `String`.

### DIFF
--- a/crates/amalthea/src/wire/comm_info_request.rs
+++ b/crates/amalthea/src/wire/comm_info_request.rs
@@ -13,7 +13,8 @@ use crate::wire::jupyter_message::MessageType;
 /// Represents a request from the frontend to show open comms
 #[derive(Debug, Serialize, Deserialize, Clone)]
 pub struct CommInfoRequest {
-    pub target_name: String,
+    #[serde(default)]
+    pub target_name: Option<String>,
 }
 
 impl MessageType for CommInfoRequest {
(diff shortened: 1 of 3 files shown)


crates/amalthea/src/wire/comm_info_request.rs
```diff
@@ -13,7 +13,8 @@ use crate::wire::jupyter_message::MessageType;
 /// Represents a request from the frontend to show open comms
 #[derive(Debug, Serialize, Deserialize, Clone)]
 pub struct CommInfoRequest {
-    pub target_name: String,
+    #[serde(default)]
```
DavisVaughan, line 16: I noticed that https://github.com/posit-dev/ark/pull/1073/changes#diff-f6c818169bdba550162747979f6b260e31e570d6ed62dc25d2d0a203aa01d114 was also adding `default` everywhere. What's up with that?

lionel-: This is needed to make the field truly optional, so it can be absent from the json.

In the other PR, even though the fields are technically required within each variant type, the protocol also notes:

> Most of the history messaging options are not used by Jupyter frontends, and many kernels do not implement them. If you’re implementing these messages in a kernel, the ‘tail’ request is the most useful; this is used by the Qt console, for example. The notebook interface does not use history messages at all.

Issue #1074, opened 2026-03-02
Migrate existing comm tests to protocol-level tests


lionel-:
In recent PRs we've started using the `DummyArkFrontend` infrastructure for protocol-level tests of comms:

- https://github.com/posit-dev/ark/pull/1052 (variables)
- https://github.com/posit-dev/ark/pull/1060 (data explorer and variables)
- https://github.com/posit-dev/ark/pull/1067 (plots)

These tests are much nicer and effective than the old comm-based legacy tests because:

- They exercise the whole kernel. This means that there is no need to simulate environment changed events after modifying the global env, which requires knowing a deeper knowledge of how internals work to produce a meaningful test (see https://github.com/posit-dev/ark/blob/239ee7e4d98101348fab833d1d668fd5a8d86b0b/crates/ark/tests/variables.rs#L116). We can let the normal operation of the kernel generate this sort of events.

- It allows testing comms in conjunction with DAP and LSP operations and state. For instance the DAP now triggers comm-visible events, and it's nice to exercise these paths.

- Inputs to the tests can be modelled after user requests, such as executing R code in the console. Compare to how we simulate global environment changes for Variables pane tests: https://github.com/posit-dev/ark/blob/239ee7e4d98101348fab833d1d668fd5a8d86b0b/crates/ark/tests/variables.rs#L109-L113

Before doing that though, we should solve https://github.com/posit-dev/ark/issues/689. This will allow comms to behave deterministically and tested with sequential assertions that are much stronger than fuzzy matching of events.

Issue #691, opened 2024-02-26
Ark: Infrastructure: Reduce concurrent R evaluations to a minimum


lionel-:
Currently there are many places where we evaluate R code / access the R API through the use of `r_task()`, which hooks into R's polled events. This makes it easy to implement analytic features but has a number of downsides:

- The R event polls might be slowed down of stuck if native code running in the R session doesn't check for interrupts sufficiently often.

- While R polls at specific times such as checks for interrupts, in practice we should consider polled events to be preemptive. This is unsafe because we might accidentally call a non-rentrant function, for instance we might force a promise that causes a `loadNamespace()`, which is not reentrant, while it is already running. It is difficult to manage these risks as the project grows and many features are added, querying R all the time.

So we should strive to reduce concurrent evaluations to the bare minimum:

- [x] https://github.com/posit-dev/positron/issues/2284
- [ ] https://github.com/posit-dev/ark/issues/689
- [ ] https://github.com/posit-dev/positron/issues/2321

Issue #689, opened 2024-01-12
RPC: Ark: Switch to blocking requests to avoid concurrent R evaluations


lionel-:
I think we need to be a bit more careful about the timing of evaluation of backend RPCs on the R side. This concerns the `callMethod()` mechanism (which makes a request from typescript via the UI comm) and more generally any requests performed by a comm. These shouldn't run concurrently with the R interpreter because we are running complex functions that have not been designed for reentrancy or preemption (more precisely interrupt-time, which happens at somewhat controlled points, but in practice we should consider it happens at any time).

RPCs are becoming our main way of calling R code from positron-r because they support serialisation of arguments whereas `execute()` only support textual arguments. Because of this we are bound to be calling complex code, for example code that loads an R package. This is potentially bad for many reasons, for instance if `loadNamespace()` is in the middle of loading another package at that moment.

Same concern for comm RPCs, for instance for the connection comm. The `.ps.connection_` API is accessible from R and might be called by a package or a user. Then a message comes in and runs connection code preemptively (at interrupt time). This means the connection R code should be reentrant but we can't reasonably expect this. I think we'll eventually end up with weird state corruption. It doesn't even require mutable state to get corruption, it just suffices for state to be split in multiple places and change preemptively between two lookups.

Thankfully I don't think we'll lose much by making these requests synchronous. They are currently made via `comm_msg` which is queued on the Shell socket. The problem comes from closing the `comm_msg` request too fast, which allows Shell to invoke the next request, possibly an `execute_request` or another comm RPC. To avoid that we just need to block during the processing of the comm message to prevent `Shell` from executing the next request while the comm is calling R code.

lionel-:
Once we've switched to blocking requests, we can make the integration tests for comms more deterministic, see https://github.com/posit-dev/ark/blob/12ec7fed4c3e3a53cd575769ccf9c441e9a3b5a8/crates/ark_test/src/dummy_frontend.rs#L72-L76

PR #1075, opened 2026-03-02
Run comms on the R thread

lionel-:
Progress towards https://github.com/posit-dev/ark/issues/691
Progress towards https://github.com/posit-dev/ark/issues/1074

Comm handlers (data explorer, variables, connections...) currently run on their own spawned threads and call R through `r_task()`, which hooks into R's polled events. In practice this makes R evaluations preemptive: a comm handler may force a promise that triggers `loadNamespace()` while `loadNamespace()` is already on the call stack. As we move more logic to comm RPCs, the surface area for these reentrancy bugs grows. The data explorer is a good example: it sorts, filters, and profiles columns by calling back into R, all from a background thread.

A second, more subtle problem is message ordering. Because comm handlers run on their own threads, IOPub events (like data-changed notifications or `comm_close`) race with Idle. Tests had to work around this with complicated buffering and polling in `DummyArkFrontend`.

This PR introduces a blocking comm path and migrates the data explorer as the first comm to demonstrate the wins:

- No more R reentrancy risk from `r_task()` at interrupt time.
- No more risk of _omitting_ `r_task()` wrappers around R-related code, which allows a much nicer development experience (and easier code reviews).
- Much simpler implementation (the dedicated thread and `select!` loop goes away).
- 10 data explorers used to spawn 10 comm threads. We now spawn 0 additional threads. In addition to reduced complexity, this reduces memory usage since each thread allocates 2mb for its stack. This reduction will also apply to plot comms, so could be significant in real sessions.
- Deterministic update ordering (environment change side effects land within the Busy/Idle window of the request that caused them). The tests become deterministic and all the event buffering test infra goes away.

### Blocking Shell while comms run on the R thread

The fix is to make comm message handling **block Shell**. When a `comm_msg` arrives on the Shell socket, instead of forwarding it to a channel and immediately moving on to the next request, we send the message to the R thread and wait for it to finish. This way no `execute_request` (or other comm RPC) can start while R code is running for a comm.

#### Amalthea side

- Two new default methods on `ShellHandler`: `handle_comm_msg()` and `handle_comm_close()`. They return `CommHandled::Handled` or `CommHandled::NotHandled`. The latter falls back to the existing `incoming_tx` path, so non-migrated comms keep working unchanged. This is the only part that touches the framework; the rest is in ark.

- In `Shell::handle_comm_msg()` / `handle_comm_close()`, the new handler is tried first. If it returns `NotHandled`, we fall through to the old channel send.

#### Ark side

**The `CommHandler` trait**. This is the new contract for comms that want to run on the R thread:

- `handle_open()`: called once on the R thread when the comm is registered.
- `handle_msg()`: incoming RPC or data message.
- `handle_close()`: frontend-initiated close.
- `handle_environment()`: notification after each top-level execution (or debug frame selection), so handlers can react to binding changes.
- `open_metadata()`: metadata for the `comm_open` IOPub message (backend-initiated comms).

A `CommHandlerContext` gives each handler access to any relevant Console state or methods, to its `outgoing_tx` for emitting messages to the frontend, and to a `close_on_exit()` mechanism for self-closing (e.g. when a watched binding disappears).

**Console as the comm registry**:

Console owns a `HashMap<String, ConsoleComm>` keyed by comm ID. The `comm_register()` method handles backend-initiated opens (creates `CommSocket`, calls `handle_open`, sends `CommEvent::Opened`). Frontend-initiated opens go through `KernelRequest::CommOpen` with a factory closure so the handler is constructed on the R thread (important because it may hold `RObject`s).

**Ordering change** in `Console::handle_execute_result()`: The old code emitted `environment_changed` *before* returning to the event loop, which meant the event ran at an unpredictable time via `r_task()`. Now the sequence is explicit:

1. Send execute result/error on IOPub
2. Call `comm_notify_environment_changed()` synchronously
3. *Then* send the execute reply (which unblocks Shell, which sends Idle)

This guarantees all comm side-effects (data explorer updates, closes) land on IOPub **within the Busy/Idle window** of the execute request that caused them. This is what makes test assertions deterministic.

### Data explorer migration

Data explorer is the first comm fully migrated to the new path. The diff is large but the transformation is mechanical:

- No more spawned thread: `RDataExplorer::start()` → `RDataExplorer::new()`. Construction returns the struct directly instead of spawning a thread with a `select!` loop.
- No more `r_task()` calls: All the `r_task(|| self.r_foo())` wrappers become direct `self.foo()` calls. The `r_` prefix indicating a method is to be called on the R thread is dropped.
- No more `RThreadSafe`: Since everything runs on the R thread, `Table` becomes a plain newtype over `RObject` (with a manual `unsafe impl Send` for the idle-task path). The `Arc<Mutex<Option<RThreadSafe<RObject>>>>` is gone.

Because data explorer messages now arrive deterministically on IOPub, the entire `DataExplorerBuffer` infrastructure in `DummyArkFrontend` is removed (~180 lines of buffering, polling, and `try_buffer_msg` logic). The integration tests now robustly assert comm events sequentially.

### DIFF
--- a/crates/ark/src/data_explorer/r_data_explorer.rs
+++ b/crates/ark/src/data_explorer/r_data_explorer.rs
@@ -538,7 +538,7 @@ impl RDataExplorer {
         let id = params.callback_id.clone();
 
         let params = ProcessColumnsProfilesParams {
-            table: self.table.clone(),
+            table: self.table.clone_for_task(),
             indices: self.filtered_indices.clone(),
             kind: self.shape.kind,
             request: params,
--- a/crates/ark/src/data_explorer/table.rs
+++ b/crates/ark/src/data_explorer/table.rs
@@ -7,23 +7,25 @@
 
 use harp::RObject;
 
-#[derive(Clone)]
-pub struct Table(RObject);
+use crate::thread::RThreadSafe;
 
-// Safety: `Table` is only accessed on the R thread (or in R idle tasks,
-// which also run on the R thread).
-unsafe impl Send for Table {}
+pub struct Table(RThreadSafe<RObject>);
 
 impl Table {
     pub fn new(data: RObject) -> Self {
-        Self(data)
+        Self(RThreadSafe::new(data))
     }
 
     pub fn get(&self) -> &RObject {
-        &self.0
+        self.0.get()
     }
 
     pub fn set(&mut self, data: RObject) {
-        self.0 = data;
+        self.0 = RThreadSafe::new(data);
+    }
+
+    /// Clone the table for use in an idle task. Must be called on the R thread.
+    pub fn clone_for_task(&self) -> Self {
+        Self(RThreadSafe::new(self.0.get().clone()))
     }
 }
--- a/crates/ark/src/thread.rs
+++ b/crates/ark/src/thread.rs
@@ -84,6 +84,15 @@ impl<T> Drop for RThreadSafe<T> {
             return;
         };
 
+        // In tests we're already on the R thread, so drop directly.
+        // Going through `spawn_interrupt` would call `block_on` which
+        // panics if we're already inside an executor (e.g. from
+        // `spawn_idle`'s test path).
+        if stdext::IS_TESTING {
+            drop(shelter);
+            return;
+        }
+
         let _span = tracing::trace_span!("async drop").entered();
 
         r_task::spawn_interrupt(async move || {


jmcphers (approved):
I didn't review this in detail, but the structure is sound. I do think it is useful to have _some_ async comms for cases wherein we don't actually need to talk to R at all and don't need to synchronize ourselves with the busy/idle groupings. But this pattern feels much better for things like the Data Explorer that are primarily interacting with R state.

crates/amalthea/src/language/shell_handler.rs
```diff
@@ -71,4 +82,35 @@ pub trait ShellHandler: Send {
     /// * `target` - The target name of the comm, such as `positron.variables`
     /// * `comm` - The comm channel to use to communicate with the frontend
     async fn handle_comm_open(&self, target: Comm, comm: CommSocket) -> crate::Result<bool>;
+
+    /// Handle an incoming comm message (RPC or data) synchronously on the
+    /// kernel's main thread. Return `CommHandled::Handled` if the message was
```
DavisVaughan: Do you really want to be talking about "kernel main thread" in amalthea docs?

crates/amalthea/src/socket/shell.rs
```diff
@@ -483,6 +484,13 @@ impl Shell {
...
 
+        // Try to dispatch the message to the new handler API
+        match shell_handler.handle_comm_msg(&msg.comm_id, &comm.comm_name, comm_msg.clone())? {
+            CommHandled::Handled => return Ok(()),
+            CommHandled::NotHandled => {},
+        }
+
+        // Fall back to old approach for compatibility while we migrate comms
         log::trace!("Sending message to comm '{}'", comm.comm_name);
         comm.incoming_tx.send(comm_msg).log_err();
```
DavisVaughan, line 494:
```suggestion
        // Try to dispatch the message to the new handler API
        match shell_handler.handle_comm_msg(&msg.comm_id, &comm.comm_name, comm_msg.clone())? {
            CommHandled::Handled => Ok(()),
            CommHandled::NotHandled => {
                // Fall back to old approach for compatibility while we migrate comms
                log::trace!("Sending message to comm '{}'", comm.comm_name);
                comm.incoming_tx.send(comm_msg).log_err();
                Ok(())
            },
        }
```

idk, seems like a clearer use of match?

crates/ark/src/console.rs
```diff
@@ -215,7 +217,7 @@ pub struct Console {
     session_mode: SessionMode,
 
     /// Channel used to send along messages relayed on the open comms.
-    comm_event_tx: Sender<CommEvent>,
+    pub(crate) comm_event_tx: Sender<CommEvent>,
```
DavisVaughan: I thought `get_comm_event_tx()` that doesn't expose this as `pub` was a nice abstraction :/

crates/ark/src/data_explorer/table.rs
```diff
@@ -1,50 +1,29 @@
...
+pub struct Table(RObject);
 
-impl Table {
-    pub fn new(data: RThreadSafe<RObject>) -> Self {
-        let table = Arc::new(Mutex::new(Some(data)));
-        Self { table }
-    }
+// Safety: `Table` is only accessed on the R thread (or in R idle tasks,
+// which also run on the R thread).
+unsafe impl Send for Table {}
```
DavisVaughan: I am very uneasy about this

We still ship a `Table` around via an `r_task::spawn_idle()` (as you mentioned).

I know that we:
- Send the task from the main R thread
- Pick the task up and run it on the main R thread

But I am still extremely nervous about providing anything outside of `RThreadSafe` that can send across threads. I just don't trust us to get it right every time.

I think I would prefer to keep this wrapped in `RThreadSafe`, because to me that is The Way to ship across threads, even if you end up running from the main R thread on both sides.

DavisVaughan: I would also like to note that this problem would very likely go away entirely if we had a variant of `r_task::spawn_idle()` that did not require a `Send` bound on `T`.

I believe we are stuck with that as long as we are using crossbeam channels, but everything happening here is all on one thread! The main R thread!

All we really want is to queue up a task _within the same thread_ so that read_console can run it at the next idle iteration. Something like `r_task::enque_idle()` maybe, just spitballing.

That shouldn't require a crossbeam channel ideally (although compatibility with a crossbeam `select!` would make it challenging probably). We maybe just need some `VecDeque` to push and pop from.

Then you should be able to ship a closure around, even if it has an `RObject` inside, without a `Send` bound.

But until then, I still like `RThreadSafe`

crates/ark/src/data_explorer/r_data_explorer.rs
```diff
@@ -163,215 +154,63 @@ pub struct RDataExplorer {
...
+        })
     }
 
     /// Check the environment bindings for updates to the underlying value
     ///
     /// Returns true if the update was processed; false if the binding has been
     /// removed and the data viewer should be closed.
-    fn update(&mut self) -> anyhow::Result<bool> {
+    fn update(&mut self, ctx: &CommHandlerContext) -> anyhow::Result<bool> {
         // No need to check for updates if we have no binding
```
DavisVaughan, line 197: Random thought. Would it be nice to have some kind of `assert_r_thread!()` macro we could put at the top of functions like this? Panic in debug mode and no-op in release mode? It would be self documenting and would help us with our invariants.

crates/ark/src/data_explorer/r_data_explorer.rs
```diff
@@ -581,9 +417,33 @@ impl RDataExplorer {
...
+        };
+        match self.update(ctx) {
+            Ok(true) => {},
+            Ok(false) => ctx.close_on_exit(),
+            Err(err) => {
+                log::error!("Error while checking environment for data viewer update: {err}");
+            },
+        }
+    }
+}
```
DavisVaughan, line 445: Neat!

crates/ark/src/console.rs
```diff
@@ -2086,6 +2104,29 @@ impl Console {
...
+                comm_id,
+                comm_name: _,
+                factory,
+                ctx,
+                done_tx,
+            } => {
+                let handler = (factory.0)();
+                self.comm_handle_open(comm_id, handler, ctx);
+                done_tx.send(()).log_err();
+            },
```
DavisVaughan: This one doesn't look to be used yet, right?

Maybe we remove it until we really need it? The factory stuff in particular leaves me feeling a little gross if we don't have an actual use case for it yet.

crates/ark/src/console.rs
```diff
@@ -2217,12 +2258,18 @@ impl Console {
...
+    /// unblocks Shell, which sends Idle.
+    fn prepare_execute_reply(
         iopub_tx: &Sender<IOPubMessage>,
         req: ActiveReadConsoleRequest,
         value: ConsoleValue,
+    ) -> (
+        amalthea::Result<ExecuteReply>,
+        Sender<amalthea::Result<ExecuteReply>>,
     ) {
         log::trace!("Completing execution after receiving prompt");
```
DavisVaughan: Returning `req.reply_tx` all feels a bit misleading. It feels like you've somehow "computed" the Sender to return the reply over here. That's not the case. You knew it in `req` all along.

You don't really need `req` in here at all. What if you just pass in `req.exec_count` as an argument instead, then at the call side you use `req.reply_tx` directly rather than making it a return value from here?

This is one of those places where I feel like we need to be fighting AI complexity with some extra human review focused on simplifiying the code.

crates/ark/src/console.rs
```diff
@@ -1187,8 +1193,6 @@ impl Console {
             self.refresh_lsp();
         }
```
DavisVaughan, line 1226: And `EVENTS` will go away completely after the variables pane migration?

Right now `EVENTS` is so misleading! It lives in lsp/ ! but we don't even use it there! 😬

crates/ark/src/console.rs
```diff
@@ -1485,27 +1489,41 @@ impl Console {
...
+        graphics_device::on_did_execute_request();
+
+        // Send execute result/error on IOPub, get back the prepared reply.
+        let (reply, reply_tx) = Self::prepare_execute_reply(&self.iopub_tx, req, value);
+
+        // Notify comm handlers about environment changes. This must happen
+        // after the execute result goes on IOPub but before the reply
+        // unblocks Shell (which sends Idle). This ensures side effects like
+        // data explorer updates and closes arrive within the Busy/Idle
+        // window of the execute request that caused them.
```
DavisVaughan: To make this more obvious, I vote we remove `iopub_tx` as an input to `prepare_execute_reply()` as well.

i.e. make `prepare_execute_reply()` a pure value function with no "sending" behavior. It should just return `reply` and `result`!

Then here you get to do all the sending, in the right order.

i.e. send out the `result` using `self.iopub_tx`, then follow that up with environment change notifications, then send out `reply`.

I really like that that would make all the "sending" behavior visible here, rather than hiding some away in `prepare_execute_reply()`

crates/ark/src/request.rs
```diff
@@ -49,9 +54,42 @@ pub fn debug_request_command(req: DebugRequest) -> String {
...
+
+    /// Register a new comm handler on the R thread (frontend-initiated comms).
+    /// Uses a factory closure so the handler (which may hold `RObject`s) is
+    /// created on the R thread rather than sent across threads.
+    CommOpen {
+        comm_id: String,
+        comm_name: String,
+        factory: CommHandlerFactory,
+        ctx: CommHandlerContext,
+        done_tx: Sender<()>,
```
DavisVaughan: This is what didn't look used to me, and the factory stuff just felt confusing if we don't have a use for it...

crates/ark/src/shell.rs
```diff
@@ -249,16 +244,65 @@ impl ShellHandler for Shell {
...
     }
+
+    fn handle_comm_msg(
+        &mut self,
+        comm_id: &str,
+        comm_name: &str,
+        msg: CommMsg,
+    ) -> amalthea::Result<CommHandled> {
+        match comm_name {
+            "positron.dataExplorer" => {
```
DavisVaughan: We have used the string `"positron.dataExplorer"` more times than I am comfortable with at this point.

Can we abstract this into a named static string that lives somewhere and reference it by name instead? A simple typo could screw it all up!

crates/ark/src/request.rs
DavisVaughan: Even if you think you'll use it for Variables, I'd be interested in delaying the addition of this to that PR so we can see / justify that we really do need this weird factory thing

crates/ark/tests/data_explorer_debug.rs
```diff
@@ -39,22 +39,20 @@ fn test_data_explorer_stable_on_frame_selection() {
     dap.evaluate(".positron_selected_frame", Some(outer_frame_id));
 
     // Data explorer should NOT have received any events
-    frontend.assert_no_data_explorer_events();
+    frontend.assert_iopub_empty();
```
DavisVaughan: Bad comment

crates/ark/src/console/console_comm.rs
```diff
@@ -0,0 +1,107 @@
...
+        for (comm_id, reg) in self.comms.iter_mut() {
+            reg.handler.handle_environment(event, &reg.ctx);
+            if reg.ctx.is_closed() {
+                closed_ids.push(comm_id.clone());
+            }
+        }
+        for comm_id in closed_ids {
+            if let Some(reg) = self.comms.remove(&comm_id) {
+                self.comm_notify_closed(&comm_id, &reg);
+            }
```
DavisVaughan: This whole `is_closed()` thing feels a bit wrong to me.

It feels like after any kind of generic message we have to check if the backend decided to close the comm? Like we do this here and in comm_handle_msg.

Should something else be handling this in a more consistent manner?

Otherwise it feels like if we add any other `comm_notify_*()` helper to this, then we are going to _also_ need to check is_closed there too, and that feels so easy to forget

crates/ark/src/comm_handler.rs
```diff
@@ -0,0 +1,157 @@
...
+}
+
+/// Why the environment changed.
+#[derive(Debug, Clone, Copy)]
+pub enum EnvironmentChanged {
+    /// A top-level execution completed (user code, debug eval, etc.).
+    Execution,
+    /// The user selected a different frame in the call stack during debugging.
+    FrameSelected,
+}
```
DavisVaughan, line 86: i like this framing!

crates/ark/src/console.rs
lionel-: You'd still need a pub crate if you call it from other files

DavisVaughan: Yes but you can't modify it from other files because you only get a reference, that's the benefit of the getter and the reason for all of the `get_` methods in here

lionel-: That makes sense, I've restored the method returning a reference, and as discussed on Slack used the submodule trick to remove `pub(crate)` for fields accessed from Console methods implemented in other files.

crates/ark/src/data_explorer/r_data_explorer.rs
lionel-: Not a bad idea but it feels superfluous to me, now that the architecture ensures that comms consistently run on the R thread.

crates/ark/src/console.rs
lionel-: That's coming up pretty soon.

crates/ark/src/console.rs
lionel-: yep

crates/ark/src/console/console_comm.rs
lionel-: I had the same feeling as you but couldn't find a better way to structure this.

crates/ark/src/request.rs
lionel-: ok let's see in the Variables PR how that comes along

crates/ark/src/data_explorer/table.rs
lionel-: yep all task spawning should eventually be done from the R thread, and then we use a simple queue data structure instead of the crossbeam channel.

I'm comfortable with the current setup but since it makes you uneasy I'll restore the RThreadSafe.

lionel-: Just a heads up that this forced me to add some IS_TESTING weirdness to RThreadSafe's drop method

PR #1076, opened 2026-03-02
Test on r-devel

lionel-:
Closes #1066 

To track upstream breaking changes to critical components of our infrastructure:

-    The way we integrate with ReadConsole (https://github.com/posit-dev/ark/pull/960) and more generally the frontend API
-   Graphics devices API

Second commit fixes sensitivity of a test on HTML produced by R:

```
        FAIL [   0.834s] (296/861) ark lsp::help::tests::test_help_from_function
  stdout ───

    running 1 test
    test lsp::help::tests::test_help_from_function ... FAILED

    failures:

    failures:
        lsp::help::tests::test_help_from_function

    test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 490 filtered out; finished in 0.82s

  stderr ───

    thread 'lsp::help::tests::test_help_from_function' (22703) panicked at crates/ark/src/lsp/help.rs:384:38:
    called `Option::unwrap()` on a `None` value
```

PR #1083, opened 2026-03-04
Compatibility with R 4.6

lionel-:
Branched from #1076 

R core _just_ removed access to a bunch of APIs, just in time to put our new r-devel build (#1076) to good use.

Screenshot 2026-03-04 at 18 11 17

The main change is that we no longer unlock our private namespaces and environments. Instead we wait until they are fully initialised to lock them. In debug builds we never lock them to allow hot-reloading when a .R file changes.

Otherwise we follow what we've been doing in rlang, lobstr, vctrs, and other packages regarding compatibility with C API. E.g. in `zap_srcref()` or `obj_size()`. We use new APIs (with polyfills for old R compats) or new workaround approaches.

I've also proactively removed usage of `ATTRIB()` and `SET_ATTRIB()` as well (except in the altrep utils) since these have been promoted from NOTE to WARNING on CRAN.

### DIFF
--- a/crates/harp/src/r.rs
+++ b/crates/harp/src/r.rs
@@ -26,11 +26,17 @@ pub fn fn_env(x: SEXP) -> SEXP {
     }
 }
 
-pub unsafe fn new_function(formals: SEXP, body: SEXP, env: SEXP) -> SEXP {
-    if libr::has::R_mkClosure() {
-        libr::R_mkClosure(formals, body, env)
-    } else {
-        compat::alloc_closure(formals, body, env)
+pub fn new_function(formals: SEXP, body: SEXP, env: SEXP) -> SEXP {
+    unsafe {
+        if libr::has::R_mkClosure() {
+            libr::R_mkClosure(formals, body, env)
+        } else {
+            let out = libr::Rf_allocSExp(libr::CLOSXP);
+            libr::SET_FORMALS(out, formals);
+            libr::SET_BODY(out, body);
+            libr::SET_CLOENV(out, env);
+            out
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


crates/harp/src/utils.rs
```diff
@@ -126,11 +126,11 @@ pub fn r_is_altrep(object: SEXP) -> bool {
 }
 
 pub fn r_is_object(object: SEXP) -> bool {
-    unsafe { libr::OBJECT(object) != 0 }
+    unsafe { libr::Rf_isObject(object) != 0 }
 }
 
 pub fn r_is_s4(object: SEXP) -> bool {
-    unsafe { libr::IS_S4_OBJECT(object) != 0 }
+    unsafe { libr::Rf_isS4(object) != 0 }
```
DavisVaughan, line 134: Can you please go through `r.rs` and remove anything we no longer need? Like I see `OBJECT` and `IS_S4_OBJECT` are still in there, and it would be nice to remove them so we don't accidentally use them again.

crates/ark/src/srcref.rs
```diff
@@ -160,10 +161,8 @@ fn generate_source(
     // Inject source references in functions. This is slightly unsafe but we
     // couldn't think of a dire failure mode.
     unsafe {
-        // First replace the body which contains expressions tagged with srcrefs
-        // such as calls to `{`. Compiled functions are a little more tricky.
```
DavisVaughan, line 164: Outdated? Dropped?

crates/harp/src/r.rs
```diff
@@ -0,0 +1,76 @@
...
+    unsafe {
+        libr::Rf_defineVar(sym, value, env);
+    }
+    if locked {
+        env_binding_lock(env, sym);
+    }
+}
+
+/// Creates a closure.
+pub unsafe fn new_function(formals: SEXP, body: SEXP, env: SEXP) -> SEXP {
```
DavisVaughan: do you want it to be unsafe?

crates/harp/src/r.rs
```diff
@@ -0,0 +1,76 @@
...
+mod compat {
+    use libr::SEXP;
+
+    pub unsafe fn alloc_closure(formals: SEXP, body: SEXP, env: SEXP) -> SEXP {
+        let out = libr::Rf_allocSExp(libr::CLOSXP);
+        libr::SET_FORMALS(out, formals);
+        libr::SET_BODY(out, body);
+        libr::SET_CLOENV(out, env);
+        out
+    }
```
DavisVaughan: it feels more useful to me to just inline this. we'd only ever use it in this 1 place probably?

crates/harp/src/r.rs
```diff
@@ -54,15 +66,60 @@ pub fn env_bind_force(env: SEXP, sym: SEXP, value: SEXP) {
...
+        }
+    }
+}
+
+/// Copies all attributes from `src` to `dst`.
+pub fn attrib_poke_from(dst: SEXP, src: SEXP) {
+    attrib_for_each(src, |tag, val| unsafe {
+        libr::Rf_setAttrib(dst, tag, val);
+    });
+}
```
DavisVaughan: Why not `DUPLICATE_ATTRIB` or `SHALLOW_DUPLICATE_ATTRIB`?

crates/harp/src/r.rs
lionel-: nope!

crates/harp/src/r.rs
lionel-: Good catch, should be shallow-duplicate as in https://github.com/r-lib/rlang/pull/1873/changes#diff-4abaa2fa8b7c6338784d3376cb679dcf2a5ad2338b416544767314699e3f9089

PR #1085, opened 2026-03-04
Add code evaluation API

jmcphers:
This change adds an API for evaluating code to Ark.

Most of the change is actually just getting smarter about type coercion, so that named character vectors get serialized into JSON objects instead of losing their names. 

Part of https://github.com/posit-dev/positron/issues/7112.

### DIFF
--- a/crates/ark/tests/evaluate-code.rs
+++ b/crates/ark/tests/evaluate-code.rs
@@ -23,11 +23,11 @@ fn evaluate_code(frontend: &DummyArkFrontend, comm_id: &str, code: &str) -> UiBa
     frontend.send_shell_comm_msg(String::from(comm_id), data);
 
     // The shell routes the message to the UI comm thread and goes busy/idle.
-    // The RPC reply arrives asynchronously from the UI comm thread after that.
+    // The RPC reply is sent from the UI comm thread and can arrive on IOPub
+    // either before or after the shell's Idle status.
     frontend.recv_iopub_busy();
-    frontend.recv_iopub_idle();
 
-    let reply = frontend.recv_iopub_comm_msg();
+    let reply = frontend.recv_iopub_comm_msg_and_idle();
     assert_eq!(reply.comm_id, comm_id);
 
     serde_json::from_value::<UiBackendReply>(reply.data).unwrap()
--- a/crates/ark_test/src/dummy_frontend.rs
+++ b/crates/ark_test/src/dummy_frontend.rs
@@ -636,6 +636,33 @@ impl DummyArkFrontend {
         }
     }
 
+    /// Receive a CommMsg and Idle from IOPub in either order.
+    ///
+    /// Some comm RPC replies race with the shell's Idle status because the
+    /// reply is sent from a separate thread (e.g. the UI comm thread). This
+    /// helper accepts both orderings and returns the CommMsg content.
+    #[track_caller]
+    pub fn recv_iopub_comm_msg_and_idle(&self) -> amalthea::wire::comm_msg::CommWireMsg {
+        let first = self.recv_iopub_next();
+        let second = self.recv_iopub_next();
+
+        let (comm_msg, idle) = match (first, second) {
+            (Message::CommMsg(comm), Message::Status(status)) => (comm, status),
+            (Message::Status(status), Message::CommMsg(comm)) => (comm, status),
+            (a, b) => panic!("Expected CommMsg and Idle in either order, got {:?} and {:?}", a, b),
+        };
+
+        assert_eq!(
+            idle.content.execution_state,
+            amalthea::wire::status::ExecutionState::Idle,
+            "Expected Idle status"
+        );
+
+        self.flush_streams_at_boundary();
+
+        comm_msg.content
+    }
+
     /// Receive from IOPub and assert CommOpen message.
     /// Automatically skips any Stream messages.
     #[track_caller]


crates/harp/src/json.rs
```diff
@@ -163,17 +175,41 @@ impl TryFrom<RObject> for Value {
                     Ok(Value::String(str))
                 },
 
-                // With multiple values, convert to a string array
+                // With multiple values, convert to a string array or object
```
lionel-: Although this is defensive regarding the presence of empty names, I'm a bit concerned that simple changes to the R representation (a name slips in) results in a completely different json representation.

Also since the mapping from named vector to json map is lossy, this could lead to surprising behaviour or bugs in extension code.

I see that this behaviour already exists for lists though, but I wonder if we should opt for a stricter approach before we get too many downstream dependencies.

crates/harp/src/json.rs
```diff
@@ -436,6 +455,27 @@ mod tests {
         })
     }
 
+    #[test]
+    #[allow(non_snake_case)]
+    fn test_json_named_character_vectors() {
```
lionel-: Missing tests:

- Empty character vector (empty array)
- Character vector of size one (still an array, not a scalar)
- Character vector with `NA` names (which core R methods sometimes introduce)
- Character vector with duplicate names, should test that last one wins.

crates/ark/src/ui/ui.rs
```diff
@@ -233,6 +236,31 @@ impl UiComm {
...
+            Value::try_from(evaluated)
+        });
+
+        match result {
+            Ok(value) => Ok(UiBackendReply::EvaluateCodeReply(value)),
+            Err(err) => {
+                let message = match &err {
+                    harp::Error::TryCatchError { message, .. } => message.clone(),
+                    harp::Error::ParseError { message, .. } => message.clone(),
+                    harp::Error::ParseSyntaxError { message } => message.clone(),
```
lionel-: I think the clones can be avoided if you match on `err` instead of `&err`

crates/harp/src/json.rs
```diff
@@ -53,6 +53,18 @@ use crate::object::RObject;
...
 /// - Named lists with duplicate keys have the values combined into an array
 ///   - e.g.: list(a = 1L, a = 2L, a = 3L) -> {"a": [1, 2, 3]}
+/// Returns the names of an R object, or `None` if the object has no names or
+/// all names are empty/NA. This is used to decide whether a vector should be
+/// serialized as a JSON object (when names are present) or as an array.
+fn non_empty_names(obj: &RObject) -> Option<Vec<Option<String>>> {
+    let names = obj.names()?;
+    let all_empty = names.iter().all(|name| match name {
+        Some(name) => name.is_empty(),
+        None => true,
```
lionel-: `NA` names become `None` here IIUC? This would be the expected behaviour, just checking (and should also be tested).

crates/amalthea/src/comm/ui_comm.rs
```diff
@@ -437,6 +450,9 @@ pub enum UiBackendReply {
 	/// The method result
 	CallMethodReply(CallMethodResult),
 
+	/// The result of evaluating the code
+	EvaluateCodeReply(serde_json::Value),
```
lionel-: By the way there's one command that I was using all the time in Emacs and that I'd like to implement in Positron. It's a version of "evaluate statement" that inserts the output of the evaluation below the statement, with #> prefix.

IIUC it's not currently possible to implement this from an extension because the evaluation needs to happen in the background (silently) but the output should still be captured and returned to the caller. It looks like that use case is not covered by https://github.com/posit-dev/positron/issues/7112.

To support this case, perhaps we could consider returning a structured value with `result` and `output` fields. The `output` field could be opt-in by language in case the kernel can't easily support it. See https://github.com/posit-dev/ark/pull/1064 for an example of how to evaluate with output and condition capture.

In any case, wrapping the return value in a struct would be more future-proof, allowing new fields to be added as needed.

crates/ark/src/ui/ui.rs
jmcphers: you're right!

crates/ark/src/ui/ui.rs
```diff
@@ -243,12 +244,38 @@ impl UiComm {
         log::trace!("Evaluating code: {}", params.code);
 
         let result = r_task(|| {
-            let evaluated = parse_eval_global(&params.code)?;
-            Value::try_from(evaluated)
+            // Set up a raw connection to capture printed output via sink().
```
lionel-: Can we use `Console::start_capture()` instead?

Warnings are a complication to be aware of, as by defaults they are emitted when R goes back to idle.

I realised in https://github.com/posit-dev/ark/pull/1064/changes#r2896192009 that the warning capture I've implemented for conditional breakpoints can be simplified by having the output capture guard set `options(warn = 1)` while output is being captured. This causes warnings to be emitted immediately instead of being buffered by R. So if you switch to the console capture mechanism, it should take care of warnings once that change is merged.

Warnings will behave more like messages and regular output in that they will be interleaved in the captured output, but I think that's acceptable.

jmcphers: Impl in https://github.com/posit-dev/ark/pull/1085/commits/9ccdd0dacaed1a32d6ca3dd8ddeaf684ad2d27b5, let me know what you think. The console isn't available in unit tests, so I added some tests closer to the integration layer (i.e. using a "UI Comm").

crates/ark/src/ui/ui.rs
```diff
@@ -233,6 +239,50 @@ impl UiComm {
...
+            } else {
+                ConsoleOutputCapture::dummy()
+            };
+
+            // Evaluate the user's code
+            let eval_result = parse_eval_global(&params.code);
+
+            // Take captured output before dropping the capture guard
+            let output = capture.take();
+            drop(capture);
```
DavisVaughan, line 260: @lionel- it might be nice to have `capture.consume()` which returns a `String` like `take()` but also consumes `self` and drops it - like a one shot capture so you don't have to think about dropping as much.

lionel-: hmm, or `capture.into_string()`?

The explicit drop is a clearer signal of the side effect though.

lionel-:
If you'd like you can rebase on main to make a test with warning capture. The Console output capture mechanism now automatically sets the option to emit warnings immediately instead of letting R buffer them.

PR #1086, opened 2026-03-05
Implement log breakpoints

lionel-:
Branched from #1064.

Adds support for DAP log breakpoints (logpoints). When a breakpoint has a log message template instead of (or in addition to) a condition, it evaluates the template at hit time, prints the result to the debug console, and does not stop execution. This is the DAP "logpoint" feature.

The log message template uses `{expression}` placeholders for interpolation, as mandated by the DAP spec. On the R side, interpolation is handled by `glue::glue()` if we detect a `{` character. This fails actionnably (message in Console output) if glue is not installed. If there is no `{` character, the template is returned as is, allowing simple messages to work without the glue dependency.

Log breakpoints respect conditions: if a breakpoint has both a condition and a log message, the condition is evaluated first and the log message only fires when the condition is met. This naturally extends to hit-count breakpoints in the future (see #1087).

Output is emitted through the same `emit_breakpoint_output` path as conditional breakpoints, using the ```` ```breakpoint ```` fenced block with a clickable file link. When there's no condition, the condition header line (`#> ...`) is omitted and only the log message content is shown.

Screenshot 2026-03-05 at 10 43 18

With an error we get:

Screenshot 2026-03-05 at 10 55 23

### QA Notes

Tested on the backend side.

To test manually:

```r
foo <- function(x) {
  y <- x * 2
  y + 1
}
for (i in 1:5) {
  foo(i)
}
```

1. Set a log breakpoint on `y <- x * 2` with message `x is {x}, y will be {x * 2}`. Run the loop — should see 5 log messages in the console without stopping.
2. Add a condition `x > 3` to the same log breakpoint. Run the loop — should only see messages for `x = 4` and `x = 5`.
3. Try a log message referencing a nonexistent variable like `{nonexistent}`: should show an inline error without stopping.
4. Without glue installed: a log message without glue syntax like `checkpoint reached` should work. A log message with glue syntax should fail actionably (error in console output).

### DIFF
--- a/crates/ark/tests/dap_breakpoints.rs
+++ b/crates/ark/tests/dap_breakpoints.rs
@@ -215,7 +215,11 @@ fn test_dap_breakpoints_unsaved_file_unverified() {
     let frontend = DummyArkFrontend::lock();
     let mut dap = frontend.start_dap();
 
-    let breakpoints = dap.set_breakpoints("/tmp/nonexistent_file_for_test.R", &[1, 5]);
+    let path = std::env::temp_dir()
+        .join("nonexistent_file_for_test.R")
+        .to_string_lossy()
+        .replace('\\', "/");
+    let breakpoints = dap.set_breakpoints(&path, &[1, 5]);
     assert_eq!(breakpoints.len(), 2);
 
     assert!(!breakpoints[0].verified);


crates/ark/src/dap/dap.rs
```diff
@@ -74,6 +74,10 @@ pub struct Breakpoint {
     /// We keep this as a string instead of a parsed expression for safety,
```
DavisVaughan, line 74: Something seems to not work right in untitled files. It works if I actually save the file. The main annoying thing is that it _looks_ active

https://github.com/user-attachments/assets/21590868-ed76-4fe5-b6bd-f48756d73228

crates/ark/src/modules/positron/eval.R
```diff
@@ -32,3 +33,14 @@ initialize_eval <- function() {
...
+
+# Interpolate a DAP log message template using `glue`. We call
+# unconditionally so that a missing glue package surfaces as an
+# actionable error rather than silently returning the raw template.
+#' @export
+.ark_eval_log_message <- function(template, env = parent.frame()) {
+    if (!grepl("{", template, fixed = TRUE)) {
+        return(template)
+    }
+    as.character(glue::glue(template, .envir = env))
```
DavisVaughan, line 24: Would it be more user friendly to detect a lack of `glue` and give a more informative user message like

```
Can't interpolate breakpoint log messages without the glue package.
```

?

crates/ark/src/dap/dap.rs
```diff
@@ -74,6 +74,10 @@ pub struct Breakpoint {
...
     /// since breakpoint state is also inspected by the DAP server thread.
     pub condition: Option<String>,
+    /// Optional log message template (DAP "logpoint"). When set, the
+    /// breakpoint evaluates `{expression}` placeholders in the template,
+    /// prints the result to the debug console, and does not stop execution.
```
DavisVaughan:
```suggestion
    /// breakpoint interpolates `{expression}` placeholders in the template using glue,
    /// prints the result to the debug console, and does not stop execution.
```

crates/ark/src/dap/dap.rs
```diff
@@ -647,10 +652,9 @@ impl Dap {
         Ok(obj.get().sexp)
     }
 
-    pub(crate) fn breakpoint_condition(&self, uri: &Url, id: i64) -> Option<(u32, &str)> {
+    pub(crate) fn get_breakpoint(&self, uri: &Url, id: i64) -> Option<&Breakpoint> {
         let (_, breakpoints) = self.breakpoints.get(uri)?;
-        let bp = breakpoints.iter().find(|bp| bp.id == id)?;
-        Some((bp.line, bp.condition.as_deref()?))
+        breakpoints.iter().find(|bp| bp.id == id)
```
DavisVaughan: That seems much nicer!

DavisVaughan (approved):

crates/ark/src/dap/dap.rs
lionel-: That's a general pre-existing issue with breakpoints due to the default way breakpoints look like in the frontend. We don't get breakpoint events with untitiled files.

PR #1087, opened 2026-03-05
Implement hit count breakpoints

lionel-:
Branched from #1086

Adds support for DAP hit count breakpoints. When a breakpoint has a `hitCondition`, the breakpoint location must be reached at least that many times before it fires. The value is interpreted as a plain integer with `>=` semantics. For more complex hit patterns, users can combine hit counts with conditional breakpoints.

Hit counts are reset when `ReadConsole` returns to a top-level (non-browser) prompt, meaning the execution that may have hit breakpoints is complete. This way counts survive across continue/step within a debug session but start fresh for new sessions.

Hit count integrates with the existing breakpoint feature chain. Hit counts with conditions require both to be satisfied, and hit counts with log messages only emit output once the threshold is reached. If the hit count can't be parsed, the error message is propagated to the Console in breakpoint fences with a clickable link.

### QA Notes

Tested on the backend side.

Hit counts should behave well on their own or in combination with logpoints and conditional breakpoints. Parse errors appear in the console.

Screenshot 2026-03-05 at 12 05 36

Screenshot 2026-03-05 at 12 05 24

### DIFF
--- a/crates/ark/src/console_debug.rs
+++ b/crates/ark/src/console_debug.rs
@@ -537,7 +537,7 @@ pub unsafe extern "C-unwind" fn ps_handle_breakpoint(
     let id: i64 = id.parse()?;
 
     let console = Console::get_mut();
-    let mut dap = console.debug_dap.lock().unwrap();
+    let dap = console.debug_dap.lock().unwrap();
 
     let enabled = dap.is_breakpoint_enabled(&uri, id);
     let bp = dap.get_breakpoint(&uri, id);
@@ -560,30 +560,12 @@ pub unsafe extern "C-unwind" fn ps_handle_breakpoint(
         return Ok(RObject::from(false).sexp);
     }
 
-    let hit_count = dap.increment_hit_count(&uri, id);
-
     // Must drop before calling back into R to avoid deadlock
     drop(dap);
 
-    if let Some(ref hit_condition) = hit_condition {
-        match hit_condition.trim().parse::<u64>() {
-            Ok(threshold) => {
-                if hit_count < threshold {
-                    return Ok(RObject::from(false).sexp);
-                }
-            },
-            Err(err) => {
-                emit_breakpoint_block(
-                    &uri,
-                    bp_line,
-                    &format!("Expected a positive integer: {err}"),
-                );
-            },
-        }
-    }
-
     // Evaluate condition first as it applies to all breakpoints, including log
-    // and hit-count breakpoints
+    // and hit-count breakpoints. Per the DAP spec, `hitCondition` should only
+    // be evaluated (and the hit count incremented) if the `condition` is met.
     let should_break = match &condition {
         None => true,
         Some(condition) => {
@@ -604,6 +586,27 @@ pub unsafe extern "C-unwind" fn ps_handle_breakpoint(
 
     if !should_break {
         return Ok(RObject::from(false).sexp);
+    }
+
+    if let Some(ref hit_condition) = hit_condition {
+        match hit_condition.trim().parse::<u64>() {
+            Ok(threshold) => {
+                let mut dap = Console::get_mut().debug_dap.lock().unwrap();
+                let hit_count = dap.increment_hit_count(&uri, id);
+                drop(dap);
+
+                if hit_count < threshold {
+                    return Ok(RObject::from(false).sexp);
+                }
+            },
+            Err(err) => {
+                emit_breakpoint_block(
+                    &uri,
+                    bp_line,
+                    &format!("Error: Expected a positive integer, {err}"),
+                );
+            },
+        }
     }
 
     // Log breakpoints evaluate the template and never stop
@@ -1084,6 +1087,22 @@ mod tests {
         insta::assert_snapshot!(result.unwrap().replace(&link, "<test.R#3>"), @r"
         ```breakpoint <test.R#3>
         Error: object 'z' not found
+        ```
+        ");
+    }
+
+    #[test]
+    fn test_format_breakpoint_block_hit_condition_error() {
+        let uri = test_uri("test.R");
+        let result = format_breakpoint_block(
+            &uri,
+            17,
+            "Error: Expected a positive integer, invalid digit found in string",
+        );
+        let link = breakpoint_label(&uri, 17);
+        insta::assert_snapshot!(result.unwrap().replace(&link, "<test.R#18>"), @r"
+        ```breakpoint <test.R#18>
+        Error: Expected a positive integer, invalid digit found in string
         ```
         ");
     }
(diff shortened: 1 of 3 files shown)


crates/ark/src/console.rs
```diff
@@ -1220,6 +1220,7 @@ impl Console {
         // here, but only containing high-level information such as `search()`
         // contents and `ls(rho)`.
         if !self.debug_is_debugging && !matches!(info.kind, PromptKind::InputRequest) {
+            self.debug_dap.lock().unwrap().reset();
```
DavisVaughan, line 1237: So we do this every time we come through read-console at top level?

Is there any way to only do it after we exit the debugger?

Just curious. I know it's probably not performance sensitive at all right now, I'm just wondering if we can only call this when we really need it.

crates/ark/src/console_debug.rs
```diff
@@ -538,23 +538,53 @@ pub unsafe extern "C-unwind" fn ps_should_break(
...
+         condition: {condition:?}, \
+         log_message: {log_message:?}",
+        hit_count = bp.map_or(0, |bp| bp.hit_count)
+    );
 
     if !enabled {
         return Ok(RObject::from(false).sexp);
     }
 
+    let hit_count = dap.increment_hit_count(&uri, id);
```
DavisVaughan: Just a random thought that this function is called `ps_should_break` and that makes me think its a simple yes/no kind of thing, but man it's doing way more now! I wonder if there is a better name we can come up with.

crates/ark/src/dap/dap_server.rs
```diff
@@ -497,6 +498,8 @@ impl<R: Read, W: Write> DapServer<R, W> {
                         injected: false,
                         condition: bp.condition.clone(),
                         log_message: bp.log_message.clone(),
+                        hit_condition: bp.hit_condition.clone(),
```
DavisVaughan, line 516: I would not be opposed to parsing `hit_condition` as an integer here so you don't have to do it at each should-break step

crates/ark/src/dap/dap_server.rs
```diff
@@ -497,6 +498,8 @@ impl<R: Read, W: Write> DapServer<R, W> {
                         injected: false,
                         condition: bp.condition.clone(),
                         log_message: bp.log_message.clone(),
```
DavisVaughan, line 515: I'd like to call out that the docs suggest

```
The expression that controls how many hits of the breakpoint are ignored.
```

To me that means if you set `hit_condition = 2`, then 2 hits should be ignored, and only on the 3rd does it stop.

But you said its a `>=` implementation, and that seems to go against that?

crates/ark/src/console_debug.rs
```diff
@@ -538,23 +538,53 @@ pub unsafe extern "C-unwind" fn ps_should_break(
...
+
     // Must drop before calling back into R to avoid deadlock
     drop(dap);
 
+    if let Some(ref hit_condition) = hit_condition {
+        match hit_condition.trim().parse::<u64>() {
+            Ok(threshold) => {
+                if hit_count < threshold {
```
DavisVaughan, line 598:
```suggestion
        match hit_condition.trim().parse::<u64>() {
            Ok(hit_condition) => {
                if hit_count < hit_condition {
```

Super nit, but using a consistent name feels easier to me to read

crates/ark/src/console_debug.rs
```diff
@@ -538,23 +538,53 @@ pub unsafe extern "C-unwind" fn ps_should_break(
...
     if !enabled {
         return Ok(RObject::from(false).sexp);
     }
 
+    let hit_count = dap.increment_hit_count(&uri, id);
+
     // Must drop before calling back into R to avoid deadlock
     drop(dap);
 
+    if let Some(ref hit_condition) = hit_condition {
```
DavisVaughan, line 591: Ooooh I believe you have an ordering error!

According to the docs

```
   * If both this property and `condition` are specified, `hitCondition` should
   * be evaluated only if the `condition` is met, and the debug adapter should
   * stop only if both conditions are met.
```

So IIUC, this whole section should come _after_ your `should_break` early exit!

i.e. if my `condition` is `x > 5` and then i have a `hitCondition` of 2, then when `x <= 5` it should not be incrementing the `hitCondition`!

This would make for a good test!

crates/ark/src/console_debug.rs
```diff
@@ -538,23 +538,53 @@ pub unsafe extern "C-unwind" fn ps_should_break(
...
+        match hit_condition.trim().parse::<u64>() {
```
DavisVaughan, line 592: I also want to call out that I'm not sure that interpreting this as a simple number is how most people implement hit conditions

https://github.com/go-delve/delve/pull/2490/changes#diff-d8611b2d0b4b5d3b08ffb7b8e3f913420120a431d6dfd81c6f2a8221065bdf70R225-R238

It looks like a common way to do this is to allow the user to supply a hit condition of `== 5`, where you fill in the LHS of the binary expression there with the current `hitCount`.

Then users could supply `> 5` as well, for "break every time after 5 hits"

You'd support a limited set of options, like `==`, `>`, `>=`, `<`, `<=`, and that would maybe be enough.

DavisVaughan: Then again, some people do what you did

https://github.com/mfussenegger/nvim-dap/blob/a9d8cb68ee7184111dc66156c4a2ebabfbe01bc5/doc/dap.txt#L808

lionel-: yep the agent initially implemented a small parser along these lines, but that seemed a bit overkill to me. I thought we'd start with a simple `>=` interpretation.

crates/ark/src/console_debug.rs
lionel-: Good catch!

crates/ark/src/console.rs
lionel-: This is the right place, for this purpose we only know a debug session has finished when we come back to a non-debug prompt.

crates/ark/src/console_debug.rs
lionel-: Good point, I had the same feeling. It's now `handle_breakpoint()`.

crates/ark/src/dap/dap_server.rs
lionel-: That'd be less user friendly because then users don't get parse error messages in a breakpoint block in the Console.

crates/ark/src/dap/dap_server.rs
lionel-: Adapters generally use `>=` or `==` (and some don't allow simple numbers). Since I've opted for requiring numbers, it makes more sense to go with the former. That's also what's most generally useful.

PR #1088, opened 2026-03-06
Support `positron.session_init` and `positron.session_reconnect` hooks

### LINKED ISSUE #697: ark: Support a session init hook
Follow-up work related to posit-dev/positron#2070 and renv support.

RStudio supports a session init hook that runs after R is fully initialized:

https://github.com/rstudio/rstudio/blob/8ebe5431ca2fb3c9719b70d4fbc45201d226c79c/src/cpp/session/modules/SessionRHooks.hpp#L19

https://github.com/rstudio/rstudio/blob/8ebe5431ca2fb3c9719b70d4fbc45201d226c79c/src/cpp/session/SessionMain.cpp#L857

https://github.com/rstudio/rstudio/blob/8ebe5431ca2fb3c9719b70d4fbc45201d226c79c/src/cpp/session/modules/SessionRHooks.hpp#L36-L39

https://github.com/rstudio/rstudio/blob/8ebe5431ca2fb3c9719b70d4fbc45201d226c79c/src/cpp/session/modules/SessionRHooks.R#L16-L31

---

Good resource on R startup: https://rstats.wtf/r-startup

---

People typically use this sort of hook to schedule startup-ish code that makes use of, for example, `readline()`, `menu()` or rstudioapi. Our immediate use case here in Positron is related to renv, where it would be nice to schedule the "hey you need to do `renv::restore()`; shall we do it? interaction for execution after R has been fully stood up.

---

GitHub search for uses of `rstudio.sessionInit`:

<https://github.com/search?q=rstudio.sessionInit&type=code>

jennybc:
Addresses https://github.com/posit-dev/positron/pull/9763
Closes #697

Companion PR in Positron: https://github.com/posit-dev/positron/pull/12312 (see there for full context, background, and QA instructions)

## Summary

Adds R session init hooks that fire after the frontend is fully connected, so that rstudioapi functions (including those that make RPCs back to the frontend, like `getActiveProject()`) work from hook code.

### Hook design (two hooks)

- **`positron.session_init(start_type)`** -- fires on new sessions and restarts
  - `start_type = "new"` for new sessions
  - `start_type = "restart"` for restarted sessions
- **`positron.session_reconnect()`** -- fires on reconnects only (e.g. window reload)

These are analogous to RStudio's [`rstudio.sessionInit`](https://rstudio.github.io/rstudioapi/articles/r-session.html). Users register hooks in `.Rprofile` via `setHook()`.

### The originator problem and fix

rstudioapi functions like `getActiveProject()` make synchronous RPCs back to the frontend. These RPCs require a Jupyter message originator so replies can be properly parented. Previously, only `ExecuteRequest` handlers set an originator -- `comm_msg` handlers did not.

Since the `frontend_ready` signal arrives as a `comm_msg` on the UI comm, hooks fired from that context couldn't make frontend RPCs. This PR threads the originator from the `comm_msg` Jupyter message through to the R thread:

- **amalthea**: `handle_comm_msg` trait method gains an `originator` parameter; `shell.rs` creates `Originator` from the `comm_msg` request header
- **ark**: `KernelRequest::CommMsg` carries the originator; `Console` temporarily sets `comm_msg_originator` during `comm_handle_msg`; `call_frontend_method()` falls back to `comm_msg_originator` when there is no `active_request`

### Other changes

- **`comm_open` data**: `handle_comm_open` trait method gains a `data` parameter. The UI comm deserializes this into a typed `UiCommOpenData` struct and reads `console_width` from it to set immediately (replacing the old `setConsoleWidth` RPC from `positron-r`).
- **`frontend_ready` handler**: Dispatches to `.ps.run_session_init_hooks(start_type)` or `.ps.run_session_reconnect_hooks()` based on `start_type`.
- **`hooks.R`**: Two new internal functions with tryCatch error isolation per hook.

### DIFF
--- a/crates/ark/tests/kernel-hooks-session.rs
+++ b/crates/ark/tests/kernel-hooks-session.rs
@@ -1,0 +1,127 @@
+//
+// kernel-hooks-session.rs
+//
+// Copyright (C) 2026 by Posit Software, PBC
+//
+//
+
+use amalthea::comm::ui_comm::UiBackendReply;
+use amalthea::fixtures::dummy_frontend::ExecuteRequestOptions;
+use ark_test::DummyArkFrontend;
+
+/// Helper to execute R code and consume the full busy/idle window including
+/// UI comm busy/prompt_state events.
+fn execute(frontend: &DummyArkFrontend, comm_id: &str, code: &str) {
+    frontend.send_execute_request(code, ExecuteRequestOptions::default());
+    frontend.recv_iopub_busy();
+    frontend.recv_iopub_execute_input();
+    frontend.recv_ui_busy(comm_id, true);
+    frontend.recv_ui_busy(comm_id, false);
+    frontend.recv_ui_prompt_state(comm_id);
+    frontend.recv_iopub_idle();
+    frontend.recv_shell_execute_reply();
+}
+
+/// A "new" session fires session_init hooks with start_type = "new".
+///
+/// The hook calls `rstudioapi::navigateToFile()`, a fire-and-forget event.
+/// We can't verify the frontend acts on it, but we can verify the
+/// `open_editor` message arrives on the UI comm.
+#[test]
+fn test_session_init_hook_new() {
+    let frontend = DummyArkFrontend::lock();
+    let comm_id = frontend.open_ui_comm();
+
+    // Create a temp file so normalizePath() inside navigateToFile() succeeds.
+    let file = tempfile::NamedTempFile::new().unwrap();
+    let path = file.path().to_str().unwrap().replace('\\', "/");
+
+    execute(
+        &frontend,
+        &comm_id,
+        &format!(
+            "setHook('positron.session_init', function(start_type) rstudioapi::navigateToFile('{path}'))"
+        ),
+    );
+
+    let data = serde_json::json!({
+        "method": "frontend_ready",
+        "params": { "start_type": "new" },
+        "id": "frontend-ready-rpc"
+    });
+    frontend.send_shell_comm_msg(String::from(&comm_id), data);
+    frontend.recv_iopub_busy();
+
+    // The hook sends an open_editor event to the frontend
+    let event = frontend.recv_iopub_comm_msg();
+    assert_eq!(event.comm_id, comm_id);
+    assert_eq!(
+        event.data.get("method").and_then(|v| v.as_str()),
+        Some("open_editor")
+    );
+
+    let reply = frontend.recv_iopub_comm_msg();
+    assert_eq!(reply.comm_id, comm_id);
+    let reply = serde_json::from_value::<UiBackendReply>(reply.data).unwrap();
+    assert_eq!(reply, UiBackendReply::FrontendReadyReply());
+    frontend.recv_iopub_idle();
+}
+
+/// A "restart" session fires session_init hooks with start_type = "restart".
+#[test]
+fn test_session_init_hook_restart() {
+    let frontend = DummyArkFrontend::lock();
+    let comm_id = frontend.open_ui_comm();
+
+    execute(
+        &frontend,
+        &comm_id,
+        "setHook('positron.session_init', function(start_type) cat(start_type))",
+    );
+
+    let data = serde_json::json!({
+        "method": "frontend_ready",
+        "params": { "start_type": "restart" },
+        "id": "frontend-ready-rpc"
+    });
+    frontend.send_shell_comm_msg(String::from(&comm_id), data);
+    frontend.recv_iopub_busy();
+    frontend.assert_stream_stdout_contains("restart");
+    let reply = frontend.recv_iopub_comm_msg();
+    assert_eq!(reply.comm_id, comm_id);
+    let reply = serde_json::from_value::<UiBackendReply>(reply.data).unwrap();
+    assert_eq!(reply, UiBackendReply::FrontendReadyReply());
+    frontend.recv_iopub_idle();
+}
+
+/// A "reconnect" fires session_reconnect hooks (not session_init hooks).
+#[test]
+fn test_session_reconnect_hook() {
+    let frontend = DummyArkFrontend::lock();
+    let comm_id = frontend.open_ui_comm();
+
+    execute(
+        &frontend,
+        &comm_id,
+        "setHook('positron.session_init', function(start_type) cat('init ran'))",
+    );
+    execute(
+        &frontend,
+        &comm_id,
+        "setHook('positron.session_reconnect', function() cat('reconnect ran'))",
+    );
+
+    let data = serde_json::json!({
+        "method": "frontend_ready",
+        "params": { "start_type": "reconnect" },
+        "id": "frontend-ready-rpc"
+    });
+    frontend.send_shell_comm_msg(String::from(&comm_id), data);
+    frontend.recv_iopub_busy();
+    frontend.assert_stream_stdout_contains("reconnect ran");
+    let reply = frontend.recv_iopub_comm_msg();
+    assert_eq!(reply.comm_id, comm_id);
+    let reply = serde_json::from_value::<UiBackendReply>(reply.data).unwrap();
+    assert_eq!(reply, UiBackendReply::FrontendReadyReply());
+    frontend.recv_iopub_idle();
+}
--- a/crates/ark/tests/ui-prompt-state.rs
+++ b/crates/ark/tests/ui-prompt-state.rs
@@ -5,38 +5,8 @@
 //
 //
 
-use amalthea::comm::ui_comm::PromptStateParams;
 use amalthea::fixtures::dummy_frontend::ExecuteRequestOptions;
 use ark_test::DummyArkFrontend;
-
-/// Receive a UI comm event from IOPub and assert it is a `busy` event.
-#[track_caller]
-fn recv_ui_busy(frontend: &DummyArkFrontend, comm_id: &str, expected: bool) {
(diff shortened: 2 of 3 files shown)


crates/ark/src/console/console_integration.rs
```diff
@@ -23,13 +24,37 @@ impl Console {
         // Create and store the sender channel
         self.ui_comm_tx = Some(UiCommSender::new(ui_comm_tx));
 
+        // Set initial console width from the comm_open data, if provided.
+        if let Some(width) = comm_open_data.get("console_width").and_then(|v| v.as_i64()) {
```
lionel-: I think here we'd try to deserialise from a struct defined in ark/ui/ui_comm.rs with `serde_json::from_value()`.

crates/ark/src/console/console_integration.rs
```diff
@@ -23,13 +24,37 @@ impl Console {
...
+            if let Err(err) = harp::exec::RFunction::from(".ps.rpc.setConsoleWidth")
+                .param("width", RObject::from(width as i32))
+                .call()
+            {
+                log::warn!("Failed to set initial console width: {err:?}");
+            }
```
lionel-: Here and below: you can use the `.log_err()` method to collapse this to a single call chain.

crates/ark/src/console/console_integration.rs
```diff
@@ -23,13 +24,37 @@ impl Console {
...
 
             ui_comm_tx.send_refresh(input_prompt, continuation_prompt);
         });
+
+        // Run session init hooks. Extract the start type from
+        // comm_open data and pass it to the hook function.
+        let start_type = comm_open_data
+            .get("start_type")
+            .and_then(|v| v.as_str())
+            .unwrap_or("new");
```
lionel-: With the struct approach you can define default values in the struct directly.

crates/ark/src/modules/positron/hooks.R
```diff
@@ -160,6 +160,25 @@ check_version <- function(pkg) {
     )
 }
 
+#' @export
```
lionel-: Does this need to be exported on the search path? Exporting should generally be a last resort.

lionel- (commented):
Looking good! Just a few style comments

crates/ark/src/console/console_integration.rs
```diff
@@ -24,7 +24,11 @@ impl Console {
         let comm = self.comms.get(self.ui_comm_id.as_deref()?)?;
         Some(UiCommRef {
             comm,
-            originator: self.active_request.as_ref().map(|r| &r.originator),
+            originator: self
+                .active_request
+                .as_ref()
+                .map(|r| &r.originator)
+                .or(self.comm_msg_originator.as_ref()),
```
jennybc, line 31: This is the key change that makes frontend RPCs work from `comm_msg` handlers.

crates/ark/src/ui/ui_comm.rs
```diff
@@ -40,10 +41,25 @@ pub const UI_COMM_NAME: &str = "positron.ui";
 pub struct UiComm {
     graphics_device_tx: AsyncUnboundedSender<GraphicsDeviceNotification>,
     working_directory: PathBuf,
+    comm_open_data: Value,
 }
 
 impl CommHandler for UiComm {
     fn handle_open(&mut self, ctx: &CommHandlerContext) {
+        // Set initial console width from the comm_open data, if provided.
```
jennybc, line 57: We had hoped to do *everything* inside `handle_open()`, i.e. fire the session init hooks. Turns out that doesn't work. But it is nice to handle the initial console width this way, so I kept that bit.

crates/amalthea/src/comm/ui_comm.rs
```diff
@@ -166,6 +166,14 @@ pub struct DidChangePlotsRenderSettingsParams {
 	pub settings: PlotRenderSettings,
 }
 
+/// Parameters for the FrontendReady method.
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+pub struct FrontendReadyParams {
+	/// The type of session start: 'new' for new sessions, 'restart' for
+	/// restarted sessions, 'reconnect' for reconnected sessions
+	pub start_type: String,
+}
```
DavisVaughan, line 175: Architectural (later):

In the future, we think `start_type` might eventually come through `comm_open()`. In that world, I think it does make sense for this to be a 3 valued string like is shown here. Because we get a `comm_open()` for all 3 types of "start".

When we move `start_type` to `comm_open()`, this method probably changes from `FrontendReady` to something more specific like `FireHooks` (or maybe the name still makes sense, idk yet?). Regardless, what definitely would change is that `start_type` would be removed from the params here, because it would have already come through in `comm_open()`.

crates/ark/src/modules/positron/hooks.R
```diff
@@ -160,6 +160,45 @@ check_version <- function(pkg) {
...
+                log_warning(sprintf(
+                    "Error in positron.session_reconnect hook: %s",
+                    conditionMessage(err)
+                ))
+            }
+        )
+    }
+
+    invisible(NULL)
+}
```
DavisVaughan, line 198: Testing (hopefully now):

I would like to see 1 test for each of the 3 hook types if possible

crates/ark/src/ui/ui_comm.rs
```diff
@@ -40,10 +41,25 @@ pub const UI_COMM_NAME: &str = "positron.ui";
 pub struct UiComm {
     graphics_device_tx: AsyncUnboundedSender<GraphicsDeviceNotification>,
     working_directory: PathBuf,
+    comm_open_data: Value,
 }
```
DavisVaughan, line 54: Architectural (later):

We noted in the call that `readline()` and `menu()` calls from a session init hook still won't work.

This requires a non trivial amount of additional work

But this was historically useful for something like renv, where it prompts the user on startup that they may need to `renv::restore()` before proceeding, and doing that at session init (rather than in .Rprofile directly) is "the right time" to talk to the user in such a way.

crates/ark/src/ui/ui_comm.rs
```diff
@@ -40,10 +41,25 @@ pub const UI_COMM_NAME: &str = "positron.ui";
...
 impl CommHandler for UiComm {
     fn handle_open(&mut self, ctx: &CommHandlerContext) {
+        // Set initial console width from the comm_open data, if provided.
+        if let Some(width) = self
+            .comm_open_data
+            .get("console_width")
+            .and_then(|v| v.as_i64())
+        {
+            if let Err(err) = RFunction::from(".ps.rpc.setConsoleWidth")
+                .param("width", RObject::from(width as i32))
```
DavisVaughan: Is it possible to avoid the double conversion of `as_i64()` and `as i32`?

crates/ark/src/ui/ui_comm.rs
```diff
@@ -145,6 +164,25 @@ impl UiComm {
...
+                log::warn!("Failed to execute session reconnect hooks: {err:?}");
+            }
+        } else {
+            if let Err(err) = RFunction::from(".ps.run_session_init_hooks")
+                .param("start_type", RObject::from(params.start_type.as_str()))
+                .call()
+            {
+                log::warn!("Failed to execute session init hooks: {err:?}");
+            }
+        }
```
DavisVaughan, line 191: Nit (later, or never i guess)

A small part of me wanted `params.start_type` to be a 3 valued enum so that you can exhaustively switch on it here.

Right now its impossible to read off the other 2 types (new / restart) from the code, but with an exhaustive switch forced by a 3 value enum, it would have been clearer.

Plus in the future if we add a 4th type then the compiler would force us to handle it at the call site here

crates/amalthea/src/comm/ui_comm.rs
jennybc: FWIW I stayed away from hard-wiring "fire the hooks!" terminology into the Positron side. Because I figured, Positron really doesn't need to know what the backend is doing with this info. It just needs to say "I am ready".

crates/ark/src/modules/positron/hooks.R
jennybc: Added in eb19361c94eef682a855694f3531d180be5492bc

crates/ark/src/ui/ui_comm.rs
jennybc: Issue opened: https://github.com/posit-dev/ark/issues/1130 (I'll come back with good example code after this get this merge done.)

crates/ark/tests/kernel-hooks-session.rs
```diff
@@ -0,0 +1,127 @@
...
+#[test]
+fn test_session_init_hook_restart() {
+    let frontend = DummyArkFrontend::lock();
+    let comm_id = frontend.open_ui_comm();
+
+    execute(
+        &frontend,
+        &comm_id,
+        "setHook('positron.session_init', function(start_type) cat(start_type))",
+    );
```
lionel-, line 80: Could also set a reconnect hook to verify that it is not fired up.

crates/amalthea/src/comm/ui_comm.rs
DavisVaughan: Yea that is also where I was starting to land with "or maybe the name still makes sense, idk yet?"

crates/ark/tests/kernel-hooks-session.rs
DavisVaughan: thats a nice idea

PR #1097, opened 2026-03-10
Add non-Send task spawning

lionel-:
- Add non-Send variants of task spawning. To keep things manageable, a single entry point `spawn()` now takes a task wrapped in an enum describing the properties of the task (idle/interrupt, send/!send). Non-send tasks panic if spawned outside the R thread.

- The `Table` field of the data explorer no longer needs to be wrapped in `RThreadSafe`. Follow-up to data explorer migration in #1075.

### DIFF
--- a/crates/ark/src/r_task.rs
+++ b/crates/ark/src/r_task.rs
@@ -327,7 +327,7 @@ impl RTask {
         F: FnOnce(ConsoleOutputCapture) -> Fut + 'static,
         Fut: Future<Output = ()> + 'static,
     {
-        RTask::Idle(Self::with_capture(fun))
+        RTask::Idle(Self::pin_with_capture(fun))
     }
 
     #[allow(unused)]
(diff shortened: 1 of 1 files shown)


crates/ark/src/dap/dap_server.rs
```diff
@@ -824,7 +824,7 @@ impl<R: Read, W: Write> DapServer<R, W> {
         let responses_tx = self.responses_tx.clone();
 
         log::trace!("DAP: Spawning idle task for evaluate");
-        spawn_idle_any_prompt(async move |mut capture| {
+        r_task::spawn(RTask::send_idle_any_prompt(async move |mut capture| {
```
DavisVaughan, line 827: It's just `RTask::idle` so why not `RTask::idle_any_prompt`?

DavisVaughan: Oh wait i see, there's `idle` and `send_idle`

crates/ark/src/r_task.rs
```diff
@@ -283,91 +286,143 @@ where
...
+        RTask::IdleAnyPrompt(Self::with_capture(fun))
     }
 
-    let tasks_tx = channels.tx();
+    fn with_capture<F, Fut>(fun: F) -> BoxFuture<'static, ()>
+    where
+        F: FnOnce(ConsoleOutputCapture) -> Fut + 'static,
+        Fut: Future<Output = ()> + 'static,
+    {
+        Box::pin(async move {
```
DavisVaughan: Maybe `pin_with_capture()` to make the `pin()` explicit in the name?

I thought at first maybe you'd need to wrap `with_capture()` in `Box::pin()` when I saw the call site and compared to `interrupt()`

crates/ark/src/r_task.rs
```diff
@@ -283,91 +286,143 @@ where
...
     };
 
-    // Send the async task to the R thread
-    let task = RTask::Async(RTaskAsync {
-        fut: Box::pin(fun()) as BoxFuture<'static, ()>,
+    let task = QueuedRTask::Async(RTaskAsync {
+        fut,
         tasks_tx: tasks_tx.clone(),
-        start_info: RTaskStartInfo::new(only_idle),
+        start_info: RTaskStartInfo::new(idle),
```
DavisVaughan: Was `only_idle` more informative?

PR #1099, opened 2026-03-10
Run UI comm on the R thread

lionel-:

- UI comm now accessible via `Console::ui_comm()`. It returns an optional reference to a UI comm view with methods to send UI comm events or requests. If the option is `None`, the UI comm is not connected. If it's `Some()`.

- The `Console::try_ui_comm()` variant returns a `Result` and is nice for easy propagation of errors when the UI comm is expected.

- Since the UI comm is frontend-initiated, added `CommOpen` plumbing. You were right to call out the complicated-looking factory handler, it was not necessary (and I can't think of a reason for it to be necessary with other comms).

- The continuation prompt field has been removed from `PromptInfo`. We now refresh the prompt from the `handle_environment()` comm method, and fetch `getOption("continue")` from there.

- Fixes the busy/idle race condition, and makes the tests both simpler, more robust, and more exhaustive.

### DIFF
--- a/crates/ark/src/console/console_comm.rs
+++ b/crates/ark/src/console/console_comm.rs
@@ -97,7 +97,7 @@ impl Console {
         self.comms.insert(comm_id, ConsoleComm { handler, ctx });
     }
 
-    pub(super) fn comm_notify_environment_changed(&mut self, event: EnvironmentChanged) {
+    pub(super) fn comm_notify_environment_changed(&mut self, event: &EnvironmentChanged) {
         for (_, comm) in self.comms.iter_mut() {
             comm.handler.handle_environment(event, &comm.ctx);
         }
--- a/crates/ark/src/console/console_repl.rs
+++ b/crates/ark/src/console/console_repl.rs
@@ -561,8 +561,8 @@ impl Console {
         };
 
         // Initial input and continuation prompts
-        let input_prompt: String = harp::get_option("prompt").try_into().unwrap();
-        let continuation_prompt: String = harp::get_option("continue").try_into().unwrap();
+        let input_prompt = harp::get_input_prompt();
+        let continuation_prompt = harp::get_continuation_prompt();
 
         let kernel_info = KernelInfo {
             version: version.clone(),
@@ -803,6 +803,11 @@ impl Console {
         let info = self.prompt_info(prompt);
         log::trace!("R prompt: {}", info.input_prompt);
 
+        // Continuation prompt for the frontend (i.e. `getOption("continue")`).
+        // Passed along with the input prompt to `handle_active_request` and
+        // forwarded to comm handlers via `EnvironmentChanged::Execution`.
+        let continuation_prompt = harp::get_continuation_prompt();
+
         let is_browser = matches!(info.kind, PromptKind::Browser);
 
         let suppress = filter_debug_output();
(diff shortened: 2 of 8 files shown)


crates/ark_test/src/dummy_frontend.rs
```diff
@@ -1094,54 +1064,28 @@ impl DummyArkFrontend {
...
-        // as CommMsg on IOPub. These can arrive in any order relative to
-        // the busy/idle. We wait for the prompt_state CommMsg as evidence
-        // that the UI comm has been established, draining everything.
-        let deadline = Instant::now() + RECV_TIMEOUT;
-        let mut got_prompt_state = false;
-        let mut idle_count = 0u32;
+        // The UI comm runs on the R thread via CommHandler. The comm_open
+        // blocks Shell while the handler's `handle_open()` runs, so events
+        // arrive deterministically within the Busy/Idle window.
+        self.recv_iopub_busy();
```
DavisVaughan, line 1082: Nice

crates/ark/src/ui/ui_comm.rs
```diff
@@ -0,0 +1,373 @@
+//
+// ui.rs
```
DavisVaughan:
```suggestion
// ui_comm.rs
```

crates/ark/src/lsp/backend.rs
```diff
@@ -107,13 +105,11 @@ fn report_crash() {
         "with full logs (see https://positron.posit.co/troubleshooting.html#python-and-r-logs)."
     );
 
+    // NOTE: This is a legit use of interrupt-time task. No R access here, and
+    // we need to go through Console since it owns the UI comm.
```
DavisVaughan, line 110: Could update this note to reflect our thoughts about the fact that this could be a crossbeam channel notification that the Console knows about, without needing an r_task

crates/ark/src/console/console_repl.rs
```diff
@@ -1209,17 +1195,6 @@ impl Console {
             return;
         };
```
DavisVaughan, line 1231: Nice!

crates/ark/src/console/console_comm.rs
```diff
@@ -67,6 +72,35 @@ impl Console {
...
+
+        if comm_name == UI_COMM_NAME {
+            if let Some(old_id) = self.ui_comm_id.take() {
+                log::info!("Replacing an existing UI comm.");
+                if let Some(mut old) = self.comms.remove(&old_id) {
+                    old.handler.handle_close(&old.ctx);
+                }
+            }
+            self.ui_comm_id = Some(comm_id.clone());
+        }
```
DavisVaughan, line 95: Something feels slightly off that here and in `comm_handle_close` you have to specially handle the `ui_comm_id` but I can't quite figure out how you'd simplify it

crates/ark/src/lsp/backend.rs
lionel-: hmm... I think I'd still see it modelled as some kind of `r_interrupt_task()` that receives a trimmed down `Console` as input.

crates/ark/src/console/console_comm.rs
lionel-: I've extracted into `comm_remove()` so all special handling is centralised.

PR #775, opened 2025-04-14
Send pre-renderings of new plots to the frontend

lionel-:
Ark side of https://github.com/posit-dev/positron/pull/7247
Progress towards https://github.com/posit-dev/positron/issues/5184
Progress towards https://github.com/posit-dev/positron/issues/6736

We now generate pre-renderings of new plots and send them over to the frontend as part of `comm_open` messages.

The current settings for the pre-renders are stored in a `Cell`. This requires a `Copy` type which is generated on the Positron side (see linked PR).

PR #1100, opened 2026-03-10
Run plot comms on the R thread

### LINKED ISSUE #791: Rewrite `on_process_idle_events()` as an async idle task

lionel-:
Branched from #1099
Addresses https://github.com/posit-dev/positron/issues/12825 (planning to close)
Closes https://github.com/posit-dev/ark/issues/791

I did not foresee all the simplifications falling out of this one!

- Plots no longer live in their own threads, they live on the R thread. That frees up memory since each background thread takes about 2mb of stack space (depending on platform), so 20 open plots would take 40mb of memory.

- `DeviceContext` moves from a standalone `thread_local!` into a regular field of `Console`. All access goes through `Console::get().device_context()`. The `CommHandlerContext` provides access to `Console` via `ctx.console()`, so plot handlers reach the device context through a controlled context chain.

- The old idle-time polling loop (`process_rpc_requests` with `crossbeam::Select`) is removed. Plot RPCs are now dispatched synchronously through the shell handler on the R thread. This improves plot latency when pre-renderings require adjustments because we no longer timeout-poll for plot updates (related to https://github.com/posit-dev/positron/issues/5184).

- The `GraphicsDeviceNotification` async channel is removed. Since the UI comm now runs on the R thread (from #1099), prerender settings are updated synchronously via `Console::get().device_context().set_prerender_settings()`. This removes quite a bit of plumbing.

- Fixes a `dev.hold()` regression I introduced with pre-renderings (https://github.com/posit-dev/ark/pull/775). We were previously sending pre-renderings unconditionally, now they are held until `dev.flush()`. This is tested by new integration tests.


crates/ark/tests/plots.rs
```diff
@@ -583,3 +583,71 @@ fn test_plot_source_context_stacking() {
...
+fn test_dev_hold_suppresses_intermediate_plots() {
+    let frontend = DummyArkFrontend::lock();
+
+    // Activate the graphics device
+    frontend.send_execute_request("plot(1:10)", ExecuteRequestOptions::default());
+    frontend.recv_iopub_busy();
+    frontend.recv_iopub_execute_input();
+    frontend.recv_iopub_display_data();
+    frontend.recv_iopub_idle();
+    frontend.recv_shell_execute_reply();
```
DavisVaughan: This smells a bit funny. I don't think you should need this if things are working correctly

crates/ark/src/console/console_repl.rs
```diff
@@ -481,17 +480,16 @@ impl Console {
...
-        // Initialize the GD context on this thread.
+        // Perform R-side graphics device initialization (register as
+        // interactive, spawn notification listener). The `DeviceContext`
+        // itself is already created as part of `Console::new()`.
+        //
         // Note that we do it after init is complete to avoid deadlocking
         // integration tests by spawning an async task. The deadlock is caused
         // by https://github.com/posit-dev/ark/blob/bd827e735970ca17102aeddfbe2c3ccf26950a36/crates/ark/src/r_task.rs#L261.
         // We should be able to remove this escape hatch in `r_task()` by
         // instantiating an `Console` in unit tests as well.
```
DavisVaughan: I think this comment no longer applies because there is no spawn_*() in init_graphics device now?

crates/ark/src/console/console_repl.rs
```diff
@@ -481,17 +480,16 @@ impl Console {
             }
         });
 
-        // Initialize the GD context on this thread.
+        // Perform R-side graphics device initialization (register as
+        // interactive, spawn notification listener). The `DeviceContext`
```
DavisVaughan: "spawn notification listener" it does not look like you do this

crates/ark/src/console/console_repl.rs
```diff
@@ -589,6 +587,8 @@ impl Console {
         dap: Arc<Mutex<Dap>>,
         session_mode: SessionMode,
     ) -> Self {
+        let device_context = DeviceContext::new(iopub_tx.clone());
```
DavisVaughan: It makes sense to me to remove `iopub_tx` from `DeviceContext` now and _only_ pull `iopub_tx` directly from `Console`

Over in `graphics_device.rs`, you can just `Console::get().iopub_tx()` anytime you need it now!

It feels like this better scopes responsibilities to me

crates/ark/src/ui/ui_comm.rs
```diff
@@ -131,11 +125,9 @@ impl UiComm {
...
         }
 
-        self.graphics_device_tx
-            .send(GraphicsDeviceNotification::DidChangePlotRenderSettings(
-                params.settings,
-            ))
-            .map_err(|err| anyhow::anyhow!("Failed to send plot render settings: {err}"))?;
+        Console::get()
+            .device_context()
+            .set_prerender_settings(params.settings);
```
DavisVaughan, line 176: Is it possible that we could now make `DeviceContext` have non-`Cell` internals?

I am disturbed by being able to call `get()` to then perform a mutable update, rather than being forced to go through `get_mut()`

crates/ark/src/comm_handler.rs
```diff
@@ -56,6 +59,11 @@ impl CommHandlerContext {
         };
         self.outgoing_tx.send(CommMsg::Data(json)).log_err();
     }
+
+    /// Access the Console singleton (the R runtime).
+    pub(crate) fn console(&self) -> &Console {
+        Console::get()
+    }
```
DavisVaughan: Nothing really stops a handler from doing `Console::get()` on their own though, right? So I'm not entirely sure I see the value of this yet

crates/ark/src/plots/graphics_device.rs
```diff
@@ -335,9 +291,15 @@ impl DeviceContext {
...
     fn hook_holdflush(&self, level: i32) {
+        let was_held = !self.should_render.get();
         // Be extra safe and check `level <= 0` rather than just `level == 0` in case
         // our shadowed device returns a negative `level`
         self.should_render.replace(level <= 0);
+
+        // Flush deferred changes on hold→release transition
+        if was_held && self.should_render.get() {
+            self.process_changes();
+        }
```
DavisVaughan:
```suggestion
        // Be extra safe and check `level <= 0` rather than just `level == 0` in case
        // our shadowed device returns a negative `level`
        let is_released = level <= 0;
        let was_held = self.should_render.replace(is_released);

        // Flush deferred changes on hold→release transition
        if was_held && is_released {
            self.process_changes();
        }
```

How about this? Since `replace()` returns the old value, and we don't need to `get()` again at the end. Also feels more readable

crates/ark/src/plots/graphics_device.rs
```diff
@@ -474,111 +436,6 @@ impl DeviceContext {
...
-        if self.is_drawing.get() {
-            log::trace!("Refusing to render due to `is_drawing`");
-            return;
-        }
-
-        // Don't try to render a plot if someone is asking us not to, i.e. `dev.hold()`
-        if !self.should_render.get() {
-            log::trace!("Refusing to render due to `should_render`");
-            return;
-        }
```
DavisVaughan, line 534: I am pretty sure these are important

crates/ark/src/plots/graphics_device.rs
```diff
@@ -474,111 +436,6 @@ impl DeviceContext {
...
-
-        // Collect existing sockets into a vector of tuples.
```
DavisVaughan, line 536: Broken regression from this PR

```r
plt2 = function() {
  layout(matrix(1:2, 2))
  plot(1, 1)
  plot(1, 1)
}
plt2()
```

DavisVaughan: In general, just send this whole chunk "all at once" rather than in separate requests

```r
  layout(matrix(1:2, 2))
  plot(1, 1)
  plot(1, 1)
```

crates/ark/src/plots/graphics_device.rs
```diff
@@ -474,111 +436,6 @@ impl DeviceContext {
...
-        // Necessary for handling Select in a clean way.
-        let sockets = {
-            // Refcell Safety: Clone the hashmap so we don't hold a reference for too long
```
DavisVaughan, line 539: Broken regression

```r
# Stars raster
library(stars)
library(terra)
r <- matrix(runif(9), nrow = 3) |>
  st_as_stars()
plot(r)
rast(r) |> plot()
```

crates/ark/src/plots/graphics_device.rs
```diff
@@ -474,111 +436,6 @@ impl DeviceContext {
...
-            let sockets = self.sockets.borrow().clone();
-            sockets.into_iter().collect::<Vec<_>>()
```
DavisVaughan, line 541: Broken regression

```r
# rpart
library(rpart)
library(rpart.plot)
model <- rpart(Species ~ ., data = iris, method = "class")
rpart.plot(model)
```

DavisVaughan: Actually something somehow seems deeply broken by this? I am getting blank new-pages fairly often and reproducibly. Run these 3 plots one at a time.

```r
# rpart
library(rpart)
library(rpart.plot)
model <- rpart(Species ~ ., data = iris, method = "class")
rpart.plot(model)

# sf
geo <- sf::st_make_grid(
  cellsize = c(1, 1),
  offset = c(0, 0),
  n = 10
)
x <- sample(c(0, 1), 100, prob = c(0.9, 0.1), replace = TRUE)
plot(sf::st_sf(x = x, geo))

# rpart again
library(rpart)
library(rpart.plot)
model <- rpart(Species ~ ., data = iris, method = "class")
rpart.plot(model)
```


https://github.com/user-attachments/assets/99299568-e452-473c-a326-73121f48742c

crates/ark/src/plots/graphics_device.rs
```diff
@@ -474,111 +436,6 @@ impl DeviceContext {
...
-        };
-
```
DavisVaughan, line 543: In general here is my "regression script" I run through one at a time (following the per example instructions of whether to run it one line at a time or as a chunk of code)

All of these should work

```r
# -------------------------
# Multiple plots at once

# Run the chunk all at once
plot(1:10)
plot(2:20)

# Run the chunk all at once
library(ggplot2)
g <- ggplot(mpg, aes(class))
# Number of cars in each class:
g + geom_bar()
# Total engine displacement of each class
g + geom_bar(aes(weight = displ))
# Map class to y instead to flip the orientation
ggplot(mpg) + geom_bar(aes(y = class))

# Inside a function
plt2 = function() {
  layout(matrix(1:2, 2))
  plot(1, 1)
  plot(1, 1)
}
plt2()

# Multiple ggplots
library(ggplot2)
library(purrr)
economics_list <- split(economics_long, economics_long$variable)
economics_list |>
  imap(\(df, nm) {
    df |>
      ggplot(aes(date, value)) +
      geom_line() +
      labs(title = nm)
  })

library(ggExtra)
library(ggplot2)
set.seed(30)
df <- data.frame(x = rnorm(500, 50, 10), y = runif(500, 0, 50))
p2 <- ggplot(df, aes(x, y)) + geom_point()
ggMarginal(p2, type = "boxplot")
ggMarginal(p2, type = "histogram")
par(mfrow = c(1, 1))

# -------------------------
# Plotting in a loop

# (Note it would be nice if we could process renders at interrupt time)

for (i in 1:5) {
  plot(i)
  Sys.sleep(1)
}

# -------------------------
# Par related examples

par(mfrow = c(2, 2))
hist(
  iris$Petal.Width[iris$Species == "virginica"],
  xlim = c(0, 3),
  breaks = 9,
  main = "Petal Width for Virginica",
  xlab = "",
  col = "green"
)
hist(
  iris$Petal.Width[iris$Species == "versicolor"],
  xlim = c(0, 3),
  breaks = 9,
  main = "Petal Width for Versicolor",
  xlab = "",
  col = "blue"
)
hist(
  iris$Petal.Width[iris$Species == "versicolor"],
  xlim = c(0, 3),
  breaks = 9,
  main = "Petal Width for Versicolor",
  xlab = "",
  col = "blue"
)
plot(1:10)
par(mfrow = c(1, 1))


par(mfrow = c(2, 1))
plot(rnorm(100))
plot(rnorm(1000))
par(mfrow = c(1, 1))

# Stars raster
library(stars)
library(terra)
r <- matrix(runif(9), nrow = 3) |>
  st_as_stars()
plot(r)
rast(r) |> plot()

# Performance models
library(performance)
data(mtcars)
m <- lm(mpg ~ gear + cyl + hp, data = mtcars)
check_model(m)

m <- lm(mpg ~ gear + cyl + hp, data = mtcars)
p <- plot(check_model(m), panel = FALSE)
p[[1]] # works
p[[2]] # works
# blank plot
patchwork::wrap_plots(p)

# rpart
library(rpart)
library(rpart.plot)
model <- rpart(Species ~ ., data = iris, method = "class")
rpart.plot(model)

# layout() related
# layout(matrix(
#   c(1, 1, 1, 1, 1, 1, 1, 1, 1, 2, 2, 2),
#   nrow = 4,
#   byrow = TRUE
# ))
# layout.show(n = 2)

# sf
geo <- sf::st_make_grid(
  cellsize = c(1, 1),
  offset = c(0, 0),
  n = 10
)
x <- sample(c(0, 1), 100, prob = c(0.9, 0.1), replace = TRUE)
plot(sf::st_sf(x = x, geo))

# biplot
dat <- mtcars
pr <- princomp(dat)
biplot(pr)

# -------------------------
# dev.hold()-ing

# Run these one at a time!
# (RStudio can't actually do this)
dev.hold()
plot(1:10)
abline(1, 2)
dev.flush()

# -------------------------
# tinyplots

library(tinyplot)

plt(Sepal.Length ~ Petal.Length, data = iris)
plt(Sepal.Length ~ Petal.Length | Species, data = iris)
plt(Sepal.Length ~ Petal.Length | Species, data = iris, legend = "bottomright")

# generate data
lambda1 <- 1.8
lambda2 <- 0.7
x <- 0:10
p1 <- dpois(x, lambda1)
p2 <- dpois(x, lambda2)
dp <- data.frame(
  x = c(x, x),
  p = c(p1, p2),
  lambda = rep(c(lambda1, lambda2), each = length(x))
)
dp$lambda <- as.factor(dp$lambda)
# plot
plt(
  p ~ x,
  facet = ~lambda,
  data = dp,
  type = "h",
  lwd = 2,
  col = "skyblue",
  draw = c(
    axis(1, at = x, labels = x),
    points(p ~ x, data = dp, col = "black")
  )
)

# -------------------------
# The dreaded graphics device swap

# Run this all at once
library(ggplot2)
p <- ggplot(mtcars, aes(x = wt, y = mpg)) + geom_point()
p
ggsave("temp.png", p)

# Run this all at once
plot(1:10)
grDevices::png()
plot(1:20)
dev.off()

# -------------------------
# Something related to big plots was not working when you resize
# (I did some perf improvements)

library(qgraph)
data(big5)
data(big5groups)
# First plot
qgraph(
  cor(big5),
  minimum = 0.25,
  groups = big5groups,
  legend = TRUE,
  borders = FALSE,
  title = "Big 5 correlations"
)
# Second identical plot
qgraph(
  cor(big5),
  minimum = 0.25,
  groups = big5groups,
  legend = TRUE,
  borders = FALSE,
  title = "Big 5 correlations"
)

# -------------------------
# Doesn't work - stepping through plots one at a time
# Not precisely sure why, because in theory we are at idle time

fit_lm <- lm(
  mpg ~ cyl + wt,
  data = mtcars
)

# Renders only plot(fit_lm, which = 5)
# Not able to scroll through the which = 1, 2, 3 plots
# that are automatically called by plot.lm().
plot(fit_lm)

# -------------------------
# Doesn't work - we don't respond to render requests in the middle of R execution
# We respond at idle time right now in `process_events()`.
# So all the plots appear at once at the end

tourr::animate_xy(tourr::flea[, 1:6])

# Simpler example of the idea
{
  plot(1:10)
  plot(2:11)
  Sys.sleep(5)
}

```

DavisVaughan (changes requested):
I have paused on review until we can get these regressions resolved

crates/ark/src/plots/graphics_device.rs
lionel-: Thanks! I'll add these as test cases

crates/ark/src/console/console_repl.rs
lionel-: I feel the opposite. It hides that the device context directly communicates with the frontend via IOPub.

Also ideally we should avoid `Console::get()` calls as these bypass Rust's ownership model. But then the issue is that the device belongs to console and not the other way around. Cloning the IOPub channel is the right ownership structure from that point of view.

crates/ark/src/ui/ui_comm.rs
lionel-: The `Cell` is actually more reassuring, especially regarding the possibility of reentrancy (flushing calls `process_changes` which can recurse into graphics device callbacks).

In general `get()` is not a signal that the state is not mutable, it's only a signal that the state is shared, and can still be mutable e.g. through Cells or other mechanisms.

It's all a bit hand wavy considering that our callbacks might produce multiple active `get_mut()` at the same time, but adding to the unsoundness would be... unsound ;)

crates/ark/src/comm_handler.rs
lionel-: The idea was to move towards explicit ownership and borrowing of context. That said, it now occurs to me that it doesn't make sense to store a reference permanently in the handler context, as that would prevent exclusive `&mut` references. So I've refactored things so that we pass `&Console` directly to handler methods, instead of storing them in each `CommHandlerContext`.

crates/ark/src/plots/graphics_device.rs
lionel-: Since we now process the requests when R is idle, `is_drawing` is structurally false, we're never mid-draw.

Regarding `should_render` (matching `dev.hold()`), it's true that holds can span across prompts, I can see that happening when developing a plot function interactively. This would require:

- Creating a plot that Positron knows about (opened comm)
- Calling `dev.hold()` in the console
- Resizing the plot in the frontend UI

So that's a very narrow case. We show the uncommitted visual changes in that case, which seems like a reasonable compromise. The alternative would be to refuse to render, producing a distorted plot in the frontend after resize.

crates/ark/src/plots/graphics_device.rs
lionel-: Should now be fixed. AFAICT all the issues came down to the new implementation failing to _record_ plot updates when a rendering was on hold. Now we do record them, and only hold _notifications_ to the frontend.

https://github.com/posit-dev/ark/pull/1100/commits/bc636c54cd963ec883e765de37e4d53347aa1106

lionel-:
I also fixed a race between `comm_open` and subsequent `comm_msg`, the tests revealed it was possible to get the latter before the former. That's because the Shell thread is in charge of sending `comm_open` on IOPub, so we had a triangular configuration where the R thread and the Shell thread were racing to send their messages on IOPub.

To fix this, I made comm opening _blocking_. There is a new `CommEvent::Barrier` variant to support this. On the Shell side, the shell thread now uses a select loop that drains comm events while a request is in flight.

crates/amalthea/src/socket/shell.rs
```diff
@@ -363,6 +372,63 @@ impl Shell {
...
+
+            let ready = sel.ready();
+
+            while let Ok(event) = self.comm_event_rx.try_recv() {
+                self.process_comm_event(event);
+            }
+
+            if ready == resp_idx {
+                break response_rx.recv().unwrap();
+            }
```
DavisVaughan: So even if `sel.ready()` returns the execute response as the first "ready" thing, you want to go off and process the `comm_event`s first?

I think I was expecting to see

```
if ready == resp_idx {
    break response_rx.recv().unwrap();
}
```

right after `let ready =`

crates/ark/src/console/console_comm.rs
```diff
@@ -57,14 +63,21 @@ impl Console {
...
+        handler.handle_open(&ctx, Console::get());
 
         self.comms
             .insert(comm_id.clone(), ConsoleComm { handler, ctx });
 
         self.comm_event_tx
             .send(CommEvent::Opened(comm, open_metadata))?;
 
+        // Block until Shell has processed the Opened event, ensuring the
+        // `comm_open` message is on IOPub before we return. Any updates
```
DavisVaughan: Just a gut check, but is this really what it ensures? In `CommEvent::Opened` we do an `iopub_tx.send()` call, but that doesn't necessarily mean that IOPub has _received_ it by the end of `CommEvent::Opened` and the following `CommEvent::Barrier` event is processed.

I'm not sure if that matters.

crates/ark/src/console/console_repl.rs
DavisVaughan: > It hides that the device context directly communicates with the frontend via IOPub

Hmm I see what you mean

Like the `DeviceContext` struct should have an "include what you use" kind of policy in the struct fields, and by not having `iopub_tx` in there, we mask what it uses

crates/ark/src/console/console_repl.rs
```diff
@@ -669,10 +663,66 @@ impl Console {
...
+    /// rather than silently hanging.
+    ///
+    /// For tests that exercise real Console behaviour (execution, comms
+    /// lifecycle, IOPub output), prefer full integration tests with
+    /// `DummyArkFrontend` which spins up a real Console.
+    ///
+    /// Idempotent per thread. Does not set `R_INIT`, so the `r_task()` escape
+    /// hatch for unit tests continues to work.
+    #[cfg(feature = "testing")]
+    pub fn test_init() {
```
DavisVaughan, line 697: Can you put this up with `start()` and call it `test_start()`?

crates/ark/src/data_explorer/r_data_explorer.rs
```diff
@@ -481,13 +481,18 @@ impl CommHandler for RDataExplorer {
         serde_json::json!({ "title": self.title, "inline_only": inline_only })
     }
 
-    fn handle_msg(&mut self, msg: CommMsg, ctx: &CommHandlerContext) {
+    fn handle_msg(&mut self, msg: CommMsg, ctx: &CommHandlerContext, _console: &Console) {
```
DavisVaughan: I'm not sure I agree with passing `console` through all these handlers

It seems like the only current use that I can see is in `impl CommHandler for PlotComm` for `handle_msg()` and `handle_close()`.

But those could just do `Console::get()`.

It doesn't seem like it helps clarify ownership very much, because so many other things in `graphics_device.rs` also just do `Console::get()` already. And it seems likely that in `handle_rpc()` (called by `handle_msg()`) we could very easily just do another `Console::get()` rather than passing through the `&Console` we got from `handle_msg()`. It seems complex to me to know when to use which pattern.

It also seems like this `_console: &Console` argument is the reason for all of the new `test_init()` infra, and overall this just seems like it adds quite a bit of complexity for (to me) not a lot of obvious benefit 😢

DavisVaughan: As an example, should `on_did_execute_request()` gain a `console: &Console` argument and thread it through `process_changes` -> `process_new_plot` -> `should_use_dynamic_plots` to the `Console::get` call there?

I don't think so, I feel like it makes sense that we just invoke `Console::get()` from there as required like we do everywhere else.

But now introducing this new pattern of `handle_msg()` and friends having a `console` passed through confuses me and makes me question how we should do things in all the other cases, and I'm not sure that's worth it

crates/ark/src/console/console_comm.rs
```diff
@@ -57,14 +63,21 @@ impl Console {
...
+        // the caller sends after this point are guaranteed to follow it.
+        let (done_tx, done_rx) = crossbeam::channel::bounded(0);
+        self.comm_event_tx.send(CommEvent::Barrier(done_tx))?;
+        done_rx.recv()?;
```
DavisVaughan, line 107: Why have `CommEvent::Barrier` rather than just providing `done` to `CommEvent::Opened`?

I realize we do `.send(CommEvent::Opened` in other places as well, but wouldn't it be useful for them to block too?

crates/ark/src/plots/graphics_device.rs
```diff
@@ -1271,11 +1137,10 @@ pub(crate) fn on_execute_request(
 #[tracing::instrument(level = "trace", skip_all)]
 pub(crate) fn on_did_execute_request() {
```
DavisVaughan: I think `on_did_execute_request()` should perhaps be a `&self` method

It is only called in `handle_active_request()`, but we have `Console` there as `self`, hence we also have `DeviceContext`, so we can just call `self.device_context().on_did_execute_request()` from there?

DavisVaughan: But I am also a bit weirded out by the ownership implications of all this (both the current state of things and my proposal)

`self.device_context().on_did_execute_request()` would call `DeviceContext::process_changes()`, but that goes through `process_new_plot()` and `should_use_dynamic_plots()` and you end up accessing the "parent" owner of `DeviceContext` from there via `Console::get()`, which feels very mind boggling!

crates/ark/src/plots/graphics_device.rs
```diff
@@ -811,14 +674,19 @@ impl DeviceContext {
...
-        }
+        let plot_comm = PlotComm {
+            id: id.clone(),
+            open_data,
+        };
 
-        // Save our new socket.
-        // Refcell Safety: Short borrows in the file.
-        self.sockets.borrow_mut().insert(id.clone(), socket);
+        match Console::get_mut().comm_open_backend(PLOT_COMM_NAME, Box::new(plot_comm)) {
```
DavisVaughan: I think this is another place where it _feeeels_ weird that `DeviceContext` is a field of `Console`, yet refers to `Console` as well

crates/ark/src/console.rs
```diff
@@ -177,7 +177,7 @@ thread_local! {
     pub static CONSOLE: RefCell<UnsafeCell<Console>> = panic!("Must access `CONSOLE` from the R thread");
 }
 
-pub(crate) struct Console {
+pub struct Console {
```
DavisVaughan, line 192: ?

crates/ark/tests/plots.rs
```diff
@@ -699,3 +699,481 @@ fn test_plot_default_size_without_metadata() {
     frontend.recv_iopub_idle();
```
DavisVaughan, line 699: Random note that a `cargo build --release` is throwing this for me right now

```
warning: unused import: `libr::SEXP`
  --> crates/ark/src/r_task.rs:19:5
   |
19 | use libr::SEXP;
   |     ^^^^^^^^^^
```

I think this type is only used in debug builds?

crates/ark/tests/plots.rs
```diff
@@ -699,3 +699,481 @@ fn test_plot_default_size_without_metadata() {
...
     frontend.recv_shell_execute_reply();
 }
+
+/// Test that `dev.hold()` suppresses intermediate plot output.
```
DavisVaughan, line 790: It looks like `plot(lm(disp ~ drat, data = mtcars))` is working again, but I dont see a backend test for it?

DavisVaughan (approved):
Two main comments to discuss, other things are minor:

- Not sure about passing `console` through `handle_msg` and friends
- Not sure about `CommEvent::Barrier` vs passing `done` through `CommEvent::Opened`

See below for more on both

My test file seems to be working fine now

lionel-:
Ok let's discuss `Console` access by comms in our next sync

lionel-:
Updates following our sync (progress towards https://github.com/posit-dev/ark/issues/1145).

**Console::get() pushed to the FFI boundary**

The graphics device callbacks are plain C function pointers - there's no way to pass Rust state through them. `Console::get()` is unavoidable there. But everything past that boundary now receives `&Console` explicitly. The C callbacks are thin shims: capture `Console::get()` once, pass it inward.

**DeviceContext no longer reaches back into Console**

Previously DeviceContext (a field of Console) called `Console::get()` internally for `should_use_dynamic_plots`, `capture_execution_context`, and `comm_open_backend`. Now:
- `session_mode` is stored on DeviceContext directly (it's static, set once at startup)
- `capture_execution_context` returns a default when no context was pushed, removing the Console fallback entirely
- `comm_open_backend` takes `&self` (via RefCell on comms), so DeviceContext calls it through the `&Console` parameter it already has

**PlotComm owns what it needs**

PlotComm is the only CommHandler that used the `console: &Console` parameter (for `device_context()`). Instead of threading Console through the entire CommHandler trait, PlotComm holds an `Rc<DeviceContext>`. The `console` parameter is gone from the trait entirely. Other handlers that need Console access (`RDataExplorer` opening sub-explorers) call `Console::get()` themselves - they're at a boundary where that's appropriate.

**Comm dispatch is `&self` throughout**

All methods in `console_comm.rs` take `&self`. Handler dispatch uses a take/remove pattern: the comm is temporarily removed from its `RefCell` storage before calling the handler, so no borrow guard is held during the call. This prevents panics if the handler reenters the RefCell (e.g. `RDataExplorer` calling `comm_open_backend`). The pattern is encapsulated in helpers (`with_ui_comm_mut`, `with_comm_mut`, etc.).

**Shell deadlock fix**

`comm_open_backend` blocks on a barrier in `CommEvent::Opened` to guarantee IOPub ordering. Previously, if a handler called `comm_open_backend` during comm_msg dispatch, Shell was blocked waiting for the handler to finish and couldn't process the barrier. Now `handle_comm_msg`, `handle_comm_close`, and `handle_comm_open` return a completion receiver instead of blocking. Shell select-loops on it, draining comm events while the handler runs - same pattern that `handle_execute_request` already used. `handle_comm_open` is also sync now (was async behind `block_on`).

**Changelog**

- Explicit `&Console` on DeviceContext methods**: `process_changes`, hooks, and their callees receive `console: &Console` as a parameter instead of calling `Console::get()` internally. C callbacks and `#[harp::register]` functions are the only place `Console::get()` appears.

- **`on_execute_request` / `on_did_execute_request` moved to Console methods** in a new `console_graphics.rs` file, replacing free functions that called `Console::get()`.

- **`DeviceContext` wrapped in `Rc`**: Console holds `Rc<DeviceContext>`. PlotComm gets its own `Rc<DeviceContext>` at construction, giving it direct access without needing `&Console`.

- **`console: &Console` removed from `CommHandler` trait**: No handler receives `&Console` anymore. PlotComm uses its `Rc<DeviceContext>`. Other handlers that need Console access call `Console::get()` themselves.

- **`session_mode` stored on `DeviceContext`**: Passed at construction time instead of queried from Console on every `should_use_dynamic_plots` call.

- **`capture_execution_context` simplified**: The Console fallback (`console.get_execution_context()`) removed. Returns `ExecutionContext::default()` when no context was pushed. `get_execution_context` method on Console removed as dead code.

- **`comms` field wrapped in `RefCell`**: Enables `comm_open_backend` to take `&self` instead of `&mut self`, so DeviceContext can call it through its `&Console` parameter.

- **UI comm stored in its own `RefCell` field**: Separated from the `comms` HashMap so `ui_comm()` can borrow independently.

- **`comm_id` added to `ConsoleComm`**: The UI comm (stored outside the HashMap) carries its own ID for message routing. Eliminated the `UiCommEntry` wrapper type.

- **All `console_comm.rs` dispatch methods take `&self`**: Use take/remove pattern via helpers (`with_ui_comm_mut`, `with_comm_mut`, `take_ui_comm`, `take_comm`, etc.) to avoid holding RefCell borrows during handler dispatch.

- **Shell deadlock fix**: `handle_comm_msg`, `handle_comm_close`, and `handle_comm_open` on the `ShellHandler` trait return a completion `Receiver<()>` instead of blocking. Amalthea's Shell select-loops on it via `drain_comm_events_until` to process `CommEvent`s while the handler runs.

- **`handle_comm_open` made sync**: Was `async` (always called via `block_on`). Now sync with completion receiver, consistent with `handle_comm_msg` / `handle_comm_close`.

- **`CommEvent::Barrier` folded into `CommEvent::Opened`**: The barrier `Sender<()>` is now an optional third field on `Opened`, eliminating a separate event variant.

- **`drain_comm_events_until` extracted as generic helper**: Used by `handle_execute_request`, `handle_comm_msg_request`, `handle_comm_close_request`, and `handle_comm_open_request` via `handle_comm_notification`.

PR #1105, opened 2026-03-11
Fix starvation of Shell requests when idle tasks are scheduled

lionel-:
Hopefully fixes:

```
  ❌ r.tibble - Record data load, basic filtering and sorting [5,051,640 rows]
[tests/data-explorer/data-explorer-performance.test.ts / tests/data-explorer/data-explorer-performance.test.ts] 
(tests/data-explorer/data-explorer-performance.test.ts)  tags: :web, :win, :data-explorer, :performance
```

https://github.com/posit-dev/positron/actions/runs/22932437514/job/66561255858

I think this is due to starvation of Shell request handling in ReadConsole when there are idle tasks in the queue (in this case spawned by the data explorer comm to compute column profiles asynchronously).

From the logs, the sequence on the frontend is:

- Open data explorer for the 5M-row dataset
- Send `get_state`, `get_schema`, `get_data_values` (initial data load)
- Send 19 individual `get_column_profiles` requests (one per column, above the 1M-row batching threshold)
- `get_column_profiles` for indices 15-18 time out after 5 seconds
- `get_schema` times out after 5 seconds
- Replies eventually arrive but listeners are already torn down: "Dropping event 'GetColumnProfilesReply'", "Dropping event 'GetSchemaReply'"
- Grid view stays blank

And on the backend:
- Receives and replies to `get_state`, `get_schema`, `get_data_values` in <1ms each
- Receives `get_column_profiles` requests one by one, acknowledges each in <1ms, spawns idle tasks for the actual computation
- The Shell thread blocks on `dispatch_kernel_request` for each subsequent request, waiting for the R thread to pick it up
- The R event loop picks idle tasks (profile computations, ~500ms-1s each on 5M rows) instead of kernel requests because `select` chooses randomly among ready channels
- `return_column_profiles` async results trickle out over ~8 seconds

I think this has started to fail following the data explorer migration to synchronous comm handling in #1075. Before, the Shell requests would immediately reply to the frontend, while a background thread would independently schedule the idle tasks for column profile computations. After the migration, the idle tasks are scheduled before sending responses to the frontend. This is correct behaviour but it likely has shifted the timing of the `get_schema` and later column profile requests, which are now in contention with all the idle tasks.

We already have a special casing for execute requests so they get absolute priority over idle tasks. The fix is to add other shell request to that special casing.

I've added integration tests that check for starvation of comm requests so that this doesn't happen again.

### DIFF
--- a/crates/ark/src/r_task.rs
+++ b/crates/ark/src/r_task.rs
@@ -454,7 +454,8 @@ pub(crate) fn spawn(task: RTask) {
 //    oneshot, letting the idle task finish and drop its capture cleanly.
 
 #[cfg(debug_assertions)]
-static PENDING_TASK_TX: Mutex<Option<futures::channel::oneshot::Sender<()>>> = Mutex::new(None);
+static TEST_PENDING_TASK_TX: Mutex<Option<futures::channel::oneshot::Sender<()>>> =
+    Mutex::new(None);
 
 /// Clone of the kernel-request channel sender, stored at startup when
 /// `IS_TESTING` is true. Allows R-callable test helpers (e.g.
@@ -481,7 +482,7 @@ unsafe extern "C-unwind" fn ps_test_spawn_pending_task() -> anyhow::Result<SEXP>
     harp::parse_eval_base("options(ark.test.task_polled = FALSE)")?;
 
     let (tx, rx) = futures::channel::oneshot::channel::<()>();
-    *PENDING_TASK_TX.lock().unwrap() = Some(tx);
+    *TEST_PENDING_TASK_TX.lock().unwrap() = Some(tx);
 
     spawn(RTask::idle(async move |_| {
         // Signal that we've been polled (capture is now active)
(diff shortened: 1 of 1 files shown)


crates/ark/src/console/console_repl.rs
```diff
@@ -987,14 +987,23 @@ impl Console {
...
 
-            // First handle execute requests outside of `select` to ensure they
-            // have priority. `select` chooses at random.
+            // First handle execute requests and kernel requests outside of
+            // `select` to ensure they have priority over idle tasks. `select`
+            // chooses at random among ready channels, including idle tasks,
+            // which could include e.g. a lot of data explorer column profile
+            // computations. The intent is for idle tasks to operate in the
+            // background while R is idle and there is no user requests to
+            // handle, so we should handle those with absolute priority.
```
DavisVaughan, line 998: Have you seen `select_biased!`?
https://docs.rs/crossbeam-channel/latest/crossbeam_channel/macro.select_biased.html

crates/ark/src/r_task.rs
```diff
@@ -395,15 +398,29 @@ where
 // 5. Test calls `.Call("ps_test_complete_pending_task")` to unblock the
 //    oneshot, letting the idle task finish and drop its capture cleanly.
 
+#[cfg(debug_assertions)]
 static PENDING_TASK_TX: Mutex<Option<futures::channel::oneshot::Sender<()>>> = Mutex::new(None);
```
DavisVaughan: Maybe a `TEST_` prefix like `TEST_KERNEL_REQUEST_TX`?

DavisVaughan (approved):

crates/ark/src/console/console_repl.rs
lionel-: I'd like to keep random selection in the main switch to avoid starvation issues.

PR #1108, opened 2026-03-11
Fix clippy notes about safety

lionel-:
- Commits 1 and 3: Remove `unsafe` from more harp functions. That fixes the "missing safety doc" lint.

- Commit 2 (best reviewed separately): Wrap `SEXP` in a type instead of a type alias, to avoid clippy requiring `#[allow(clippy::not_unsafe_ptr_arg_deref)]` on every function taking and dereferencing a `SEXP`. The type is tagged with `#[repr(transparent)]` to ensure FFI compatibility.

  Wrapping a raw pointer in this way with transparent representation is the standard approach for type safety, see e.g. `Pin` and `NonNull`. https://doc.rust-lang.org/reference/type-layout.html#the-transparent-representation

### DIFF
--- a/crates/ark/src/debug.rs
+++ b/crates/ark/src/debug.rs
@@ -32,6 +32,7 @@ static _ARK_DISPLAY_VALUE: unsafe extern "C" fn(x: libr::SEXP) -> *const ffi::c_
 // Implementations for entry points in `debug.c`.
 
 #[cfg_attr(not(test), no_mangle)]
+#[allow(clippy::not_unsafe_ptr_arg_deref)]
 pub extern "C-unwind" fn ark_print_rs(x: libr::SEXP) -> *const ffi::c_char {
     capture_console_output(|| {
         unsafe { libr::Rf_PrintValue(x) };
@@ -81,6 +82,7 @@ pub extern "C-unwind" fn ark_trace_back_rs() -> *const ffi::c_char {
 }
 
 #[cfg_attr(not(test), no_mangle)]
+#[allow(clippy::not_unsafe_ptr_arg_deref)]
 pub extern "C-unwind" fn ark_display_value_rs(x: libr::SEXP) -> *const ffi::c_char {
     let value = unsafe {
         let kind = tidy_kind(r_typeof(x));
--- a/crates/ark/src/sys/unix/console.rs
+++ b/crates/ark/src/sys/unix/console.rs
@@ -155,6 +155,7 @@ pub extern "C-unwind" fn r_cleanup_for_tests(_save_act: i32, _status: i32, _run_
 }
 /// On Unix, we assume that the buffer to write to the console is
 /// already in UTF-8
+#[allow(clippy::not_unsafe_ptr_arg_deref)]
 pub fn console_to_utf8(x: *const c_char) -> anyhow::Result<String> {
     let x = unsafe { CStr::from_ptr(x) };
 
--- a/crates/ark/src/variables/variable.rs
+++ b/crates/ark/src/variables/variable.rs
@@ -409,6 +409,7 @@ impl WorkspaceVariableDisplayType {
     /// - value: The R object to create the display type and type info for.
     /// - include_length: Whether to include the length of the object in the
     ///   display type.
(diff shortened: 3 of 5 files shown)


crates/libr/src/types.rs
```diff
@@ -50,7 +50,37 @@ pub const FUNSXP: u32 = 99;
...
-pub type SEXP = *mut SEXPREC;
+
+/// SEXP as a newtype rather than a type alias for `*mut SEXPREC`. This way
+/// clippy doesn't see it as a raw pointer, avoiding
+/// `clippy::not_unsafe_ptr_arg_deref` on every public function that takes a
+/// SEXP. `#[repr(transparent)]` guarantees the same ABI as `*mut SEXPREC`
+/// across FFI boundaries.
+#[repr(transparent)]
+#[derive(Copy, Clone, PartialEq, Eq, Hash)]
+pub struct SEXP(pub *mut SEXPREC);
```
DavisVaughan: I'd advocate for not making `pub *mut SEXPREC` `pub`

It seems like it compiles fine for me without this

crates/harp/src/weak_ref.rs
```diff
@@ -88,21 +88,19 @@ impl RWeakRef {
...
-        let key = unsafe { libr::R_WeakRefKey(self.weak_ref.sexp) };
+        let key = libr::R_WeakRefKey(self.weak_ref.sexp);
         if key == RObject::null().sexp {
             return None;
         }
 
-        Some(RObject::new(unsafe {
-            libr::R_WeakRefValue(self.weak_ref.sexp)
-        }))
+        Some(RObject::new({ libr::R_WeakRefValue(self.weak_ref.sexp) }))
```
DavisVaughan: When I compile locally I get a note about extra braces

Image

crates/libr/src/constant_globals.rs
```diff
@@ -41,8 +41,8 @@ macro_rules! generate {
                 paste::paste! {
                     $(#[doc=$doc])*
                     $(#[cfg($cfg)])*
-                    pub unsafe fn $name() -> bool {
-                        [<$name _has>]
+                    pub fn $name() -> bool {
+                        unsafe { [<$name _has>] }
```
DavisVaughan: I do not agree with making everything exposed from libr safe.

These are by definition unsafe. I mean, it's as unsafe as you can get. You're calling the R API directly.

I'd like to talk about this if you feel strongly that we should do this, it feels very wrong to me and like a very large change. We have typically been "re-exposing" the unsafe R API with safer Rust wrappers.

DavisVaughan (changes requested):

crates/harp/src/weak_ref.rs
DavisVaughan: this seems to get removed in https://github.com/posit-dev/ark/pull/1111

crates/libr/src/constant_globals.rs
lionel-: We can talk about it, but many of our "safe" wrappers are strictly just as safe as the libr entry points (not very).

With clippy you now get lints about dereference of other raw pointer types such as `* char`. So we still have this safety net, which is more targeted.

Unless you wrap the R API with fully vetted Rust types and methods, the harp wrappers are still unsafe by nature, you have to know what you're doing to use them safely. I think making the libr entry points safe or unsafe is mostly inconsequential from this point of view.

I'll try and go back to unsafe to see what this actually buys us though, might be easy to keep it unsafe.

lionel-: yep this was just about missing Safety doc! No real need to make it safe, it was just for consistency with our harp wrappers. I agree it also makes sense for this lower level crate to provide unsafe entry points.

I've added an opt out for this lint in libr.

crates/libr/src/types.rs
lionel-: That would make it harder to convert a `*void` to a `SEXP`, which can be needed e.g. for callbacks. We could add a From method for this case though.

I'm not sure we need to prevent access on the raw pointer, which is as (un)safe as our wrapper.
I'll make it private for now though.

PR #1110, opened 2026-03-11
Util for getting active request

jmcphers:
I'm preparing to do some work to add more metadata to the execute request, and am adding this to make it easy to inspect it. 

image

### DIFF
--- a/crates/ark/src/console/console_repl.rs
+++ b/crates/ark/src/console/console_repl.rs
@@ -2794,7 +2794,7 @@ unsafe extern "C-unwind" fn ps_get_virtual_document(uri: SEXP) -> anyhow::Result
 /// Returns the currently active execute request as an R named list,
 /// or NULL if no execute request is in flight.
 #[harp::register]
-pub unsafe extern "C-unwind" fn ps_active_request() -> anyhow::Result<SEXP> {
+pub unsafe extern "C-unwind" fn ps_get_active_request() -> anyhow::Result<SEXP> {
     let Some(req) = Console::get().get_active_execute_request() else {
         return Ok(libr::R_NilValue);
     };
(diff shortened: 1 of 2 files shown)


crates/ark/src/execute_request.rs
```diff
@@ -0,0 +1,25 @@
+//
```
lionel-: I think it's a bit much to have a whole file just for this, as we're trying to keep the repo easily browsable. Can you please move this below `get_active_execute_request()` in console_repl.rs?

In the future we _could_ move these to `console/console_callbacks.rs`. That would allow seeing at a glance everything that's exported to the R side.

crates/ark/src/modules/positron/utils.R
```diff
@@ -56,6 +56,11 @@
     ark_version
 }
 
+#' @export
+.ps.active_request <- function() {
```
lionel-: Since this is just for development (IIUC), I'd remove the `@export` and `.ps.` prefix and add a comment that suggests using:

```r
.ps.internal(active_request())
```

crates/ark/src/console/console_repl.rs
```diff
@@ -2785,3 +2790,16 @@ unsafe extern "C-unwind" fn ps_get_virtual_document(uri: SEXP) -> anyhow::Result
         None => Ok(RObject::null().sexp),
     }
 }
+
+/// Returns the currently active execute request as an R named list,
+/// or NULL if no execute request is in flight.
+#[harp::register]
+pub unsafe extern "C-unwind" fn ps_active_request() -> anyhow::Result<SEXP> {
+    let Some(req) = Console::get().get_active_execute_request() else {
```
DavisVaughan:
```suggestion
pub unsafe extern "C-unwind" fn ps_get_active_execute_request() -> anyhow::Result<SEXP> {
    let Some(req) = Console::get().get_active_execute_request() else {
```

Optional nit to try and keep consistent naming

DavisVaughan (approved):

crates/ark/src/console/console_repl.rs
jmcphers: updated!

Issue #1104, opened 2026-03-11
Enable clippy linting in repo


lionel-:
https://github.com/rust-lang/rust-clippy

We use it in air: https://github.com/posit-dev/air/blob/main/clippy.toml

And it's used in most major Rust projects:
- https://github.com/biomejs/biome/blob/main/clippy.toml
- https://github.com/rust-lang/rust-analyzer/blob/master/clippy.toml
- https://github.com/astral-sh/ruff/blob/main/clippy.toml
- https://github.com/zed-industries/zed/blob/main/clippy.toml

In addition to linting towards conventional style conventions, it also makes it possible to disallow use of certain methods, functions, or types, which could be handy in the future (see clippy.toml files above for examples of this). We could disallow `.unwrap()` and `.expect()` in favour of `log_err` or explicit `panic` for instance.

PR #1111, opened 2026-03-12
Manual clippy fixes

lionel-:
Branched from #1108 
Progress towards #1104 

Main lints fixed:

- Too many arguments
- Unnecessary try_ conversions (can just use non-failing conversions)
- Large enum variants. Sometimes fixed by boxing, sometimes by disabling the lint.
- Use `&str` instead of `&String`, `&Path` instead of `&PathBuf`, `&[]` instead of `&Vec`
- Clippy doesn't like uppercase variants like `EOF`, so they've become e.g. `EndOfFile`
- Removed common prefixes/suffixes from enum variants. I added an opt out for the LSP notification enum so we can keep the `Did` prefix, e.g. `KernelNotification::DidOpenVirtualDocument`.
- `into_` methods must take `self`. Renamed those that didn't, e.g. `Dap::make_variables()` which is not really an "into" method strictly speaking.
- Simplified unnecessary uses of `match`
- Use `.is_some()` or `is_err()` methods when applicable
- Use `?` instead of explicit `return` when applicable
- Build regex in lazy globals

### DIFF
--- a/crates/libr/src/functions_variadic.rs
+++ b/crates/libr/src/functions_variadic.rs
@@ -45,7 +45,7 @@ macro_rules! generate {
                     $(#[doc=$doc])*
                     $(#[cfg($cfg)])*
                     pub unsafe fn $name() -> bool {
-                        [<$name _opt>].is_some()
+                        (*std::ptr::addr_of!([<$name _opt>])).is_some()
                     }
                 }
             )+
--- a/crates/libr/src/functions.rs
+++ b/crates/libr/src/functions.rs
@@ -43,7 +43,7 @@ macro_rules! generate {
                     $(#[doc=$doc])*
                     $(#[cfg($cfg)])*
                     pub unsafe fn $name() -> bool {
-                        [<$name _opt>].is_some()
+                        (*std::ptr::addr_of!([<$name _opt>])).is_some()
                     }
                 }
             )+


crates/amalthea/src/kernel.rs
```diff
@@ -51,29 +51,52 @@ pub enum StreamBehavior {
     None,
 }
 
+/// Handler implementations provided by the language runtime.
+pub struct Handlers {
```
lionel-, line 55: These new structs fix the "too many arguments" lint.

crates/amalthea/src/socket/comm.rs
```diff
@@ -35,6 +35,7 @@ impl CommOutgoingTx {
     }
 
     /// Send an outgoing comm message through IOPub.
+    #[allow(clippy::result_large_err)]
```
DavisVaughan: This is probably one I would perma-allow in Cargo.toml

DavisVaughan: Like this https://github.com/astral-sh/ruff/blob/3f96e100c8b1d5d026405db3ff5a7642c2d14763/Cargo.toml#L237

crates/amalthea/src/comm.rs
```diff
@@ -11,8 +11,13 @@
 
 pub mod base_comm;
 pub mod comm_channel;
+
+// The large enum variant comes from code generated from the frontend. For now
+// just silence it.
 #[rustfmt::skip]
+#[allow(clippy::large_enum_variant)]
 pub mod data_explorer_comm;
```
DavisVaughan: As of late last year, clippy has gained `expect()` rather than `allow()` which I think is quite nice.
https://users.rust-lang.org/t/dont-allow-expect-and-be-reasonable/136078

`allow()` says "_if_ there is a problem, ignore it"

`expect()` says "there _is_ a problem, ignore it, but also let me know _if it isnt there anymore_", which allows you to remove the expect() if it goes away at some point in the future

crates/amalthea/src/kernel.rs
```diff
@@ -51,29 +51,52 @@ pub enum StreamBehavior {
...
+        control_handler,
+        server_handlers,
+    } = handlers;
+    let ConnectionChannels {
+        iopub_tx,
+        iopub_rx,
+        comm_event_rx,
+        stdin_request_rx,
+        stdin_reply_tx,
+    } = channels;
```
DavisVaughan: To me, packing and immediately unpacking is more noise, and I'd just ignore clippy here.

crates/ark/src/console/console_repl.rs
```diff
@@ -585,6 +586,7 @@ impl Console {
         R_INIT.set(()).expect("`R_INIT` can only be set once");
     }
 
+    #[allow(clippy::too_many_arguments)]
```
DavisVaughan: I would personally perma ignore, I can decide if it is too many arguments or not 😛

crates/ark/src/data_explorer/format.rs
```diff
@@ -393,9 +392,9 @@ fn pad_exponent(x: String) -> String {
 #[derive(Clone)]
 enum FormattedValue {
     Unkown,
-    NULL,
-    NA,
-    NaN,
+    Null,
+    Na,
+    Nan,
```
DavisVaughan, line 403: Booo I like mimicking the R values!

crates/ark/src/lsp/traits/cursor.rs
```diff
@@ -62,15 +62,15 @@ pub trait TreeCursorExt {
...
-    /// For now, use `node.find_closest_node_for_point()`.
+    // /// Move this cursor to the node that is at or closest to (but not after)
+    // /// the requested point.
+    // ///
+    // /// Returns `false` if the cursor is after the `point`.
+    // ///
+    // /// TODO: It would be nice to use this, but there is a bug in tree sitter
+    // /// that makes it impossible right now:
+    // /// https://github.com/tree-sitter/tree-sitter/issues/2767
+    // /// For now, use `node.find_closest_node_for_point()`.
```
DavisVaughan, line 73: I think you meant to remove a `/` not add more

crates/ark/src/lsp/traits/point.rs
```diff
@@ -21,6 +21,8 @@ fn compare(lhs: Point, rhs: Point) -> i32 {
     }
 }
 
+// Point is a small Copy type, so it's fine to pass by value
+#[allow(clippy::wrong_self_convention)]
```
DavisVaughan: it is a little weird for `is_` methods though

crates/ark/src/lsp/diagnostics.rs
```diff
@@ -284,10 +284,10 @@ fn recurse(
             recurse_subset_or_subset2(node, context, diagnostics)
         },
         NodeType::Call => recurse_call(node, context, diagnostics),
-        NodeType::UnaryOperator(op) => match op {
-            UnaryOperatorType::Tilde => recurse_formula(node, context, diagnostics),
-            _ => recurse_default(node, context, diagnostics),
+        NodeType::UnaryOperator(UnaryOperatorType::Tilde) => {
+            recurse_formula(node, context, diagnostics)
         },
+        NodeType::UnaryOperator(_) => recurse_default(node, context, diagnostics),
```
DavisVaughan, line 290: im not really sure if i like that

crates/harp/src/size.rs
```diff
@@ -66,7 +64,6 @@ fn obj_size_tree(
     sizeof_node: usize,
     sizeof_vector: usize,
     seen: &mut HashSet<SEXP>,
-    depth: u32,
```
DavisVaughan, line 69: Should this have been used?

crates/libr/src/functions_variadic.rs
```diff
@@ -45,7 +46,7 @@ macro_rules! generate {
                     $(#[doc=$doc])*
                     $(#[cfg($cfg)])*
                     pub fn $name() -> bool {
-                        unsafe { matches!([<$name _opt>], Some(_)) }
+                        unsafe { (*std::ptr::addr_of!([<$name _opt>])).is_some() }
```
DavisVaughan: What is clippy upset about here? This seems fine to me.

Is it possible to leave this and `functions.rs` as is for now?

crates/amalthea/src/comm.rs
lionel-: oh that's nice!

crates/amalthea/src/kernel.rs
lionel-: I don't know, if feels a bit more tidy to me.

Although the unpacking is unnecessary, I'll remove it

crates/ark/src/lsp/traits/cursor.rs
lionel-: nope I actually commented out the docstring

crates/amalthea/src/socket/comm.rs
lionel-: yep I was thinking the same actually. I'll go ahead and unbox everything.

IIUC the lint is useful in repos where you create lots of vectors of data structures, and since the enum is as large as the largest variant, boxing improves memory usage. Well... We might do a lot more of this in the future in the LSP actually! hmmm

lionel-: Looks like individual crates can re-opt into lints disabled at workspace level if needed, so that approach sounds good to me

crates/amalthea/src/comm.rs
lionel-: Switching to `expect()` already allowed finding some dangling opt-outs for deref lints, following the SEXP repr PR!

crates/ark/src/console/console_repl.rs
lionel-: I don't feel strongly either way so I'll perma-ignore.

I do like packed arguments in the kernel connection entry points as it makes call sites easier to read and understand. But I can see cases where that doesn't make sense.

crates/harp/src/size.rs
lionel-: In lobstr it's only used in debugging print statements that are commented out.

crates/libr/src/functions_variadic.rs
lionel-: The current versions are accessing `static mut` references, which the Rust community has been trying to phase out (https://github.com/rust-lang/rust/issues/53639). The new ones are creating raw pointers instead, to be accessed by unsafe code.

Since you wrote these macros I've reverted, but I'd have kept the new versions I think.

lionel-: Actually the switch to `is_some()` creates a reference that is linted by rustc, not just clippy. I'd keep the new versions.

PR #1113, opened 2026-03-12
Fix remaining clippy issues with `--all-targets`

lionel-:
Branched from #1112 
Progress towards #1104

### DIFF
--- a/crates/libr/src/lib.rs
+++ b/crates/libr/src/lib.rs
@@ -5,6 +5,8 @@
 //
 //
 
+#![allow(clippy::missing_safety_doc)]
+
 mod constant_globals;
 mod functions;
 mod functions_variadic;


crates/stdext/src/lib.rs
```diff
@@ -51,7 +51,10 @@ macro_rules! assert_match {
     };
 
     ($expression:expr, $pattern:pat_param) => {
-        assert!(matches!($expression, $pattern))
+        #[allow(clippy::redundant_pattern_matching)]
```
DavisVaughan, line 55: do you know what this one is about?

lionel-: That prevents clippy from complaining that we could use `.is_some()` instead of `Some(_)`

PR #1119, opened 2026-03-16
Respect plot sizing metadata when present

jmcphers:
This change adds the ability to change the way plots are drawn using four new values in the `positron` field of `execute_request`

- `fig-width` / `fig-height` from Quarto (takes precedence), from the cell options
- `output_width_px` and `output_pixel_ratio` from the notebook and/or source editor. These are device metrics that can change at any time.

Used together, these values cause most plots in notebooks and Quarto documents to be drawn at the user's preferred size, and at the scaling appropriate for their display. 

Quarto:

image

Jupyter:

image

### DIFF
--- a/crates/ark/src/plots/graphics_device.rs
+++ b/crates/ark/src/plots/graphics_device.rs
@@ -655,7 +655,7 @@ impl DeviceContext {
                             Some(intrinsic) => intrinsic.to_plot_size(),
                             None => {
                                 return Err(anyhow!(
-                                    "Intrinsically sized plots are not yet supported."
+                                    "No size provided for plot {id} and no intrinsic size available"
                                 ));
                             },
                         }


crates/amalthea/src/wire/execute_request.rs
```diff
@@ -41,9 +41,23 @@ pub struct ExecuteRequest {
...
+
+    /// Figure height in inches, set by Quarto
+    #[serde(rename = "fig-height")]
+    pub fig_height: Option<f64>,
+
+    /// Output area width in pixels
+    pub output_width_px: Option<f64>,
+
+    /// Device pixel ratio of the output area
+    pub output_pixel_ratio: Option<f64>,
```
lionel-, line 60: Although the dependency direction would be a little weird, I'd love if we could reuse `PlotRenderSettings` from the plot comm for this. Looks like we only need to add `output_width_px`.

At the very least, duplicating the data structure if we can't reuse it would mean that the plot settings are nicely namespaced in their own struct and keep the parent struct tidy.

crates/ark/src/console/console_repl.rs
```diff
@@ -1288,12 +1288,25 @@ impl Console {
...
+                    let has_sizing = p.fig_width.is_some() ||
+                        p.fig_height.is_some() ||
+                        p.output_width_px.is_some();
+                    has_sizing.then_some(graphics_device::PlotSizingMetadata {
+                        fig_width: p.fig_width,
+                        fig_height: p.fig_height,
+                        output_width_px: p.output_width_px,
+                        output_pixel_ratio: p.output_pixel_ratio,
+                    })
+                });
```
lionel-: This is a bit domain-specific for this file. Perhaps make a `from_execute_request()` method on `PlotSizingMetadata`?

crates/ark/src/plots/graphics_device.rs
```diff
@@ -757,7 +867,14 @@ impl DeviceContext {
             self.iopub_tx.clone(),
         );
 
-        let settings = self.prerender_settings.get();
+        // Use sizing metadata for prerender if available, otherwise fall back to default
```
lionel-: Would be helpful to mention that we're merging any overrides available for this plot (possibly not exhaustive) into current frontend defaults.

crates/ark/src/plots/graphics_device.rs
```diff
@@ -677,6 +717,86 @@ impl DeviceContext {
...
+    const DEFAULT_DPI: f64 = 96.0;
+
+    /// Default aspect ratio (width:height) used when only output_width_px is provided.
+    const DEFAULT_ASPECT_RATIO: f64 = 4.0 / 3.0;
+
+    /// Convert an intrinsic size (in inches) to a logical-pixel-based `PlotSize`.
+    ///
+    /// Returns dimensions in CSS/logical pixels. The R rendering layer handles
+    /// physical pixel scaling via the separate `pixel_ratio` parameter.
+    fn intrinsic_size_to_plot_size(intrinsic: &IntrinsicSize) -> PlotSize {
```
lionel-: `IntrinsicSize` belongs to Amalthea but you could make an extension trait `IntrinsicSizeExt` in Ark and make this `IntrinsicSizeExt::to_plot_size()`?

crates/ark/src/plots/graphics_device.rs
```diff
@@ -677,6 +717,86 @@ impl DeviceContext {
...
+            },
+        }
+    }
+
+    /// Build an `IntrinsicSize` from plot sizing metadata, if applicable.
+    ///
+    /// If `fig_width`/`fig_height` are both set, returns an intrinsic size in inches
+    /// with source "Quarto". Otherwise returns `None` (output_width_px affects render
+    /// size but is not an intrinsic size).
+    fn intrinsic_size_from_metadata(sizing: &PlotSizingMetadata) -> Option<IntrinsicSize> {
```
lionel-: `PlotSizingMetadata::to_intrinsic_size()`

crates/ark/src/plots/graphics_device.rs
```diff
@@ -677,6 +717,86 @@ impl DeviceContext {
...
+    ///
+    /// Returns the plot size (in logical/CSS pixels) and pixel ratio to use for
+    /// pre-rendering. The R rendering layer handles physical pixel scaling
+    /// internally by multiplying dimensions and DPI by `pixel_ratio`, so the
+    /// sizes here must be in logical pixels to avoid double-scaling.
+    ///
+    /// If `fig_width`/`fig_height` are set, converts inches to logical pixels.
+    /// If only `output_width_px` is set, uses it as width with a default aspect ratio.
+    /// Otherwise returns `None` (use the default prerender settings).
+    fn prerender_overrides_from_metadata(sizing: &PlotSizingMetadata) -> Option<(PlotSize, f64)> {
```
lionel-: `PlotSizingMetadata::to_prerender_overrides()`

crates/ark/src/plots/graphics_device.rs
```diff
@@ -189,6 +207,9 @@ struct DeviceContext {
     /// Mapping of plot ID to its metadata (captured at creation time)
     metadata: RefCell<HashMap<PlotId, PlotMetadata>>,
 
+    /// Mapping of plot ID to its intrinsic size (set when Quarto specifies fig-width/fig-height)
+    intrinsic_sizes: RefCell<HashMap<PlotId, IntrinsicSize>>,
```
lionel-: We've now got 3 hashmaps over `PlotId`. Is it time for a struct?

crates/amalthea/src/wire/execute_request.rs
DavisVaughan: Yea at the very least it feels like this should be something like

```rust
pub plot_render_settings: Option<PlotRenderSettings>
```

But if @lionel- is suggesting that perhaps this should not be a part of `ExecuteRequestPositron` at all, then I'd also be fine with that (I can't quite tell)

crates/ark/src/plots/graphics_device.rs
```diff
@@ -128,12 +130,28 @@ struct WrappedDeviceCallbacks {
 #[derive(Debug, Clone, Hash, PartialEq, Eq)]
 struct PlotId(String);
 
+/// Plot sizing metadata from the execute request's `positron` field.
+/// When `fig_width`/`fig_height` are set, they take priority (Quarto-specified size in inches).
+/// Otherwise, `output_width_px` is used as the render width.
+#[derive(Debug, Clone)]
+pub(crate) struct PlotSizingMetadata {
```
DavisVaughan: Yea this does feel very close to `PlotRenderSettings`!

crates/ark/src/plots/graphics_device.rs
```diff
@@ -617,9 +644,21 @@ impl DeviceContext {
...
+                let size = match plot_meta.size {
+                    Some(size) => size,
+                    None => {
+                        // No explicit size requested — use intrinsic size if available
+                        let intrinsic = self.intrinsic_sizes.borrow().get(id).cloned();
+                        match intrinsic {
+                            Some(intrinsic) => Self::intrinsic_size_to_plot_size(&intrinsic),
+                            None => {
+                                return Err(anyhow!(
+                                    "Intrinsically sized plots are not yet supported."
```
DavisVaughan: Is that the right message? Or is it just that we somehow are missing an intrinsic size for this id?

crates/ark/src/plots/graphics_device.rs
```diff
@@ -677,6 +717,86 @@ impl DeviceContext {
         }
     }
 
+    /// Default DPI for converting inches to pixels (matches R's default on macOS).
+    const DEFAULT_DPI: f64 = 96.0;
```
DavisVaughan: FWIW, in `graphics.R`

```r
#' Default OS resolution in PPI (pixels per inch)
#'
#' Thomas thinks these are "more correct than any other numbers." Specifically,
#' macOS uses 96 DPI for its internal scaling, but this is user definable on
#' Windows.
#'
#' This corresponds to a scaling factor that tries to make things that appear
#' "on screen" be as close to the size in which they are actually printed at,
#' which has always been tricky.
default_resolution_in_pixels_per_inch <- function() {
    if (Sys.info()[["sysname"]] == "Darwin") {
        96L
    } else {
        72L
    }
}
```

crates/ark/src/plots/graphics_device.rs
```diff
@@ -677,6 +717,86 @@ impl DeviceContext {
...
+    /// If `fig_width`/`fig_height` are both set, returns an intrinsic size in inches
+    /// with source "Quarto". Otherwise returns `None` (output_width_px affects render
+    /// size but is not an intrinsic size).
+    fn intrinsic_size_from_metadata(sizing: &PlotSizingMetadata) -> Option<IntrinsicSize> {
+        match (sizing.fig_width, sizing.fig_height) {
+            (Some(w), Some(h)) if w > 0.0 && h > 0.0 => Some(IntrinsicSize {
+                width: w,
+                height: h,
+                unit: PlotUnit::Inches,
+                source: String::from("Quarto"),
```
DavisVaughan: Calling out that hardcoding Quarto doesn't feel quite right to me

crates/ark/src/plots/graphics_device.rs
```diff
@@ -914,14 +1045,26 @@ impl DeviceContext {
             .log_err();
     }
 
-    fn create_display_data_plot(&self, id: &PlotId) -> Result<serde_json::Value, anyhow::Error> {
-        // TODO: Take these from R global options? Like `ark.plot.width`?
```
DavisVaughan, line 918: This TODO should stay

DavisVaughan: For example, how would, say, Jupyterlab or Zed set this?

crates/ark/src/plots/graphics_device.rs
```diff
@@ -677,6 +717,86 @@ impl DeviceContext {
...
+    /// Otherwise returns `None` (use the default prerender settings).
+    fn prerender_overrides_from_metadata(sizing: &PlotSizingMetadata) -> Option<(PlotSize, f64)> {
+        let pixel_ratio = sizing.output_pixel_ratio.unwrap_or(1.0);
+
+        if let (Some(w), Some(h)) = (sizing.fig_width, sizing.fig_height) {
+            if w > 0.0 && h > 0.0 {
+                return Some((
+                    PlotSize {
+                        width: (w * Self::DEFAULT_DPI).round() as i64,
+                        height: (h * Self::DEFAULT_DPI).round() as i64,
```
DavisVaughan: So, IIUC, your goal is to massage the sizing information into pixels for the R side of the graphics engine

```r
#' ...
#' @param width The plot width, in pixels.
#' @param height The plot height, in pixels.
#' @param pixel_ratio The device pixel ratio (e.g. 1 for standard displays, 2
#'   for retina displays)
#' ...
with_graphics_device <- function(
```

...even if on the R side we possibly invert them back from pixels -> inches depending on the graphics device

```r
finalize_device_arguments <- function(format, width, height, pixel_ratio) {
    if (format == "png" || format == "jpeg" || format == "tiff") {
        # These devices require `width` and `height` in pixels, which is what
        # they are provided in already. For pixel based devices, all relevant
        # values are upscaled by `pixel_ratio`.
        #
        # `res` is nominal resolution specified in pixels-per-inch (ppi).
        return(list(
            res = default_resolution_in_pixels_per_inch() * pixel_ratio,
            width = width * pixel_ratio,
            height = height * pixel_ratio
        ))
    }

    if (format == "svg" || format == "pdf") {
        # These devices require `width` and `height` in inches, but they are
        # provided to us in pixels, so we have to perform a conversion here.
        # For vector based devices, providing the size in inches implicitly
        # tells the device the relative size to use for things like text,
        # since that is the absolute unit (pts are based on inches).
        #
        # Thomas says the math for `width` and `height` here are correct, i.e.
        # we don't also multiply `default_resolution_in_pixels_per_inch()` by
        # `pixel_ratio` like we do above, which would have made it cancel out of
        # the equation below.
        #
        # There is no `res` argument for these devices.
        return(list(
            res = NULL,
            width = width *
                pixel_ratio /
                default_resolution_in_pixels_per_inch(),
            height = height *
                pixel_ratio /
                default_resolution_in_pixels_per_inch()
        ))
    }

    stop("Internal error: Unknown plot `format`.")
}
```

I think that probably makes sense, but if that is the case I do think we should definitely make an attempt to keep `default_resolution_in_pixels_per_inch()` in sync with `DEFAULT_DPI` (could even give them the same name with a comment that links them together, or go all the way to making the R one just return the Rust variable)

crates/ark/src/plots/graphics_device.rs
DavisVaughan: That feels kind of wrong to me? i.e. for plot specific metadata to affect global state in this way

crates/ark/tests/plots.rs
```diff
@@ -6,6 +6,25 @@ use amalthea::wire::execute_request::JupyterPositronRange;
...
+        .or_else(|_| base64::engine::general_purpose::STANDARD_NO_PAD.decode(&cleaned))
+        .expect("Failed to decode base64 PNG data");
+    // Validate PNG signature and minimum size for IHDR
+    let png_signature: [u8; 8] = [137, 80, 78, 71, 13, 10, 26, 10];
+    assert!(bytes.len() >= 24);
+    assert_eq!(bytes[..8], png_signature);
+    // PNG IHDR: 8-byte signature, 4-byte chunk length, 4-byte "IHDR", then width (4) and height (4)
+    let width = u32::from_be_bytes([bytes[16], bytes[17], bytes[18], bytes[19]]);
+    let height = u32::from_be_bytes([bytes[20], bytes[21], bytes[22], bytes[23]]);
+    (width, height)
```
DavisVaughan, line 35: That is wild

crates/amalthea/src/wire/execute_request.rs
lionel-: It's fine in the execute request, just needs some consolidation as in your snippet.

crates/ark/src/plots/graphics_device.rs
lionel-: The global state is copied locally from the Cell, then the overrides are merged in the copy and used in the pre-render. Not global side effect IIUC.

crates/amalthea/src/wire/execute_request.rs
jmcphers: Consolidated in https://github.com/posit-dev/ark/pull/1119/commits/bed45b05f190ad9ca1b43fac4c75fae34701adbd

crates/ark/src/plots/graphics_device.rs
jmcphers: Well, while I'm in here I guess I could just make it read from global options :-) https://github.com/posit-dev/ark/pull/1119/commits/11542ad89751c63a98383ff7ac3579b95cf4009a

crates/ark/src/plots/graphics_device.rs
jmcphers: Decided to just use `IntrinsicSize` directly

crates/ark/src/plots/graphics_device.rs
jmcphers: Good question! The one that tracks sockets has a different lifetime, but I've combined the other two.

crates/ark/src/plots/graphics_device.rs
jmcphers: Refactored out of `PlotSizingMetadata`

crates/ark/src/plots/graphics_device.rs
jmcphers: Removed in favor of `PlotRenderSettings`!

crates/ark/src/plots/graphics_device.rs
jmcphers: yep! https://github.com/posit-dev/ark/pull/1119/commits/1bd395056d904a4752eaf043a4d1177bccac11ec

crates/ark/src/plots/graphics_device.rs
jmcphers: It does feel a LITTLE weird to go through inches -> px -> inches here! But don't think there's any arrangement of this that isn't awkward somewhere.

I've updated it so that Rust is the keeper of the default DPI setting and R fetches it from there.

crates/ark/tests/plots.rs
jmcphers: You can thank Claude for this one, its initial run at test coverage for this feature was 200 lines of tests to be sure that DPI -> inch conversion was correct (yes Claude good job I am worried that someone might break the _division operator_ in a future change).

I asked it to do a more meaningful higher-level test instead and it came up with the idea of checking the actual size of the emitted PNG which I would never have thought to have done!

crates/ark/src/plots/graphics_device.rs
jmcphers: Good catch, it's the latter one. I've updated the message.

crates/ark/src/plots/graphics_device.rs
jmcphers: There's no other way for these values to be set AFAIK... I think the only way to avoid this would be to give this intrinsic size a generic name (like "document"? "external"?), or to add another field to the metadata naming the source of the figure size values that Positron could set. Does either of those seem right to you?

crates/amalthea/src/wire/execute_request.rs
lionel-: Nice! In this case though, I meant collecting all the plot setting fields in a nested struct in the `execute_request.positron` struct.

DavisVaughan, line 60: I believe I would be ok with keeping these nested 1 layer deeper under `execution_metadata: Option<ExecutionMetadata>`, mirroring your initial implementation here https://github.com/posit-dev/positron/pull/12513

IIUC, we could frame it as:

- `code_location` is "officially supported" by Positron and is Positron provided
- `execution_metadata` is "passed through" by Positron and may contain fields that any extension feels like passing through in their `session.execute()` calls.

On the Ark side, we'd just "happen to recognize" some specific `execution_metadata` fields of `fig_width`, `fig_height`, etc, and ignore (probably?) anything else that comes over the wire that we don't recognize.

But I like the idea of containing these extension-specific-fields-that-ark-happens-to-recognize scoped to `execution_metadata`

crates/ark/src/plots/graphics_device.rs
DavisVaughan: So I'm looking at 

```
pub struct IntrinsicSize {
	/// The width of the plot
	pub width: f64,

	/// The height of the plot
	pub height: f64,

	/// The unit of measurement of the plot's dimensions
	pub unit: PlotUnit,

	/// The source of the intrinsic size e.g. 'Matplotlib'
	pub source: String
}
```

and mentally I'm understanding `source` here as "the thing that provided this width/height/unit override"

So to me it would make sense for this proposed `ExecutionMetadata` to have a `source?: String` field that is "the source of all metadata overrides", which Quarto would set to itself when it calls `session.execute()` and would flow through to ark.

That feels kind of right to me, and may be useful for other things in the future?

If `source` is not set in `ExecutionMetadata`, then as a fallback you could probably say that the `source` for the `IntrinsicSize` is something like `"Metadata"`, but I also feel like it would be kind of reasonable for `IntrinsicSize`'s `source` field to be optional, and we'd just not provide anything at all if `ExecutionMetadata` lacks the `source` field but does happen to provide `fig_height` and `fig_width`. Sort of a recognition of "we got these things but we don't know where they originally came from". And if Positron displays this information somehow to the user, it would just not show the `source` field or would show `???` if it had to show something.

Does that make any sense?

crates/amalthea/src/wire/execute_request.rs
lionel-: hmm I think there is still an agreed upon protocol here. In this case, it seems it's one that is owned by Ark: Ark knows about these fields and Quarto is using them.

So, should it be in `execute_request.ark` rather than `execute_request.positron`? And then we can nest in `plot_render_settings`: this is no longer pass through, this is part of an established and documented data structure.

On the frontend side, passthrough metadata is taken with a namespace string (in this case set to `'ark'`) that represents the field to inject in the execute request.

wdyt @jmcphers @DavisVaughan?

lionel-: Or the other way around, `execute_request.quarto`.

> On the frontend side, passthrough metadata is taken with a namespace string (in this case set to 'ark') that represents the field to inject in the execute request.

There could be multiple namespaced sets of params, so I guess it's just a passthrough object and caller makes sure the params are appropriately namespaced (like `{ quarto: { ...}, ark: { ... } }`, then Positron injects these in the execute request.

jmcphers:
@lionel- @DavisVaughan I agree that it'd be better to namespace the execution metadata in some way to clarify the provenance of the fields, but as @lionel- noted the right nesting structure isn't immediately obvious, plus will require us to sort out things upstream in Positron as well. I'm going to merge this as is to get the functionality working e2e and file a followup issue to decide on how we will nest/group metadata.

PR #1123, opened 2026-03-21
Set options lazily and conditionally

lionel-:
~Addresses https://github.com/posit-dev/positron/issues/8450~

Set our options after the user R profiles have been sourced, and only if the options haven't been set by the user already. We detect user overrides in one of two ways:

- If R sets the option to a non-null value by default, we require the user to wrap the option in `AsIs`, which we take as a sign it was overridden in the Rprofile.

- For all other options, if we see a non-null value it means it was overridden.

I've taken the opportunity to consolidate all options settings in one place (the lazily invoked init function), and document our overrides in `doc/configuration.md`.

To launch Shiny applications in the system browser you need the following configuration:

```r
options(
  browser = I(Sys.getenv("R_BROWSER")),
  shiny.launch.browser = TRUE
)
```

An alternative to this approach is to recommend using the new hook that @jennybc is implementing in #1088. The approach implemented in this PR has the advantage that the user configuration also works in the R cli without any duplication.

### DIFF
--- a/crates/ark/tests/options.rs
+++ b/crates/ark/tests/options.rs
@@ -1,0 +1,77 @@
+use std::io::Write;
+
+use ark_test::DummyArkFrontendRprofile;
+
+// These tests verify that options set in `.Rprofile` interact correctly with
+// `initialize_options()` during startup. Each test needs its own process
+// because `DummyArkFrontendRprofile` can only be locked once, which is
+// handled by nextest.
+
+#[test]
+fn test_override_option_replaces_user_value() {
+    let mut file = tempfile::NamedTempFile::new().unwrap();
+    writeln!(file, "options(max.print = 500)").unwrap();
+    unsafe { std::env::set_var("R_PROFILE_USER", file.path()) };
+
+    let frontend = DummyArkFrontendRprofile::lock();
+
+    frontend.execute_request("getOption('max.print')", |result| {
+        assert_eq!(result, "[1] 1000");
+    });
+}
+
+#[test]
+fn test_protected_override_option_keeps_user_value() {
+    let mut file = tempfile::NamedTempFile::new().unwrap();
+    writeln!(
+        file,
+        "options(max.print = 500, positron.protected_options = 'max.print')"
+    )
+    .unwrap();
+    unsafe { std::env::set_var("R_PROFILE_USER", file.path()) };
+
+    let frontend = DummyArkFrontendRprofile::lock();
+
+    frontend.execute_request("getOption('max.print')", |result| {
+        assert_eq!(result, "[1] 500");
+    });
+}
+
+#[test]
+fn test_default_option_keeps_user_value() {
+    let mut file = tempfile::NamedTempFile::new().unwrap();
+    writeln!(file, "options(help_type = 'text')").unwrap();
+    unsafe { std::env::set_var("R_PROFILE_USER", file.path()) };
+
+    let frontend = DummyArkFrontendRprofile::lock();
+
+    frontend.execute_request("getOption('help_type')", |result| {
+        assert_eq!(result, "[1] \"text\"");
+    });
+}
+
+#[test]
+fn test_default_option_sets_when_null() {
+    let mut file = tempfile::NamedTempFile::new().unwrap();
+    writeln!(file).unwrap();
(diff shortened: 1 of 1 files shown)


crates/ark/src/modules/positron/options.R
```diff
@@ -5,47 +5,111 @@
...
+        return(invisible())
+    }
+    do.call(options, set_names(list(value), name))
+}
+
+# Set an option only when it is currently NULL (i.e. the user hasn't set it),
+# also respects `positron.protected_options`
+set_when_null <- function(name, value) {
+    if (is_protected(name)) {
+        return(invisible())
```
DavisVaughan: since `set_when_null()` also respected the protected option, maybe a single helper like this is more self documenting

```
set_unless_protected <- function(name, value, only_set_if_null = FALSE)
```

doc/configuration.md
```diff
@@ -16,32 +16,74 @@ If `TRUE` and if ragg \>=1.4.0 is installed, then ragg will be used for graphics
...
 
 If `TRUE`, errors will be entraced by `rlang::entrace()` if rlang is installed. This often results in more informative backtraces.
+
+# R options
+
+Ark overrides a number of R options during startup, *after* sourcing the user's `.Rprofile`. When an option is `NULL`, we know the user hasn't set it and we apply our default. However, a subset of base R options have non-`NULL` defaults, which makes it impossible to reliably detect user overrides. For those options, you can list them in `positron.protected_options` to tell Positron to preserve the existing value (either R's default or the value set in Rprofile).
+
+## `positron.protected_options`
+
+A character vector of option names that Positron should leave alone. Set this in your `.Rprofile` to prevent Positron from overriding specific options. Accepted values are the option names listed under *Set unless protected* below. It can also be used for any of the options listed under *Set if not `NULL`* and *Package options* to preserve their `NULL` default. For example:
```
DavisVaughan: I think you should say something like 

"Note that if you override these, some Positron behavior may no longer work. For example, overriding `browser` means that URLs will no longer open in Positron's integrated Viewer pane."

doc/configuration.md
```diff
@@ -16,32 +16,74 @@ If `TRUE` and if ragg \>=1.4.0 is installed, then ragg will be used for graphics
...
+### R options
+
+- **`help_type`** — Set to `"html"`. Displays help pages as HTML rather than plain text.
+- **`viewer`** — Displays HTML content in Positron's Viewer pane. Used by htmlwidgets, Shiny gadgets, etc.
+
+### Package options
+
+- **`shiny.launch.browser`** — Set to a function that shows the Shiny app URL in Positron's Viewer pane. Set `options(shiny.launch.browser = TRUE)` in your `.Rprofile` to open Shiny apps in the system browser instead. Note that `TRUE` causes Shiny to use the `browser` option, which defaults to Positron's viewer, so you may also want to protect `browser`.
+- **`plumber.docs.callback`** — Set to a function that shows Plumber API docs in Positron's Viewer pane.
+- **`profvis.print`** — Set to a function that renders profvis output and displays it in Positron's Viewer pane.
```
DavisVaughan: I think you can collapse these two sections to be more symmetric with `## Set unless protected`

It's pretty obvious that they are R vs Package options imo

crates/ark/src/modules/positron/viewer.R
```diff
@@ -5,7 +5,7 @@
 #
 #
```
DavisVaughan, line 6: This all seems like something that we could very easily add a few tests for!

It would be nice to confirm that "keep the user's value" is working as intended via some tests

DavisVaughan (approved):

doc/configuration.md
lionel-: Good point, I definitely planned to mention this in the positron-level setting (and categorise it in advanced settings).

crates/ark/src/modules/positron/options.R
lionel-: I see what you mean, but I'd rather keep the two different helpers for the two strategies.

I've instead renamed to `set_override()` (for options for which we override the value) and `set_default()` (for options that we only set if not `NULL`). This way the names do not convey wrong implications about protection or lack thereof.

PR #1124, opened 2026-03-24
Add inline data explorers to Ark

jmcphers:
This change updates the R kernel to supply inline data explorers in notebook mode (which encompasses both Quarto documents and Positron notebooks). 

image

The approach is very symmetric to how the Python kernel provides the same feature (see https://github.com/posit-dev/positron/pull/11732); when a data frame-like object is printed, and we're in a Positron notebook, we open up a new data explorer comm and send it to the front end with a special MIME type.

### DIFF
--- a/crates/ark/src/console/console_integration.rs
+++ b/crates/ark/src/console/console_integration.rs
@@ -179,7 +179,7 @@ impl Console {
     ) -> anyhow::Result<serde_json::Value> {
         let data = RObject::new(value);
 
-        // `source` is the R class family (e.g. "tibble", "data.table",
+        // `source` is the R class family (e.g. "tbl_df", "data.table",
         // "data.frame"), following the Python kernel convention where `source`
         // is the library name ("pandas", "polars").
         let source = data


crates/ark/src/console/console_repl.rs
```diff
@@ -1131,24 +1138,84 @@ impl Console {
...
-                match to_html(value) {
-                    Ok(html) => {
-                        data.insert("text/html".to_string(), json!(html));
+        // Include HTML representation of data.frame and optionally open an
+        // inline data explorer in Positron notebook mode. Only do this when
+        // there is visible output (autoprint produced text/plain).
+        let Ok(value) = harp::environment::last_value() else {
+            return data;
+        };
+        if !data.is_empty() && r_is_data_frame(value.sexp) {
```
lionel-: nit: early return for `data.is_empty()` case? Then the following branches (only data frames for now) are for special data types.

crates/ark/src/console/console_repl.rs
```diff
@@ -1131,24 +1138,84 @@ impl Console {
...
+                }
             }
         }
 
         data
     }
 
+    /// Open an inline data explorer for a data frame value and return the MIME
+    /// type payload to include in the execute result.
+    fn open_inline_data_explorer(&mut self, value: SEXP) -> anyhow::Result<serde_json::Value> {
```
lionel-: I think I'd move this to `console_integrations.rs`, since it's comm-related?

crates/ark/src/console/console_repl.rs
```diff
@@ -1131,24 +1138,84 @@ impl Console {
...
+        let data = RObject::new(value);
+
+        // `source` is the R class family (e.g. "tibble", "data.table",
+        // "data.frame"), following the Python kernel convention where `source`
+        // is the library name ("pandas", "polars").
+        let source = harp::utils::r_classes(value)
+            .and_then(|classes| classes.get_unchecked(0).map(|s| s.to_string()))
+            .unwrap_or_else(|| String::from("data.frame"));
```
lionel-: `RObject::class()` is likely more ergonomic

crates/ark/src/console/console_repl.rs
```diff
@@ -1131,24 +1138,84 @@ impl Console {
...
 
         data
     }
 
+    /// Open an inline data explorer for a data frame value and return the MIME
+    /// type payload to include in the execute result.
+    fn open_inline_data_explorer(&mut self, value: SEXP) -> anyhow::Result<serde_json::Value> {
+        let data = RObject::new(value);
+
+        // `source` is the R class family (e.g. "tibble", "data.table",
```
lionel-:
```suggestion
        // `source` is the R class family (e.g. "tbl_df", "data.table",
```

crates/ark/src/data_explorer/r_data_explorer.rs
```diff
@@ -173,6 +200,7 @@ impl RDataExplorer {
         title: String,
         data: RObject,
         binding: Option<DataObjectEnvInfo>,
+        inline_only: bool,
```
lionel-: Suggestion: take an enum so the call site is self-documenting

```rust
enum DataExplorerMode {
    Inline,
    Full,
}
```

crates/ark/src/data_explorer/r_data_explorer.rs
```diff
@@ -416,13 +449,25 @@ impl RDataExplorer {
             DataExplorerBackendRequest::SuggestCodeSyntax => Ok(
                 DataExplorerBackendReply::SuggestCodeSyntaxReply(self.suggest_code_syntax()),
             ),
+
+            DataExplorerBackendRequest::SetDatasetImportOptions(_) => {
+                Err(anyhow!("Data Explorer: Not yet supported"))
+            },
+
+            DataExplorerBackendRequest::OpenDataExplorer => {
```
lionel-, line 463: Worth a comment to mention that this is promoting an inline data explorer to a full explorer.

That's also an example of how changing `false` to `DataExplorerMode::Full` would help understand the context / flow.

crates/ark/tests/kernel-notebook-data-explorer.rs
```diff
@@ -0,0 +1,110 @@
...
+use ark_test::DummyArkFrontendNotebook;
+
+/// Open the UI comm if it hasn't been opened yet. Since tests in this file
+/// share a single kernel process and run serially (via `lock()`), the UI comm
+/// only needs to be opened once. Returns the UI comm ID.
+fn ensure_ui_comm(frontend: &DummyArkFrontendNotebook) -> String {
+    use std::sync::OnceLock;
+    static UI_COMM_ID: OnceLock<String> = OnceLock::new();
+    UI_COMM_ID.get_or_init(|| frontend.open_ui_comm()).clone()
+}
```
lionel-: Just noting that this is not necessary since with nextest we run each test in its own process. Agents seem to have trouble recognising this, I'll try and add some colour about this in AGENTS.md.

crates/ark/tests/kernel-notebook-data-explorer.rs
```diff
@@ -0,0 +1,110 @@
...
+
+/// Drain the UI comm messages that arrive during execution (busy=true,
+/// busy=false, prompt_state). These are CommMsg messages on the UI comm's
+/// channel that interleave with the execute result on IOPub.
+fn drain_ui_comm_msgs(frontend: &DummyArkFrontendNotebook, ui_comm_id: &str) {
```
lionel-: In #1100 I changed the UI comm in tests to ignore busy/idle messages unless specifically requested, to avoid having to deal with this. I'll update this file too. Just noting it.

crates/ark/tests/kernel-notebook-data-explorer.rs
```diff
@@ -0,0 +1,110 @@
...
+    assert!(de_data["title"].as_str().is_some());
+
+    // prompt_state arrives after execute_result
+    drain_ui_comm_prompt_state(&frontend, &ui_comm_id);
+
+    frontend.recv_iopub_idle();
+    frontend.recv_shell_execute_reply();
+
+    // The comm_open for the inline data explorer arrives after Idle
+    // (it goes through Shell's comm event channel)
```
lionel-, line 72: That's changing with #1100, the R thread will be able to open new comms while an execute request is in flight. This improves UX when opening multiple comms/widgets in a long running expression. I'll adjust the ordering if you merge before.

crates/ark/tests/kernel-notebook-data-explorer.rs
```diff
@@ -0,0 +1,110 @@
...
+    let comm_open = frontend.recv_iopub_comm_open();
+    assert_eq!(comm_open.target_name, "positron.dataExplorer");
+    assert_eq!(comm_open.data["inline_only"], true);
+    assert_eq!(comm_open.comm_id, de_data["comm_id"].as_str().unwrap());
+}
+
+#[test]
+fn test_notebook_no_inline_data_explorer_for_non_data_frame() {
```
lionel-, line 80: This test is missing:

```rust
    unsafe { std::env::set_var("POSITRON", "1") };
```

Maybe it's worth setting in a new constructor, like `DummyArkPositronNotebook::lock()`?

crates/ark/src/console/console_repl.rs
jmcphers: done!

crates/ark/src/console/console_repl.rs
jmcphers: moved it.

crates/ark/src/console/console_repl.rs
jmcphers: yep!

crates/ark/src/console/console_repl.rs
jmcphers: Made this change in `console_integration.rs` after moving the function.

crates/ark/src/data_explorer/r_data_explorer.rs
jmcphers: Oh, that does read nicer! Done.

crates/ark/src/data_explorer/r_data_explorer.rs
jmcphers: done!

crates/ark/tests/kernel-notebook-data-explorer.rs
jmcphers: Cleaned up. There _is_ a note in `AGENTS.md` but the agent indeed thought this was necessary...

crates/ark/tests/kernel-notebook-data-explorer.rs
jmcphers: 👍

crates/ark/tests/kernel-notebook-data-explorer.rs
jmcphers: sure! done in https://github.com/posit-dev/ark/pull/1124/commits/dbbac19ff94c8b6a5c4b9e45e86e87d268e05111

PR #1125, opened 2026-03-30
Fix restoration of debugger environment on error

lionel-:
Addresses https://github.com/posit-dev/positron/issues/12642

This fixes a bug in the round-trip section of the ReadConsole state machine. We need to perform round trips (yield back to R with an invisible evaluation):

- After an error: reset R state (recovery round-trip)
- When debugging: capture the evaluation environment (env-capture round-trip)

Previously the recovery round-trip happened after the env-capture one, which had an early return in case recovery was needed. But that guard also blocked capture on the next call once the recovery roundtrip completed, because `last_error` was still `Some()`.

Instead of fixing the guard, I restructured the state machine to make the transistions clearer. We now have an explicit error recovery state which we handle first before attempting any environment capture.

### DIFF
--- a/crates/ark/src/console/console_repl.rs
+++ b/crates/ark/src/console/console_repl.rs
@@ -256,6 +256,19 @@ pub(super) enum ReadConsolePendingAction {
     #[default]
     None,
 
+    /// An error longjumped out of our eval. Give R one REPL iteration to
+    /// restore internal state (`R_EvalDepth`, `R_PPStackTop`, etc.) before
+    /// we do anything else.
+    ///
+    /// Technically this also resets time limits (see
+    /// `base::setTimeLimit()`) but these aren't supported in Ark
+    /// because they cause errors when we poll R events.
+    ///
+    /// References:
+    /// - R_PPStackTop: https://github.com/r-devel/r-svn/blob/74cd0af4/src/main/main.c#L227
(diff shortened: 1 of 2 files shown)


crates/ark/src/console/console_repl.rs
```diff
@@ -2491,6 +2506,17 @@ pub extern "C-unwind" fn r_read_console(
             R_ENVS.global.into()
         },
 
+        ReadConsolePendingAction::RecoverFromError => {
+            // Evaluate last value so that `base::.Last.value` remains the same
+            Console::on_console_input(
```
DavisVaughan, line 2554:
```suggestion
            // Evaluate last value so that `base::.Last.value` remains the same
            // throughout the recovery process
            Console::on_console_input(
```

Just a little more on the "why"

crates/ark/tests/kernel-execute-error.rs
```diff
@@ -154,7 +154,6 @@ options(error = function() menu("ouch"))
 
     assert!(frontend.recv_iopub_execute_error().contains("foo"));
 
-    frontend.assert_stream_stdout_contains("Enter an item from the menu, or 0 to exit");
```
DavisVaughan, line 157: Purposeful change?

DavisVaughan: Seems like we should get this?

lionel-: We now get this because we fail as soon as the prompt arrives

 Image

PR #1127, opened 2026-04-01
Flush autoprint before readline

lionel-:
Addresses https://github.com/posit-dev/positron/issues/12688

### DIFF
--- a/crates/ark/tests/kernel-stdin.rs
+++ b/crates/ark/tests/kernel-stdin.rs
@@ -154,3 +154,40 @@ fn test_stdin_from_menu() {
 
     assert_eq!(frontend.recv_shell_execute_reply(), input.execution_count);
 }
+
+#[test]
+fn test_stdin_readline_during_autoprint() {
+    let frontend = DummyArkFrontend::lock();
+
+    let options = ExecuteRequestOptions {
+        allow_stdin: true,
+        ..Default::default()
+    };
+
+    // The print method writes output with `cat()` then calls `readline()`.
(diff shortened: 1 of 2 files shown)


crates/ark/tests/kernel-stdin.rs
```diff
@@ -154,3 +154,40 @@ fn test_stdin_from_menu() {
...
+
+    // The print method writes output with `cat()` then calls `readline()`.
+    // Since the object is the last expression, R auto-prints it via an
+    // inlined `print()` call. `is_auto_printing()` returns true so `cat()`
+    // output gets buffered in `autoprint_output`. That buffer must be
+    // flushed as stream stdout before the input request is sent to the
+    // frontend, otherwise the user sees only the bare prompt without the
+    // preceding question text (https://github.com/posit-dev/positron/issues/12688).
+    let code =
+        "print.test_input <- function(x, ...) { cat('question?\\n'); readline('prompt>') }\nstructure(1, class = 'test_input')";
```
DavisVaughan, line 175: I also think usethis does this A LOT because the pattern there is often:

```
inform("Menu header")
menu(options)
```

Where they dont use the `title =` argument of `menu()` and instead use `inform()`

DavisVaughan: See `usethis:::ui_yeah`

jennybc: That would explain why this problem sounds so familiar to me, yet I couldn't quite remember why 🤪

PR #1139, opened 2026-04-03
Rename `positron.error_entrace` to `ark.error_entrace`

lionel-:
Breaking change, but for a niche option so I'm not too worried.

This option is relevant in Jupyter apps too so the right prefix is `ark`.

### DIFF
--- a/doc/configuration.md
+++ b/doc/configuration.md
@@ -4,6 +4,12 @@ This guide details Ark's configuration.
 
 # Ark options
 
+## `ark.error_entrace`
+
+A boolean, with a default value of `TRUE`.
+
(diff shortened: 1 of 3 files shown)


DavisVaughan (approved):

doc/configuration.md
```diff
@@ -42,7 +42,7 @@ A boolean, with a default value of `FALSE`.
 
 If `TRUE`, the special `.Last.value` will be shown in Positron's Variables pane.
 
-## `positron.error_entrace`
+## `ark.error_entrace`
```
DavisVaughan: move this to ark options section

PR #1140, opened 2026-04-07
Support notifications from frontend

lionel-:
Part of https://github.com/posit-dev/positron/issues/7448
Backend side of https://github.com/posit-dev/positron/pull/12859

- "Plot render settings" and "frontend ready" are now notifications instead of RPCs with ignored results.
- New `handle_comm_message()` helper that takes care of unserialisation and dispatch to RPC and event paths.


jmcphers (approved):
Works great with the Positron changes.

Issue #1141, opened 2026-04-08
Implement the semantic index


lionel-:
Part of https://github.com/posit-dev/ark/issues/1117
Part of https://github.com/posit-dev/positron/issues/2321

To support static analysis in the LSP, we plan to introduce a semantic index modelled after ty's implementation (https://github.com/astral-sh/ruff/tree/main/crates/ty_python_semantic/src/semantic_index). The index contains in particular:

- A tree of scopes with a symbol table for every scope. Symbol flags indicate for each scope whether a symbol is used or bound.
- Definitions indexed by syntax node (`definitions_by_node`), each carrying a `DefinitionKind` that classifies the syntactic construct (assignment, parameter, for-variable, etc.). This is the input to type inference: given a definition, inspect the syntax node to determine the type. As such, definitions bridge the semantic index and type checking.
- Use-def maps which track which definitions reach each use through control flow.

This data structure will be built in a similar fashion to how the current diagnostics for e.g. unused variables work (https://github.com/posit-dev/ark/blob/main/crates/ark/src/lsp/diagnostics.rs), with recursive descent over the syntax tree. Whereas our current diagnostics path mixes scope analysis and diagnostics emission, we plan to separate the concerns in the future by first building the semantic index, then using it to emit diagnostics.

The index will provide the foundation for new features (like rename and jump-to-def) and existing ones:

- Rename: scope-aware, within a file https://github.com/posit-dev/positron/issues/4525
- Go-to-definition: jump to binding site https://github.com/posit-dev/positron/issues/8631
- Completions: symbols in scope at cursor
- Outline / document symbols: top-level bindings
- Diagnostics: "symbol not in scope" with proper scoping

It will also be the input for our first type inference (is this binding a function, supporting diagnostics like "object of type closure is not subsettable").

Design-wise, the index is a nice middle ground regarding the thoughts in https://github.com/rust-lang/rust-analyzer/issues/8713 and https://hackmd.io/@matklad/rJzQhvk2u.

- There is no separate IR or lossy lowering as in Rust-Analyzer, instead the index is a flat overlay on the syntax tree (arenas of metadata laid out in preorder and keyed by `TextRange`).

[...]

PR #1143, opened 2026-04-09
Draft semantic index

lionel-:
Part of #1141 

Implements scopes and symbol tables. Prelude to implementing use-def maps.

To get oriented:
- See comments documenting the new data structures in `src/semantic_index.rs`
- See `tests/builder.rs` documenting the behaviour of the recursive descent in `src/builder.rs`.

### DIFF
--- a/crates/oak_index/src/builder.rs
+++ b/crates/oak_index/src/builder.rs
@@ -112,8 +112,35 @@ impl SemanticIndexBuilder {
     }
 
     fn add_binding(&mut self, name: &str, flags: SymbolFlags, range: TextRange) {
-        let symbol = self.symbol_tables[self.current_scope].intern(name, flags);
-        self.bindings[self.current_scope].push(Binding { symbol, range });
+        self.add_binding_in_scope(self.current_scope, name, flags, range);
+    }
+
+    fn add_binding_in_scope(
+        &mut self,
+        scope: ScopeId,
+        name: &str,
+        flags: SymbolFlags,
+        range: TextRange,
+    ) {
+        let symbol = self.symbol_tables[scope].intern(name, flags);
+        self.bindings[scope].push(Binding { symbol, range });
+    }
+
+    /// Walk from `current_scope` up through ancestors looking for a scope
+    /// that already has a binding for `name`. Returns the file scope if
+    /// no existing binding is found (matching R's runtime `<<-` semantics).
+    fn resolve_super_target(&self, name: &str) -> ScopeId {
+        let file = ScopeId::from(0);
+        let mut scope = self.scopes[self.current_scope].parent;
+        while let Some(id) = scope {
+            if let Some(sym) = self.symbol_tables[id].get(name) {
+                if sym.flags().contains(SymbolFlags::IS_BOUND) {
+                    return id;
+                }
+            }
+            scope = self.scopes[id].parent;
+        }
+        file
     }
 
     fn add_use(&mut self, name: &str, range: TextRange) {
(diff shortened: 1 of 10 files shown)


crates/oak_index/tests/builder.rs
DavisVaughan, line 1: I'm surprised you are using a separate `tests/` folder rather than using unit tests in `src/builder.rs` directly

crates/oak_index/src/index_vec.rs
DavisVaughan, line 1: I think I still mostly prefer `index-vec.rs` as a name if we are keeping `IndexVec` as the type name

crates/oak_index/src/builder.rs
```diff
@@ -0,0 +1,471 @@
...
+            },
+
+            AnyRExpression::RBogusExpression(_) => {},
+
+            // Generic fallback: walk over descendant nodes and collect their
+            // `AnyRExpression` children, letting `collect_expression`
+            // handle their contents. This covers `RIfStatement`,
+            // `RWhileStatement`, `RRepeatStatement`, `RUnaryExpression`,
+            // `RParenthesizedExpression`, `RReturnExpression`, literals, and
+            // any future expression types without needing explicit arms.
```
DavisVaughan, line 250: I think I would not mind explicitly listing all arms that fall through to here

So if or when we add more, we are forced to confront them via the compiler and can ensure they don't need special casing.

Plus, I just like seeing what goes through the fallthrough. Like, in the comment you already listed like 7 different types that come through here, you may as well just encode them in the code instead and remove the comment!

crates/oak_index/src/builder.rs
```diff
@@ -0,0 +1,471 @@
...
+            },
+        }
+    }
+
+    // Walk descendant nodes of `expr`, collecting the outermost
+    // `AnyRExpression` nodes and recursing into them via `collect_expression`.
+    // This skips intermediate wrapper nodes (e.g. `RElseClause`) while
+    // correctly stopping at expression boundaries.
+    fn collect_descendants(&mut self, node: &RSyntaxNode) {
+        let mut preorder = node.preorder();
```
DavisVaughan, line 267: well that turned out nice to be able to reuse that

crates/oak_index/src/semantic_index.rs
```diff
@@ -0,0 +1,454 @@
...
+// by function declarations resolved via salsa queries).
+#[derive(Debug)]
+pub struct Scope {
+    pub(crate) parent: Option<ScopeId>,
+    pub(crate) kind: ScopeKind,
+    pub(crate) range: TextRange,
+    // Scopes are allocated in preorder, so a scope's descendants always
+    // occupy a contiguous range of IDs. This lets `child_scopes` and
+    // `scope_at` work by range arithmetic instead of pointer chasing.
+    pub(crate) descendants: Range<ScopeId>,
```
DavisVaughan, line 158: Oh descendants being a Range of indices is so cool

DavisVaughan (approved):

crates/oak_index/tests/builder.rs
lionel-: I wouldn't write a giant suite of test alongside the implementation. If unit tests are needed instead of integration tests (for scoping reasons), I'd probably do `src/tests/builders.rs`.

crates/oak_index/src/builder.rs
lionel-: That's what I had initially but that means that in addition to all that explicit walking code you need comprehensive unit tests to make sure you're following through all children, and that an upstream change will not cause you to miss any. I think that additional code would be distracting: it would hint as something special going on, or at least would cause readers to wonder if it's fully correct.

Compared to ruff that doesn't use rowan, we are lucky to have a generic node type that can be iterated with a simple Rust iterator. Correctness falls out of that rather than from the code not making any mistakes. I think that's really nice.

PR #1144, opened 2026-04-10
Implement use-def maps

lionel-:
Part of #1141.

Provides a map of definitions for every uses (references to a variable) encountered in code. Unlocks accurate file-local features like rename, goto definition, find references.

See module-level documentation in `use_def_maps.rs` to get oriented. In particular the discussions regarding complications created by loops and `<<-`.

### DIFF
--- a/crates/oak_index/src/builder.rs
+++ b/crates/oak_index/src/builder.rs
@@ -73,6 +73,9 @@ impl SemanticIndexBuilder {
             descendants: ScopeId::from(1)..ScopeId::from(1),
         });
 
+        // All `ScopeId`-indexed vecs must be pushed in lockstep so they stay
+        // the same length. The `push_scope()` method is in charge of
+        // guaranteeing that invariant after construction.
         symbol_tables.push(SymbolTableBuilder::new());
         definitions.push(IndexVec::new());
         uses.push(IndexVec::new());
@@ -188,9 +191,8 @@ impl SemanticIndexBuilder {
     // found (mirroring R's assignment to the global environment).
     fn resolve_super_target(&self, name: &str) -> ScopeId {
         let file_scope = ScopeId::from(0);
-        let mut scope = match self.scopes[self.current_scope].parent {
-            Some(parent) => parent,
-            None => return file_scope,
+        let Some(mut scope) = self.scopes[self.current_scope].parent else {
+            return file_scope;
         };
 
         loop {
@@ -203,10 +205,10 @@ impl SemanticIndexBuilder {
                     return scope;
                 }
             }
-            scope = match self.scopes[scope].parent {
-                Some(parent) => parent,
-                None => return file_scope,
+            let Some(parent) = self.scopes[scope].parent else {
+                return file_scope;
             };
+            scope = parent;
         }
     }
 
--- a/crates/oak_index/src/builder.rs
+++ b/crates/oak_index/src/builder.rs
@@ -4,6 +4,7 @@ use aether_syntax::AnyRValue;
 use aether_syntax::RArgumentList;
 use aether_syntax::RBinaryExpression;
 use aether_syntax::RExpressionList;
+use aether_syntax::RForStatement;
 use aether_syntax::RFunctionDefinition;
 use aether_syntax::RParameter;
 use aether_syntax::RParameters;
@@ -27,6 +28,7 @@ use crate::semantic_index::ScopeId;
 use crate::semantic_index::ScopeKind;
 use crate::semantic_index::SemanticIndex;
 use crate::semantic_index::SymbolFlags;
+use crate::semantic_index::SymbolId;
 use crate::semantic_index::SymbolTableBuilder;
 use crate::semantic_index::Use;
 use crate::semantic_index::UseId;
@@ -286,7 +288,10 @@ impl SemanticIndexBuilder {
                 let pre_loop = self.current_use_def.snapshot();
 
                 if let Ok(body) = stmt.body() {
+                    let first_use = self.uses[self.current_scope].next_id();
+                    let loop_header = self.build_loop_header(body.syntax());
                     self.collect_expression(&body);
+                    self.finish_loop_header(&loop_header, first_use);
                 }
 
                 self.current_use_def.merge(pre_loop);
@@ -328,7 +333,10 @@ impl SemanticIndexBuilder {
                 let pre_loop = self.current_use_def.snapshot();
 
                 if let Ok(body) = stmt.body() {
+                    let first_use = self.uses[self.current_scope].next_id();
+                    let loop_header = self.build_loop_header(body.syntax());
                     self.collect_expression(&body);
+                    self.finish_loop_header(&loop_header, first_use);
                 }
 
                 // Body may not execute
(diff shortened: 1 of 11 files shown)


crates/oak_index/src/use_def_map.rs
```diff
@@ -0,0 +1,404 @@
...
+
+/// The set of definitions that can reach a particular point in control flow,
+/// plus whether the symbol may be unbound (no definition on some path).
+///
+/// Definitions are stored sorted by ID so that merge is a linear merge-join. The
+/// `SmallVec<[DefinitionId; 2]>` avoids heap allocation for the common case of
+/// 1-2 live definitions.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct Bindings {
+    definitions: SmallVec<[DefinitionId; 2]>,
```
DavisVaughan, line 155: Using a SmallVec here makes so much sense!

My immediate thought while looking at 

```
    // Per-scope flow-sensitive map from each use site to the set of definitions
    // that can reach it. Built alongside the other arrays during the tree walk.
    use_def_maps: IndexVec<ScopeId, UseDefMap>,
```

over in SemanticIndex was "wow i wonder how much memory that could end up using since each variable use gets a Map" but chances are there aren't many definitions for each variable, so it should be fairly cheap this way.

crates/oak_index/src/index_vec.rs
```diff
@@ -63,6 +63,15 @@ impl<I: Idx, V> FromIterator<V> for IndexVec<I, V> {
...
 }
 
+impl<I: Idx, V: Clone> Clone for IndexVec<I, V> {
+    fn clone(&self) -> Self {
+        Self {
+            raw: self.raw.clone(),
+            _phantom: PhantomData,
+        }
+    }
+}
```
DavisVaughan: I went back and looked at ruff and they can `derive(Clone)` on their `IndexVec`

There's nothing really different with ours right? Can we simplify to that?

crates/oak_index/src/builder.rs
```diff
@@ -70,12 +76,16 @@ impl SemanticIndexBuilder {
         symbol_tables.push(SymbolTableBuilder::new());
         definitions.push(IndexVec::new());
         uses.push(IndexVec::new());
+        use_def_maps.push(UseDefMap::empty());
```
DavisVaughan: IIUC, it is critically important that everything we index with `ScopeId` must all be the same length, i.e. you must push() to them all simultaneously.

I see we repeat this push() dance in 2 places in this file

Could we extract out a `self.push_scope()` helper that pushes a single `Scope` and also does a `push()` for all of these other fields? 

It would also give us a chance to document that you must keep them all in sync, which is not knowledge that I had before now

DavisVaughan: Oh hmm on closer inspection i see you exactly have push_scope() down below

But in `new()` here you don't have that, so you must "set it up" with initial pushes manually.

I think that's fine then

crates/oak_index/src/builder.rs
```diff
@@ -120,17 +141,103 @@ impl SemanticIndexBuilder {
...
+            Some(parent) => parent,
+            None => return file_scope,
+        };
+
+        loop {
+            if let Some(id) = self.symbol_tables[scope].id(name) {
+                if self.symbol_tables[scope]
+                    .symbol(id)
+                    .flags()
+                    .contains(SymbolFlags::IS_BOUND)
```
DavisVaughan, line 185: What if the parent scope has an id that IS_SUPER_BOUND? Still keep going?

DavisVaughan: Might be a good test candidate if we dont have one

crates/oak_index/src/builder.rs
```diff
@@ -120,17 +141,103 @@ impl SemanticIndexBuilder {
...
+        let builder = self.use_def_builder_mut(target_scope);
+        builder.ensure_symbol(target_symbol);
+        builder.record_deferred_definition(target_symbol, target_def_id);
+    }
+
+    // Walk up from the parent scope looking for a scope where `name` already
+    // has `IS_BOUND`. Returns that scope, or the file scope if no binding is
+    // found (mirroring R's assignment to the global environment).
+    fn resolve_super_target(&self, name: &str) -> ScopeId {
+        let file_scope = ScopeId::from(0);
```
DavisVaughan, line 175: Feels like it might be useful to have `ScopeId::from(0)` in a const `FILE_SCOPE` variable, which is probably possible since it just wraps an integer?

crates/oak_index/src/builder.rs
```diff
@@ -120,17 +141,103 @@ impl SemanticIndexBuilder {
...
+        let mut scope = match self.scopes[self.current_scope].parent {
+            Some(parent) => parent,
+            None => return file_scope,
+        };
```
DavisVaughan: To me this kind of hid the control flow of bailing when we try to go past the file scope by _returning_ with a "max" limit of the file scope itself. This might be clearer? It feels clearer to me at least, but this is admittedly a nit

```
let Some(mut scope) = self.scopes[self.current_scope].parent else {
    return file_scope;
}
```

And same below at the end of this function

crates/oak_index/src/builder.rs
```diff
@@ -120,17 +141,103 @@ impl SemanticIndexBuilder {
...
+                }
+            }
+            scope = match self.scopes[scope].parent {
+                Some(parent) => parent,
+                None => return file_scope,
+            };
+        }
+    }
+
+    fn use_def_builder_mut(&mut self, target: ScopeId) -> &mut UseDefMapBuilder {
```
DavisVaughan: Maybe `find_use_def_builder_mut()`? or `get_`?

Intent wasn't immediately clear I guess

crates/oak_index/src/builder.rs
```diff
@@ -120,17 +141,103 @@ impl SemanticIndexBuilder {
...
+        while scope != target {
+            scope = match self.scopes[scope].parent {
+                Some(parent) => parent,
+                None => panic!("Target scope {target:?} is not an ancestor of current scope"),
+            };
+            steps += 1;
+        }
+
+        let index = self.use_def_stack.len() - steps;
+        &mut self.use_def_stack[index]
```
DavisVaughan: Ok I must be crazy but to me this looks like it could be simplified to just `self.use_def_maps[scope]` after you confirm that the `target` is a parent scope of the current scope?

Then you'd remove `use_def_stack` alltogether? It can't be that simple...

DavisVaughan: I have given this a shot in https://github.com/posit-dev/ark/pull/1156, my gut feeling may have been right? Needs a close review, and if I am wrong then we should add a test to prove why I am wrong here. Otherwise if I am right then I think this greatly simplifies things!

DavisVaughan (approved):
Main comment is to look into #1156 paired with one of my comments below

crates/oak_index/src/builder.rs
lionel-: yep we keep going (that's expected), good call on adding a test

crates/oak_index/src/builder.rs
lionel-: hmm I don't know, I feel like this hides the nice structural property

crates/oak_index/src/builder.rs
DavisVaughan: fine as is based on the rust convention that `get_` returns an Option or Result, and unprefixed helpers return the object in a nonfallible way

crates/oak_index/src/builder.rs
DavisVaughan: oh the more i see `from(0)` in future PRs, the more i want `FILE_SCOPE`

i would also be fine with `ScopeId::file()`

PR #1146, opened 2026-04-10
Add `PackageCache` for populating a package source cache

DavisVaughan:
https://github.com/posit-dev/positron/issues/2286
Progress towards https://github.com/posit-dev/positron/issues/2749

Two tiered approach to caching R package sources, in the following priority order:
- Source references on already installed package (requires sidecar R session)
- CRAN download (requires internet)
    - Future Bioconductor download (not yet, but Gabor has code for this in pkgdepends)

## API

```
PackageCache::new(r_script_path: PathBuf, r_libpaths: Vec<PathBuf>)
PackageCache::get(&mut self, package: &str) -> Option<PathBuf>
```

## Source references

When a package is installed with source references, `attr(attr(pkg::fn, "srcref"), "srcfile")$original$lines` contains something like this:

```
[1] ".packageName <- \"vctrs\""
[2] "#line 1 \"/private/var/folders/by/4wf2qrmn4_j93s5k5k5fzrx00000gp/T/Rtmp5mAZjD/R.INSTALL164782a40400c/vctrs/R/aaa.R\""                  
   [3] "replace_from <- function(what, pkg, to = topenv(caller_env())) {"                                                                      
   [4] "  if (what %in% getNamespaceExports(pkg)) {"                                                                                           
   [5] "    env <- ns_env(pkg)"                                                                                                                
   [6] "  } else {"                                                                                                                            
   [7] "    env <- to"                                                                                                                         
   [8] "  }"                                                                                                                                   
   [9] "  env_get(env, what, inherit = TRUE)"                                                                                                  
  [10] "}"                                                                                                                                     
  [11] ""                                                                                                                                      
  [12] "# nocov start"                                                                                                                         
  [13] ""                                                                                                                                      
  [14] "# Useful for micro-optimising default arguments requiring evaluation,"                                                                 
  [15] "# such as `param = c(\"foo\", \"bar\")`. Buys about 0.6us on my desktop."                                                              
  [16] "fn_inline_formals <- function(fn, names) {"                                                                                            
  [17] "  stopifnot(typeof(fn) == \"closure\")"                                                                                                
  [18] ""                                                                                                                                      
  [19] "  fmls <- formals(fn)"                                                                                                                 
  [20] "  fmls[names] <- lapply(fmls[names], eval)"                                                                                            
  [21] ""                                                                                                                                      
  [22] "  formals(fn) <- fmls"                                                                                                                 
  [23] "  fn"                                                                                                                                  
  [24] "}"                                                                                                                                     
  [25] ""                                                                                                                                      
  [26] "# nocov end"                                                                                                                           
  [27] "#line 1 \"/private/var/folders/by/4wf2qrmn4_j93s5k5k5fzrx00000gp/T/Rtmp5mAZjD/R.INSTALL164782a40400c/vctrs/R/altrep-lazy-character.R\""
  [28] "#' Lazy character vector"                                                                                                              
  [29] "#'"                                                                                                                                    
  [30] "#' `new_lazy_character()` takes a function with no arguments which must return"                                                        
  [31] "#' a character vector of arbitrary length. The function will be evaluated"
```

These source references are generated from a number of sources:

- `devtools::install(build = FALSE, keep_source = TRUE)`
    - Absolute file paths will point to the user's local package folder
- `devtools::install(build = TRUE, keep_source = TRUE)`
    - Absolute file paths will point to a nonexistent temp directory
- `pak::pak(<github source>)`
    - Only with `KeepSource: true` or envvar `R_KEEP_PKG_SOURCE=yes`
    - Absolute file paths will point to a nonexistent temp directory
- CRAN binary download
    - From package bundle generated with `KeepSource: true`
    - Absolute file paths will point to a nonexistent CRAN machine directory

For the most part, the absolute paths are useless to us. But the _relative package structure_ is incredibly useful. i.e. with `<install-path>/vctrs/R/altrep-lazy-character.R` we can rebuild `altrep-lazy-character.R` - comments and all!

This unfortunately requires R to yank out this srcref information.

## CRAN download

Given a package name and version, we can go out and download its sources from CRAN. They retain historical archives for all package versions.

## What does this get us?

We can `get("vctrs")` and be returned a `PathBuf` that points to a directory that is guaranteed to contain:
- `DESCRIPTION`
- `NAMESPACE`
- `R/` with full comments and file structure

We can use this for further static analysis.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,7 +42,7 @@ blake3 = "1.8.2"
 bus = "2.3.0"
 cc = "1.1.22"
 cfg-if = "1.0.0"
-chrono = "0.4.23"
+chrono = "0.4.44"
 crossbeam = { version = "0.8.4", features = ["crossbeam-channel"] }
 crypto-common = "0.1.6"
 ctor = "0.1.26"
@@ -52,6 +52,8 @@ dirs = "4.0.0"
 ego-tree = "0.6.2"
 embed-resource = "2.5.0"
 env_logger = "0.10.0"
+etcetera = "0.11.0"
+flate2 = "1.1.1"
 futures = "0.3.30"
 generic-array = "0.14.6"
 harp = { path = "crates/harp" }
@@ -70,17 +72,20 @@ mime_guess = "2.0.4"
 nix = { version = "0.26.2", features = ["signal"] }
 notify = "6.0.0"
 oak_core = { path = "crates/oak_core" }
+oak_fs = { path = "crates/oak_fs" }
 oak_index = { path = "crates/oak_index" }
 oak_package = { path = "crates/oak_package" }
+oak_r_process = { path = "crates/oak_r_process" }
+oak_sources = { path = "crates/oak_sources" }
 once_cell = "1.17.1"
 parking_lot = "0.12.3"
 paste = "1.0.14"
 quote = "1.0.42"
 rand = "0.8.5"
 regex = "1.10.0"
-reqwest = { version = "0.12.5", default-features = false, features = ["json"] }
-reqwest-middleware = "0.3.3"
-reqwest-retry = "0.6.1"
+reqwest = { version = "0.13.2", default-features = false, features = ["blocking", "json", "rustls"] }
+reqwest-middleware = "0.5.1"
+reqwest-retry = "0.9.1"
 rust-embed = "8.2.0"
 rustc-hash = "2.1.1"
 scraper = "0.15.0"
@@ -96,6 +101,7 @@ streaming-iterator = "0.1.9"
 strum = "0.26.2"
 strum_macros = "0.26.2"
 syn = { version = "2.0.111", features = ["full"] }
+tar = "0.4.44"
 tempfile = "3.13.0"
 tokio = { version = "1.26.0", features = ["full"] }
 # For https://github.com/ebkalderon/tower-lsp/pull/428
@@ -109,6 +115,7 @@ tree-sitter-r = { git = "https://github.com/r-lib/tree-sitter-r", rev = "95aff09
 url = "2.5.7"
 uuid = { version = "1.3.0", features = ["v4"] }
 walkdir = "2"
+windows-sys ="0.61.2"
 winsafe = { version = "0.0.19", features = ["kernel"] }
 xdg = "2.5.2"
 yaml-rust = "0.4.5"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -103,7 +103,6 @@ strum_macros = "0.26.2"
 syn = { version = "2.0.111", features = ["full"] }
 tar = "0.4.44"
 tempfile = "3.13.0"
-threadpool = "1"
 tokio = { version = "1.26.0", features = ["full"] }
 # For https://github.com/ebkalderon/tower-lsp/pull/428
 tower-lsp = { branch = "bugfix/patches", git = "https://github.com/lionel-/tower-lsp" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -103,6 +103,7 @@ strum_macros = "0.26.2"
 syn = { version = "2.0.111", features = ["full"] }
 tar = "0.4.44"
 tempfile = "3.13.0"
+threadpool = "1"
 tokio = { version = "1.26.0", features = ["full"] }
 # For https://github.com/ebkalderon/tower-lsp/pull/428
 tower-lsp = { branch = "bugfix/patches", git = "https://github.com/lionel-/tower-lsp" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,15 +10,7 @@
 resolver = "2"
 
 members = [
-    "crates/amalthea",
-    "crates/ark",
-    "crates/echo",
-    "crates/harp",
-    "crates/libr",
-    "crates/oak_core",
-    "crates/oak_index",
-    "crates/oak_package",
-    "crates/stdext",
+    "crates/*"
 ]
 
 [workspace.package]
--- a/crates/oak_package/src/package_description.rs
+++ b/crates/oak_package/src/package_description.rs
@@ -38,8 +38,15 @@ pub struct Description {
     /// `Depends` field. Currently doesn't contain versions.
     pub depends: Vec<String>,
 
+    pub repository: Option<Repository>,
+
     /// Raw DCF fields
     pub fields: Dcf,
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum Repository {
+    CRAN,
 }
 
 impl Description {
@@ -70,10 +77,18 @@ impl Description {
             })
             .unwrap_or_default();
 
+        let repository = fields.get("Repository").and_then(|repository| {
+            if repository == "CRAN" {
+                return Some(Repository::CRAN);
+            }
+            None
+        });
+
         Ok(Description {
             name,
             version,
             depends,
+            repository,
             fields,
         })
     }
@@ -180,6 +195,26 @@ Description: This is a long description
         let parsed = Description::parse(desc).unwrap();
         assert_eq!(parsed.name, "mypackage");
         assert_eq!(parsed.version, "1.0.0");
+    }
+
+    #[test]
+    fn parses_description_with_known_repository() {
+        let desc = r#"Package: mypackage
+Version: 1.0.0
+Title: My Package
+Repository: CRAN"#;
+        let parsed = Description::parse(desc).unwrap();
+        assert_eq!(parsed.repository, Some(Repository::CRAN));
+    }
+
+    #[test]
+    fn parses_description_with_unknown_repository() {
+        let desc = r#"Package: mypackage
+Version: 1.0.0
+Title: My Package
+Repository: notCRAN"#;
+        let parsed = Description::parse(desc).unwrap();
+        assert_eq!(parsed.repository, None);
     }
 
     #[test]
--- a/crates/oak_sources/scripts/srcrefs.R
+++ b/crates/oak_sources/scripts/srcrefs.R
@@ -1,5 +1,8 @@
 # Extract source code from an installed R package using srcref metadata.
 #
+# See also `lobstr::src()`:
+# https://github.com/r-lib/lobstr/blob/85dd18d43e4612f98b05bb93019a7b12fb0bbf6b/R/src.R#L205
+#
 # Arguments (via commandArgs):
 #   1. package - Package name
 #   2. version - Expected package version
--- a/crates/oak_sources/scripts/srcrefs.R
+++ b/crates/oak_sources/scripts/srcrefs.R
@@ -1,0 +1,99 @@
+# Extract source code from an installed R package using srcref metadata.
+#
+# Arguments (via commandArgs):
+#   1. package - Package name
+#   2. version - Expected package version
+#
+# Library paths have been set ahead of time via the `R_LIBS` environment
+# variable.
+#
+# On success: prints the concatenated source lines (with `#line` directives) to
+# stdout.
+#
+# On failure: exits with code 1 (package unexpectedly not installed, version
+# mismatch) or 2 (no srcrefs).
+
+args <- commandArgs(trailingOnly = TRUE)
+
+if (length(args) != 2L) {
+    message("'package' and 'version' are required arguments.")
+    quit(status = 1L)
+}
+
+package <- args[[1L]]
+version <- args[[2L]]
+
+# Make sure package is installed
+path <- find.package(package, quiet = TRUE)
+if (length(path) == 0L) {
+    message(paste0("Package '", package, "' is not installed"))
+    quit(status = 1L)
+}
+
+# Make sure installed version matches the requested version
+installed_version <- as.character(packageVersion(package))
+if (installed_version != version) {
+    message(paste0(
+        "Version mismatch for '",
+        package,
+        "': ",
+        "installed ",
+        installed_version,
+        ", requested ",
+        version
+    ))
+    quit(status = 1L)
+}
+
+# Get the package namespace and find a function object.
+# Functions will contain the srcrefs if srcrefs have been kept, and are what we
+# can back out the full file structure from.
+ns <- asNamespace(package)
+exports <- getNamespaceExports(ns)
+
+fn <- NULL
+for (name in exports) {
+    candidate <- get0(name, envir = ns, mode = "function")
+    if (!is.null(candidate)) {
+        fn <- candidate
+        break
+    }
+}
+
+if (is.null(fn)) {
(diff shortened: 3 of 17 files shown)


Cargo.toml
```diff
@@ -82,7 +78,7 @@ paste = "1.0.14"
 quote = "1.0.42"
 rand = "0.8.5"
 regex = "1.10.0"
-reqwest = { version = "0.12.5", default-features = false, features = ["json"] }
+reqwest = { version = "0.12.5", default-features = false, features = ["blocking", "json", "rustls-tls"] }
```
DavisVaughan: Not excited about a rustls-tls dependency, which I think brings in some heavy stuff. But it seems required to be able to download anything from the internet.

crates/ark/src/modules/positron/debug.R
```diff
@@ -793,19 +794,24 @@ replace_non_parseable <- function(x) {
...
+# are multiple of them in a single deparsed object.
 non_parseable_pattern_infos <- function() {
     list(
-        non_parseable_pattern_info("<S4 object of class .*>", "...S4..."),
-        non_parseable_pattern_info("<promise: .*>", "...PROMISE..."),
-        non_parseable_pattern_info("<pointer: .*>", "...POINTER..."),
+        non_parseable_pattern_info("<S4 object of class .*?>", "...S4..."),
+        non_parseable_pattern_info("<promise: .*?>", "...PROMISE..."),
+        non_parseable_pattern_info("<pointer: .*?>", "...POINTER..."),
         non_parseable_fixed_info("<environment>", "...ENVIRONMENT..."),
```
DavisVaughan, line 807: I realized this was incorrect after some further testing

DavisVaughan:
It's also worth it to talk about what was considered but isn't implemented

## Deparse approach

Using the same approach as the deparse engine used at package load time (and the one used in the debugger fallback), we could create a single collated `namespace.R` file that holds all of the functions from a package.

This has a number of bleh side effects:
- We have to patch up non parseable objects, like environments, which means that sources aren't accurate
- For inline things like S7 and R6 classes, deparsing creates a huge mess that doesn't provide useful sources

We decided not to have this fallback

crates/ark/src/modules/positron/debug.R
```diff
@@ -793,19 +794,24 @@ replace_non_parseable <- function(x) {
...
+# a first `<promise:` all the way to the LAST `>`, which is a problem if there
+# are multiple of them in a single deparsed object.
 non_parseable_pattern_infos <- function() {
     list(
-        non_parseable_pattern_info("<S4 object of class .*>", "...S4..."),
-        non_parseable_pattern_info("<promise: .*>", "...PROMISE..."),
-        non_parseable_pattern_info("<pointer: .*>", "...POINTER..."),
+        non_parseable_pattern_info("<S4 object of class .*?>", "...S4..."),
+        non_parseable_pattern_info("<promise: .*?>", "...PROMISE..."),
+        non_parseable_pattern_info("<pointer: .*?>", "...POINTER..."),
```
DavisVaughan, line 806: Note to self to move this to another PR since we no longer do deparsing as a fallback, so this "related change" is no longer "related"

crates/oak_sources/src/srcref.rs
```diff
@@ -0,0 +1,319 @@
...
+
+/// Extracts an R package's source files from srcref metadata if possible and adds
+/// them to the cache at `destination`
+///
+/// Launches a sidecar R session to read the srcrefs from the installed package.
+pub(crate) fn cache_srcref(
+    package: &str,
+    version: &str,
+    destination: &Path,
+    r_script_path: &Path,
```
DavisVaughan: Lets see if we can make this `r_bin_path` instead

crates/oak_sources/src/fs.rs
```diff
@@ -0,0 +1,33 @@
...
+
+use etcetera::BaseStrategy;
+
+fn cache_dir() -> anyhow::Result<PathBuf> {
+    // Can fail if the home directory can't be found
+    Ok(etcetera::choose_base_strategy()?.cache_dir().join("oak"))
+}
+
+pub(crate) fn sources_dir() -> anyhow::Result<PathBuf> {
+    Ok(cache_dir()?.join("sources").join("v1"))
```
DavisVaughan, line 13: Maybe consider putting version 1 in `.metadata` at `sources/` level

crates/oak_sources/src/lib.rs
```diff
@@ -0,0 +1,457 @@
...
+///
+/// # On disk layout
+///
+/// The on disk cache at `<cache_dir>/oak/sources/v1/` is the source of truth
+/// across all sessions. Each entry is a flat folder:
+///
+/// ```text
+/// <cache_dir>/oak/sources/v1/
+///     .lock
+///     {package}_{version}_libpath-{hash}_description-{hash}/
```
DavisVaughan, line 45: talk about why `description-{hash}` makes a package unique (Built in particular)

crates/oak_sources/src/lib.rs
```diff
@@ -0,0 +1,457 @@
...
+///         DESCRIPTION
+///         NAMESPACE
+///         R/
+/// ```
+///
+/// Files other than `.lock` and `.metadata` are marked read only. `.metadata`
+/// is written last and is the sole completion sentinel. An entry without it
+/// is considered garbage and will be wiped by the next writer for that key.
+///
+/// # Locking
```
DavisVaughan, line 68: Maybe reference that this is the cargo model too

crates/oak_sources/src/lib.rs
```diff
@@ -0,0 +1,457 @@
...
+        // - The same R package across multiple libpaths
+        // - Reinstalling a dev R package without changing the version (0.1.0.9000)
+        let key =
+            format!("{package}_{version}_libpath-{libpath_hash}_description-{description_hash}");
+
+        // In memory cache check
+        if let Some(status) = self.cache.get(&key) {
+            match status {
+                Status::NoSources => return Ok(None),
+                Status::Sources(sources) => return Ok(Some(sources.clone())),
```
DavisVaughan: ONLY cache that "we have tried to get this package that is installed, but there are NoSources"

Otherwise, just fall through to on disk cache paths

DavisVaughan (commented):
From pair review with @lionel-

crates/oak_sources/src/fs.rs
DavisVaughan: I'm actually leaning towards keeping `v1/` as a separate directory

I'm imagining a case where a user installs a new version of oak that crosses the `v1/` to `v2/` version boundary. They might realize almost immediately that they need to roll oak back to some older version due to some kind of issue. It would stink if oak just nuked their whole `v1/` cache before they could roll back.

It seems like maybe we should keep `v1/` and `v2/` as separate folders, and if we are on oak v2 we can drop a `.metadata` file in `v1/` with a timestamp. After 1 week goes by, oak v2 deletes `v1/` to clean it out. But this at least gives a little grace period.

I'm not super worried about the path getting long. `v1` is quite short and the final folder path with `vctrs_0.7.3_libpath123_description123/R/aaa.R` is going to be the more useful part anyways

Cargo.toml
```diff
@@ -10,15 +10,7 @@
...
-    "crates/ark",
-    "crates/echo",
-    "crates/harp",
-    "crates/libr",
-    "crates/oak_core",
-    "crates/oak_index",
-    "crates/oak_package",
-    "crates/stdext",
+    "crates/*"
 ]
```
DavisVaughan, line 14: This seemed easier to manage

Cargo.toml
```diff
@@ -78,17 +72,20 @@ mime_guess = "2.0.4"
...
 paste = "1.0.14"
 quote = "1.0.42"
 rand = "0.8.5"
 regex = "1.10.0"
-reqwest = { version = "0.12.5", default-features = false, features = ["json"] }
-reqwest-middleware = "0.3.3"
-reqwest-retry = "0.6.1"
+reqwest = { version = "0.13.2", default-features = false, features = ["blocking", "json", "rustls"] }
+reqwest-middleware = "0.5.1"
+reqwest-retry = "0.9.1"
```
DavisVaughan, line 88: The version bump is required due to some rustls details. They used to use the ring crate dependency, which sucked to install. It now uses aws-ls, which still sucks to install, but slightly less so. Both are cryptographic libraries with lots of C/C++

crates/oak_package/src/package_description.rs
```diff
@@ -70,20 +77,21 @@ impl Description {
             })
             .unwrap_or_default();
 
+        let repository = fields.get("Repository").and_then(|repository| {
```
lionel-, line 80: add some context about why we don't support e.g. bioc right away?

crates/oak_sources/scripts/srcrefs.R
```diff
@@ -0,0 +1,99 @@
+# Extract source code from an installed R package using srcref metadata.
```
lionel-, line 1: Include link to `?src`? https://github.com/r-lib/lobstr/blob/85dd18d43e4612f98b05bb93019a7b12fb0bbf6b/R/src.R#L217

crates/oak_sources/src/cran.rs
```diff
@@ -0,0 +1,124 @@
+use std::io::Cursor;
+use std::path::Path;
+
+use flate2::read::GzDecoder;
+use tar::Archive;
+
+/// Assumes `destination` exists and is file locked by the caller
```
lionel-: Can we encode this in the type by moving the lock in and out (or sharing it)?

just a thought

crates/oak_sources/src/cran.rs
```diff
@@ -0,0 +1,124 @@
...
+    std::fs::create_dir(&destination)?;
+
+    extract(package, response, destination.as_path())?;
+
+    Ok(true)
+}
+
+fn download(package: &str, version: &str) -> anyhow::Result<reqwest::blocking::Response> {
+    // Try released version
+    let url = format!("https://cran.r-project.org/src/contrib/{package}_{version}.tar.gz");
```
lionel-: Probably worth adding https://cran.rstudio.com as fallback?

crates/oak_sources/src/cran.rs
```diff
@@ -0,0 +1,124 @@
...
+
+fn extract(
+    package: &str,
+    response: reqwest::blocking::Response,
+    destination: &Path,
+) -> anyhow::Result<()> {
+    let bytes = response.bytes()?;
+    let cursor = Cursor::new(bytes);
+    let gz = GzDecoder::new(cursor);
+    let mut archive = Archive::new(gz);
```
lionel-: Maybe mention this is doing the heavy lifting of writing gzipped files to disk?

crates/oak_sources/src/fs.rs
```diff
@@ -0,0 +1,39 @@
+use std::io;
+use std::path::Path;
+use std::path::PathBuf;
+
+use etcetera::BaseStrategy;
```
lionel-, line 5: Love the name!

crates/oak_sources/src/fs.rs
lionel-: I think I'd still remove the versioning but if you prefer to keep it that's totally fine

crates/oak_sources/src/srcref.rs
```diff
@@ -0,0 +1,296 @@
+use std::path::Path;
+use std::path::PathBuf;
+
+const SCRIPT: &str = include_str!("../scripts/srcrefs.R");
+
+/// Extracts an R package's source files from srcref metadata if possible and adds
+/// them to the cache at `destination`
+///
+/// Launches a sidecar R session to read the srcrefs from the installed package.
```
lionel-, line 11: Not sure if we need to worry about the ergonomics and performance of launching many processes versus reusing one. The latter is trickier because of lifecycle management, especially when something goes wrong. We could launch `ark` as a sidecar process to send tasks as execute requests. And once oak is independent from ark, we can also use the in-process R for this. Just noting these thoughts, the current approach is simple and clearly sufficient for now. And has the big benefit of being self-contained.

crates/oak_sources/src/srcref.rs
```diff
@@ -0,0 +1,296 @@
...
+        assert_eq!(parse_line_directive("#line 1 missing-quotes"), None);
+    }
+
+    /// Requires R on the PATH and internet access
+    ///
+    /// Installs source version of {generics} from CRAN into a temporary library, then
+    /// extracts the R source files via srcref metadata. We use {generics} because it
+    /// is very easy to install from source and lightweight.
+    #[test]
+    fn test_srcref_extraction() {
```
lionel-, line 198: Nice integration test

crates/oak_sources/src/lib.rs
```diff
@@ -0,0 +1,461 @@
...
+    generated_at: DateTime<Utc>,
+}
+
+impl PackageCache {
+    pub fn new(r: PathBuf, r_libpaths: Vec<PathBuf>) -> anyhow::Result<Self> {
+        let cache_root = file_lock::Filesystem::new(crate::fs::sources_dir()?);
+        cache_root.create_dir()?;
+
+        // Try to clean the cache. Only works if no other processes hold a shared root lock.
+        if let Some(root_lock) = cache_root.try_open_rw_exclusive_create(LOCK_FILENAME)? {
```
lionel-: It's so neat the kind of ergonomic tooling we have access to with the Rust ecosystem

crates/oak_sources/src/lib.rs
```diff
@@ -0,0 +1,461 @@
...
+            source_unavailable: RwLock::new(HashSet::new()),
+        })
+    }
+
+    /// Get a package's cached source folder
+    ///
+    /// May spawn an R subprocess or download from CRAN (in a blocking manner) to
+    /// generate the sources, so keep that in mind when calling this.
+    pub fn get(&self, package: &str) -> Option<PathBuf> {
+        match self.get_impl(package) {
```
lionel-:
```suggestion
        match self.get_result(package) {
```

lionel- (approved):
Looks great!

crates/oak_package/src/package_description.rs
DavisVaughan: Bioconductor uses some weird `biocViews` field in DESCRIPTION, not Repository, I've added some notes in https://github.com/posit-dev/ark/pull/1146/commits/17f16b9cd8c4dd3972b38cabaa7e5fae615735c3

crates/oak_sources/src/cran.rs
DavisVaughan: We can! Done in https://github.com/posit-dev/ark/pull/1146/commits/d936d3ef36f6f631d7dbc6ad55205e32001676b3

crates/oak_sources/src/srcref.rs
DavisVaughan: Yea it might require some iteration!

PR #1147, opened 2026-04-13
Implement enclosing snapshots

lionel-:
Branched from #1144 
Part of https://github.com/posit-dev/ark/issues/1141

From the module-level doc in `use-def-maps.rs`:

```rs
// ## Enclosing snapshots
//
// Use-def maps are per-scope, so a free variable in a nested function
// gets `{ definitions: [], may_be_unbound: true }` locally. To resolve
// it, we need the enclosing scope's bindings. Enclosing snapshots
// bridge this gap.
//
// A snapshot is registered when `add_use` detects `may_be_unbound` in
// the nested scope. The builder walks up the scope chain to find the
// ancestor where the symbol is bound and records a snapshot in that
// ancestor's `UseDefMapBuilder`. The snapshot captures what's live at
// the nested scope's definition point, then a watcher accumulates
// subsequent definitions. We take the union of all subsequent definitions
// because we can't tractably know exactly when the function is called. In the
// following example, `x`'s use sees both B and C even though from this
// particular snippet it can only see B.
//
// Providing more accurate answers in the general case would require whole
// program analysis (e.g. `f` may be passed down to other functions). For now we
// accept the over-approximation of taking the union of subsequent definitions,
// although we could improve on that for simple cases in the future (tracking
// simple calls, and falling back to over-approximation in the other cases).
//
// ```r
// x <- 0              # def A
// x <- 1              # def B (shadows A)
// f <- function() x   # snapshot initialized: {B}
// f()
// x <- 2              # watcher fires → snapshot: {B, C}
// ```
//
// Note that the snapshot is {B, C}, not {A, B, C}, because def A was already
// shadowed when `f` was first defined. The rule: what was live at the
// definition point (prior shadowing applied) plus every definition added after.
//
// When `may_be_unbound` is true due to a conditional local definition, the
// snapshot is also consulted (unlike Python, R has no name-binding rule, so
// conditional local definitions don't prevent fallthrough):
//
// ```r
// x <- 1
// f <- function(cond) {
//     if (cond) x <- 2
//     x                  # local: {x <- 2, may_be_unbound: true}
// }                      # enclosing snapshot: {x <- 1}
// ```
//
// The consumer combines both: the local bindings and the enclosing
// snapshot give the full picture of what `x` could be.
//
// For eager NSE scopes (e.g. `local()`), the snapshot will be even more
// precise: since the body executes at the call site, the snapshot is a
// point-in-time capture with no watcher, reflecting exactly the linear state.
// No union over-approximation needed.
```

To support this:

- We now do a pre-scan that allows us to see the full list of defined names before recursing into nested scopes. This way we're able to see that `x` is bound in:

  ```r
  f <- function() x
  x <- 1
  ```

- When a use may be unbound, we register an "enclosing snapshot" in the ancestor scope that captures what's live at the definition point and accumulates subsequent definitions via watchers.

### DIFF
--- a/crates/oak_index/src/builder.rs
+++ b/crates/oak_index/src/builder.rs
@@ -16,17 +16,22 @@ use biome_rowan::AstSeparatedList;
 use biome_rowan::SyntaxNodeCast;
 use biome_rowan::TextRange;
 use biome_rowan::WalkEvent;
+use rustc_hash::FxHashMap;
+use smallvec::SmallVec;
 
 use crate::index_vec::Idx;
 use crate::index_vec::IndexVec;
 use crate::semantic_index::Definition;
 use crate::semantic_index::DefinitionId;
 use crate::semantic_index::DefinitionKind;
+use crate::semantic_index::EnclosingSnapshotId;
+use crate::semantic_index::EnclosingSnapshotKey;
 use crate::semantic_index::Scope;
 use crate::semantic_index::ScopeId;
 use crate::semantic_index::ScopeKind;
 use crate::semantic_index::SemanticIndex;
 use crate::semantic_index::SymbolFlags;
+use crate::semantic_index::SymbolId;
 use crate::semantic_index::SymbolTableBuilder;
 use crate::semantic_index::Use;
 use crate::semantic_index::UseId;
@@ -36,6 +41,7 @@ use crate::use_def_map::UseDefMapBuilder;
 pub fn build(root: &RRoot) -> SemanticIndex {
     let range = root.syntax().text_trimmed_range();
     let mut builder = SemanticIndexBuilder::new(range);
+    builder.pre_scan_scope(root.syntax());
     builder.collect_expression_list(&root.expressions());
     builder.finish()
 }
@@ -50,6 +56,9 @@ struct SemanticIndexBuilder {
     uses: IndexVec<ScopeId, IndexVec<UseId, Use>>,
     use_def_maps: IndexVec<ScopeId, UseDefMapBuilder>,
     current_scope: ScopeId,
+    current_pre_scan: PreScanScope,
+    pre_scan_stack: Vec<PreScanScope>,
+    enclosing_snapshots: FxHashMap<EnclosingSnapshotKey, (ScopeId, EnclosingSnapshotId)>,
 }
 
 impl SemanticIndexBuilder {
@@ -85,6 +94,9 @@ impl SemanticIndexBuilder {
             uses,
             use_def_maps,
             current_scope: file,
+            current_pre_scan: PreScanScope::new(),
+            pre_scan_stack: Vec::new(),
+            enclosing_snapshots: FxHashMap::default(),
         }
     }
 
@@ -109,6 +121,9 @@ impl SemanticIndexBuilder {
         self.uses.push(IndexVec::new());
         self.use_def_maps.push(UseDefMapBuilder::new());
 
+        let parent_pre_scan = std::mem::replace(&mut self.current_pre_scan, PreScanScope::new());
+        self.pre_scan_stack.push(parent_pre_scan);
+
         id
     }
 
@@ -119,6 +134,11 @@ impl SemanticIndexBuilder {
         self.current_scope = match self.scopes[id].parent {
             Some(parent) => parent,
             None => panic!("`pop_scope()` called on the file scope"),
+        };
+
+        self.current_pre_scan = match self.pre_scan_stack.pop() {
+            Some(pre_scan) => pre_scan,
+            None => panic!("`pop_scope()` called with empty pre-scan stack"),
         };
     }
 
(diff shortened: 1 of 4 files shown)


crates/oak_index/src/builder.rs
```diff
@@ -238,6 +259,71 @@ impl SemanticIndexBuilder {
...
+                // Intern with empty flags: we just need a stable `SymbolId` for
+                // the lookup key. If found via `found_by_flag`, the symbol
+                // already exists with `IS_BOUND`. If found via pre-scan only,
+                // the later `add_definition` call during the full walk will set
+                // `IS_BOUND`.
+                let enclosing_symbol_id =
+                    self.symbol_tables[current_scope].intern(name, SymbolFlags::empty());
+
+                let key = EnclosingSnapshotKey {
+                    nested_scope: self.current_scope,
```
DavisVaughan: `self.current_scope` or `current_scope`?

DavisVaughan: smells like a missing test case

crates/oak_index/src/builder.rs
```diff
@@ -459,12 +557,60 @@ impl SemanticIndexBuilder {
...
     }
 
+    /// Pre-scan a scope to collect all definition names (skipping nested
+    /// function bodies). Runs before the full walk so that enclosing
+    /// snapshot registration can find where free variables are bound,
+    /// even when the walk in the parent scope hasn't reached the
+    /// definition yet. Must stay in sync with the full walk's definition
+    /// handling: any construct that calls `add_definition` should have a
+    /// corresponding entry here.
+    fn pre_scan_scope(&mut self, node: &RSyntaxNode) {
```
DavisVaughan, line 522: IIUC you have to leave the depth-first search world a little bit to be able to do enclosing snapshot registration? And that's the idea here? Can you write me an example of why that is required?

crates/oak_index/src/semantic_index.rs
```diff
@@ -58,6 +62,10 @@ pub struct SemanticIndex {
     // Per-scope flow-sensitive map from each use site to the set of definitions
     // that can reach it. Built alongside the other arrays during the tree walk.
     use_def_maps: IndexVec<ScopeId, UseDefMap>,
+
+    // For each free variable in a nested scope, maps to the enclosing scope and
+    // snapshot where that symbol is bound.
+    enclosing_snapshots: FxHashMap<EnclosingSnapshotKey, (ScopeId, EnclosingSnapshotId)>,
```
DavisVaughan, line 68: Not an IndexVec?

DavisVaughan: Oh I see its indexed by EnclosingSnapshotKey not EnclosingSnapshotId

crates/oak_index/src/use_def_map.rs
```diff
@@ -130,24 +135,96 @@ use crate::semantic_index::UseId;
...
+// accept the over-approximation of taking the union of subsequent definitions,
+// although we could improve on that for simple cases in the future (tracking
+// simple calls, and falling back to over-approximation in the other cases).
+//
+// ```r
+// x <- 0              # def A
+// x <- 1              # def B (shadows A)
+// f <- function() x   # snapshot initialized: {B}
+// f()
+// x <- 2              # watcher fires → snapshot: {B, C}
```
DavisVaughan, line 167: Oh hmm so the defns of the `x` in `f <- function() x` are `x <- 1` and `x <- 2` due to the enclosing scope bits? Hmm

DavisVaughan: > The rule: what was live at the definition point (prior shadowing applied) plus every definition added after.

hmmmmmmmm i think that makes sense

although this is interesting to think about

```r
x <- 1 
fn <- function() x
fn()
x <- 2
fn()
```

I guess your argument is that it is based on where `fn` is _defined_ not _called_

DavisVaughan (approved):

crates/oak_index/src/use_def_map.rs
lionel-: Right, without call analysis (which can range from easy to very complicated depending how far we want to go), the only thing we know for sure is the point at which the function was defined in a given scope.

crates/oak_index/src/builder.rs
lionel-: It's keyed by use, so the current code is correct.

To make it more obvious, `register_enclosing_snapshot()` now requires the caller to create the key, and to pass it to a `use_key` parameter.

crates/oak_index/src/builder.rs
lionel-: See module-level doc in `use_def_maps.rs`

crates/oak_index/src/use_def_map.rs
lionel-: In static analysis this is called an "over-approximation". The notion of enclosing snapshot is straight from ruff as Python has the same issue with closures.

crates/oak_index/src/builder.rs
DavisVaughan: yea i understood this after reading more

PR #1151, opened 2026-04-14
Add external symbol resolution infrastructure

lionel-:
Branched from https://github.com/posit-dev/ark/pull/1150
Part of https://github.com/posit-dev/ark/issues/1148

- Adds data structures for the scope chains of binding sources
- The semantic index now tracks `library()` and `require()` calls and creates scope chains from them.
- New helpers to resolve unbound symbols from external definitions (scope chains).

For future work:

- Create scope chains for files in a package, respecting the collation order (`aaa.R` will have a larger chain than `zzz.R`).
- Resolve symbol at point in a file from the current scope chain tracked in the index (e.g. `library()` calls), potentially concatenated to the global scope chain.


crates/oak_index/src/external.rs
```diff
@@ -0,0 +1,113 @@
...
+                }
+            },
+
+            BindingSource::PackageExports(pkg_name) => {
+                let Some(pkg) = library.get(pkg_name) else {
+                    continue;
+                };
+                if pkg
+                    .exported_symbols
+                    .binary_search(&name.to_string())
```
DavisVaughan: It might be nice to make a small wrapper type for `pub exported_symbols: Vec<String>,` so that we can implement a `get()` method that does binary search under the hood

I think the fact that you have to "know" it is sorted under the hood when using it externally smells a little unsafe

crates/oak_index/src/external.rs
```diff
@@ -0,0 +1,113 @@
...
+        range: TextRange,
+    },
+
+    /// Found in an installed package (via `importFrom()`, `library()`, etc.).
+    Package { package: String, name: String },
+}
+
+/// Walk the scope chain front-to-back, returning the first match.
+pub fn resolve_external_name(
+    library: &Library,
```
DavisVaughan, line 44: I have some weird gut feeling about `library` being an external input to the system

Like it feels like `BindingSource::PackageExports` should already have all the info it needs to resolve the symbol name

I don't feel strongly enough to request a change, just pointing out something that smells kinda maybe off

crates/oak_index/tests/external.rs
```diff
@@ -0,0 +1,358 @@
...
+}
+
+#[test]
+fn test_resolve_first_match_wins() {
+    let library = test_library(vec![("stats", vec!["filter"])]);
+
+    let scope = vec![
+        file_exports("utils.R", vec![("filter", range(0, 6))]),
+        BindingSource::PackageExports("stats".to_string()),
+    ];
```
DavisVaughan, line 182: Maybe also a test for two packages that have bindings that shadow each other?

I think that last package wins in that case?

DavisVaughan (approved):

crates/oak_index/src/external.rs
lionel-: yep how we handle library is definitely up in the air

crates/oak_index/tests/external.rs
lionel-: yep last one wins

PR #1152, opened 2026-04-14
Implement package root layers with file collation support

lionel-:
Branched from #1151 
Part of https://github.com/posit-dev/ark/issues/1148

### DIFF
--- a/crates/oak_index/src/external.rs
+++ b/crates/oak_index/src/external.rs
@@ -111,12 +111,20 @@ pub fn file_layers(file: Url, index: &SemanticIndex) -> Vec<BindingSource> {
 
 /// Build the root layers for a package from its NAMESPACE.
 ///
-/// These go at the back of every file's scope chain. Currently includes
-/// `PackageExports` layers from `import()` directives. `importFrom()`
-/// names are not yet included because the NAMESPACE parser doesn't
-/// preserve which package each name came from.
+/// These go at the back of every file's scope chain:
+/// - `PackageImports` from `importFrom()` directives (name → package)
+/// - `PackageExports` from `import()` directives
 pub fn package_root_layers(namespace: &Namespace) -> Vec<BindingSource> {
     let mut layers = Vec::new();
+
+    if !namespace.imports.is_empty() {
+        let map = namespace
+            .imports
+            .iter()
+            .map(|(name, pkg)| (name.clone(), pkg.clone()))
+            .collect();
+        layers.push(BindingSource::PackageImports(map));
+    }
 
     for pkg in &namespace.package_imports {
         layers.push(BindingSource::PackageExports(pkg.clone()));
--- a/crates/oak_index/src/external.rs
+++ b/crates/oak_index/src/external.rs
@@ -121,7 +121,7 @@ pub fn package_root_layers(namespace: &Namespace) -> Vec<BindingSource> {
         let map = namespace
             .imports
             .iter()
-            .map(|(name, pkg)| (name.clone(), pkg.clone()))
+            .map(|imp| (imp.name.clone(), imp.package.clone()))
             .collect();
         layers.push(BindingSource::PackageImports(map));
     }
--- a/crates/oak_index/src/external.rs
+++ b/crates/oak_index/src/external.rs
@@ -2,6 +2,7 @@ use std::collections::HashMap;
 
 use biome_rowan::TextRange;
 use oak_package::library::Library;
+use oak_package::package_namespace::Namespace;
 use url::Url;
 
 use crate::semantic_index::DirectiveKind;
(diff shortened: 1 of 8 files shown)


crates/oak_index/src/external.rs
```diff
@@ -111,3 +112,27 @@ pub fn file_layers(file: Url, index: &SemanticIndex) -> Vec<BindingSource> {
 
     layers
 }
+
+/// Build the root layers for a package from its NAMESPACE.
+///
+/// These go at the bottom of every file's scope chain:
+/// - `PackageImports` from `importFrom()` directives (name → package)
+/// - `PackageExports` from `import()` directives
```
DavisVaughan:
```suggestion
/// These go at the back of every file's scope chain:
/// - `PackageImports` from `importFrom()` directives (name → package)
/// - `PackageExports` from `import()` directives
```

You say things like `Walk the scope chain front-to-back` elsewhere. Do you mean "back"? Because they are the last resort matches?

crates/oak_index/tests/external.rs
```diff
@@ -356,3 +357,121 @@ fn test_chained_scope_predecessor_files() {
         })
     );
 }
+
+// --- root_layers ---
+
+#[test]
+fn test_root_layers_from_namespace_imports() {
```
DavisVaughan, line 400: Do you want a test that ensures that the importFrom layer comes before the package exports layer?

crates/oak_index/tests/external.rs
```diff
@@ -356,3 +357,121 @@ fn test_chained_scope_predecessor_files() {
...
+    let mut scope = Vec::new();
+    scope.extend(file_layers(file_url("b.R"), &index_b));
+    scope.extend(file_layers(file_url("a.R"), &index_a));
+    scope.extend(package_root_layers(&ns));
+
+    // Predecessor file export
+    let result = resolve_external_name(&library, &scope, "helper_b");
+    assert!(matches!(
+        result,
+        Some(ExternalDefinition::ProjectFile { .. })
```
DavisVaughan: Probably good to check it comes from `"b.R"`?

crates/oak_index/tests/external.rs
```diff
@@ -356,3 +357,121 @@ fn test_chained_scope_predecessor_files() {
...
+    let ns = Namespace {
+        package_imports: vec!["rlang".to_string()],
+        ..Default::default()
+    };
+
+    let mut scope = Vec::new();
+    scope.extend(file_layers(file_url("utils.R"), &index));
+    scope.extend(package_root_layers(&ns));
+
+    // File export shadows the rlang root layer
```
DavisVaughan, line 535: I was also thinking it would be useful to have a test where you have 2 `index_source()`s and one shadows the other, to test the ordering (front to back)

crates/oak_package/src/package_namespace.rs
```diff
@@ -11,8 +11,9 @@ use oak_core::syntax_ext::RIdentifierExt;
 pub struct Namespace {
     /// Names of objects exported with `export()`
     pub exports: Vec<String>,
-    /// Names of objects imported with `importFrom()`
-    pub imports: Vec<String>,
+    /// Names of objects imported with `importFrom()`, with their source package.
+    /// Each entry is `(name, package)`.
+    pub imports: Vec<(String, String)>,
```
DavisVaughan:
```rust
pub struct Import {
    name: String,
    package: String
}
```

feels like time?

crates/oak_package/src/package_namespace.rs
```diff
@@ -53,8 +54,22 @@ impl Namespace {
...
+                    };
+                    let pkg_name = pkg_ident.name_text();
+
+                    for item in arg_iter {
+                        let Ok(arg) = item else { continue };
+                        let Some(AnyRExpression::RIdentifier(ident)) = arg.value() else {
+                            continue;
+                        };
+                        imports.push((ident.name_text(), pkg_name.clone()));
+                    }
```
DavisVaughan: `collect_imports()`? would definitely make reading the control flow nicer

crates/oak_package/src/package_namespace.rs
```diff
@@ -67,8 +82,8 @@ impl Namespace {
         // this but for now just be defensive.
         exports.sort();
         exports.dedup();
-        imports.sort();
-        imports.dedup();
+        imports.sort_by(|a, b| a.0.cmp(&b.0));
+        imports.dedup_by(|a, b| a.0 == b.0);
```
DavisVaughan: yea `Import` would certainly make this clearer!

crates/oak_package/src/collation.rs
```diff
@@ -0,0 +1,91 @@
+use std::collections::HashSet;
+
+use crate::package_description::Description;
+
+/// Determine the order in which R source files should be loaded.
+///
+/// If the DESCRIPTION has a `Collate` field, its whitespace-separated
+/// file list is used. Otherwise files are sorted in C locale order
+/// (byte-wise, same as R's default).
+pub fn collation_order(description: &Description, files: &[String]) -> Vec<String> {
```
DavisVaughan, line 10: This feels like it could be a method on `Description`, like `description.collate() -> Option<Vec<String>>` (we don't have to collect Collate in `new()`, it could be an on demand collection)

And the caller would handle the sorted files part, which feels like it should be quite separate, or at the very least this helper function here could do `description.collate().or_else(|| <do sorted files>)`?

DavisVaughan (approved):

PR #1153, opened 2026-04-15
Implement goto-definition with Oak

lionel-:
Branched from #1152 
Part of #1151 and #1148

Progress towards https://github.com/posit-dev/positron/issues/8631

- Introduces the `oak_ide` crate. Like in rust-analyser and ty, the _ide crate implements IDE functionality decoupled from LSP considerations (LSP positions, WorldState, etc).
- Wire up symbol resolution to our LSP handler: `oak_index` (I'd like to rename it to `oak_semantic`) → `oak_ide` → `ark` (we'll move the worldstate and LSP handlers to a new crate later on).
- Finalise cross-file resolution at WorldState level, switching between package scope and search path scope depending on project type.

For packages, I realised that the collation approach laid out in #1148 needed a significant refinement: only top-level symbols are sensitive to the collation order. Symbols in lazy scopes (e.g. inside function definitions) see the whole finalised namespace. Because of this, `FileScope::at()` takes a position in the file so it can select the right scope. You can also request the full lazy scope without a position.

Currently, scripts are fully isolated from other files. I'm working on adding support for `source()` directives in the semantic index. Then goto-definition will work more like in RStudio and only see symbols you've explicitly sourced (https://github.com/posit-dev/positron/issues/11112).

With this PR, we're ready to integrate with #1146 to jump at the source of package exports.


crates/oak_index/src/semantic_index.rs
```diff
@@ -146,6 +146,26 @@ impl SemanticIndex {
...
+
+    /// Find the use site at `offset`, if any.
+    pub fn use_at_offset(&self, offset: TextSize) -> Option<(ScopeId, UseId)> {
+        let scope = self.scope_at(offset);
+        let use_id = self
+            .uses(scope)
+            .iter()
+            .find_map(|(id, u)| u.range().contains(offset).then_some(id));
+        Some((scope, use_id?))
+    }
```
DavisVaughan, line 167: These felt a bit wrong to me because at the call site in goto_definition:
- You look up the scope twice (once for defn, once for use)
- You have the Definition and Use right here, but don't return them, then need to look them up again in goto_definition by their Id

Check out https://github.com/posit-dev/ark/pull/1159 for an alternate approach that tries to simplify this

crates/oak_ide/src/lib.rs
```diff
@@ -0,0 +1,83 @@
...
+use biome_rowan::TextSize;
+pub use goto_definition::goto_definition;
+use oak_index::external::BindingSource;
+use oak_index::semantic_index::ScopeKind;
+use oak_index::semantic_index::SemanticIndex;
+use url::Url;
+
+/// The external scope chain for a file, determined by its project context.
+#[derive(Debug)]
+pub enum FileScope {
```
DavisVaughan, line 13: This feels worth of a `file_scope.rs` file

crates/oak_ide/src/lib.rs
```diff
@@ -0,0 +1,83 @@
...
+    /// namespace is fully populated before any function runs.
+    Package {
+        top_level: Vec<BindingSource>,
+        lazy: Vec<BindingSource>,
+    },
+
+    /// Script or file outside a package. The scope chain is the R
+    /// search path: `library()` attachments from the file itself,
+    /// default packages (stats, graphics, etc.), and base.
+    SearchPath(Vec<BindingSource>),
```
DavisVaughan, line 28: ...and any source() calls?

crates/oak_ide/src/lib.rs
```diff
@@ -0,0 +1,83 @@
...
+
+    /// Return the scope chain appropriate for the given offset. For
+    /// packages, top-level scope uses predecessors only while lazy
+    /// (function) scopes see all files. For scripts, the same search
+    /// path applies everywhere.
+    pub fn at(&self, index: &SemanticIndex, offset: TextSize) -> &[BindingSource] {
+        match self {
+            Self::Package { top_level, lazy } => {
+                let scope = index.scope_at(offset);
+                match index.scope(scope).kind() {
```
DavisVaughan, line 54: Oh, like in https://github.com/posit-dev/ark/pull/1159 what if you make `scope_at()` return `(ScopeId, Scope)`? You've got it _right there_ in `scope_at()`, so looking it up twice feels like a crime!

crates/oak_ide/src/lib.rs
```diff
@@ -0,0 +1,83 @@
...
+                    ScopeKind::File => top_level,
+                    ScopeKind::Function => lazy,
+                }
+            },
+            Self::SearchPath(layers) => layers,
+        }
+    }
+
+    /// The full scope for lazy contexts. Useful for features that don't
+    /// have a cursor position (e.g. completions, workspace symbols).
```
DavisVaughan, line 64: Completions definitely have a cursor position right?

crates/oak_core/src/file.rs
```diff
@@ -0,0 +1,19 @@
+use std::path::Path;
+use std::path::PathBuf;
+
+pub fn is_r_file(path: &Path) -> bool {
+    path.extension()
+        .and_then(|e| e.to_str())
+        .is_some_and(|e| e.eq_ignore_ascii_case("r"))
+}
+
+pub fn list_r_files(dir: &Path) -> Vec<PathBuf> {
```
DavisVaughan, line 10:
```suggestion
pub fn collect_r_files(dir: &Path) -> Vec<PathBuf> {
```

maybe more in line with `collect()`?

crates/oak_ide/src/goto_definition.rs
```diff
@@ -0,0 +1,138 @@
...
+            .collect()
+    };
+
+    let external_targets = || resolve_external(library, scope_chain, symbol_name);
+
+    if !definitions.is_empty() {
+        let mut targets = local_targets(scope_id, definitions);
+        if bindings.may_be_unbound() {
+            targets.extend(external_targets());
+        }
```
DavisVaughan, line 96: Is this case something like

```
fn <- function() {
    x <<- 1
    x<here>
}
```

Where `x` has a definition in this scope but may be unbound in this scope because its really bound in the parent scope?

Confusing!

crates/oak_ide/src/goto_definition.rs
```diff
@@ -0,0 +1,138 @@
...
+        return targets;
+    }
+
+    // No local definitions. If we're in a nested scope, check enclosing
+    // bindings (the symbol might be defined in an outer function scope).
+    if let Some((enclosing_scope, enclosing_bindings)) = index.enclosing_bindings(scope_id, use_id)
```
DavisVaughan, line 102: I have a gut feeling that `enclosing_bindings()` should take an `EnclosingSnapshotKey` (or a `scope_id` and its `use_site` needed to make the key) because if you merge in https://github.com/posit-dev/ark/pull/1159 then you already have the `use_site` and you should not force `enclosing_bindings()` to re-look it up

crates/oak_ide/src/goto_definition.rs
```diff
@@ -0,0 +1,138 @@
...
+
+    external_targets()
+}
+
+fn resolve_external(
+    library: &Library,
+    scope_chain: &[BindingSource],
+    symbol_name: &str,
+) -> Vec<NavigationTarget> {
+    let Some(external) = resolve_external_name(library, scope_chain, symbol_name) else {
```
DavisVaughan, line 122: I think `resolve_external_name()` currently only knows how to resolve `mutate()` to dplyr by walking the scopes? But we should have some kind of shortcut for prefixed `dplyr::mutate()`. Maybe that comes in the next PR related to `::`!

crates/ark/src/lsp/goto_definition.rs
```diff
@@ -0,0 +1,505 @@
...
+) -> anyhow::Result<LocationLink> {
+    let doc = if let Some(open) = state.documents.get(&target.file) {
+        open
+    } else {
+        let path = target
+            .file
+            .to_file_path()
+            .map_err(|_| anyhow!("Can't convert URI to path: {}", target.file))?;
+        let contents = std::fs::read_to_string(&path)?;
+        &Document::new(&contents, None)
```
DavisVaughan, line 59: Oooooh because we can jump to files outside the project?

crates/ark/src/lsp/goto_definition.rs
```diff
@@ -0,0 +1,505 @@
...
+    );
+
+    if targets.is_empty() {
+        return Ok(None);
+    }
+
+    let links = targets
+        .into_iter()
+        .map(|target| nav_target_to_link(target, state))
+        .collect::<anyhow::Result<Vec<_>>>()?;
```
DavisVaughan, line 42: Do you want to bail on any failures? Or try and at least return partial results?

crates/oak_ide/src/goto_definition.rs
```diff
@@ -0,0 +1,138 @@
...
+            file: file.clone(),
+            name: name.to_string(),
+            full_range: def.range(),
+            focus_range: def.range(),
+        }];
+    }
+
+    // Use site: resolve through use-def map, enclosing scopes, external.
+    if let Some((scope_id, use_id)) = index.use_at_offset(offset) {
+        let scope_chain = scope.at(index, offset);
```
DavisVaughan, line 53: Would you mind pushing this `scope.at(index, offset)` call all the way down into `resolve_external()`?

It isn't used when resolving local targets and often isn't needed at all.

It confused me a bit because I was following how the LSP side `file_scope()` flows through to here, and I thought that in the script case the `default_search_path()` was the ONLY thing being searched, but really that's the only thing being searched in the script case when we need to find an "external" symbol (i.e. we also resolve all uses within this file before looking externally).

IIUC, a `library()` call at the top of a script is resolved "internally"? Which is why `default_search_path()` only includes the base packages? Which are "just there" as an external fallback regardless of any library calls?

crates/oak_ide/src/lib.rs
```diff
@@ -0,0 +1,83 @@
...
+use oak_index::external::BindingSource;
+use oak_index::semantic_index::ScopeKind;
+use oak_index::semantic_index::SemanticIndex;
+use url::Url;
+
+/// The external scope chain for a file, determined by its project context.
+#[derive(Debug)]
+pub enum FileScope {
+    /// File inside an R package. The scope chain includes layers from other
+    /// package files (ordered by collation, later files shadow earlier ones)
```
DavisVaughan, line 15: I was pretty confused by this name until I realized it was ONLY for _external_ resolution

Could we add that in the name? `ExternalScopeChain` or something? And then change `scope: &FileScope,` to `scope_chain: &ExternalScopeChain` in `oak_ide::goto_definition()`? I think that would have clarified a lot for me!

crates/ark/src/lsp/state.rs
```diff
@@ -80,6 +88,114 @@ impl WorldState {
...
+        let r_dir = pkg.path.join("R");
+        let r_dir_url = Url::from_directory_path(&r_dir).ok();
+
+        let mut filenames = HashSet::new();
+
+        // Discover open documents
+        if let Some(ref dir_url) = r_dir_url {
+            for uri in self.documents.keys() {
+                if uri.as_str().starts_with(dir_url.as_str()) {
+                    if let Some(name) = uri.path().rsplit('/').next() {
```
DavisVaughan, line 116: Aren't you really going for something like `uri.to_file_path().unwrap().file_name()`? But without the unwrap, etc

crates/ark/src/lsp/state.rs
```diff
@@ -80,6 +88,114 @@ impl WorldState {
...
+
+        // Then files on disk that aren't already known from open documents.
+        for path in list_r_files(r_dir.as_ref()) {
+            if let Some(name) = path.file_name().and_then(|n| n.to_str()) {
+                filenames.insert(name.to_string());
+            }
+        }
+
+        let filenames: Vec<String> = filenames.into_iter().collect();
+        let ordered = collation_order(&pkg.description, &filenames);
```
DavisVaughan, line 131: Can we flip the ordering of the `collation_order()` args around? IIUC it 

"takes a vector of file names and returns a vector of file names sorted by collation order"

So it feels like `files` should be first

Also, it subtly will drop any `files` that aren't in the Collate field, which I guess is right

crates/ark/src/lsp/state.rs
```diff
@@ -80,6 +88,114 @@ impl WorldState {
...
+                    }
+                }
+            }
+        }
+
+        // Then files on disk that aren't already known from open documents.
+        for path in list_r_files(r_dir.as_ref()) {
+            if let Some(name) = path.file_name().and_then(|n| n.to_str()) {
+                filenames.insert(name.to_string());
+            }
```
DavisVaughan, line 127: IIUC, you're filtering out any open files that aren't under `R/`

So you might as well not discover open documents at all, because they will be in `R/` anyways.

Just load all files under `R/`?

The collation order handles the traversal ordering if it matters, otherwise I think it shouldn't matter for a package (i.e. I don't see how open files in a package that doesn't have a Collate field should get any special treatment)

---

Actually, `collation_order()` even _sorts_ the `filenames` anyways, so we lose any potential ordering anyways in favor of alphabetical order.

Hmm, I think this whole function needs a once over review to make sure its really working as intended, I have a gut feeling it can be simplified

crates/ark/src/lsp/state.rs
```diff
@@ -80,6 +88,114 @@ impl WorldState {
...
+        }
+
+        let filenames: Vec<String> = filenames.into_iter().collect();
+        let ordered = collation_order(&pkg.description, &filenames);
+
+        let filename = file.path().rsplit('/').next().unwrap_or_default();
```
DavisVaughan, line 133: again, probably want file_name() via a `Path`

crates/ark/src/lsp/state.rs
```diff
@@ -80,6 +88,114 @@ impl WorldState {
...
+        for name in ordered.iter().rev() {
+            if name.as_str() == filename {
+                past_current = true;
+                continue;
+            }
+
+            let path = r_dir.join(name);
+            let Some(uri) = Url::from_file_path(&path).log_err() else {
+                continue;
+            };
```
DavisVaughan, line 151: I think something is kinda wasteful here.

When we iterate over `r_dir` with `list_r_files()` we get a vector of PathBufs already, then we are converting them to strings in `filenames`/`ordered`, and now back to `PathBuf`s. 

It seems likely we can just keep them as `PathBuf`s the whole time, and teach `collation_order()` how to sort a vector of PathBuf instead.

crates/ark/src/lsp/state.rs
```diff
@@ -80,6 +88,114 @@ impl WorldState {
...
+
+            // Use the open document if available, otherwise read from disk.
+            // TODO: Store non-opened workspace documents in VFS.
+            let doc = if let Some(open) = self.documents.get(&uri) {
+                open
+            } else {
+                let Ok(contents) = std::fs::read_to_string(&path) else {
```
DavisVaughan, line 158: `.log_err()`?

crates/ark/src/lsp/state.rs
```diff
@@ -80,6 +88,114 @@ impl WorldState {
...
+                    continue;
+                };
+                &Document::new(&contents, None)
+            };
+
+            let layers = file_layers(uri, &doc.semantic_index());
+            lazy.extend(layers.clone());
+            if past_current {
+                top_level.extend(layers);
```
DavisVaughan, line 167: Tricky little use of `past_current` here!

I guess this also relies on the fact that `collation_order()` sorts in alphabetical order even without a Collate field, which is also the order that R sources them?

crates/ark/src/lsp/state.rs
```diff
@@ -80,6 +88,114 @@ impl WorldState {
...
+            }
+        }
+
+        top_level.extend(root_layers.clone());
+        top_level.push(BindingSource::PackageExports("base".to_string()));
+        lazy.extend(root_layers);
+        lazy.push(BindingSource::PackageExports("base".to_string()));
```
DavisVaughan, line 174: Not utils/tools and friends because in a package they require an Import, right?

DavisVaughan (approved):
Two main things

- https://github.com/posit-dev/ark/pull/1159/ to try and simplify some lookups
- `file_scope()` feels like it could use another pass, found quite a few things there that smell off

crates/oak_ide/src/lib.rs
DavisVaughan: Update: See i was confused by the name! I didn't realize this was only for external things

crates/oak_ide/src/lib.rs
lionel-: An array lookup isn't exactly a crime but yep sounds good

crates/oak_core/src/file.rs
lionel-: hmm I would expect from `collect_` to read from the files?

crates/oak_ide/src/goto_definition.rs
lionel-: It's more about conditional definitions:

```r
if (cond) {
  x <- 1
}
use(x)
```

The variable is resolved either locally (if cflow went through the branch) or externally (if it didn't).

crates/oak_ide/src/goto_definition.rs
lionel-: yes that's the next PR

crates/ark/src/lsp/goto_definition.rs
lionel-: It's for files in the project but not opened as document (i.e. workspace files). I feel like I've abstracted that in a method, maybe in an ulterior PR.

crates/ark/src/lsp/goto_definition.rs
lionel-: Good idea.

And look how simple the pattern is with `log_err()`!

```rs
    let links: Vec<_> = targets
        .into_iter()
        .filter_map(|target| nav_target_to_link(target, state).log_err())
        .collect();
```

crates/oak_core/src/file.rs
DavisVaughan: oh that's reasonable

crates/oak_ide/src/goto_definition.rs
lionel-: > Would you mind pushing this scope.at(index, offset) call all the way down into resolve_external()?

Good idea!

> IIUC, a library() call at the top of a script is resolved "internally"?

Actually no, it's part of external resolution but that wasn't wired in for scripts. Now fixed.

crates/oak_ide/src/lib.rs
lionel-: Ah I see the confusion. "File" here is meant like "TopLevel", but really it's about what's _outside_ the file.

I'll rename to `ExternalScope`, and `BindingSource` to `ScopeLayer`.

crates/ark/src/lsp/state.rs
lionel-: I've added a comment:

```rs
        // This silently discards files not listed in the Collation field. TODO:
        // We should lint when that is the case, mirroring how R CMD INSTALL
        // errors out.
```

crates/ark/src/lsp/state.rs
lionel-: hmm right, it's about file names here, not contents, and there can't be e.g. an unsaved file that doesn't have an on-disk version, since those are untitled.

crates/ark/src/lsp/state.rs
lionel-: Right only the base namespace is in scope for packages

lionel-: Let's move this to `package_root_layers()` (with doc)

PR #1154, opened 2026-04-15
Add plumbing for resolving `::` and `:::`

lionel-:
Branched from #1153 

Currently unused in goto-definition but in the future will map to package source provided by #1146 

Supported by new `oak_ide::Identifier` struct with `classify()` constructor.

### DIFF
--- a/crates/ark/src/lsp/goto_definition.rs
+++ b/crates/ark/src/lsp/goto_definition.rs
@@ -25,9 +25,11 @@ pub(crate) fn goto_definition(
     )?;
 
     let index = document.semantic_index();
+    let root = document.syntax();
     let targets = oak_ide::goto_definition(
         offset,
         &uri,
+        &root,
         &index,
         &state.file_scope(&uri),
         &state.library,
--- a/crates/ark/src/lsp/goto_definition.rs
+++ b/crates/ark/src/lsp/goto_definition.rs
@@ -24,9 +24,11 @@ pub(crate) fn goto_definition(
     )?;
 
     let index = document.semantic_index();
+    let root = document.syntax();
     let targets = oak_ide::goto_definition(
         offset,
         &uri,
+        &root,
         &index,
         &state.file_scope(&uri),
         &state.library,
--- a/crates/ark/src/lsp/goto_definition.rs
+++ b/crates/ark/src/lsp/goto_definition.rs
@@ -25,7 +25,7 @@ pub(crate) fn goto_definition(
     )?;
 
     let index = document.semantic_index();
-    let root = document.syntax();
+    let root = document.syntax()?;
     let targets = oak_ide::goto_definition(
         offset,
         &uri,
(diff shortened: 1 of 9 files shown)


crates/oak_index/src/external.rs
```diff
@@ -84,6 +84,26 @@ pub fn resolve_external_name(
     None
 }
 
+/// Resolve a name in a specific package's exported symbols.
+pub fn resolve_in_package(
```
DavisVaughan, line 88: Use in `resolve_external_name()` for the `BindingSource::PackageExports` branch ?

crates/ark/src/lsp/goto_definition.rs
```diff
@@ -24,9 +24,11 @@ pub(crate) fn goto_definition(
     )?;
 
     let index = document.semantic_index();
+    let root = document.syntax();
```
DavisVaughan, line 27: hmmmmm something something...what happens when there are parse errors?

crates/oak_ide/src/goto_definition.rs
```diff
@@ -26,35 +29,45 @@ use crate::NavigationTarget;
...
 
-        return vec![NavigationTarget {
-            file: file.clone(),
-            name: name.to_string(),
-            full_range: def.range(),
-            focus_range: def.range(),
-        }];
-    }
+    match ident {
+        Identifier::Definition { scope_id, def_id } => {
```
DavisVaughan, line 50: I think this may have to be reworked on ideas from https://github.com/posit-dev/ark/pull/1159, but I do think that is probably worth it, because it reduces the number of lookups

crates/oak_ide/src/identifier.rs
```diff
@@ -0,0 +1,82 @@
...
+    pub fn classify(root: &RSyntaxNode, index: &SemanticIndex, offset: TextSize) -> Option<Self> {
+        if let Some((scope_id, def_id)) = index.definition_at_offset(offset) {
+            return Some(Identifier::Definition { scope_id, def_id });
+        }
+
+        if let Some((scope_id, use_id)) = index.use_at_offset(offset) {
+            return Some(Identifier::Use { scope_id, use_id });
+        }
+
+        classify_namespace(root, offset)
```
DavisVaughan, line 46:
```suggestion
        if let Some(namespace_access) = classify_namespace(root, offset) {
            return Some(namespace_access);
        }

        None
```

this feels a bit more clear to me that we might add a number of additional conditions to this, and if none hit then we fall through to `None`

DavisVaughan (approved):

crates/ark/src/lsp/goto_definition.rs
lionel-: you know what, I'll go ahead and make this return a Result ;)

DavisVaughan: For Document::syntax() that is probably a good idea!

PR #1156, opened 2026-04-15
Try `use_def_maps: IndexVec<ScopeId, UseDefMapBuilder>`

DavisVaughan:
Merges into #1144 

My attempt to get rid of `current_use_def` and `use_def_stack` in favor of changing

```
# from
use_def_maps: IndexVec<ScopeId, UseDefMap>

# to
use_def_maps: IndexVec<ScopeId, UseDefMapBuilder>
```

Which is exactly how `symbol_tables` works.

Give this a close review, because I'm going off what my gut was telling me from reading #1144 closely, but I don't have a strong mental model yet.

All tests pass for me, which at the very least means we should add a test to #1144 to prove why what I've changed here _doesn't_ work.

### DIFF
--- a/crates/oak_index/src/builder.rs
+++ b/crates/oak_index/src/builder.rs
@@ -30,7 +30,6 @@ use crate::semantic_index::SymbolFlags;
 use crate::semantic_index::SymbolTableBuilder;
 use crate::semantic_index::Use;
 use crate::semantic_index::UseId;
-use crate::use_def_map::UseDefMap;
 use crate::use_def_map::UseDefMapBuilder;
 
 /// Build a [`SemanticIndex`] from a parsed R file.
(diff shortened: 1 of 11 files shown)


crates/oak_index/src/builder.rs
```diff
@@ -210,34 +192,14 @@ impl SemanticIndexBuilder {
...
-            scope = match self.scopes[scope].parent {
-                Some(parent) => parent,
-                None => panic!("Target scope {target:?} is not an ancestor of current scope"),
-            };
-            steps += 1;
-        }
-
-        let index = self.use_def_stack.len() - steps;
-        &mut self.use_def_stack[index]
-    }
```
DavisVaughan: We lose this whole helper but I don't think it ends up actually mattering anymore based on its 1 usage

lionel-:
Much better, thanks! Cherry picked into parent.

PR #1157, opened 2026-04-16
Add support for `source()` directive in symbol resolution

### LINKED ISSUE #1148: Cross-file symbol resolution

lionel-:
Branched from #1154 

Closes https://github.com/posit-dev/ark/issues/1148

Adds support for `source()` directives in scripts, allowing for cross-file symbol resolution. Like in RStudio, sourced files are resolved from the workspace directory. Effects on the LSP:

- Top-level definitions in the sourced files are made available below the `source()` call.
- Side effects from top-level `library()` and `require()` calls are also made available after the `source()` call.
- Sourced files can transitively source other files.

For now this only affects goto-definition, but should later also affect diagnostics such as unused symbols.

Both `source()` and `library()` are allowed in lazy nested scopes such as function bodies. In that case the side effects are available inside that scope (and nested children), but not outside, even though the effects are global. That's because outside that scope, code can't make assumptions about the availability of the symbols.

With a `helpers.R` file containing:

```r
foo <- function() {}
library(dplyr)
```

We have at top-level:

```r
# Both undefined
print(foo)
print(mutate)

source("helpers.R")

# Now defined
print(foo)
print(mutate)
```

In nested scopes:

```r
# Both undefined
print(foo)
print(mutate)

function() {
  source("helpers.R")
  # Now defined
  print(foo)
  print(mutate)
}

# Still undefined
print(foo)
print(mutate)
```

At top-level, it doesn't matter whether `local` is true or false, the sourced definitions overwrite any local ones. We have a new `DefinitionKind::Sourced` variant for these definitions:

```r
foo <- "local"
source("helpers.R")
print(foo) # Resolves to helper's definition
```

In nested scopes, if `local = FALSE`, the sourced definitions are made available via cross-file directives instead, that are only used for unbound symbols. This allows proper resolution of local symbols:

```r
function() {
  foo <- 1
  source("helpers.R")
  print(foo) # Still refers to the local definition
}
```

If `local` is true, sourced definitions are instead added to the local scope.

### DIFF
--- a/crates/ark/src/lsp/goto_definition.rs
+++ b/crates/ark/src/lsp/goto_definition.rs
@@ -504,4 +504,157 @@ mod tests {
         // `is.null` is missing from the INDEX-based export list.
         assert_eq!(result, None);
     }
+
+    // --- source() directive in scripts ---
+
+    #[test]
+    fn test_script_source_directive_resolves() {
+        // script.R has `source("helpers.R")` then uses `helper`.
+        // WorldState::file_scope() should resolve the source() directive
+        // and make helpers.R's exports visible via the search path.
+        let script_dir = std::env::temp_dir().join("test_script_source");
+
+        let helpers_doc = Document::new("helper <- function() 1\n", None);
+        let helpers_uri = lsp_types::Url::from_file_path(script_dir.join("helpers.R")).unwrap();
+
+        let script_doc = Document::new("source(\"helpers.R\")\nhelper\n", None);
+        let script_uri = lsp_types::Url::from_file_path(script_dir.join("script.R")).unwrap();
+
+        let mut state = WorldState::default();
+        state.documents.insert(helpers_uri.clone(), helpers_doc);
+        state
+            .documents
+            .insert(script_uri.clone(), script_doc.clone());
+
+        // Cursor on `helper` (line 1, col 0)
+        let params = make_params(script_uri, 1, 0);
+        assert_matches!(
+            goto_definition(&script_doc, params, &state).unwrap(),
+            Some(GotoDefinitionResponse::Link(ref links)) => {
+                assert_eq!(links[0].target_uri, helpers_uri);
+                assert_eq!(
+                    links[0].target_range,
+                    lsp_types::Range {
+                        start: lsp_types::Position::new(0, 0),
+                        end: lsp_types::Position::new(0, 6),
+                    }
+                );
+            }
+        );
+    }
+
+    #[test]
+    fn test_script_source_directive_resolves_nested_library() {
+        // helpers.R has `library(dplyr)` and defines `helper`.
+        // script.R sources helpers.R then uses `mutate` (from dplyr).
+        // The nested library() directive should be visible through
+        // the source() resolution.
+        let Some(library) = r_library() else {
+            eprintln!("skipping: R not found");
+            return;
+        };
+
+        let script_dir = std::env::temp_dir().join("test_script_source_nested");
+
+        let helpers_doc = Document::new("library(dplyr)\nhelper <- function() 1\n", None);
+        let helpers_uri = lsp_types::Url::from_file_path(script_dir.join("helpers.R")).unwrap();
+
+        let script_doc = Document::new("source(\"helpers.R\")\nmutate\nhelper\n", None);
+        let script_uri = lsp_types::Url::from_file_path(script_dir.join("script.R")).unwrap();
+
+        let mut state = make_state(&script_uri, &script_doc);
+        state.library = library;
+        state.documents.insert(helpers_uri.clone(), helpers_doc);
+
+        // `mutate` (line 1) resolves via dplyr, attached by helpers.R's library() call.
+        // Package symbol, no NavigationTarget.
+        let params = make_params(script_uri.clone(), 1, 0);
+        let result = goto_definition(&script_doc, params, &state).unwrap();
+        assert_eq!(result, None);
+
+        // `helper` (line 2) resolves to helpers.R's definition
+        let params = make_params(script_uri, 2, 0);
+        assert_matches!(
+            goto_definition(&script_doc, params, &state).unwrap(),
+            Some(GotoDefinitionResponse::Link(ref links)) => {
+                assert_eq!(links[0].target_uri, helpers_uri);
+                assert_eq!(
+                    links[0].target_range,
+                    lsp_types::Range {
+                        start: lsp_types::Position::new(1, 0),
+                        end: lsp_types::Position::new(1, 6),
+                    }
+                );
+            }
+        );
+    }
+
+    #[test]
+    fn test_script_source_resolves_from_disk() {
+        // helpers.R exists on disk but is NOT in state.documents.
+        // script.R sources it. The resolver should read from disk.
+        let script_dir = tempfile::tempdir().unwrap();
+
+        let helpers_path = script_dir.path().join("helpers.R");
+        std::fs::write(&helpers_path, "helper <- function() 1\n").unwrap();
+        let helpers_uri = lsp_types::Url::from_file_path(&helpers_path).unwrap();
+
+        let script_doc = Document::new("source(\"helpers.R\")\nhelper\n", None);
+        let script_uri =
+            lsp_types::Url::from_file_path(script_dir.path().join("script.R")).unwrap();
+
+        let mut state = WorldState::default();
+        state
+            .documents
+            .insert(script_uri.clone(), script_doc.clone());
+        // helpers.R is intentionally NOT inserted into state.documents
+
+        let params = make_params(script_uri, 1, 0);
+        assert_matches!(
+            goto_definition(&script_doc, params, &state).unwrap(),
+            Some(GotoDefinitionResponse::Link(ref links)) => {
+                assert_eq!(links[0].target_uri, helpers_uri);
+                assert_eq!(
+                    links[0].target_range,
+                    lsp_types::Range {
+                        start: lsp_types::Position::new(0, 0),
+                        end: lsp_types::Position::new(0, 6),
+                    }
+                );
+            }
+        );
+    }
+
+    #[test]
+    fn test_script_file_scope_from_disk() {
+        // The script itself is on disk, not in state.documents.
+        // file_scope should still read it and resolve its directives.
+        let script_dir = tempfile::tempdir().unwrap();
+
+        let helpers_path = script_dir.path().join("helpers.R");
+        std::fs::write(&helpers_path, "helper <- function() 1\n").unwrap();
+
+        let script_path = script_dir.path().join("script.R");
+        std::fs::write(&script_path, "source(\"helpers.R\")\nhelper\n").unwrap();
+        let script_uri = lsp_types::Url::from_file_path(&script_path).unwrap();
+
+        let script_doc = Document::new("source(\"helpers.R\")\nhelper\n", None);
+
+        // Neither file is in state.documents
+        let state = WorldState::default();
+
+        let params = make_params(script_uri, 1, 0);
+        assert_matches!(
+            goto_definition(&script_doc, params, &state).unwrap(),
+            Some(GotoDefinitionResponse::Link(ref links)) => {
+                assert_eq!(
+                    links[0].target_range,
+                    lsp_types::Range {
+                        start: lsp_types::Position::new(0, 0),
+                        end: lsp_types::Position::new(0, 6),
+                    }
+                );
+            }
+        );
+    }
 }
--- a/crates/ark/src/lsp/goto_definition.rs
+++ b/crates/ark/src/lsp/goto_definition.rs
@@ -984,4 +984,33 @@ mod tests {
         let outer_chain = file_scope.at(&index, outer_offset);
         assert!(!has_dplyr(&outer_chain));
     }
+
+    #[test]
+    fn test_script_source_later_shadows_earlier() {
+        // When two sourced files define the same name, the later
+        // source() call shadows the earlier one.
+        let dir = tempfile::tempdir().unwrap();
+
+        std::fs::write(dir.path().join("a.R"), "foo <- 1\n").unwrap();
+        std::fs::write(dir.path().join("b.R"), "foo <- 2\n").unwrap();
+
+        let script_doc = Document::new("source(\"a.R\")\nsource(\"b.R\")\nfoo\n", None);
+        let script_uri = lsp_types::Url::from_file_path(dir.path().join("script.R")).unwrap();
+
+        let mut state = WorldState::default();
+        state
+            .documents
+            .insert(script_uri.clone(), script_doc.clone());
+
+        let b_uri = lsp_types::Url::from_file_path(dir.path().join("b.R")).unwrap();
+
+        // `foo` (line 2) should resolve to b.R (later source shadows earlier)
+        let params = make_params(script_uri, 2, 0);
+        assert_matches!(
+            goto_definition(&script_doc, params, &state).unwrap(),
+            Some(GotoDefinitionResponse::Link(ref links)) => {
+                assert_eq!(links[0].target_uri, b_uri);
+            }
+        );
+    }
 }
--- a/crates/ark/src/lsp/goto_definition.rs
+++ b/crates/ark/src/lsp/goto_definition.rs
@@ -23,16 +23,10 @@ pub(crate) fn goto_definition(
         document.position_encoding,
     )?;
 
-    let index = document.semantic_index();
+    let (index, file_scope) = state.file_analysis(&uri, document);
     let root = document.syntax();
-    let targets = oak_ide::goto_definition(
-        offset,
-        &uri,
-        &root,
-        &index,
-        &state.file_scope(&uri),
-        &state.library,
-    );
+    let targets =
+        oak_ide::goto_definition(offset, &uri, &root, &index, &file_scope, &state.library);
 
     if targets.is_empty() {
         return Ok(None);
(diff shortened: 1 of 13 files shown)


crates/ark/src/lsp/state.rs
```diff
@@ -89,15 +94,39 @@ impl WorldState {
...
+    /// TODO: Replace with a proper VFS so non-opened workspace documents
+    /// are cached rather than re-read on every query.
+    fn workspace_document(&self, uri: &Url) -> Option<Document> {
+        if let Some(doc) = self.documents.get(uri) {
+            return Some(doc.clone());
+        }
+        let path = uri.to_file_path().log_err()?;
+        let contents = std::fs::read_to_string(&path).log_err()?;
+        Some(Document::new(&contents, None))
+    }
```
DavisVaughan, line 109: I'm a bit worried about getting in too deep with something that _clones_ an already open document on every retrieval. That sounds like a nasty performance hit.

It definitely feels like we want to return a reference here, we rarely if ever want to own the `Document` at the use site

I realize that makes things difficult due to the create-the-doc-on-the-fly path. I guess we really need the VFS as a place to "put" the document that we can then return a ref from.

crates/ark/src/lsp/state.rs
```diff
@@ -149,16 +178,8 @@ impl WorldState {
...
-            let doc = if let Some(open) = self.documents.get(&uri) {
-                open
-            } else {
-                let Ok(contents) = std::fs::read_to_string(&path) else {
-                    continue;
-                };
-                &Document::new(&contents, None)
+            let Some(doc) = self.workspace_document(&uri) else {
+                continue;
             };
```
DavisVaughan, line 183: yea see we are trading a ref for an owned Document here, which doesn't feel great to me

crates/ark/src/lsp/state.rs
```diff
@@ -173,7 +194,98 @@ impl WorldState {
...
+
+    fn script_file_analysis(&self, file: &Url, doc: &Document) -> (SemanticIndex, FileScope) {
+        // Resolve `source()` paths relative to the workspace root,
+        // matching RStudio's behaviour of setting the working directory
+        // to the project root. Fall back to the file's own directory
+        // when no workspace folder is open.
+        let file_dir = file
+            .to_file_path()
+            .ok()
+            .and_then(|p| p.parent().map(|d| d.to_path_buf()));
```
DavisVaughan, line 208: This is eagerly computing file_dir when it isn't required most of the time

crates/oak_index/src/builder.rs
```diff
@@ -43,16 +47,46 @@ use crate::use_def_map::UseDefMapBuilder;
...
     builder.collect_expression_list(&root.expressions());
     builder.finish()
 }
 
+/// The result of resolving a `source()` call. Returned by the resolver
+/// callback passed to the builder.
+pub struct SourceResolution {
+    /// Definitions to inject as synthetic bindings in the calling scope.
+    /// Each entry is (name, file_url, range_in_source_file).
+    pub definitions: Vec<(String, Url, TextRange)>,
```
DavisVaughan, line 78: `SourceResolutionDefinition` might be a reasonable struct here (if i feel like i need to document the fields, i feel like i just need a struct)

crates/ark/src/lsp/state.rs
```diff
@@ -173,7 +194,98 @@ impl WorldState {
...
+        stack: &mut HashSet<Url>,
+    ) -> Option<SourceResolution> {
+        let resolved = base_dir.join(path);
+        let url = Url::from_file_path(&resolved).log_err()?;
+
+        if !stack.insert(url.clone()) {
+            return None;
+        }
+
+        let sourced_doc = self.workspace_document(&url)?;
```
DavisVaughan, line 251: This early exit feels like a problem

- You add to stack
- Doc doesn't exist, so workspace_document bails early
- You never remove from the stack

In general it feels like we need to ensure the stack is popped on any exit from this function

crates/ark/src/lsp/state.rs
```diff
@@ -173,7 +194,98 @@ impl WorldState {
...
+                oak_index::semantic_index::DirectiveKind::Attach(pkg) => {
+                    packages.push(pkg.clone());
+                },
+                oak_index::semantic_index::DirectiveKind::Source {
+                    file: source_file,
+                    exports,
+                } => {
+                    for (name, range) in exports {
+                        definitions.push((name.clone(), source_file.clone(), *range));
+                    }
```
DavisVaughan, line 278: Is this for when a `source()`d file also `source()`'s another file?

It just extends the list of definitions present in the 1st sourced file? Does the ordering matter? Like what if that 1st sourced file has `source("helpers.R")` in it, but also defines `fn <- function() {}`. The order of where those are placed in that 1st sourced file probably matters, but I'm not sure if that is reflected here

crates/ark/src/lsp/goto_definition.rs
```diff
@@ -504,4 +498,659 @@ mod tests {
...
+    #[test]
+    fn test_script_source_resolves_from_workspace_root() {
+        // source() paths are resolved relative to the workspace root,
+        // not the sourcing file's directory. Here script.R is in a
+        // subdirectory but sources "helpers.R" which lives at the root.
+        let dir = tempfile::tempdir().unwrap();
+        let subdir = dir.path().join("subdir");
+        std::fs::create_dir(&subdir).unwrap();
+
+        std::fs::write(dir.path().join("helpers.R"), "helper <- function() 1\n").unwrap();
```
DavisVaughan, line 513: we might want to document this somewhere?

crates/ark/src/lsp/goto_definition.rs
```diff
@@ -504,4 +498,659 @@ mod tests {
...
+        std::fs::write(
+            dir.path().join("a.R"),
+            "source(\"b.R\")\nfrom_a <- function() 2\n",
+        )
+        .unwrap();
+
+        let script_doc = Document::new("source(\"a.R\")\nfrom_a\nfrom_b\nmutate\n", None);
+        let script_uri = lsp_types::Url::from_file_path(dir.path().join("script.R")).unwrap();
+
+        let Some(library) = r_library() else {
```
DavisVaughan, line 706: I think this example gets at the heart of one of my other questions about how `source("b.R")` is loaded in.

If `b.R` also has `from_a`, does this ordering in `a.R` matter? I think it should? But I'm not sure it currently does

```
source("b.R") # has from_a
from_a <- function() {}
```

```
from_a <- function() {}
source("b.R") # has from_a
```

crates/oak_ide/src/lib.rs
```diff
@@ -25,14 +28,27 @@ pub enum FileScope {
...
-    /// search path: `library()` attachments from the file itself,
+    /// search path: `library()` and `source()` directives from the file
+    /// itself (position-stamped, only active after their call site),
     /// default packages (stats, graphics, etc.), and base.
-    SearchPath(Vec<BindingSource>),
+    SearchPath {
+        /// Layers from the file's own top-level directives, each stamped with
+        /// the offset of the directive that produced them. Only layers with
+        /// offset <= cursor position are active.
+        directive_layers: Vec<(TextSize, ScopeId, BindingSource)>,
```
DavisVaughan, line 38: It might be nice if `directives: Vec<Directive>` over in `SemanticIndex` and `SemanticIndexBuilder` were really built as a `BTreeSet` sorted on `offset`

So then when you build `directive_layers` from that, you get another `BTreeSet` you can use `BTreeSet::range()` on to quickly grab all directives `<= cursor`.

Just a thought

crates/oak_index/src/external.rs
```diff
@@ -117,15 +123,34 @@ pub fn file_layers(file: Url, index: &SemanticIndex) -> Vec<BindingSource> {
...
+                layers.push((offset, directive.scope(), BindingSource::FileExports {
+                    file: file.clone(),
+                    exports: exports.clone(),
+                }));
             },
         }
     }
-
     layers
 }
```
DavisVaughan, line 155: This function is used in quite a few places, but are we sure we really need it?

It doesn't look like it adds anything new that isn't already in `Directive` itself, it just...rearranges and clones the fields?

If the `BindingSource` is the important added info, maybe a `Directive` should just have a `binding_source() -> BindingSource` method on it that can return its source?

DavisVaughan: I also like that `Directive` is a named object, because seeing the unnamed `(TextSize, ScopeId, BindingSource)` floating around in a few places gets a little confusing, im left thinking "what is all this again?"

DavisVaughan: In general there seems to be a lot of overlap between `BindingSource` and `DirectiveKind`, making me wonder if both are required

crates/ark/src/lsp/state.rs
DavisVaughan: Oh can we `Cow` the `Document` in the meantime?

crates/oak_ide/src/lib.rs
```diff
@@ -25,14 +28,27 @@ pub enum FileScope {
...
+        // FIXME: Redundant with `ScopeId(0)` check but avoids importing
+        // the index type at construction time. Remove once we have salsa.
+        /// Always-visible base layers (default packages, base).
+        base: Vec<BindingSource>,
```
DavisVaughan, line 42: Maybe `default_search_path()` should move into `oak_ide` or `oak_index` and be a global variable accessible at any time for purposes like this where we know we always have `base`?

It seems a little odd that it currently lives over in the lsp

crates/oak_index/src/semantic_index.rs
```diff
@@ -119,13 +121,36 @@ impl SemanticIndex {
         let symbols = &self.symbol_tables[file_scope];
         self.definitions[file_scope]
             .iter()
+            .filter(|(_id, def)| !matches!(def.kind(), DefinitionKind::Sourced { .. }))
```
DavisVaughan, line 124: Is it better to be explicit about what you are filtering _for_ rather than what to filter _out_? i.e. if you add another DefinitionKind in the future, it would be nice if it forced a diff here to force us to acknowledge if it is a file export or not

crates/oak_index/src/semantic_index.rs
```diff
@@ -486,6 +512,11 @@ pub enum DefinitionKind {
     SuperAssignment(RSyntaxNode),
     Parameter(RSyntaxNode),
     ForVariable(RSyntaxNode),
+    /// Injected from a `source()` call. The definition lives in an external
+    /// file; `range` on the `Definition` gives the name's range in that file.
+    Sourced {
+        file: Url,
+    },
```
DavisVaughan, line 519: This somehowwww feels wrong to me. Like it needed to be a `ExternalDefinition` struct that allowed for this

I have somewhat strong gut feelings that a `Definition` should not be allowed to jump across files, but I can't quite explain why

crates/oak_index/src/semantic_index.rs
```diff
@@ -528,13 +559,18 @@ impl Use {
...
 
-// TODO: `Source()` directives
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum DirectiveKind {
     /// `library(pkg)` or `require(pkg)`: attaches a package to the search path.
     Attach(String),
+    /// `source(file)`: brings exports from another file into scope.
+    Source {
+        file: Url,
+        exports: HashMap<String, TextRange>,
```
DavisVaughan, line 572: I find it somewhat interesting that the package exports are "lazy" here but the source exports are not

crates/oak_index/src/builder.rs
```diff
@@ -663,26 +703,33 @@ impl SemanticIndexBuilder {
...
+    fn collect_attach_directive(&mut self, call: &aether_syntax::RCall) {
         let Ok(args) = call.arguments() else {
             return;
         };
         let mut items = args.items().iter();
 
-        // For now, only recognise exactly one unnamed argument. We'll do
-        // argument matching later (`character.only` unquoting is another
-        // complication).
         let Some(Ok(first_arg)) = items.next() else {
```
DavisVaughan, line 733: should probably keep this

crates/oak_index/src/builder.rs
```diff
@@ -703,12 +749,173 @@ impl SemanticIndexBuilder {
...
+    }
+
+    /// Call the source resolver for `path`, temporarily taking it out of
+    /// `self` to avoid borrow conflicts.
+    fn resolve_source(&mut self, path: &str) -> Option<SourceResolution> {
+        let mut resolver = self.source_resolver.take()?;
+        let result = resolver(path);
+        self.source_resolver = Some(resolver);
+        result
+    }
```
DavisVaughan, line 917:
```suggestion
    fn resolve_source(&mut self, path: &str) -> Option<SourceResolution> {
        let source_resolver = self.source_resolver.as_mut()?;
        source_resolver(path)
    }
```

crates/oak_index/src/builder.rs
```diff
@@ -703,12 +749,173 @@ impl SemanticIndexBuilder {
...
+
+        if is_local || !in_nested_scope {
+            // `local = TRUE` or at file scope: inject into the
+            // use-def map so sourced definitions shadow locals.
+            for (name, file, range) in resolution.definitions {
+                self.add_definition(
+                    &name,
+                    SymbolFlags::IS_BOUND,
+                    DefinitionKind::Sourced { file },
+                    range,
```
DavisVaughan, line 875: Maybe something that throws me off is that when you collect a `source()` directive, the `range` doesn't correspond to the range of `source()`, which feels wrong to me for ordering things like 

```
source("helpers.R") # has fn()
fn <- function() {}
```

Like, IIUC the "definition" underlying `source()` no longer knows where it originated from in the file?

DavisVaughan (commented):
Something feels a little off here but I can't quite put my finger on it. Like, concretely I am weirded out by the fact that our `definitions` table can include a `range` into a different file. That somehow feels like we aren't quite modeling that correctly.

It seems like one other way you could model this is:

- `SemanticIndex` has a `dependencies: Vec<Dependency>` field
- `Dependency { range: Range, index_id: SemanticIndexId }`

The idea being that when you encounter a `source("bar.R")` call while building the semantic index for `foo.R`, you build the semantic index for `bar.R` and then store it somewhere (im hand waving here about where it is stored, maybe a Salsa db?) _without modification or simplification_. And then you _reference it_ from `foo.R`'s semantic index by ID as being included at some `Range` in the file (the location of the `source("bar.R")` call).

Then your methods on `SemanticIndex` would have to know how to layer in `dependencies` when you ask for things like `file_all_definitions()`, and maybe even also for things as simple as `SemanticIndex::symbols()`, BUT I think that Salsa would probably help with that, because it would be computed once and then as long as the file underlying `SemanticIndexId` doesn't change, then `symbols()` just returns a cached result.

It also means it would be quite easy for us to _limit_ `symbols()` to only return `foo.R`'s direct symbols if we wanted to. Right now it gets blurred with the ones from `bar.R`.

---

I think the biggest source of weirdness in my head right now is that the model for `SemanticIndex` gets a bit blurry with this PR. It used to be that `SemanticIndex` modeled symbols and definitions found in _exactly one file_. Now `SemanticIndex::symbol_tables` and `SemanticIndex::definitions` can _cross files_ and that feels quite weird to me, and breaks my mental model of how to use this thing.

If `SemanticIndex` instead just _referenced another semantic index_ at some `Range`, then that feels more correct again, because its own internal `symbol_tables` and `definitions` would still be _only_ about itself.

I think that would _also_ make it much easier to update if, say, `bar.R` is updated by the user. You get to change ABSOLUTELY NOTHING about `foo.R`'s `SemanticIndex` object. All of the `foo.R` fields are still completely valid because nothing in `foo.R` changed, and the fields only model `foo.R`. And then the _reference_ to `bar.R`'s `SemanticIndex` would pull in the updated symbols/definitions from `bar.R` post-change, but Salsa would help us know when it gets invalidated.

lionel-:
> I think that would also make it much easier to update if, say, bar.R is updated by the user. You get to change ABSOLUTELY NOTHING about foo.R's SemanticIndex object.

ha! But in the next PR that introduces NSE, updating `bar.R` can invalidate the ENTIRE scope structure of `foo.R`! Fun times ahead ;) In that branch we do some fixpoint iteration, just like rust-analyzer has to for macros.

Without this upcoming NSE work, I don't think I'd have introduced support for `local = TRUE` because, as you noticed, it introduces external resolution in the index builder. But for NSE support, external resolution is a requirement. `identity({ ... })` might or might not create a new scope.

Regarding text ranges, I think it's more that we need definitions to feature some notion of file origin, like in ty's index. That said I'll take a deeper look and think more about your comments when I get back from vacations.

crates/ark/src/lsp/goto_definition.rs
lionel-: Good point. We should use the same approach as `local = TRUE` and create new definitions in the semantic index, instead of resolving through the cross-file path. This way we can have precise ordering of the symbol definitions across source calls.

Edit: Done

crates/ark/src/lsp/state.rs
lionel-: I expect all of these performance issues to eventually be tackled as part of the switch to Salsa, where documents (opened or workspace) will be referred to with IDs. But it's a good idea to use a Cow in the meantime.

crates/oak_index/src/builder.rs
lionel-: We're no longer storing definitions from the external file in the resolution. Instead, our imported definitions point into the external file (with file id + name), and we re-resolve the target definitions on demand (which will be cheap with Salsa).

crates/ark/src/lsp/state.rs
lionel-: This should now be much better:

- Sourced symbols are injected as `Import`-kind definitions in a scope.
- We're no longer eagerly resolving the definitions. Instead the imported definitions point at the file, and the resolver uses the local index to resolve at top-level.

crates/ark/src/lsp/goto_definition.rs
lionel-: There's a new `source_root()` method that documents this.

crates/oak_ide/src/lib.rs
lionel-: Although linear scans might be faster.

crates/oak_index/src/external.rs
lionel-: Good call, removing this made things much simpler.

crates/oak_index/src/semantic_index.rs
lionel-: All lazy now

crates/oak_index/src/semantic_index.rs
lionel-: This is now resolved in a more elegant way:

```rs
        // Definitions with empty ranges (e.g. imports) are naturally excluded
        // here since they can't contain the offset
        if let Some((def_id, _def)) = index.definitions(scope_id).contains(offset) {
            return Some(Identifier::Definition { scope_id, def_id });
        }
```

PR #1159, opened 2026-04-17
Look up all of `scope_id`, `def`, and `use` exactly once

DavisVaughan:
Merges into https://github.com/posit-dev/ark/pull/1153


lionel-:
The alternative is to take `scope_id` in the def-at-offset / use-at-offset routines, but I like the use of ty's `Ranged` trait approach here. I imagine we'll scale it up to other types. Cherry-picked into parent, thanks.

DavisVaughan:
> but I like the use of ty's Ranged trait

Ha, that was a Davis idea not a ty idea but I'm glad we ended up in the same place 😛

Issue #1145, opened 2026-04-10
Improvements to internal ownership semantics


lionel-:
- Get rid of `Console::get_mut()`. Use interior mutability on specific fields. Maybe with `ArkRefCell` that's only panicking in debug builds.

- Some of the channels consumed by the readconsole event loop shouldn't be clonable. ReadConsole is reentrant, and cloning a receiver means only one of the clones gets a message. We currently clone them, but luckily the reentrant path shuts down the event loop beforehand. We should make an unclonable wrapper to enforce the property. https://github.com/posit-dev/ark/blob/74efb5294878454ff3d618debcd4fd528cd6f0b7/crates/ark/src/console/console_repl.rs#L963-L969

Not urgent (Ark is not exploding left and right) but would be nice to converge towards a safer setup.

DavisVaughan:
Worth noting that we should not focus on these at all right now

```rust
 // Cloning is necessary to avoid a double mutable borrow error 
 let r_request_rx = self.r_request_rx.clone(); 
 let stdin_reply_rx = self.stdin_reply_rx.clone(); 
 let kernel_request_rx = self.kernel_request_rx.clone(); 
 let tasks_interrupt_rx = self.tasks_interrupt_rx.clone(); 
 let tasks_idle_rx = self.tasks_idle_rx.clone(); 
 let tasks_idle_any_rx = self.tasks_idle_any_rx.clone(); 
```

If we transition `Console` fields that require mutable access to `Cell` / `DebugRefCell` and all `Console` methods to `&self` rather than `&mut self`, then the double borrow goes away and we can just remove these clones too.

---

Also, here's the suggestion by Claude about what the final `CONSOLE` could look like after we no longer need container level mutability, which would be pretty nice

```rust
  thread_local! {
      /// The `Console` singleton.
      ///
      /// Leaked to obtain a genuine `'static` reference. The `Console` lives
      /// for the entire process on the R thread, so leaking is a non-issue.
      /// Only ever accessed from the R thread; `None` before initialization.
      static CONSOLE: Cell<Option<&'static Console>> = const { Cell::new(None) };
  }

  //   Init (replaces CONSOLE.set(UnsafeCell::new(...)) at both sites):
  CONSOLE.set(Some(Box::leak(Box::new(Console::new(...)))));

  //  Access (replaces both get() and get_mut()):
  pub fn get() -> &'static Self {
      CONSOLE.with(|console| {
          console
              .get()
[...]

PR #1161, opened 2026-04-22
Review `Console` reference management

lionel-:
Branched from https://github.com/posit-dev/ark/pull/1100
Progress towards https://github.com/posit-dev/ark/issues/1145

- New `Console::with()` that passes a Console ref and catches panics, transforming them to `Result`.

- `Console::get()` pushed to the FFI boundary and abstracted with `[ark::register]`. This new macro wraps `[harp::register]` and calls `Console::with()`. Any panics due to e.g. RefCell issues gets converted to an R error.

- The graphics callbacks are migrated to `[ark::register]`, but all Ark FFI entries should use it in the future.

- `DeviceContext` no longer reaches back into `Console`. It stores all needed state on itself.

- `PlotComm` also owns what it needs. I was wrong to pass `&Console` to all handlers. `PlotComm` is the only `CommHandler` that used `&Console`, just to get the `device_context()`. It now holds an `Rc<DeviceContext>` passed at construction time.

- All methods in `console_comm.rs` now take `&self`. Supported by a take/remove pattern. Also the comms are now stored in a `DebugRefCell`, a new type that maps to `RefCell` in debug builds and to `UnsafeCell` (with logging) in release builds. This allows us to transition to more Rust safety without causing panics in release builds, and without having to handle panics in awkward places.


Cargo.toml
```diff
@@ -38,6 +39,7 @@ aether_syntax = { git = "https://github.com/posit-dev/air", package = "air_r_syn
 amalthea = { path = "crates/amalthea" }
 anyhow = "1.0.100"
 ark = { path = "crates/ark" }
+ark_macros = { path = "crates/ark/ark_macros" }
```
DavisVaughan: Image

Can we please avoid the crate-in-a-crate design?

crates/stdext/src/cell.rs
```diff
@@ -0,0 +1,302 @@
...
+pub struct DebugRefCell<T: ?Sized> {
+    #[cfg(debug_assertions)]
+    inner: std::cell::RefCell<T>,
+
+    #[cfg(not(debug_assertions))]
+    inner: UnsafeCell<T>,
+    /// Borrow state: 0 = unused, positive = number of shared borrows,
+    /// -1 = exclusive borrow.
+    #[cfg(not(debug_assertions))]
+    borrow_count: Cell<isize>,
```
DavisVaughan: I feel like a design based on this might be easier to reason about

```rust
enum BorrowState {
    Unused,
    Shared(isize),
    Exclusive
}
```

i think it would be cool that you can really easily `match` on the states and, for example, in `Drop` you could match on `Shared(count) if count == 1 => Unused` to show the state transition to unused

crates/stdext/src/cell.rs
```diff
@@ -0,0 +1,302 @@
...
+            }
+        }
+        #[cfg(not(debug_assertions))]
+        {
+            let count = self.borrow_count.get();
+            let tracked = if count < 0 {
+                log::error!(
+                    "INTERNAL ERROR (DebugRefCell): immutable borrow while mutably borrowed (at {})",
+                    std::panic::Location::caller(),
+                );
```
DavisVaughan: like it would make this immediately self documenting, because you'd check `if matches(state, Exclusive)`

right now i have to hold a lot of context about `count` in my head to understand this

crates/amalthea/src/language/shell_handler.rs
```diff
@@ -81,26 +81,30 @@ pub trait ShellHandler: Send {
...
-        &self,
+    /// Returns `(true, None)` if the comm was opened synchronously.
+    /// Returns `(false, None)` if the comm was not handled.
+    fn handle_comm_open(
+        &mut self,
         target: Comm,
         comm: CommSocket,
         data: serde_json::Value,
-    ) -> crate::Result<bool>;
+    ) -> crate::Result<(bool, Option<Receiver<()>>)>;
```
DavisVaughan, line 97: what kind of comm_event_rx comm events do you need to drain while opening a comm? i guess i dont understand this use case

crates/amalthea/src/socket/shell.rs
```diff
@@ -429,34 +401,99 @@ impl Shell {
...
+                &req.content,
+            )
+        })
+    }
+
+    /// Wrap a comm handler in busy/idle status and drain comm events while
+    /// the handler runs. The handler returns a result and an optional
+    /// completion receiver; if present, Shell select-loops on it to process
+    /// comm events (e.g. barriers from `comm_open_backend`).
+    fn handle_comm_notification<T: ProtocolMessage>(
```
DavisVaughan: `handle_comm_request`? its used for 3 requests above

crates/amalthea/src/language/shell_handler.rs
DavisVaughan: yea after more reading i just dont get the use case for drain_comm_events_until(). can you explain it to me with an example?

crates/ark/src/comm_handler.rs
```diff
@@ -198,36 +192,32 @@ pub fn handle_rpc_request<Reqs, Reps>(
...
+    match serde_json::from_value::<Reqs>(data) {
         Ok(m) => match handler(m) {
             Ok(reply) => match serde_json::to_value(reply) {
                 Ok(value) => value,
                 Err(err) => {
-                    let message = format!(
-                            "Failed to serialise reply for {comm_name} request: {err} (request: {data})"
-                        );
+                    let message =
+                        format!("Failed to serialise reply for {comm_name} request: {err}");
```
DavisVaughan, line 208: that seems like it was useful!

crates/ark/src/lib.rs
```diff
@@ -5,6 +5,10 @@
 //
 //
 
+extern crate self as ark;
+
+pub use ark_macros::register;
```
DavisVaughan, line 10: it seems like one of the reasons to put it in another crate is so that it doesnt get recompiled every time something in ark changes, because proc macros are slow (at least that is what i remember)

i still think it should just go in a top level ark_macros/ crate that this imports, if possible

DavisVaughan: although now i do see the prior art with harp-macros. hmm. i guess my preference would have been to make that a top level crate too

crates/ark/src/main.rs
```diff
@@ -392,31 +393,32 @@ fn main() -> anyhow::Result<()> {
...
-            }
-        };
+        let msg: String;
+        if let Some(s) = info.downcast_ref::<&str>() {
+            msg = s.to_string();
+        } else if let Some(s) = info.downcast_ref::<String>() {
+            msg = s.clone();
+        } else {
+            msg = String::from("No contextual information.");
+        }
```
DavisVaughan, line 403: is this mostly just a simplification?

crates/ark/tests/kernel-hooks-session.rs
```diff
@@ -37,7 +37,7 @@ fn test_session_init_hook_new() {
         &frontend,
         &comm_id,
         &format!(
-            "setHook('positron.session_init', function(start_type) rstudioapi::navigateToFile('{path}'))"
+            "setHook('positron.session_init', function(start_type) .ps.ui.navigateToFile('{path}'))"
```
DavisVaughan, line 40: are you sure you want to change this? this is Jenny's test and tests a very particular use case

crates/ark/src/console/console_integration.rs
```diff
@@ -226,14 +231,18 @@ impl Console {
...
 pub(crate) struct UiCommRef<'a> {
-    comm: &'a ConsoleComm,
+    guard: DebugRef<'a, Option<ConsoleComm>>,
     originator: Option<&'a Originator>,
     stdin_request_tx: &'a Sender<StdInRequest>,
 }
 
 impl UiCommRef<'_> {
+    fn comm(&self) -> &ConsoleComm {
+        self.guard.as_ref().unwrap()
```
DavisVaughan: The `unwrap()` feels a bit unsafe

Is there no way to turn `guard` into `DebugRef<ConsoleComm>`?

You did a `is_none()` check up above before constructing a `UiCommRef`, but that is so disconnected from this `comm()` function, it's hard to connect them without any context

DavisVaughan (commented):
temporary state because github is being very flaky with my review comments

crates/ark/src/comm_handler.rs
lionel-: We no longer own the data. No big deal, raw messages are logged elsewhere

crates/ark/src/lib.rs
lionel-: yes I'll move harp too (with underscore)

crates/ark/src/main.rs
lionel-: Nope, this bit is new:

```rs
        // `Console::with()` catches panics with `catch_unwind` in release
        // builds. Return early so the catch handler can convert the panic
        // to an `anyhow::Error`. The backtrace is already logged above.
        if catching_panics() {
            return;
        }
```

crates/ark/tests/kernel-hooks-session.rs
lionel-: IIRC I didn't have rstudioapi installed and it was failing. It's a bit more integrated going through RStudio, but the core test coverage is identical I think.

crates/ark/src/console/console_integration.rs
lionel-: I've added `map()` and `filter_map()` methods, just like those for `cell::Ref`, that make it safer to deal with optional fields.

crates/ark/src/main.rs
DavisVaughan: yea i saw that but everything else was the same right?

crates/amalthea/src/language/shell_handler.rs
lionel-: Added this comment:

```
    /// Used by execute requests and comm handlers (msg/close/open) to prevent
    /// deadlocks when the R thread calls `comm_open_backend` during handling.
    /// `comm_open_backend` sends `CommEvent::Opened` with a barrier back to
    /// Shell and blocks until Shell processes it (to guarantee `comm_open` is
    /// on IOPub before the caller continues). If Shell were blocked on `rx`
    /// without draining, both sides would deadlock.
```

crates/ark/src/console/console_repl.rs
```diff
@@ -673,6 +689,37 @@ impl Console {
...
+    /// where a panic would unwind through R's C call stack. Panics are
+    /// caught and converted to `anyhow::Error`, which `harp::register`'s
+    /// `r_unwrap()` then surfaces as a clean R error.
+    pub fn with<T>(f: impl FnOnce(&Console) -> anyhow::Result<T>) -> anyhow::Result<T> {
+        if cfg!(debug_assertions) {
+            return f(Console::get());
+        }
+
+        CATCHING_PANICS.set(true);
+        let result = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| f(Console::get())));
```
DavisVaughan, line 712: Rather than introducing yet another global, how would you feel about an approach that:

- does `old = panic::take_hook()`
- does `panic::set_hook()` to a tweaked version of our handler that logs but doesn't explode
- calls `f()`
- reinstates old
- handles result

i like that it makes what we are trying to do here very clear and avoids the global

crates/ark/src/plots/graphics_device.rs
```diff
@@ -277,14 +294,13 @@ impl DeviceContext {
...
+    /// Should plot events be sent over comm channels to the frontend?
     ///
     /// This allows plots to be dynamically resized by their `id`. Only possible if the UI
     /// comm is connected (i.e. we are connected to Positron) and if we are in
     /// [SessionMode::Console] mode.
-    fn should_use_dynamic_plots(&self) -> bool {
-        let console = Console::get();
-        console.ui_comm().is_some() && console.session_mode() == SessionMode::Console
+    fn should_use_dynamic_plots(&self, console: &Console) -> bool {
+        self.session_mode == SessionMode::Console && console.ui_comm().is_some()
```
DavisVaughan: `DeviceContext` doesn't even need `session_mode` now that you always have `console` accessible!

crates/ark/src/plots/graphics_device.rs
```diff
@@ -368,37 +384,13 @@ impl DeviceContext {
...
-
-        ExecutionContext {
-            execution_id,
-            code,
-            code_location: None,
-            render_settings: None,
-            intrinsic_size: None,
-        }
+        // No execution context was pushed. This can happen for plots created
+        // outside of an execute request (e.g. during startup).
```
DavisVaughan: i think this comment is wrong

crates/ark/src/plots/graphics_device.rs
```diff
@@ -368,37 +384,13 @@ impl DeviceContext {
...
+        self.execution_context.borrow().clone().unwrap_or_default()
```
DavisVaughan, line 377: There was a fallback path of `// Fall back to getting context from Console (for edge cases).` that checked Console directly. I do think we should just rely on what `graphics_on_execute_request()` should have given us, so I am in favor of removing that. Is that also what you were thinking?

crates/ark/src/plots/graphics_device.rs
```diff
@@ -677,9 +669,10 @@ impl DeviceContext {
         let plot_comm = PlotComm {
             id: id.clone(),
             open_data,
+            device_context: console.device_context_rc(),
```
DavisVaughan: From `DeviceContext`, you have access to `console` which allows you to give someone else access to yourself? That sounds weirdly unsafe 😬

DavisVaughan: I guess thats the tradeoff to not having `Console` in `handle_msg()` and friends

crates/ark/src/plots/graphics_device.rs
```diff
@@ -1092,58 +1090,8 @@ pub(crate) fn compute_plot_overrides(
...
-    log::trace!("Entering on_did_execute_request");
-    let dc = Console::get().device_context();
-    dc.process_changes();
-    dc.clear_execution_context();
-    dc.clear_pending_origin();
-}
-
-/// Activation callback
+/// Run a closure with `&Console` and `&DeviceContext`, catching any panic at
+/// the FFI boundary. Graphics device callbacks are invoked from R
```
DavisVaughan: out of place comment?

DavisVaughan (approved):

crates/ark/src/console/console_repl.rs
lionel-: The big advantage of the flag is that it's thread-local, while the panic hook is fully global. So with the `take_hook()` approach, other threads might be silently panicking instead of aborting if they happen to panic while `Console::with()` is running.

crates/ark/src/plots/graphics_device.rs
lionel-: ohh nice!

crates/ark/src/plots/graphics_device.rs
lionel-: yep, `graphics_on_execute_request()` is now the only source of context. Outside an execute request you get the default, same as the old fallback path returned.

crates/ark/src/plots/graphics_device.rs
lionel-: I switched the relevant methods to `self: &Rc<Self>`, with entry points doing an explicit `Rc::clone(console.device_context())`. The device no longer reaches through `Console` for an `Rc` to itself, it can clone itself.

DavisVaughan: neat!

Issue #1162, opened 2026-04-22
Automate dependency bumps


lionel-:
We're currently upgrading Rust dependencies only when needed, which is not the norm in the Rust community. For instance both ruff/ty and biome use renovate to automatically update dependencies weekly:

- https://github.com/biomejs/biome/blob/main/.github/renovate.json5
- https://github.com/astral-sh/ruff/blob/main/.github/renovate.json5

We've accumulated security alerts and delaying version updates causes a lot of maintenance work to update to new APIs down the road. I think we should follow community practice and enable automatic PRs to update dependencies.

There are two popular options, dependabot and renovate. The latter is more configurable and seems to be the community choice.

- `schedule`: Determines periodicity. Can conveniently set the day of the week where the PR is sent. I suggest every Thursday, this way we always get the PR at least a day after release cut.

- `rangeStrategy`: Determines whether/how to bump `cargo.toml`. Ruff uses `update-lockfile` which never updates the toml file. Biome uses `bump` which does both.

  One issue with proactive bumping is that for crates meant to be used by other projects, this forces all consumers to bump as well. We don't really have any public crate as of now, but many of our crates have at least the aspiration to be used by other projects (e.g. amalthea, libr). Also it would be nice to use the same updating strategy everywhere, including in posit-dev/aether which will hold our public crate libraries.

  So I propose that we use the `replace` strategy: only bumps toml for major changes, otherwise lets the minor and patch versions drift. That's a middle ground and it's not perfect. When we have an actual dependency on a minor/patch version (e.g. new API in that version), that should be reflected in the toml file. But we can deal with these manually.

- `minimumReleaseAge`: Sets a "cooldown" waiting time before updating to avoid supply-chain attacks. Ruff sets it to 7 days.

- `automerge`: Should be false, otherwise PR is merged automatically when CI passes. One of us should take a look before merging.

Before setting this up though, we need to update to a recent baseline with `cargo update` / `cargo upgrade`.

PR #1163, opened 2026-04-22
Start updating dependencies

lionel-:
Progress towards #1162 and fixing security alerts.

- `cargo update`: bumps versions in the lockfile
- `cargo upgrade`: bumps manifest versions within current semranges
- Then a few incompatible bumps (`dashmap 6`, `embed-resource 3`, `env_logger 0.11`,`itertools 0.14`, `libloading 0.9`, `strum 0.28`, `winsafe 0.0.27`,`xdg 3`)
- Switched to rust-yaml2 as rust-yaml is unmaintained
- More updates (base64 0.21 to 0.22, dirs 4 to 6, http 0.2 removed, rand 0.8 to 0.10,


crates/ark/build.rs
```diff
@@ -65,7 +65,9 @@ fn main() {
     let resource = Path::new("resources")
         .join("manifest")
         .join("ark-manifest.rc");
-    embed_resource::compile_for_everything(resource, embed_resource::NONE);
+    embed_resource::compile_for_everything(resource, embed_resource::NONE)
+        .manifest_optional()
+        .unwrap();
```
DavisVaughan, line 70: That seems wrong. `manifest_required` seems right. Read about the difference here https://crates.io/crates/embed-resource

crates/harp/build.rs
```diff
@@ -23,5 +23,7 @@ fn main() {
     let resource = Path::new("resources")
         .join("manifest")
         .join("harp-manifest.rc");
-    embed_resource::compile_for_everything(resource, embed_resource::NONE);
+    embed_resource::compile_for_everything(resource, embed_resource::NONE)
+        .manifest_optional()
```
DavisVaughan: required

PR #1165, opened 2026-04-22
Return description and author for installed packages

bricestacey:
Extend .ps.rpc.pkg_list to return `description` and `author` alongside the existing fields so Positron's Packages pane card view can render them. Both are normalized for display: descriptions collapse whitespace so multi-line text fits on one line, and author strips trailing `<email>` markers from Maintainer values.

All three methods (pak / base / renv) populate the new fields. base and renv share a helper that wraps utils::installed.packages() so the lib lookup only differs by `lib.loc`.

See posit-dev/positron#12925

### DIFF
--- a/crates/ark/src/modules/positron/packages_pane.R
+++ b/crates/ark/src/modules/positron/packages_pane.R
@@ -8,6 +8,49 @@
 # This file contains RPC functions for the packages pane.
 # These functions are called via callMethod from the Positron R extension.
 
+# Normalize a DESCRIPTION field value for display: collapse whitespace/newlines
+# into single spaces so a multi-line Description fits on one line in the UI.
+.ps.pkg_description_text <- function(value) {
+    if (is.null(value) || is.na(value)) {
+        return("")
+    }
+    value <- as.character(value)
+    value <- gsub("\\s+", " ", value, perl = TRUE)
+    trimws(value)
+}
+
+# Normalize a Maintainer/Author value for display: collapse whitespace and strip
+# trailing email addresses in angle brackets (e.g., "Hadley Wickham <h@posit.co>"
+# becomes "Hadley Wickham").
+.ps.pkg_author_text <- function(value) {
+    if (is.null(value) || is.na(value)) {
+        return("")
+    }
+    value <- as.character(value)
+    value <- gsub("\\s+", " ", value, perl = TRUE)
+    value <- gsub("\\s*<[^>]+>", "", value, perl = TRUE)
+    trimws(value)
+}
+
+# Build the package list via utils::installed.packages(), optionally scoped to a
+# specific library path (used by the renv method).
+.ps.pkg_list_installed <- function(lib.loc = NULL) {
+    ip <- utils::installed.packages(
+        lib.loc = lib.loc,
+        fields = c("Description", "Maintainer")
+    )
+    lapply(seq_len(nrow(ip)), function(i) {
+        list(
+            id = paste0(ip[i, "Package"], "-", ip[i, "Version"]),
+            name = ip[i, "Package"],
+            displayName = ip[i, "Package"],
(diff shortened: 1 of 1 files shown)


jennybc (changes requested):
Some high-level thoughts:

1. It's not immediately obvious to me why a pak variant is needed when listing installed packages and perhaps enriching with some information from DESCRIPTION. My instinct is that someone choosing or accepting a pak lifestyle re: packages pane operations is really (and probably only?) about *installation*. I suspect this can be simplified to using something like `.ps.pkg_list_installed()` for pak/base/renv and just feeding the `lib.loc` in the renv case.
2. I'd lean more into R's native vectorization. I think there's a lot of explicit iteration that can be eliminated. I gather some sort of row-by-row move is needed eventually to get this into the needed shape, but I think we can do that all at once at the last moment. Something along these lines:

```r
.ps.pkg_list_installed <- function(lib.loc = NULL) {
    ip <- utils::installed.packages(
        lib.loc = lib.loc,
        fields = c("Description", "Maintainer")
    )

    pkgs <- ip[, "Package"]
    vers <- ip[, "Version"]
    description <- trimws(gsub(
        "\\s+",
        " ",
        ifelse(is.na(ip[, "Description"]), "", ip[, "Description"]),
        perl = TRUE
    ))
    maintainer <- trimws(gsub(
        "\\s*<[^>]+>",
        "",
        gsub(
            "\\s+",
            " ",
            ifelse(is.na(ip[, "Maintainer"]), "", ip[, "Maintainer"]),
            perl = TRUE
        ),
        perl = TRUE
    ))
    ids <- paste0(pkgs, "-", vers)

    Map(
        function(id, pkg, ver, description, maintainer) {
            list(
                id = id,
                name = pkg,
                displayName = pkg,
                version = ver,
                description = description,
                maintainer = maintainer
            )
        },
        ids,
        pkgs,
        vers,
        description,
        maintainer
    )
}

```

jennybc:
Here's a slightly cleaner version of my suggestion:

```r
.ps.pkg_list_installed <- function(lib.loc = NULL) {
    ip <- utils::installed.packages(
        lib.loc = lib.loc,
        fields = c("Description", "Maintainer")
    )

    name <- ip[, "Package"]
    version <- ip[, "Version"]
    id <- paste0(name, "-", version)
    description <- trimws(gsub(
        "\\s+",
        " ",
        ifelse(is.na(ip[, "Description"]), "", ip[, "Description"]),
        perl = TRUE
    ))
    maintainer <- trimws(gsub(
        "\\s*<[^>]+>",
        "",
        gsub(
            "\\s+",
            " ",
            ifelse(is.na(ip[, "Maintainer"]), "", ip[, "Maintainer"]),
            perl = TRUE
        ),
        perl = TRUE
    ))

    Map(
        list,
        id = id,
        name = name,
        displayName = name,
        version = version,
        description = description,
        maintainer = maintainer
    )
}
```

The `Map()` could be even cleaner if `name` and `displayName` did not derive from the same intermediate object, i.e. we wouldn't need `this = that` stuff at all. Are `name` and `displayName` ever different? And if so, would that ever happen here in ark or might that be a Positron/typescript matter?

bricestacey:
> The Map() could be even cleaner if name and displayName did not derive from the same intermediate object, i.e. we wouldn't need this = that stuff at all. Are name and displayName ever different? And if so, would that ever happen here in ark or might that be a Positron/typescript matter?

We have `name` and `displayName` for the sake of consistency with the API. The implementation is slightly different for python, where a name can be distinct from its normalized name..

https://packaging.pypa.io/en/stable/utils.html#packaging.utils.canonicalize_name

bricestacey:
I am closing this PR for now as it is not ready. Can reopen or create a new PR for adding these additional properties.

PR #1166, opened 2026-04-23
Add `PackageCache` support for base packages

DavisVaughan:
The base packages aren't in the typical location, i.e. not in https://cran.r-project.org/src/contrib/.

Instead you must download the _whole R tarball_ located at https://cran.r-project.org/src/base/R-4/, dig into it at `src/library/{base_package}/R` and copy over the `R/` sources from there.

Because the R tarball is a whopping ~40mb, if we detect that we need to do this for any base package, then we go ahead and cache all ~14 base packages.

---

As a side note, Recommended packages like MASS are in the standard download and `Archive/` locations, so we already supported those.

### DIFF
--- a/crates/oak_sources/src/installed_package.rs
+++ b/crates/oak_sources/src/installed_package.rs
@@ -1,0 +1,98 @@
+use std::fs::read_to_string;
+use std::path::Path;
+use std::path::PathBuf;
+
+use oak_package::package_description::Description;
+
+pub(crate) struct InstalledPackage {
+    key: String,
+    name: String,
+    library_path: PathBuf,
+    description: Description,
+    description_hash: String,
+}
+
+impl InstalledPackage {
+    pub(crate) fn find(package: &str, library_paths: &[PathBuf]) -> anyhow::Result<Option<Self>> {
+        let mut library_path = None;
+
+        for library_path_candidate in library_paths {
+            if library_path_candidate.join(package).exists() {
+                library_path = Some(library_path_candidate);
+                break;
+            }
+        }
+
+        let Some(library_path) = library_path else {
+            // Not installed
+            return Ok(None);
+        };
+
+        let package_path = library_path.join(package);
+
+        let description_path = package_path.join("DESCRIPTION");
+        let description_contents = read_to_string(&description_path)?;
+        let description = Description::parse(&description_contents)?;
+
+        let library_path_hash = crate::hash::hash(library_path.to_string_lossy().as_ref());
+        let description_hash = crate::hash::hash(&description_contents);
+
+        // Flat key unique enough to handle:
+        // - The same R package across multiple libpaths
+        // - Reinstalling a dev R package without changing the version (0.1.0.9000)
+        let key = format!(
+            "{name}_{version}_libpath-{library_path_hash}_description-{description_hash}",
+            name = package,
+            version = &description.version,
+            library_path_hash = &library_path_hash,
+            description_hash = &description_hash
+        );
+
+        Ok(Some(Self {
+            key,
+            name: package.to_string(),
+            library_path: library_path.clone(),
+            description,
+            description_hash,
+        }))
+    }
+
+    pub(crate) fn name(&self) -> &str {
+        &self.name
+    }
+
+    pub(crate) fn version(&self) -> &str {
+        &self.description().version
+    }
+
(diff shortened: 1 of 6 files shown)


crates/oak_sources/src/lib.rs
```diff
@@ -229,116 +204,204 @@ impl PackageCache {
...
+            package.description_path(),
+            destination_lock.parent().join("DESCRIPTION"),
+        )?;
+
+        // The `base` package itself has no NAMESPACE, for now we generate an empty
+        // NAMESPACE, but eventually we will want to fully populate it with a
+        // pseudo-NAMESPACE.
+        if package.name() == "base" {
+            std::fs::write(destination_lock.parent().join("NAMESPACE"), "")?;
+            crate::fs::set_readonly(destination_lock.parent().join("NAMESPACE"))?;
```
DavisVaughan, line 303: Good enough for the moment, but we will need to fake it somehow. I couldn't come up with a solid plan for this yet.

crates/oak_sources/src/download.rs
```diff
@@ -0,0 +1,29 @@
...
+) -> anyhow::Result<reqwest::blocking::Response> {
+    if mirrors.is_empty() {
+        panic!("`mirrors` can't be empty.");
+    }
+
+    let mut out = None;
+
+    for mirror in mirrors {
+        let url = format!("{mirror}/{suffix}");
+        let response = reqwest::blocking::get(&url)?;
```
DavisVaughan: The download of the R tarball from `r-project.org` can be somewhat slow, normally taking around 10 seconds for me.

It seems to be faster from the `rstudio.com` mirror (I tried switching the ordering so RStudio's mirror comes first).

Note that if we hit 30 seconds for the download time, then reqwest will cancel the download and return an error here. I have seen that happen exactly one time. We may want to consider that case in more detail. I'm not sure what the right thing to do is here, maybe try to up it to 1 minute? Much longer than that would be a horrible user experience.

crates/oak_sources/src/installed_package.rs
```diff
@@ -0,0 +1,98 @@
...
+
+use oak_package::package_description::Description;
+
+pub(crate) struct InstalledPackage {
+    key: String,
+    name: String,
+    library_path: PathBuf,
+    description: Description,
+    description_hash: String,
+}
```
DavisVaughan, line 13: Some details on how to create this struct have been extracted out because for the base package download I have to create one of these for each of the base packages.

crates/oak_sources/src/download.rs
DavisVaughan: Update: We have a 10 second timeout on connection and a 40 second timeout for the download. If that times out we try the other CRAN mirror (which can be much faster, as mentioned above)

crates/oak_sources/src/lib.rs
```diff
@@ -230,116 +204,202 @@ impl PackageCache {
...
+        // Download the R sources in their entirety
+        let Some(bytes) = crate::base::download(version)? else {
+            log::trace!("No R source tarball on CRAN for version {version}");
+            return Ok(false);
+        };
+
+        // Populate all base packages from the download
+        for package in crate::base::BASE_PACKAGES {
+            let Some(package) = InstalledPackage::find(package, &self.r_libpaths)? else {
+                // It would be very odd to not find a base package
```
lionel-, line 258: Probably worth a warning?

crates/oak_sources/src/lib.rs
lionel-: Yep if we use semantic indexing (looking at top-level definitions) to populate it, it's a bit round-about.

Maybe we should mark the directory as "prepped" but not "finalised", and have the caller use oak_index to finalise it?

lionel- (approved):

crates/oak_sources/src/lib.rs
DavisVaughan: Hmm, it's literally going to only be for base, and I'm also not totally convinced we should be caching DESCRIPTION and NAMESPACE in here at all - i am starting to think we should use the installed package location as our source of truth for those

In which case the cache would only hold `R/` (makes sense to me) and then `Library::get("base")` is where the "smarts" would be to specially construct a `NAMESPACE` for `"base"` in particular

PR #1169, opened 2026-04-24
Prefer explicitly specified dimensions

jmcphers:
Currently, the `ark.plot.width` / `ark.plot.height` options are fallbacks and only consulted when we do not have viewport info from render settings. This makes it difficult to use these options in Positron notebooks since the notebook dimensions take precedence.

This change simply flips the order of precedence so that if the options are set, they are preferred over the more implicit render settings. 

Part of https://github.com/posit-dev/positron/issues/9378.

### DIFF
--- a/crates/ark/src/plots/graphics_device.rs
+++ b/crates/ark/src/plots/graphics_device.rs
@@ -963,7 +963,7 @@ impl DeviceContext {
         id: &PlotId,
         ctx: &ExecutionContext,
     ) -> Result<serde_json::Value, anyhow::Error> {
-        let base = ctx.render_settings.unwrap_or_else(|| PlotRenderSettings {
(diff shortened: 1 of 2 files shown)


DavisVaughan (approved):

DavisVaughan:
Except need a clippy fix

```
error: unnecessary closure used to substitute value for `Option::None`
   --> crates/ark/src/plots/graphics_device.rs:966:20
    |
966 |           let base = ctx.render_settings.unwrap_or_else(|| PlotRenderSettings {
    |  ____________________^
967 | |             size: PlotSize {
968 | |                 width: 800,
969 | |                 height: 600,
...   |
972 | |             format: PlotRenderFormat::Png,
973 | |         });
    | |__________^
```

PR #1170, opened 2026-04-25
Abstract DAP handler

lionel-:
Progress towards https://github.com/posit-dev/positron/issues/12104

We're going to have two handlers for DAP messages, one through the TCP connection for Console debugging, and one through a Jupyter connection (Control socket) for Notebook debugging.

This PR abstracts out DAP handling to make it agnostic to the transport: a `DapHandler` takes a DAP message, handles it, and returns a `Result<DapHandlerOutput>`. `dispatch()` is in charge of transforming error cases to a DAP response (success or error). The caller (transport layer) then propagates via appropriate transport.

Evaluate remains tied to the TCP transport because it needs to run R code asynchronously, and Jupyter's Control socket requires quick synchronous replies.

This PR is pure reorganisation, no functional difference.

### DIFF
--- a/crates/ark/src/dap/dap_server.rs
+++ b/crates/ark/src/dap/dap_server.rs
@@ -80,16 +80,20 @@ pub struct DapOutput {
 }
 
 impl DapOutput {
-    pub fn new(response: Response) -> Self {
+    pub(crate) fn success(req: Request, output: DapHandlerOutput) -> Self {
         Self {
-            response,
-            dap_events: vec![],
-            console_events: vec![],
+            response: req.success(output.body),
+            dap_events: output.dap_events,
+            console_events: output.console_events,
         }
     }
 
     pub fn error(req: Request, message: &str) -> Self {
-        Self::new(req.error(message))
+        Self {
+            response: req.error(message),
+            dap_events: vec![],
+            console_events: vec![],
+        }
     }
 }
 
@@ -100,16 +104,6 @@ pub(crate) struct DapHandlerOutput {
     pub body: ResponseBody,
     pub dap_events: Vec<Event>,
     pub console_events: Vec<DapConsoleEvent>,
-}
-
-impl DapHandlerOutput {
-    fn new(body: ResponseBody) -> Self {
-        Self {
-            body,
-            dap_events: vec![],
-            console_events: vec![],
-        }
-    }
 }
 
 /// Transport-agnostic handler for DAP requests. Translates each request
@@ -168,11 +162,7 @@ impl DapHandler {
         };
 
         match result {
-            Ok(output) => DapOutput {
-                response: req.success(output.body),
-                dap_events: output.dap_events,
-                console_events: output.console_events,
-            },
+            Ok(output) => DapOutput::success(req, output),
             Err(err) => DapOutput::error(req, &format!("{err}")),
         }
     }
@@ -250,11 +240,13 @@ impl DapHandler {
             Ok(uri) => uri,
             Err(err) => {
                 log::warn!("Can't set breakpoints for non-file path: '{path}': {err}");
-                return Ok(DapHandlerOutput::new(ResponseBody::SetBreakpoints(
-                    SetBreakpointsResponse {
+                return Ok(DapHandlerOutput {
+                    body: ResponseBody::SetBreakpoints(SetBreakpointsResponse {
                         breakpoints: vec![],
-                    },
-                )));
+                    }),
+                    dap_events: vec![],
+                    console_events: vec![],
+                });
             },
         };
 
@@ -280,9 +272,11 @@ impl DapHandler {
                     })
                     .collect();
 
-                return Ok(DapHandlerOutput::new(ResponseBody::SetBreakpoints(
-                    SetBreakpointsResponse { breakpoints },
-                )));
+                return Ok(DapHandlerOutput {
+                    body: ResponseBody::SetBreakpoints(SetBreakpointsResponse { breakpoints }),
+                    dap_events: vec![],
+                    console_events: vec![],
+                });
             },
         };
 
@@ -432,11 +426,13 @@ impl DapHandler {
 
         drop(state);
 
-        Ok(DapHandlerOutput::new(ResponseBody::SetBreakpoints(
-            SetBreakpointsResponse {
+        Ok(DapHandlerOutput {
+            body: ResponseBody::SetBreakpoints(SetBreakpointsResponse {
                 breakpoints: response_breakpoints,
-            },
-        )))
+            }),
+            dap_events: vec![],
+            console_events: vec![],
+        })
     }
 
     fn handle_attach(&self, _args: AttachRequestArguments) -> anyhow::Result<DapHandlerOutput> {
@@ -477,14 +473,16 @@ impl DapHandler {
     // All servers must respond to `Threads` requests, possibly with
     // a dummy thread as is the case here
     fn handle_threads(&self) -> anyhow::Result<DapHandlerOutput> {
-        Ok(DapHandlerOutput::new(ResponseBody::Threads(
-            ThreadsResponse {
+        Ok(DapHandlerOutput {
+            body: ResponseBody::Threads(ThreadsResponse {
                 threads: vec![Thread {
                     id: THREAD_ID,
                     name: String::from("R console"),
                 }],
-            },
-        )))
+            }),
+            dap_events: vec![],
+            console_events: vec![],
+        })
     }
 
     fn handle_set_exception_breakpoints(
@@ -495,14 +493,16 @@ impl DapHandler {
             let mut state = self.state.lock().unwrap();
             state.exception_breakpoint_filters = args.filters;
         }
-        Ok(DapHandlerOutput::new(
-            ResponseBody::SetExceptionBreakpoints(SetExceptionBreakpointsResponse {
+        Ok(DapHandlerOutput {
+            body: ResponseBody::SetExceptionBreakpoints(SetExceptionBreakpointsResponse {
                 // This field is only useful for reporting problems with
                 // individual filters. Since we always accept all filters,
                 // `None` is fine here.
                 breakpoints: None,
             }),
-        ))
+            dap_events: vec![],
+            console_events: vec![],
+        })
     }
 
     fn handle_stacktrace(&self, args: StackTraceArguments) -> anyhow::Result<DapHandlerOutput> {
@@ -543,12 +543,14 @@ impl DapHandler {
         };
         let stack = stack[start..end].to_vec();
 
-        Ok(DapHandlerOutput::new(ResponseBody::StackTrace(
-            StackTraceResponse {
+        Ok(DapHandlerOutput {
+            body: ResponseBody::StackTrace(StackTraceResponse {
                 stack_frames: stack,
                 total_frames: Some(total_frames),
-            },
-        )))
+            }),
+            dap_events: vec![],
+            console_events: vec![],
+        })
     }
 
     fn handle_source(&self, _args: SourceArguments) -> anyhow::Result<DapHandlerOutput> {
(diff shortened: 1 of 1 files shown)


crates/ark/src/dap/dap_server.rs
```diff
@@ -58,325 +58,127 @@ const THREAD_ID: i64 = -1;
...
+            dap_events: vec![],
+            console_events: vec![],
+        }
+    }
 
-            // Terminate the events thread
-            let _ = done_tx.send(true);
-        });
+    pub fn error(req: Request, message: &str) -> Self {
+        Self::new(req.error(message))
```
DavisVaughan: Based on how this is used below, I don't think either of these methods really make sense as is

```
        match result {
            Ok(output) => DapOutput {
                response: req.success(output.body),
                dap_events: output.dap_events,
                console_events: output.console_events,
            },
            Err(err) => DapOutput::error(req, &format!("{err}")),
```

Like I think I'd rather either see:

- `DapOutput::response()` / `success()` + `DapOutput::error()`
- Or both constructed here with `DapOutput {` syntax

Mixing both like this is a bit confusing

crates/ark/src/dap/dap_server.rs
```diff
@@ -58,325 +58,127 @@ const THREAD_ID: i64 = -1;
...
+    pub body: ResponseBody,
+    pub dap_events: Vec<Event>,
+    pub console_events: Vec<DapConsoleEvent>,
+}
 
-            // Break the loop and terminate the thread
-            recv(done_rx) -> _ => { return; },
-        )
+impl DapHandlerOutput {
+    fn new(body: ResponseBody) -> Self {
```
DavisVaughan: Same kind of comment here, I see we have a mix of `new()` and `DapHandlerOutput {` construction below. In this case I feel like I'd just prefer to see the `DapHandlerOutput {` construction everywhere? I like that it shows at the call site that we are explicitly not sending any events along with the body of this output

crates/ark/src/dap/dap_server.rs
```diff
@@ -608,219 +410,587 @@ impl<R: Read, W: Write> DapServer<R, W> {
...
+        if matches!(&req.command, Command::Evaluate(_)) {
+            let Command::Evaluate(args) = req.command.clone() else {
+                unreachable!()
+            };
+            if let Err(err) = self.handle_evaluate(req, args) {
+                log::warn!("DAP: Handler failed, closing connection: {err:?}");
+                return false;
+            }
+            return true;
+        }
```
DavisVaughan:
```suggestion
        if let Command::Evaluate(args) = &req.command {
            let args = args.clone();
            if let Err(err) = self.handle_evaluate(req, args) {
                log::warn!("DAP: Handler failed, closing connection: {err:?}");
                return false;
            }
            return true;
        }
```

Seems cleaner

crates/ark/src/dap/dap_server.rs
```diff
@@ -58,325 +58,127 @@ const THREAD_ID: i64 = -1;
...
-            Command::SetBreakpoints(args) => self.handle_set_breakpoints(req, args),
-            Command::SetExceptionBreakpoints(args) => {
-                self.handle_set_exception_breakpoints(req, args)
-            },
-            Command::StackTrace(args) => self.handle_stacktrace(req, args),
-            Command::Source(args) => self.handle_source(req, args),
-            Command::Scopes(args) => self.handle_scopes(req, args),
-            Command::Variables(args) => self.handle_variables(req, args),
-            Command::Evaluate(args) => self.handle_evaluate(req, args),
+            Command::Initialize(args) => self.handle_initialize(args),
```
DavisVaughan, line 133: I would be extremely tempted to leave `Command::Evaluate(args)` in here with a specific panic of

```
panic!("Expected `Evaluate` to be handled via transport-specific delivery: {args}")
```

Which gets to take the place of your doc comment in a more meaningful way

crates/ark/src/dap/dap_server.rs
```diff
@@ -58,325 +58,127 @@ const THREAD_ID: i64 = -1;
...
+                self.handle_step(args, DebugRequest::StepIn, ResponseBody::StepIn)
             },
             Command::StepOut(args) => {
-                self.handle_step(req, args, DebugRequest::StepOut, ResponseBody::StepOut)
+                self.handle_step(args, DebugRequest::StepOut, ResponseBody::StepOut)
             },
-            Command::Pause(args) => self.handle_pause(req, args),
+            Command::Pause(args) => self.handle_pause(args),
             _ => {
                 log::warn!("DAP: Unknown request");
```
DavisVaughan: it might be nice to log some details about the request?

crates/ark/src/dap/dap_server.rs
```diff
@@ -58,325 +58,127 @@ const THREAD_ID: i64 = -1;
...
-                let rsp = req.error("Ark DAP: Unknown request");
-                self.respond(rsp)
+                return DapOutput::error(req, "Ark DAP: Unknown request");
```
DavisVaughan, line 160: Alternatively just let an `anyhow!()` error pass through here rather than doing an early return so it falls through to the standard `DapOutput::error` machinery below that everything else uses?

crates/ark/src/dap/dap_server.rs
```diff
@@ -608,219 +410,587 @@ impl<R: Read, W: Write> DapServer<R, W> {
...
+        })
+    }
+
+    fn handle_disconnect(&self, _args: DisconnectArguments) -> anyhow::Result<DapHandlerOutput> {
+        // Only send `Q` if currently in a debugging session.
+        let is_debugging = { self.state.lock().unwrap().is_debugging };
+        let console_events = if is_debugging {
+            vec![DapConsoleEvent::DebugCommand(DebugRequest::Quit)]
+        } else {
+            vec![]
```
DavisVaughan, line 455: My main comment is on event ordering like this

Previously:
- `send_command()` fired first
- Then `respond()`

Now we always fire:
- `respond()` first
- Then dap events
- Then console events

And I'm not sure if that matters or not

crates/ark/src/dap/dap_server.rs
```diff
@@ -608,219 +410,587 @@ impl<R: Read, W: Write> DapServer<R, W> {
...
+
+        Ok(DapHandlerOutput {
+            body: ResponseBody::Disconnect,
+            dap_events: vec![],
+            console_events,
+        })
+    }
+
+    fn handle_restart<T>(&self, _args: T) -> anyhow::Result<DapHandlerOutput> {
+        Ok(DapHandlerOutput {
```
DavisVaughan, line 466: We dropped the `// If connected to Positron` message here that felt kind of important. It should end up in the big match below I think

crates/ark/src/dap/dap_server.rs
```diff
@@ -608,219 +410,587 @@ impl<R: Read, W: Write> DapServer<R, W> {
...
+
+            let object = object.get();
+            object_variables(object.sexp)
+        });
+
+        variables
+    }
+
+    fn make_variables(&self, variables: Vec<RVariable>) -> Vec<Variable> {
+        self.state.lock().unwrap().make_variables(variables)
```
DavisVaughan, line 632: After seeing `// Wait until we're in the r_task() to lock` above my alarms immediately went off here

We might want a comment like `// make_variables() doesn't require R, so locking outside of an r_task() is fine here, unlike in collect_r_variables()`

crates/ark/src/dap/dap_server.rs
```diff
@@ -608,219 +410,587 @@ impl<R: Read, W: Write> DapServer<R, W> {
...
+    fn handle_pause(&self, _args: PauseArguments) -> anyhow::Result<DapHandlerOutput> {
+        self.state.lock().unwrap().is_interrupting_for_debugger = true;
+
+        log::info!("DAP: Received request to pause R, sending interrupt");
+
+        Ok(DapHandlerOutput {
+            body: ResponseBody::Pause,
+            dap_events: vec![],
+            console_events: vec![DapConsoleEvent::Interrupt],
+        })
```
DavisVaughan, line 657: Another case where it used to look like

```
    fn handle_pause(&mut self, req: Request, _args: PauseArguments) -> Result<(), ServerError> {
        self.state.lock().unwrap().is_interrupting_for_debugger = true;

        log::info!("DAP: Received request to pause R, sending interrupt");
        crate::sys::control::handle_interrupt_request();

        let rsp = req.success(ResponseBody::Pause);
        self.respond(rsp)
    }
```

i.e. `crate::sys::control::handle_interrupt_request();` fired first, then we respond()

im a little nervous about that? idk how important that is.

crates/ark/src/dap/dap_server.rs
```diff
@@ -608,219 +410,587 @@ impl<R: Read, W: Write> DapServer<R, W> {
...
+    pub fn new(
+        reader: BufReader<R>,
+        writer: BufWriter<W>,
+        state: Arc<Mutex<Dap>>,
+        r_request_tx: Sender<RRequest>,
+        comm_tx: CommOutgoingTx,
+        responses_tx: Sender<Response>,
+    ) -> Self {
+        let server = Server::new(reader, writer);
+        let output = server.output.clone();
```
DavisVaughan, line 892: Why is the `output` held separately? 🤔

crates/ark/src/dap/dap_server.rs
```diff
@@ -608,219 +410,587 @@ impl<R: Read, W: Write> DapServer<R, W> {
...
-            stack_frames: stack,
-            total_frames: Some(total_frames),
-        }));
+        for event in output.console_events {
+            self.handle_console_event(event);
+        }
 
-        self.respond(rsp)
+        true
     }
```
DavisVaughan, line 946: So inability to respond() or send_event() is immediate cause for `serve()` to also bail and return `false`, causing the whole DAP to disconnect. I think that feels right.

DavisVaughan: But notably handle_console_event could fail and not bring down the whole DAP right now

crates/ark/src/dap/dap_server.rs
```diff
@@ -608,219 +410,587 @@ impl<R: Read, W: Write> DapServer<R, W> {
...
-            end_line: None,
-            end_column: None,
-        }];
-
-        let rsp = req.success(ResponseBody::Scopes(ScopesResponse { scopes }));
-
-        drop(state);
-        self.respond(rsp)
+    fn respond(&mut self, rsp: Response) -> Result<(), ServerError> {
+        log::trace!("DAP: Responding to request: {rsp:#?}");
```
DavisVaughan:
```suggestion
        log::trace!("DAP: Sending response: {rsp:#?}");
```

I think this makes more sense? It looks like you are about to show the _request_ in the logs, but that's not the case. This is also symmetric with the `send_event()` log message now

DavisVaughan (approved):

crates/ark/src/dap/dap_server.rs
lionel-: Even if we added "fill-in" support to the Jupyter handler right away, I would be very uneasy about panicking here.

I think returning a DAP error makes more sense and is more defensive.

crates/ark/src/dap/dap_server.rs
lionel-: I think it doesn't really matter because the recipients (external client and internal console) are fully independent, and so the delivery is async either way. That said it's safer to preserve the existing ordering, so I've swapped them out.

crates/ark/src/dap/dap_server.rs
lionel-: This is pre-existing. It's a shareable handle that can be handed off separately. E.g. signature:

```rs
pub struct DapServer<R: Read, W: Write> {
    server: Server<R, W>,
    pub output: Arc<Mutex<ServerOutput<W>>>,
```

PR #1171, opened 2026-04-25
Add Jupyter DAP handler

### LINKED ISSUE #572: Emit debug prompts on StdIn in notebook mode
The current experience is confusing:

Image

If these prompts were emitted on StdIn, a debug session would look more like this:

Image

lionel-:
Branched from https://github.com/posit-dev/ark/pull/1170
Backend side of https://github.com/posit-dev/positron/pull/13205
Part of https://github.com/posit-dev/positron/issues/12104
Closes https://github.com/posit-dev/ark/issues/572

Implements notebook debugging via the Jupyter Debug Protocol. DAP messages flow through `debug_request`/`debug_reply` on the Control socket, with events forwarded as `debug_event` on IOPub.

Since notebook cells aren't files on disk, the Jupyter Debug Protocol uses `dumpCell` to write cell source code to deterministic temp files (path derived from a MurmurHash2 of the code). The frontend sends `dumpCell` before `setBreakpoints`, so breakpoints reference these temp files.

When execute requests contain a `cellId`, breakpoints and line directives are mapped to the temp files written by `dumpCell`. This allows users to execute and inject breakpoints at the same time, outside of a debugging session. Otherwise they'd need to start a debugging session with Debug Cell, then go back and reexecute cells containing breakpoints.

In addition:

- Interrupts in notebook mode now quit the debugger with `Q` if active. This fits the notebook UX because, unlike in Console mode, the kernel remains busy for the duration of the debugging session, and the notebook UI shows an Interrupt button next to the cell. If we don't quit the debugger, the cell will remain in busy state after interrupting, which is confusing.

- Debugging prompts created via `browser()` or `debug()` (as opposed to breakpoints) map to StdIn. Not the best UI in Positron (stdin prompts show via command palette) but much better than doing nothing. The alternative would be to initiate a full debug session from the backend, but this would require non-standard frontend changes.

https://github.com/user-attachments/assets/8ebbc213-5601-427e-9248-e63d56021e91


crates/amalthea/src/fixtures/dummy_frontend.rs
```diff
@@ -247,6 +253,56 @@ impl DummyFrontend {
...
+            code: String::from(code),
+            silent: false,
+            store_history: true,
+            user_expressions: serde_json::Value::Null,
+            allow_stdin: options.allow_stdin,
+            stop_on_error: false,
+            positron: options.positron,
+        };
+        let mut message = JupyterMessage::create(content, None, &self.session);
+        message.metadata = metadata;
```
DavisVaughan: Seems like `create_with_metadata()` might be useful

DavisVaughan: And then it is probably cleaner to add

```
fn send_with_metadata<T: ProtocolMessage>(socket: &Socket, session: &Session, msg: T, metadata: Value)
```

to this file alongside the preexisting

```
fn send<T: ProtocolMessage>(socket: &Socket, session: &Session, msg: T)
```

and use that here?

crates/ark/src/shell.rs
```diff
@@ -144,14 +145,20 @@ impl ShellHandler for Shell {
...
+        let mut supported_features = vec![String::from("debugger")];
+        if matches!(kernel_info.session_mode, SessionMode::Notebook) {
+            supported_features.push(String::from("proactive breakpoints"));
+        }
+
         Ok(KernelInfoReply {
             status: Status::Ok,
             banner: kernel_info.banner.clone(),
-            debugger: false,
+            debugger: true,
```
DavisVaughan, line 156: interesting, i guess we just send it anyways

```
    # A boolean flag which tells if the kernel supports debugging in the notebook.
    # Default is False.
    # Deprecated as replaced by 'supported_features'=['debugger'] (see below).
    'debugger': bool,
```

crates/ark/src/dap/dap_state.rs
```diff
@@ -152,10 +162,75 @@ pub struct DapExceptionEvent {
...
+}
+
 pub struct Dap {
     /// Whether the REPL is stopped with a browser prompt.
     pub is_debugging: bool,
 
+    /// Whether R is stopped at an unexpected `browser()` prompt in notebook
+    /// mode (no active debug session). Mutually exclusive with `is_debugging`.
+    /// Used by the interrupt handler to decide whether to send a "Q" command.
+    /// `id_debugging` and `is_stopped_at_browser` could be folded into a single
```
```suggestion
    /// `is_debugging` and `is_debugging_stdin` could be folded into a single
```

is this what you meant?

crates/ark/src/control.rs
```diff
@@ -60,9 +83,29 @@ impl ControlHandler for Control {
...
         log::info!("Received interrupt request");
+
+        // When an interrupt is sent while debugging in notebook mode, we quit
+        // the debugger. The difference is justified by how the Console stays
+        // busy while debugging, showing a spinning wheel to the user. Quitting
+        // debugging on interrupt is natural UX in that context.
+        if matches!(self.session_mode, SessionMode::Notebook) {
+            let dap = self.dap.lock().unwrap();
+            if dap.is_debugging || dap.is_debugging_stdin {
+                drop(dap);
```
DavisVaughan, line 98: is the explicit drop() meaningful?

crates/ark/src/control.rs
```diff
@@ -1,39 +1,62 @@
...
+use crate::console::SessionMode;
+use crate::dap::dap_jupyter_handler::DapJupyterHandler;
+use crate::dap::Dap;
 use crate::request::RRequest;
 
 pub struct Control {
     r_request_tx: Sender<RRequest>,
+    dap: Arc<Mutex<Dap>>,
+    session_mode: SessionMode,
+    dap_handler: DapJupyterHandler,
```
DavisVaughan: Since all of these are only used for notebooks, I was kind of confused that they are here at all for console mode.

It might be nice to wrap these up in an optional struct, like

```
struct JupyterDap {
    dap: Arc<Mutex<Dap>>,
    dap_handler: DapJupyterHandler,
}
```

which is then `Option<JupyterDap>` on `Control`

and then you can use the precence of that field rather than `matches!(self.session_mode, SessionMode::Notebook)` to make decisions in other parts of the codebase

crates/amalthea/src/socket/control.rs
```diff
@@ -169,11 +179,8 @@ impl Control {
             error!("Failed to send interrupt request: {:?}", err);
         }
 
-        // Lock the control handler object on this thread
-        let control_handler = self.handler.lock().unwrap();
-
         let reply = unwrap!(
-            block_on(control_handler.handle_interrupt_request()),
+            self.handler.handle_interrupt_request(),
```
DavisVaughan, line 183: What happened to allow us to remove all of this and the `async` ness of the methods? Have we never needed it?

crates/amalthea/src/socket/control.rs
```diff
@@ -156,6 +154,18 @@ impl Control {
...
+    fn handle_debug_request(&self, req: JupyterMessage<DebugRequest>) -> Result<(), Error> {
+        log::trace!("Received debug request: {:?}", req);
+
+        let Some(reply) = self.handler.handle_debug_request(&req.content).log_err() else {
+            return Ok(());
+        };
+
+        req.send_reply(reply, &self.socket).log_err();
+
+        Ok(())
```
DavisVaughan, line 166: It's interesting how we never actually return an `Error` from these methods...

crates/amalthea/src/socket/shell.rs
```diff
@@ -111,6 +111,29 @@ impl Shell {
...
+
+        #[cfg(not(all(target_os = "windows", target_arch = "aarch64")))]
+        self.listen_blocking();
+    }
+
+    /// On Windows ARM, zmq::poll with a non-zero timeout blocks forever
+    /// and inproc notification sockets may not wake the poll. Use
+    /// non-blocking poll with unconditional comm checks and a short sleep.
+    #[cfg(all(target_os = "windows", target_arch = "aarch64"))]
+    fn listen_polling(&mut self) {
```
DavisVaughan, line 125: I would have been ok with just calling both `listen()`

crates/ark/src/console/console_repl.rs
```diff
@@ -868,6 +876,18 @@ impl Console {
...
+            {
+                self.debug_dap.lock().unwrap().is_debugging_stdin = true;
+                let result = self.handle_input_request(&info, buf, buflen);
+                self.debug_dap.lock().unwrap().is_debugging_stdin = false;
+                return result;
+            }
+
             // Similarly, if we have pending inputs, we're about to immediately
             // continue with the next expression. Don't emit debug notifications
             // for these intermediate browser prompts.
```
DavisVaughan: Should you deal with `has_pendinng` below first? If not then a comment about why seems needed

crates/ark/src/console/console_repl.rs
```diff
@@ -868,6 +876,18 @@ impl Console {
                 return result;
             }
 
+            // In notebook mode, `browser()` or `debug()` calls causing a
+            // browser prompt outside of an active debugging session are routed
+            // to stdin so the user can type debug commands in an input box.
+            if self.session_mode == SessionMode::Notebook &&
+                !self.debug_dap.lock().unwrap().is_connected
+            {
+                self.debug_dap.lock().unwrap().is_debugging_stdin = true;
```
DavisVaughan, line 885:
```suggestion
                self.debug_dap.lock().unwrap().is_debugging_outside_session = true;
```

that also kind of makes sense to me?

crates/ark/src/console/console_repl.rs
```diff
@@ -974,12 +1001,17 @@ impl Console {
...
+        // expecting. We either wait for `input_reply` messages on StdIn, or for
+        // `execute_request` on Shell. We also listen for R requests, including
+        // while waiting on StdIn. Such requests are only expected in Notebook
+        // mode when debugging via StdIn. The interrupt handler may send us a
+        // Quit command to terminate the debugging session. For simplicity we
+        // listen for these requests unconditionally, including in Console sessions.
+        let r_request_index = select.recv(&r_request_rx);
+        let stdin_reply_index = match wait_for {
+            WaitFor::ExecuteRequest => None,
+            WaitFor::InputReply => Some(select.recv(&stdin_reply_rx)),
```
DavisVaughan, line 1014: That is so confusing 😢

crates/ark/src/console/console_repl.rs
```diff
@@ -1003,15 +1035,15 @@ impl Console {
...
-            // request prompt, e.g. `readline()`, we need to propagate
-            // the interrupt to the R stack. This needs to happen before
-            // `process_idle_events()`, particularly on Windows, because it
-            // calls `R_ProcessEvents()`, which checks and resets
-            // `UserBreak`, but won't actually fire the interrupt b/c
-            // we have them disabled, so it would end up swallowing the
-            // user interrupt request.
-            if matches!(info.kind, PromptKind::InputRequest) && interrupts_pending() {
+            // If an interrupt was signaled and we are waiting for user
+            // input (readline, or browser-as-stdin in notebook mode), we
```
DavisVaughan, line 1039: Okay. High level thought.

Honestly I feel like the browser-as-stdin UI experience that comes with `browser()` and `debug()` is very unenjoyable looking.

And now I'm looking at the implementation and it feels like it adds A LOT of complexity for something that is pretty gross from a user perspective.

What if we just...didn't allow it at all.

Like, what if the only way to debug in a Notebook was with breakpoints? Disallow `browser()` and `debug()` alltogether, since that UI is pretty bleh and makes things massively complicated in here.

I bet you'd capture the 95% case of people who want to do any debugging in a notebook (which is already a very small population) while avoiding adding all of this complexity to ark, which to me is a huge win.

crates/ark/src/console/console_repl.rs
```diff
@@ -1367,6 +1420,23 @@ impl Console {
...
+                // a `CodeLocation` pointing to the temp file that `dumpCell`
+                // wrote. This gives `annotate_input()` the file URI it needs
+                // for the `#line` directive and breakpoint injection.
+                let (uri_id, loc) = if cell_id.is_some() {
+                    match dap_notebook::notebook_code_location(&code) {
+                        Some(notebook_loc) => {
+                            let id = UrlId::from_url(notebook_loc.uri.clone());
+                            (Some(id), Some(notebook_loc))
+                        },
+                        None => (uri_id, loc),
```
DavisVaughan, line 1434: Should you even try to keep them here?

DavisVaughan: Or maybe this is just passing `None`s through

crates/ark/src/dap/dap_state.rs
```diff
@@ -307,27 +388,50 @@ impl Dap {
         let was_debugging = self.is_debugging;
         self.is_debugging = false;
 
-        if was_debugging && self.is_connected {
+        if was_debugging && (self.is_connected || self.iopub_tx.is_some()) {
```
DavisVaughan: It seems like `self.iopub_tx.is_some()` is signaling some kind of state, like "im in a notebook", which should be some kind of named state, because I don't really know what this means without any context

crates/ark/src/dap/dap_state.rs
```diff
@@ -525,47 +631,48 @@ impl Dap {
...
+                    id: bp.id,
+                    line: bp.line,
+                    verified: true,
+                    message: None,
+                });
             }
         }
+
+        for event in events {
+            self.send_backend_event(event);
```
DavisVaughan, line 646: Wouldn't it be cleaner to just wrap `DapBackendEvent::BreakpointState {` in `self.send_backend_event()`?

DavisVaughan: You avoid the `events` vec alltogether, I think

crates/ark/src/dap/dap_state.rs
```diff
@@ -525,47 +631,48 @@ impl Dap {
...
     /// Finds the breakpoint with the given ID for the URI and marks it as verified
     /// if it was previously unverified. Sends a `BreakpointVerified` event.
     pub fn verify_breakpoint(&mut self, uri: &UrlId, id: &str) {
-        let Some((_, bp_list)) = self.breakpoints.get_mut(uri) else {
-            return;
-        };
-        let Some(bp) = bp_list.iter_mut().find(|bp| bp.id.to_string() == id) else {
-            return;
-        };
+        let event = {
```
DavisVaughan, line 655: This seems like a very unnecessary extra `{` scope?

crates/ark/src/dap/dap_state.rs
```diff
@@ -578,42 +685,40 @@ impl Dap {
...
+                };
+                Some(DapBackendEvent::BreakpointState {
+                    id: bp.id,
+                    line: bp.line,
+                    verified: false,
+                    message: Some(reason.message().to_string()),
+                })
             })
-            .log_err();
+            .collect();
```
DavisVaughan, line 720: could probably avoid the collect() and just do `.into_iter()` on the filter_map result and loop over that?

crates/ark/src/dap/dap_notebook.rs
```diff
@@ -0,0 +1,187 @@
...
+    })
+}
+
+/// MurmurHash2 implementation for computing deterministic temp file paths
+/// from cell source code.
+///
+/// This must match the client-side `PathEncoder` in
+/// `positron-runtime-debugger` so that the notebook location mapper can
+/// correlate cell URIs with runtime source paths.
+pub fn murmur2(data: &[u8], seed: u32) -> u32 {
```
DavisVaughan, line 98: we could also just use someone else's?

crates/ark/src/dap/dap_notebook.rs
```diff
@@ -0,0 +1,187 @@
...
+/// instability. On macOS, R may unset `TMPDIR` during startup, causing
+/// `std::env::temp_dir()` to return `/tmp` instead of the per-session
+/// `/var/folders/.../T/` directory.
+static TMP_FILE_PREFIX: LazyLock<String> = LazyLock::new(|| {
+    let mut tmp_dir = std::env::temp_dir();
+    let pid = std::process::id();
+    tmp_dir.push(format!("ark-debug-{pid}"));
+    // Trailing separator so the prefix can be concatenated directly with
+    // the hash and suffix (e.g. `{prefix}{hash}.r`).
+    format!("{}/", tmp_dir.display())
```
DavisVaughan, line 34: I also like the `tempdir` crate for this kind of thing?

DavisVaughan (approved):
I did not give `dap_notebook.rs` or `dap_jupyter_handler.rs` a too close read, but I did read everything else.

Mostly im not really sure i agree with the tradeoff of all of the added weird complexity that comes with debug-via-stdin in the `browser()` or `debug()` case, vs the straightforward support we have for breakpoints here. I think the UI for the former is not very good, and it might be better to just do nothing in that case and encourage people to use breakpoints in notebooks instead. To me the complexity increase is not worth it for the _miniscule_ amount of people who might use that feature. (I have another comment on this below)

crates/ark/src/console/console_repl.rs
lionel-: It's much more enjoyable in Jupyter lab and other frontends, where you get a prompt nested in a cell.
It's not enjoyable in Positron but at least it now no longer hangs and _is_ usable.

Besides, skipping browser prompts reliably requires some logic that we don't have yet (but would like to implement to prevent dropping in the debugger while an Ark stack is running).

lionel-:
> Mostly im not really sure i agree with the tradeoff of all of the added weird complexity that comes with debug-via-stdin in the browser() or debug() case, vs the straightforward support we have for breakpoints here. 

I think it's worth it because it fixes a longstanding limitation of Ark in Jupyter apps, where browser prompts have to be nested within a given execute request (i.e. a cell). Also I do think it can be useful in Positron too, albeit for advanced cases.

I would also not overstate the added complexity of that stdin support, I'm comfortable with it especially with the test coverage.


https://github.com/user-attachments/assets/d40668e1-3b63-4241-b203-d01163228472

crates/ark/src/console/console_repl.rs
lionel-: Indeed, but I think stdin is clearer

crates/ark/src/shell.rs
lionel-: yep we're sending both

crates/ark/src/control.rs
lionel-: It seems like good practice

crates/ark/src/control.rs
lionel-: I'm intending to get breakpoint sync via the jupyter handler even in Console mode.

For now I made the handler optional, but in the future we'll have a specialised handler for consoles too.

crates/amalthea/src/socket/control.rs
lionel-: We've never needed amalthea handler methods to be async, and should move towards sync over time. I made that clean up for Control as I was in the neighborhood.

crates/amalthea/src/socket/shell.rs
lionel-: Me too, although I like the named calls too.

crates/ark/src/console/console_repl.rs
lionel-: Sorry I'm not sure what comment to write. The notebook path fully diverges here for stated reason (stdin handling). So anything DAP related is not relevant to that path, as encoded by the early exit.

crates/ark/src/console/console_repl.rs
lionel-: Agreed that this is subtle.

crates/ark/src/dap/dap_state.rs
lionel-: Good point, I've renamed it to `notebook_iopub_tx`.

crates/ark/src/dap/dap_state.rs
lionel-: I've added a comment explaining why it's structured this way:

```rs
        // Collect events first: `bp_list` borrows from `self.breakpoints`,
        // which prevents calling `&mut self` methods like `send_backend_event()`.
```

crates/ark/src/dap/dap_state.rs
lionel-: borrow-checker

crates/ark/src/dap/dap_state.rs
lionel-: borrow-checker

crates/ark/src/dap/dap_notebook.rs
lionel-: Here we're creating a directory with a very specific pattern that remains valid for the whole session.

crates/ark/src/dap/dap_notebook.rs
lionel-: We also implement it on the frontend side. We could use the murmur2 crate, or just keep this one. I'm going with the latter since it's tested already.

crates/ark/src/dap/dap_state.rs
DavisVaughan: nope

```
    pub fn verify_breakpoint(&mut self, uri: &UrlId, id: &str) {
        let Some((_, bp_list)) = self.breakpoints.get_mut(uri) else {
            return;
        };
        let Some(bp) = bp_list.iter_mut().find(|bp| bp.id.to_string() == id) else {
            return;
        };

        if !matches!(bp.state, BreakpointState::Unverified) {
            return;
        }

        bp.state = BreakpointState::Verified;

        let event = DapBackendEvent::BreakpointState {
            id: bp.id,
            line: bp.line,
            verified: true,
            message: None,
        };

        self.send_backend_event(event);
    }
```

crates/ark/src/dap/dap_state.rs
DavisVaughan: I'm getting that feeling that `iopub_seq` should be wrapped in a `Cell` or something like that so it can be interiorly mutable allowing `send_backend_event()` to just be `&self`, because it really doesn't feel like you should need `&mut self` to do a `send_backend_event()`, and the counter feels like an internal detail

crates/ark/src/dap/dap_notebook.rs
DavisVaughan: I think you can do all that with the `tempfile` crate (not `tempdir`, sorry)

But regardless, it looks like its default is to just use `std::env::temp_dir` as the starting location to put the tempdir in, so I don't think it really matters

But based on your comments im now interested in the timing - do we need to trigger this lazylock before we start R up?

PR #1172, opened 2026-04-27
Integrate `PackageCache` with goto definition

DavisVaughan:
Branched from https://github.com/posit-dev/ark/pull/1166

Some initial setup changes:
- Threaded `r_home: PathBuf` through to both the LSP and `Console` by calling `r_home_setup()` way earlier
- `oak_index_vec` is now its own crate, spun out from `oak_index`. It just holds `IndexVec`, a useful concept outside of the `SemanticIndex` (this should probably just be `oak_index` after we rename the existing `oak_index` to `oak_semantic`)

---

`Library` now holds a `dyn PackageSources` object for retrieving package sources. It also gains a `definitions() -> Option<PackageDefinitions>` method for returning a package worth of definitions that support goto definition. This is currently an expensive method until we have salsa.

We make `PackageDefinitions` for a package by doing the following for each package file: reading it, parsing it, creating a semantic index, categorizing the `file_exports()`. Then, for now, we just throw away the semantic index for that file.

`goto_definition()` can use them to determine definition locations for external packages.

---

Some questions / challenges along the way:

- Currently the `PackageCache` caches `DESCRIPTION`, `NAMESPACE`, and `INDEX` alongside the `R/` sources, which felt right at the time but now I am not so sure! I am starting to think that instead `PackageCache` should only have the `R/` sources, and we should treat the `.libPaths()` as our source of truth for those other files, accessed via `Library`, rather than having them in two places.
    - I think originally my thought was that we could recreate a "source package like" structure in the cache (with `R/`, `DESCRIPTION`, and `NAMESPACE` all present), but I no longer think we need to do that, and should instead just keep 1 source of truth for the metadata files.

- I had originally wanted `PackageDefinitions` to really be something like `PackageSemanticIndexes` where it would hold a package worth of `SemanticIndex` objects for reuse across LSP features. But this is not easy! `SemanticIndex` is not `Send + Sync` due to `DefinitionKind` holding `RSyntaxNode`, so that means we can't put a `PackageSemanticIndexes` in the `WorldState`, since it needs to be snapshotted and shuffled between threads. I think we are going to need to resolve this eventually for Salsa purposes, but for now I just cache the `PackageDefinition` objects extracted from the semantic index instead, which are send+sync.


crates/harp/src/command.rs
DavisVaughan, line 1: I'm thinking this whole file should move out of harp into something like `aether_r_process` eventually

crates/ark/src/lsp/state.rs
```diff
@@ -65,6 +66,10 @@ pub(crate) struct WorldState {
     /// Map of package name to package metadata for installed libraries. Lazily populated.
     pub(crate) library: Library,
 
+    /// Map of package name to package definitions for installed libraries.
+    /// Only usable when a package cache is available.
+    pub(crate) library_definitions: Option<LibraryDefinitions>,
```
DavisVaughan: WorldState now gets library_definitions if we have a cache available

crates/ark/src/console.rs
```diff
@@ -178,6 +178,8 @@ thread_local! {
 }
 
 pub(crate) struct Console {
+    r_home: PathBuf,
```
DavisVaughan, line 181: Unused, but I think will be nice for the DAP + PackageCache ideas

crates/oak_index/src/external.rs
```diff
@@ -0,0 +1,111 @@
...
+
+use crate::scope_layer::ScopeLayer;
+
+/// The result of resolving a name against the external scope chain.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ExternalDefinition {
+    file: Url,
+    name: String,
+    range: TextRange,
+}
```
DavisVaughan, line 17: This used to be an enum that knew if the definition came from a project file vs a package, but now that I'm actually filling out the package side of things, I think they really boil down to the same structure, so I've flattened it

crates/oak_index/src/external.rs
```diff
@@ -0,0 +1,111 @@
...
+
+    let file = package_definitions.file(definition.file_id());
+    let file = Url::from_file_path(file).log_err()?;
+
+    Some(ExternalDefinition {
+        file,
+        name: name.to_string(),
+        range: definition.range(),
+    })
+}
```
DavisVaughan, line 108: The real crux of goto-definition for a package export

crates/oak_package_metadata/src/namespace.rs
```diff
@@ -5,12 +5,13 @@ use biome_rowan::AstNodeList;
...
 use biome_rowan::SyntaxResult;
 use oak_core::syntax_ext::RIdentifierExt;
+use stdext::SortedVec;
 
 /// Parsed NAMESPACE file
 #[derive(Default, Clone, Debug)]
 pub struct Namespace {
     /// Names of objects exported with `export()`
-    pub exports: Vec<String>,
+    pub exports: SortedVec<String>,
```
DavisVaughan, line 14: As we build `PackageDefinitions`, we categorize top level definitions as `Internal` vs `Exported` based on whether or not they exist as `exports` in the `Namespace` (this helps with `::` vs `:::` goto definition resolution).

Because we search `exports` for every top level definition in the package, it seemed useful to make this a sorted vec we can binary search in

DavisVaughan: I have left `imports` and `package_imports` alone for the moment

crates/oak_index/src/package_definitions.rs
DavisVaughan, line 1: The core file for "given a `R/` directory, collect all top level definitions"

crates/oak_package_definitions/src/library_definitions.rs
```diff
@@ -0,0 +1,78 @@
...
+use oak_package::package::Package;
+use oak_sources::traits::PackageCache;
+
+use crate::definitions::PackageDefinitions;
+
+#[derive(Clone, Debug)]
+pub struct LibraryDefinitions {
+    /// Cache used for looking up package sources. `dyn` to allow easy swapping to
+    /// `TestPackageCache` in test files.
+    cache: Arc<dyn PackageCache>,
```
DavisVaughan: We have both `struct PackageCache` and `struct TestPackageCache` that can be slotted in here, where the latter is very nice for testing in goto-definition.rs and externals.rs, since `PackageCache` itself requires a live R session and internet access.

We do this with a small `trait PackageCache` that requires a simple `get()` method.

It's `dyn` rather than using a generic like `LibraryDefinitions<C: PackageCache>` because that "leaks" the generic cache details way up the chain of callers (all the way up through goto-definition, it got gross). 

I had extensive conversations with Claude about this design and we both decided this was a nice way to handle this kind of thing, and that Zed also does this kind of thing as well (with a "fake" `Filesystem` for similar reasons).

The vtable hit required by using `dyn` should be negligible vs the overhead of the cache itself

crates/oak_index/src/external.rs
```diff
@@ -0,0 +1,102 @@
...
```
DavisVaughan, line 108: The real crux of goto-definition for a package export

crates/oak_ide/tests/goto_definition.rs
```diff
@@ -868,9 +970,12 @@ fn test_fixme_namespace_access_cursor_on_package_name() {
         &ExternalScope::default(),
         &library,
     );
-    // FIXME: Cursor on the package name still classifies as NamespaceAccess,
-    // resolves `mutate` in `dplyr` — but package symbols have no target yet
-    assert!(targets.is_empty());
+    // FIXME: Cursor on the package name still classifies as NamespaceAccess
```
lionel-, line 991: On second thought, should this take us to the DESCRIPTION file instead? That'd be kind of cool, it allows inspecting exports etc. Maybe we could provide multiple definitions: DESCRIPTION, NAMESPACE, and `"_PACKAGE"`? So you can navigate to metadata of interest for a given package.

crates/oak_ide/tests/goto_definition.rs
```diff
@@ -831,14 +920,19 @@ fn test_namespace_access_unknown_package() {
 }
 
 #[test]
-fn test_fixme_namespace_access_triple_colon() {
+fn test_namespace_access_triple_colon() {
```
lionel-, line 923: I think this is missing a test for: `pkg:::external`, i.e. accessing an exported symbol from `:::`.

crates/oak_index/src/external.rs
```diff
@@ -1,43 +1,35 @@
...
-#[derive(Debug, Clone)]
-pub enum ScopeLayer {
-    /// Bindings from a project file's top-level definitions.
-    /// When a name is defined multiple times, the last definition wins.
-    FileExports {
-        file: Url,
-        exports: HashMap<String, TextRange>,
-    },
+/// The result of resolving a name against the external scope chain.
+#[derive(Debug, Clone, PartialEq, Eq)]
```
lionel-, line 12: I'm thinking we might be able to fold this into `Definition` in the future (which will gain a file id field as discussed). It's fine to keep as is, especially given the Send/Sync issues you mentioned. Once we switch to Salsa the thread sharing issue will go away and we'll be able to merge.

```suggestion
/// TODO: Try to fold into `Definition` after switch to Salsa
#[derive(Debug, Clone, PartialEq, Eq)]
```

crates/oak_index/src/external.rs
```diff
@@ -80,68 +78,25 @@ pub fn resolve_external_name(
...
-            .collect();
-        layers.push(ScopeLayer::PackageImports(map));
-    }
+    // FIXME: Slow without salsa!
+    let package_definitions = library.definitions(package)?;
+    let definition = package_definitions.get(name)?;
 
-    for pkg in &namespace.package_imports {
-        layers.push(ScopeLayer::PackageExports(pkg.clone()));
+    if visibility != definition.visibility() {
```
lionel-: if visibility is internal, this should also match exported definitions, not just internal ones.

crates/oak_index/src/external.rs
```diff
@@ -60,15 +52,21 @@ pub fn resolve_external_name(
...
-                        name: name.to_string(),
-                    });
+                    if let Some(def) = resolve_in_package(
+                        name,
+                        pkg,
+                        PackageDefinitionVisibility::Exported,
+                        library,
+                    ) {
+                        return Some(def);
+                    }
```
lionel-, line 64: This didn't use to fallthrough. Looks like it should? E.g. if package sources are not available, we did our best but can't find a definition for that symbol. By falling through, we could match an unrelated definitions if unlucky.

lionel- (approved):
Looks good! I think the architecture we landed on makes sense. I'll rename oak_index to oak_semantic in the future to reflect the larger scope of the crate.

I'm also hopeful we can remove the `PackageDefinition` struct at some point, which does not feel totally right. Since packages provide files with semantic indices and `Definition`, and since `Definition` are meant to be shared (they should gain a file ID), we might be able to reuse it here.

crates/oak_ide/tests/goto_definition.rs
DavisVaughan: I actually think the current behavior is right and maximally useful. 

Like, if you hover over `dplyr` in `dplyr::across()` we show you `across()` help

Similarly I think if you cmd+click anywhere in `dplyr::across()` it should take you to an `across()` definition

crates/oak_index/src/external.rs
DavisVaughan: We might be able to rework some things in the future to make the _location_ optional

i.e., statically we should be able to confirm that we found the package export by looking at `NAMESPACE`, even without internet access. We might not be able to tell you _which file_ it came from if we don't have sources, but we can at least say we found the relevant package it comes from, rather than falling through like here?

crates/oak_ide/tests/goto_definition.rs
lionel-: I would say that although maybe right (convenient), it's clearly not maximally useful since it's less functional than the alternative discussed above.

DavisVaughan: As a normal data scientist i would be very disconcerted if i cmd+clicked on `dplyr::across()` in the wrong spot and didn't get taken to `across()`'s docs 🤷 , that's what i meant by maximally useful

lionel-: Agreed, so maybe in the future we make two behaviours, one for analysts and the other for package developers.

It feels like it'd be nice to provide ways of accessing package metadata right from your IDE. Could be via navigation on a package name (library call or lhs of ::), or via a command. I think navigation would be rad.

DavisVaughan: What about a 💡 Code Action when your cursor is on `dplyr` of `dplyr::across` or `dplyr` of `library(dplyr)`?

- Open `DESCRIPTION`
- Open `NAMESPACE`

DavisVaughan: opened https://github.com/posit-dev/ark/issues/1203

PR #1176, opened 2026-04-28
Avoid port race condition in help proxy startup

DavisVaughan:
I was looking at the help proxy code and noticed the "classic" port race condition issue that we've fought elsewhere

- `HelpProxy::get_os_assigned_port()` bound to port `0` to get a `port` but _released_ its binding, freeing it back to the OS
- Later `HttpServer` would `bind(port)`, but its entirely possibly that the OS took that port for something else in the meantime

We've seen this happen in practice often enough that it's worth it to remove the race condition alltogether.

There have been some rare instances where my first call to `?` doesn't work for unknown reasons, and I wonder if this was related.

Now we pass `bind()` itself port `0`, letting the OS assign us a port _that we never release_, and then we just query it and send it back to the caller using a crossbeam channel

### DIFF
--- a/crates/ark/src/help_proxy.rs
+++ b/crates/ark/src/help_proxy.rs
@@ -94,7 +94,12 @@ pub fn start(target_port: u16) -> anyhow::Result<u16> {
     });
 
     // Wait for the returned port with an extensive timeout
-    Ok(port_rx.recv_timeout(Duration::from_secs(20))?)
+    match port_rx.recv_timeout(Duration::from_secs(20)) {
+        Ok(port) => Ok(port),
+        Err(err) => Err(anyhow::anyhow!(
+            "Help proxy server timed out while waiting for a port: {err:?}"
+        )),
+    }
 }
 
 // Proxies a request.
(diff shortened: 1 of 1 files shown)


crates/ark/src/help_proxy.rs
```diff
@@ -40,11 +41,39 @@ struct PreviewRdParams {
...
 }
 
+struct AppState {
+    target_port: u16,
+}
+
 // Starts the help proxy.
 pub fn start(target_port: u16) -> anyhow::Result<u16> {
-    let source_port = HelpProxy::get_os_assigned_port()?;
+    let (port_tx, port_rx) = crossbeam::channel::bounded::<u16>(1);
```
DavisVaughan, line 49: I tried to avoid the channel by creating the `server` outside of the `spawn!`, plucking off its `port`, and then `move`-ing it into the thread, but it isn't `Send` so you can't do that.

lionel- (approved):
It's good to fix that race, and also a nice simplification

PR #1177, opened 2026-04-28
Report attached status from pkg_list

bricestacey:
The Positron Packages pane renders a read-only attached indicator (filled green dot when attached, empty ring when not) using a single new field:

```
attached = paste0("package:", name) %in% search()
```

The field is named `attached` rather than `loaded` because R's `loadedNamespaces()` is a strict superset of `search()`: a package can be loaded as a transitive dependency without being attached. The new name reflects what we actually report.

The pak/base/renv method branches collapse to a single `utils::installed.packages()` call. renv overrides `.libPaths()` in an active project, so the same call works there too. The `method` parameter stays in the signature for RPC compatibility but is unused.

See posit-dev/positron#13184

### DIFF
--- a/crates/ark/src/modules/positron/packages_pane.R
+++ b/crates/ark/src/modules/positron/packages_pane.R
@@ -8,48 +8,34 @@
 # This file contains RPC functions for the packages pane.
 # These functions are called via callMethod from the Positron R extension.
 
-# Return a list of installed packages
+# Return a list of installed packages.
+#
+# The `method` parameter is accepted for compatibility with the existing
+# Positron R extension RPC contract but is unused: base R's
+# installed.packages() reads from .libPaths(), which renv overrides in an
+# active project, so the same call works for pak/base/renv callers.
 #' @export
 .ps.rpc.pkg_list <- function(method = c("pak", "base", "renv")) {
-    method <- match.arg(method)
-    switch(
-        method,
-        pak = {
-            old_opt <- options(pak.no_extra_messages = TRUE)
-            on.exit(options(old_opt), add = TRUE)
-            pkgs <- pak::lib_status()
-            lapply(seq_len(nrow(pkgs)), function(i) {
-                list(
-                    id = paste0(pkgs$package[[i]], "-", pkgs$version[[i]]),
-                    name = pkgs$package[[i]],
-                    displayName = pkgs$package[[i]],
-                    version = as.character(pkgs$version[[i]])
-                )
-            })
-        },
-        base = {
-            ip <- utils::installed.packages()
-            lapply(seq_len(nrow(ip)), function(i) {
-                list(
-                    id = paste0(ip[i, "Package"], "-", ip[i, "Version"]),
-                    name = ip[i, "Package"],
-                    displayName = ip[i, "Package"],
-                    version = ip[i, "Version"]
-                )
-            })
-        },
-        renv = {
-            ip <- utils::installed.packages(lib.loc = renv::paths$library())
-            lapply(seq_len(nrow(ip)), function(i) {
-                list(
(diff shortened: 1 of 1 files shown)


bricestacey:
This PR collapses the implementation between the three package manages into a single implementation. This was the recommendation from https://github.com/posit-dev/ark/pull/1165 and it feels like a good idea.

This PR is also a minimalist change so as to avoid any distractions trying to land code for the next release. So, this only focuses on the loaded prop.

crates/ark/src/modules/positron/packages_pane.R
```diff
@@ -8,48 +8,26 @@
...
+    lapply(seq_len(nrow(ip)), function(i) {
+        name <- ip[i, "Package"]
+        list(
+            id = paste0(name, "-", ip[i, "Version"]),
+            name = name,
+            displayName = name,
+            version = ip[i, "Version"],
+            loaded = paste0("package:", name) %in% attached_packages
+        )
+    })
```
jennybc: I think the above produces a correct result. But similar to a suggestion in #1165, it's more idiomatic for R and its native vectorization to make these objects first. Then reshape:

```r
    ip      <- utils::installed.packages()
    name    <- ip[, "Package"]
    version <- ip[, "Version"]
    id      <- paste0(name, "-", version)
    loaded  <- paste0("package:", name) %in% search()

    Map(list, id = id, name = name, displayName = name,
        version = version, loaded = loaded)
```

So, this suggestion is mostly about style, but I'll make it anyway. I guess it might even be a bit more efficient to get the `%in%` out of the loop, but that's not my main motivation.

crates/ark/src/modules/positron/packages_pane.R
```diff
@@ -8,48 +8,26 @@
...
+    ip <- utils::installed.packages()
+    attached_packages <- search()
+    lapply(seq_len(nrow(ip)), function(i) {
+        name <- ip[i, "Package"]
+        list(
+            id = paste0(name, "-", ip[i, "Version"]),
+            name = name,
+            displayName = name,
+            version = ip[i, "Version"],
+            loaded = paste0("package:", name) %in% attached_packages
```
jennybc: Should we call this "attached"? I'm just thinking to head off any confusion about attached vs loaded. Because we clearly mean attached here.

I'm getting at the difference between what `search()` reports versus `loadedNamespaces()`.

jennybc (approved):
I made 2 comments, but I'm happy for you to resolve and get this feature in.

crates/ark/src/modules/positron/packages_pane.R
bricestacey: Done in 2acdb3e1 — vectorized as suggested. One subtlety: `Map(list, id = id, ...)` produces a list named by `id`'s values (mapply's `USE.NAMES = TRUE` on a character first arg uses the values themselves), and a named R list serializes to a JSON object instead of the array the frontend expects. Wrapped the call in `unname()` and left a note explaining the gotcha.

crates/ark/src/modules/positron/packages_pane.R
bricestacey: ~Fair point on attached vs. loaded — `search()` is unambiguously the attached set. I'd like to defer this one: the field name `loaded` is also what the Python kernel reports and what the Positron UI keys on (`pkg.loaded`), so renaming only the R kernel would silently break the indicator for R packages. Renaming everywhere is a wider cross-runtime change that's outside this PR's minimalist scope. Happy to file a follow-up to do it consistently across the API surface if you'd like.~

Bad Claude! Actually, I think this is a good idea, as there are also some nuance and overloaded terminology for the python extension.. I am going to update everything to be attached.

jennybc (approved):

PR #1179, opened 2026-04-29
Override `utils::timestamp()` on Windows and explain why

DavisVaughan:
Addresses https://github.com/posit-dev/positron/issues/13261

With A LOT of explanation about `.Rhistory` and what we might want to do in the future with all of this

Yay

Screenshot 2026-04-29 at 4 46 06 PM

### DIFF
--- a/.github/workflows/test-linux.yml
+++ b/.github/workflows/test-linux.yml
@@ -18,13 +18,20 @@ jobs:
       fail-fast: false
       matrix:
         config:
+          # Test on all supported R versions
+          # (Particularly important to do this on at least one OS for
+          # version dependent R hook override checks, like `utils::View()`)
           - { rust: 'stable',  r: 'release' }
-          # Oldest supported R version
+          - { rust: 'stable',  r: '4.5' }
+          - { rust: 'stable',  r: '4.4' }
+          - { rust: 'stable',  r: '4.3' }
           - { rust: 'stable',  r: '4.2' }
-          # Nightly rust
-          - { rust: 'nightly', r: 'release' }
+
           # Track upstream breaking changes
           - { rust: 'stable',  r: 'devel' }
+
+          # Nightly rust
+          - { rust: 'nightly', r: 'release' }
     timeout-minutes: 40
     env:
       GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/test-linux.yml
+++ b/.github/workflows/test-linux.yml
@@ -27,11 +27,9 @@ jobs:
           - { rust: 'stable',  r: '4.3' }
           - { rust: 'stable',  r: '4.2' }
 
-          # Track upstream breaking changes
-          - { rust: 'stable',  r: 'devel' }
-
-          # Nightly rust
-          - { rust: 'nightly', r: 'release' }
+          # Early warning signal for upstream changes,
+          # both in Rust and R-devel
+          - { rust: 'nightly',  r: 'devel' }
     timeout-minutes: 40
     env:
       GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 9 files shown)


crates/ark/src/modules/positron/hooks.R
```diff
@@ -23,6 +23,10 @@ initialize_hooks <- function() {
         namespace = TRUE
     )
     register_getHook_hook()
+
+    if (is_windows()) {
+        rebind("utils", "timestamp", windows_timestamp, namespace = TRUE)
```
lionel-, line 28: It occurs to me that it would be helpful to have a debug-build-only check that tests that our functions have the same signature as the base R functions they replace. It would panic if not, allowing our r-devel CI checks to detect any upstream changes.

lionel- (approved):

.github/workflows/test-linux.yml
```diff
@@ -18,13 +18,20 @@ jobs:
...
       matrix:
         config:
+          # Test on all supported R versions
+          # (Particularly important to do this on at least one OS for
+          # version dependent R hook override checks, like `utils::View()`)
           - { rust: 'stable',  r: 'release' }
-          # Oldest supported R version
+          - { rust: 'stable',  r: '4.5' }
+          - { rust: 'stable',  r: '4.4' }
+          - { rust: 'stable',  r: '4.3' }
```
lionel-, line 27: slightly worried about making our CI more sensitive to flakes but that still feels like the right call.

.github/workflows/test-linux.yml
```diff
@@ -18,13 +18,20 @@ jobs:
...
           - { rust: 'stable',  r: '4.2' }
-          # Nightly rust
-          - { rust: 'nightly', r: 'release' }
+
           # Track upstream breaking changes
           - { rust: 'stable',  r: 'devel' }
+
+          # Nightly rust
+          - { rust: 'nightly', r: 'release' }
```
lionel-: Should we merge these?

crates/ark/src/modules/positron/hooks.R
```diff
@@ -68,18 +70,44 @@ pkg_bind <- function(pkg, name, value) {
         stop(msg, call. = FALSE)
     }
 
+    # Check formals, but if we are wrong, only panic in debug builds.
```
lionel-, line 73: I'm happy we're being more careful with these hooks.

crates/ark/src/modules/positron/hooks_source.R
```diff
@@ -31,9 +31,15 @@ make_ark_source <- function(original_source) {
         encoding = getOption("encoding"),
         continue.echo = getOption("continue"),
         skip.echo = 0,
-        keep.source = getOption("keep.source"),
-        ...
-    ) {
+        keep.source = getOption("keep.source")
+    )
+
+    # Remove arguments that are not yet supported
```
lionel-, line 37: oh that's a nice way of managing the versions

crates/ark/src/modules/positron/utils.R
```diff
@@ -141,6 +141,14 @@ node_poke_cdr <- function(node, cdr) {
     .ps.Call("ark_node_poke_cdr", node, cdr)
 }
 
+# Detect if ark is running under a debug build or not
```
lionel-, line 144: suggestion: A better home for this might be in harp

lionel- (approved):
yay!

crates/ark/src/modules/positron/utils.R
DavisVaughan: i thought you can't access harp rust code from ark R modules?

.github/workflows/test-linux.yml
DavisVaughan: It does seem unlikely that we'd have issues with both r-devel and nightly rust all at once, and it reduces the number of workers by 1, so that's probably worth it yea

crates/ark/src/modules/positron/utils.R
lionel-: but you can access harp R code from Ark R modules

lionel-: update: no you cannot :P

PR #1180, opened 2026-04-29
Assert that yes, `polled_events()` is really not run on Windows

DavisVaughan:
Follow up to https://github.com/posit-dev/ark/pull/1175 now that I have my real Windows machine up. Without this we get:

Screenshot 2026-04-29 at 4 57 33 PM

@lionel- is it concerning that `polled_events()` never runs on Windows?

I see we do `self.handle_task_interrupt()` elsewhere inside `run_event_loop()` so those tasks probably get to run eventually even if they can't run at interrupt time.

But this part of `polled_events()` feels kind of important?

```
        // Check stream filter timeout to handle long computations between
        // WriteConsole calls. Timeout means we didn't reach ReadConsole to
        // confirm debug output within a reasonable amount of time, so
        // accumulated content is emitted. This allows user code to produce
        // output that looks like debug lines emitted by R without them getting
        // filtered out or held up too long.
        if let Some(text) = self.debug_filter.check_timeout() {
            self.emit_stdout(text);
        }
```

### DIFF
--- a/crates/ark/src/console.rs
+++ b/crates/ark/src/console.rs
@@ -60,7 +60,6 @@ use harp::environment::r_ns_env;
 use harp::environment::Environment;
 use harp::environment::R_ENVS;
 use harp::exec::exec_with_cleanup;
-use harp::exec::r_check_stack;
 use harp::exec::r_peek_error_buffer;
 use harp::exec::r_sandbox;
 use harp::exec::with_calling_error_handler;
--- a/crates/ark/src/console/console_filter.rs
+++ b/crates/ark/src/console/console_filter.rs
@@ -263,6 +263,7 @@ impl ConsoleFilter {
     /// Check for timeout and handle state transitions.
     /// Timeout means we didn't reach ReadConsole to confirm debug output,
     /// so we emit the accumulated content back to the user.
+    #[cfg(any(unix, test))]
     pub(super) fn check_timeout(&mut self) -> Option<String> {
         self.drain_on_timeout()
     }
--- a/crates/ark/src/console/console_repl.rs
+++ b/crates/ark/src/console/console_repl.rs
@@ -2354,6 +2354,7 @@ impl Console {
     }
 
     /// Invoked by the R event loop
+    #[cfg(unix)]
     fn polled_events(&mut self) {
         // Don't process tasks until R is fully initialized
         if !Self::is_initialized() {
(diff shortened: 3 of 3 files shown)


lionel- (commented):
yep it would be safer to run polled events on Windows too.

I think we could do it via the `ProcessEvents` hook which is called here at interrupt time:

- https://github.com/r-devel/r-svn/blob/e35e930cc8cae8882ef3aca9101b4e2f1ca7a4a3/src/main/errors.c#L157

- Via https://github.com/r-devel/r-svn/blob/e35e930cc8cae8882ef3aca9101b4e2f1ca7a4a3/src/gnuwin32/system.c#L156

We currently set it to a no-op: https://github.com/posit-dev/ark/blob/d1e6263fd1a8e3e9cfd7363deb0689dd94b39d0c/crates/ark/src/sys/windows/console.rs#L213

DavisVaughan:
I think this is going to require a little thinking because right now we have 

```
    fn process_idle_events() {
        // Process regular R events. We're normally running with polled
        // events disabled so that won't run here. We also run with
        // interrupts disabled, so on Windows those won't get run here
        // either (i.e. if `UserBreak` is set), but it will reset `UserBreak`
        // so we need to ensure we handle interrupts right before calling
        // this.
        unsafe { R_ProcessEvents() };
```

i.e. we have some expectation that tasks only run at "polled events" time not at "process events" time, and we have `RLocalPolledEventsSuspended` to even suspend polled events, i.e. to suspend tasks from getting run.

If we start running tasks on Windows at "process events" time by overriding `Rp->Callback`, which sets `ptr_ProcessEvents`, then `RLocalPolledEventsSuspended` doesn't make much sense anymore as a guard and things get a bit wonky.

We should probably consider if we can consistently use "process events" as a cross-OS consistent place to run tasks. On Unix you can hook into `ptr_ProcessEvents` directly, on windows you can set it at startup via `Rp->Callback`. The issue on Windows though is that you can't _unset_ it at any arbitrary time, so I'm not sure how we'd have something like `RLocalPolledEventsSuspended` there.

It does get called at interrupt time though https://github.com/wch/r-source/blob/8e54c98da1168a2369fe4ec95190bb24df44e101/src/main/errors.c#L141

DavisVaughan:
Opened https://github.com/posit-dev/ark/issues/1187 to consider further

PR #1181, opened 2026-04-29
Create arm64/x64 releases

jmcphers:
Progress towards https://github.com/posit-dev/positron/issues/8423; sets up architecture-specific versions of Ark so we don't have to consume a universal binary in Positron.

### DIFF
--- a/.github/workflows/release-macos.yml
+++ b/.github/workflows/release-macos.yml
@@ -24,6 +24,7 @@ jobs:
 
         env:
             GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


DavisVaughan (approved):
Nice yea I saw your issue and had taken a quick look and this is also what I was thinking we'd need to tweak

So for the moment we will just continue publishing Universal builds as well until Positron is fully off them?

jmcphers:
Yep, universals stay for now. It might be nice to leave them around in case any downstream projects need them? In Positron I've advocated heavily to get us off universal binaries because the cost of maintaining them is so high, but they've not been expensive in ark.

PR #1185, opened 2026-04-30
Allow build-time override of embedded version string

jmcphers:
Adds an `ARK_BUILD_VERSION` env var that, when set, replaces `CARGO_PKG_VERSION` as the version reported by the binary (`--version`, `.ps.ark.version()`, the Jupyter kernel info banner, and the LSP `server_info`). Defaults to `CARGO_PKG_VERSION` so existing builds are unchanged.

This is the upstream hook for Positron's planned move from a prebuilt ark dependency to a git submodule (posit-dev/positron#13148). Positron prebuilds will set `ARK_BUILD_VERSION=<release>+<distance>` (e.g. `0.1.251+10`) so a user can tell at runtime which submodule commit they are running against.

### DIFF
--- a/crates/ark/src/lsp/state_handlers.rs
+++ b/crates/ark/src/lsp/state_handlers.rs
@@ -133,7 +133,7 @@ pub(crate) fn initialize(
     Ok(InitializeResult {
         server_info: Some(ServerInfo {
             name: "Ark R Kernel".to_string(),
-            version: Some(env!("ARK_BUILD_VERSION").to_string()),
+            version: Some(crate::BUILD_VERSION.to_string()),
         }),
         capabilities: ServerCapabilities {
             // Currently hard-coded to UTF-16, but we might want to allow UTF-8 frontends
--- a/crates/ark/src/lsp/state_handlers.rs
+++ b/crates/ark/src/lsp/state_handlers.rs
@@ -133,7 +133,7 @@ pub(crate) fn initialize(
     Ok(InitializeResult {
         server_info: Some(ServerInfo {
             name: "Ark R Kernel".to_string(),
-            version: Some(env!("CARGO_PKG_VERSION").to_string()),
+            version: Some(env!("ARK_BUILD_VERSION").to_string()),
         }),
         capabilities: ServerCapabilities {
             // Currently hard-coded to UTF-16, but we might want to allow UTF-8 frontends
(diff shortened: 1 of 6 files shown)


crates/ark/src/lsp/state_handlers.rs
```diff
@@ -133,7 +133,7 @@ pub(crate) fn initialize(
     Ok(InitializeResult {
```
DavisVaughan, line 133: See https://github.com/posit-dev/ark/pull/1188

crates/ark/src/shell.rs
```diff
@@ -157,7 +157,7 @@ impl ShellHandler for Shell {
             help_links: Vec::new(),
             language_info: info,
             implementation: String::from("ark"),
-            implementation_version: String::from(env!("CARGO_PKG_VERSION")),
+            implementation_version: String::from(env!("ARK_BUILD_VERSION")),
```
DavisVaughan: The Jupyter Spec actually seems to request a specific version form 
https://jupyter-client.readthedocs.io/en/stable/messaging.html

```
'implementation_version': 'X.Y.Z',
```

They typically just write `str` when they want a string

Hopefully this doesn't affect anything. At the very least it should only affect Positron builds of ark, and it doesn't seem like they will be parsing the `implementation_version` and doing anything with it that would require the `X.Y.Z` form.

DavisVaughan (approved):
See https://github.com/posit-dev/ark/pull/1188 for a few consistency tweaks

We also don't need rerun-if-changed as long as we access env vars with `env!` and `option_env!`, so that's nice https://doc.rust-lang.org/cargo/reference/build-scripts.html#rerun-if-env-changed, I removed in #1188 

I feel slightly weird about allowing an external system to modify the internal ark version, but I could not think of anything better.

crates/ark/src/shell.rs
jmcphers: That's a good find. It's not clear to me whether that's an example or normative in some way? At any rate I don't think we use this value in any special way so better to err on the side of safety; I've reverted this value to `CARGO_PKG_VERSION`.

crates/ark/src/lsp/state_handlers.rs
jmcphers: Thanks for this improvement!

PR #1190, opened 2026-05-01
Progress notification infrastructure

DavisVaughan:
Not completely hooked up - no one is sending events to the package sources event loop yet, we are waiting on Salsa support before doing that.

But I feel like the two main core changes are nice:
- New package sources event loop for managing requests to populate sources without blocking the main loop
- New progress reporting support added to the auxiliary event loop, as long as the client has indicated that it supports progress updates

You can treat these two features as orthogonal, where the package sources event loop is a user of the progress reporter

### DIFF
--- a/crates/ark/src/lsp/backend.rs
+++ b/crates/ark/src/lsp/backend.rs
@@ -15,6 +15,7 @@ use amalthea::comm::server_comm::ServerStartMessage;
 use amalthea::comm::server_comm::ServerStartedMessage;
 use anyhow::Context;
 use crossbeam::channel::Sender;
+use oak_index::library::Library;
 use serde_json::Value;
 use stdext::result::ResultExt;
 use tokio::net::TcpListener;
@@ -559,12 +560,52 @@ pub(crate) fn start_lsp(
         let (shutdown_tx, mut shutdown_rx) = tokio::sync::mpsc::channel::<()>(1);
 
         let init = |client: Client| {
-            let auxiliary = AuxiliaryState::new(client.clone());
-            let (package_sources, package_sources_event_tx) = PackageSourcesState::new();
+            let r = harp::command::r_executable(&r_home);
+
+            // FIXME: We shouldn't call R code in the kernel to figure this out
+            let library_paths = r_task(|| -> anyhow::Result<Vec<String>> {
+                Ok(harp::RFunction::new("base", ".libPaths")
+                    .call()?
+                    .try_into()?)
+            });
+            let library_paths = match library_paths {
+                Ok(library_paths) => library_paths,
+                Err(err) => {
+                    log::error!("Can't evaluate `libPaths()`: {err:?}");
+                    Vec::new()
+                },
+            };
+            let library_paths: Vec<PathBuf> =
+                library_paths.into_iter().map(PathBuf::from).collect();
+
+            // If reader/writer generation fails, all features related to source references are disabled,
+            // but we don't bring the whole system down. We don't expect this to ever occur in practice.
+            let (package_cache_reader, package_cache_writer) =
+                match oak_sources::new_cache_pair(r, library_paths.clone()).log_err() {
+                    Some((reader, writer)) => (Some(reader), Some(writer)),
+                    None => (None, None),
+                };
+
+            let package_sources = package_cache_reader
+                .map(|reader| Arc::new(reader) as Arc<dyn oak_sources::PackageSources>);
+
+            // Package cache reader goes to `Library`
+            let library = Library::new(library_paths, package_sources);
+
+            // Package cache writer goes to `PackageSourcesState` event loop
+            let (package_sources_state, package_sources_event_tx) = match package_cache_writer {
+                Some(writer) => {
+                    let (state, event_tx) = PackageSourcesState::new(writer);
+                    (Some(state), Some(event_tx))
+                },
+                None => (None, None),
+            };
+
+            let auxiliary_state = AuxiliaryState::new(client.clone());
 
             let (state, events_tx) = GlobalState::new(
                 client,
-                r_home,
+                library,
                 console_notification_tx,
                 package_sources_event_tx,
             );
@@ -588,10 +629,12 @@ pub(crate) fn start_lsp(
 
             // Spawn latency-sensitive auxiliary loop. Must be first to initialise
             // global transmission channel.
-            tokio_tasks_handle.spawn(async move { auxiliary.start().await });
+            tokio_tasks_handle.spawn(async move { auxiliary_state.start().await });
 
             // Spawn package sources event loop
-            tokio_tasks_handle.spawn(async move { package_sources.start().await });
+            if let Some(package_sources_state) = package_sources_state {
+                tokio_tasks_handle.spawn(async move { package_sources_state.start().await });
+            }
 
             // Spawn main loop
             tokio_tasks_handle.spawn(async move { state.start().await });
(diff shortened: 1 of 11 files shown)


crates/ark/src/lsp/main_loop.rs
```diff
@@ -202,67 +202,23 @@ impl GlobalState {
...
-        };
-
-        let library_paths: Vec<PathBuf> = library_paths.into_iter().map(PathBuf::from).collect();
-
-        let r = harp::command::r_executable(&r_home);
-        let package_sources = r
-            .and_then(|r| PackageCache::new(r, library_paths.clone()).log_err())
-            .map(|cache| Arc::new(cache) as Arc<dyn oak_sources::PackageSources>);
-
-        let library = Library::new(library_paths, package_sources);
```
DavisVaughan, line 227: I have lifted all of this up out of the constructor because:
- It is needed in the new package sources event loop
- We eventually want to get `library_paths` construction out of this constructor anyways (see the FIXME in here)

crates/ark/src/lsp/main_loop.rs
```diff
@@ -202,67 +202,23 @@ impl GlobalState {
...
-    pub(crate) fn start(self) -> tokio::task::JoinSet<()> {
-        let mut set = tokio::task::JoinSet::<()>::new();
-
-        // Spawn latency-sensitive auxiliary loop. Must be first to initialise
-        // global transmission channel.
-        let aux = AuxiliaryState::new(self.client.clone());
-        set.spawn(async move { aux.start().await });
-
-        // Spawn main loop
-        set.spawn(async move { self.main_loop().await });
```
DavisVaughan, line 256: It always felt weird to me that the GlobalState controlled the spawning of everything

Really we have 3 event loops:
- GlobalState::start()
- AuxiliaryState::start()
- PackageSourceState::start()

`start_lsp()` now manages all 3 more directly

(It was also kind of required to do it this way due to the fact that `PackageSourceState::new()` needs a few things that aren't available in GlobalState, so you can't start it from GlobalState)

crates/ark/src/lsp/progress.rs
DavisVaughan, line 1: All AuxiliaryState support for arbitrary progress reporting lives here

crates/ark/src/lsp/package_sources.rs
DavisVaughan, line 1: Package sources event loop lives here

crates/harp/src/command.rs
```diff
@@ -94,18 +94,19 @@ pub fn r_home_setup() -> anyhow::Result<PathBuf> {
...
 
         if candidate.exists() {
-            return Some(candidate);
+            return candidate;
         }
     }
 
-    None
+    // We fully expect to be able to find an executable, it is required for many tasks
+    panic!("Can't find R executable relative to {r_home:?}")
```
DavisVaughan, line 109: It was getting annoying for this to be optional, and it seems like critical info

crates/oak_sources/src/cache.rs
```diff
@@ -28,14 +28,25 @@ const METADATA_FILENAME: &str = ".metadata";
...
+/// # Reader / Writer split
+///
+/// The cache is split into [PackageCacheReader] and [PackageCacheWriter].
+///
+/// - [PackageCacheReader::get] reads from the cache and returns instantly. If the
+///   package isn't in the cache, it simply returns [None].
+///
+/// - [PackageCacheWriter::insert] writes into the cache if the package sources aren't
+///   already in the cache. Insertion can be very expensive, and takes place on a
+///   dedicated tokio task to avoid blocking the main loop.
```
DavisVaughan, line 48: Most important change

crates/oak_sources/src/cache.rs
```diff
@@ -106,35 +117,44 @@ impl crate::PackageSources for PackageCache {
...
+    /// we don't attempt expensive source generation again.
+    source_unavailable: HashSet<String>,
+}
+
+/// Immutable data shared between reader and writer
+///
+/// Both carry the shared cache root lock, because both look at the cache and hand out
+/// paths into it.
+#[derive(Debug)]
+struct PackageCacheShared {
```
DavisVaughan, line 145: This might be useful in the future as well

If the `Reader` needs to check if there are packages in the queue (i.e. if you just wait a moment they will be ready) then maybe we can store some shared state about them in an RwLock here, which is data shared between a Reader/Writer pair. That was my eventual plan, at least.

crates/ark/src/lsp/main_loop.rs
```diff
@@ -144,12 +144,15 @@ pub(crate) struct GlobalState {
     /// LSP client shared with tower-lsp and the log loop
     client: Client,
 
-    /// Event channels for the main loop. The tower-lsp methods forward
+    /// Event channel for recieving on the main loop. The tower-lsp methods forward
```
lionel-, line 147:
```suggestion
    /// Event channel for receiving on the main loop. The tower-lsp methods forward
```

crates/ark/src/lsp/backend.rs
```diff
@@ -557,11 +560,55 @@ pub(crate) fn start_lsp(
...
+                match oak_sources::new_cache_pair(r, library_paths.clone()).log_err() {
+                    Some((reader, writer)) => (Some(reader), Some(writer)),
+                    None => (None, None),
+                };
+
+            let package_sources = package_cache_reader
+                .map(|reader| Arc::new(reader) as Arc<dyn oak_sources::PackageSources>);
+
+            // Package cache reader goes to `Library`
+            let library = Library::new(library_paths, package_sources);
```
lionel-, line 593: With https://github.com/posit-dev/ark/pull/1221 `Library` becomes a Salsa input that you can set with `db.set_library_paths()`.

crates/oak_sources/src/cache.rs
lionel-: That seems much cleaner and easier to reason about / work with.

crates/oak_sources/src/cache.rs
```diff
@@ -106,35 +117,44 @@ impl crate::PackageSources for PackageCache {
...
+pub struct PackageCacheWriter {
+    shared: Arc<PackageCacheShared>,
+
     /// Path to an R executable
     r: PathBuf,
 
+    /// Set of packages which are installed, but we failed to populate their sources (from
+    /// CRAN or srcrefs). If we request sources for one of these packages a second time,
+    /// we don't attempt expensive source generation again.
+    source_unavailable: HashSet<String>,
```
lionel-, line 137: It now seems to me this sort of defensive measure should be handled one level higher. I.e. what if the caller/owner decides it wants to retry?

Now that we no longer try to do too much in `oak_sources` (e.g. populat sources from a `get()` call), the crate feels lower level than before.

crates/ark/src/lsp/package_sources.rs
```diff
@@ -0,0 +1,79 @@
...
+        match event {
+            PackageSourcesEvent::Populate(populate) => self.handle_populate(populate),
+        }
+    }
+
+    fn handle_populate(&mut self, populate: Populate) {
+        // We don't populate packages concurrently, so we don't need per package ids
+        let id = String::from("package-sources");
+
+        // TODO: This currently reports progress when its already in the cache (i.e. dplyr
```
lionel-, line 59: I think we'll be able to set up the driver to avoid requesting sources when they already exist. It should maintain a set of packages that either already had sources or sent a populate request.

lionel- (approved):
Looks good!

I've gathered some thoughts about how to integrate it with the Salsa database at https://github.com/posit-dev/ark/issues/1234

PR #1195, opened 2026-05-04
Update to zeromq 4.3.5 to solve polling on Windows arm64

lionel-:
In https://github.com/posit-dev/ark/pull/1033 and https://github.com/posit-dev/ark/pull/1171, I've had to add ugly workarounds for Windows arm64, very deep in polling internals of Ark. On that platform, polling on sockets with a timeout hanged indefinitely instead of resolving after the timeout. The workaround was to sample-poll with short timeouts manually.

While investigating the cause of this issue, I noticed that ZMQ 4.3.5 did not have this issue. We're currently on 4.3.4, which hasn't been released to crates.io yet (https://github.com/erickt/rust-zmq/issues/420). The ZMQ update is on main branch though, so this PR switches us to the latest commit on the repo. An alternative would be to use that fork that is more actively maintained, we could consider that in the future if the community converges there: https://github.com/rdaum/rust-zmq/

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -119,3 +119,8 @@ winsafe = { version = "0.0.27", features = ["kernel"] }
 xdg = "3.0.0"
 yaml-rust2 = "0.11"
 zmq = "0.10.0"
+
+[patch.crates-io]
+# We need ZMQ 4.3.5 to fix polling issues on Windows arm64
+zmq = { git = "https://github.com/erickt/rust-zmq", rev = "5d78967" }
+zmq-sys = { git = "https://github.com/erickt/rust-zmq", rev = "5d78967" }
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -119,3 +119,8 @@ winsafe = { version = "0.0.27", features = ["kernel"] }
 xdg = "3.0.0"
 yaml-rust2 = "0.11"
 zmq = "0.10.0"
+
+[patch.crates-io]
+# We need ZMQ 4.3.5 to fix polling issues on Windows arm64
+zmq = { git = "https://github.com/erickt/rust-zmq", rev = "5d78967" }
+zmq-sys = { git = "https://github.com/erickt/rust-zmq", rev = "5d78967" }
```
DavisVaughan: Why list this in `patch`? 

zmq itself is a direct dependency, I think this should just be listed there instead

lionel-: Because I started with a tweak of zeromq-src-rs, then zmq-sys, at which point I thought zmq should be made consistent since it's in the same repo. But then as you note we can just tweak zmq directly.

DavisVaughan (approved):
Im happy with using `main` of the official repo for now

PR #1199, opened 2026-05-05
Add hard fail on unsupported R version

### LINKED ISSUE #696: Ark: Fail hard on unsupported R versions (i.e. if launched through Jupyter)
We need to make sure that ark itself is explicit about the fact that we only support R >=4.2.0, so if we detect that we are being launched with an older version of R, we should have some way of failing early with some informative error in the log.

Positron itself is good about not allowing users to start R < 4.2.0, but that won't be the case in general if ark is used as a kernel.

thomasp85:
Fix #696

This PR inserts a version check in the start method, right after R detection. It does not dive into checking R package versions as suggested since these are not hard requirements AFAICT and Ark gracefully handles it already


crates/ark/src/version.rs
```diff
@@ -91,3 +106,63 @@ pub unsafe extern "C-unwind" fn ps_ark_version() -> anyhow::Result<SEXP> {
...
+            major: 4,
+            minor: 1,
+            patch: 0,
+            r_home: "".to_string(),
+        };
+        assert!(!version.is_supported());
+    }
+
+    #[test]
+    fn test_accept_excact() {
```
lionel-:
```suggestion
    fn test_accept_exact() {
```

crates/ark/src/version.rs
```diff
@@ -91,3 +106,63 @@ pub unsafe extern "C-unwind" fn ps_ark_version() -> anyhow::Result<SEXP> {
...
+            major: 4,
+            minor: 2,
+            patch: 0,
+            r_home: "".to_string(),
+        };
+        assert!(version.is_supported());
+    }
+
+    #[test]
+    fn test_reject_accept_high_minor() {
```
lionel-:
```suggestion
    fn test_accept_high_minor() {
```

crates/ark/src/version.rs
```diff
@@ -91,3 +106,63 @@ pub unsafe extern "C-unwind" fn ps_ark_version() -> anyhow::Result<SEXP> {
...
+            major: 4,
+            minor: 4,
+            patch: 0,
+            r_home: "".to_string(),
+        };
+        assert!(version.is_supported());
+    }
+
+    #[test]
+    fn test_reject_high_major() {
```
lionel-:
```suggestion
    fn test_high_major() {
```

crates/ark/src/version.rs
```diff
@@ -91,3 +106,63 @@ pub unsafe extern "C-unwind" fn ps_ark_version() -> anyhow::Result<SEXP> {
     let result = RObject::from(info);
     Ok(result.sexp)
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_reject_low_major() {
```
lionel-: I would prefix with `test_version_` to give some hints of what this is about in the test output

crates/ark/src/version.rs
```diff
@@ -55,13 +65,18 @@ pub fn detect_r() -> anyhow::Result<RVersion> {
...
     } else {
         anyhow::bail!("Failed to extract R version");
     }
 }
 
+// Like detect_r() but calls r_home_setup() to get the path
+pub fn detect_r_setup() -> anyhow::Result<RVersion> {
+    detect_r(&r_home_setup()?)
+}
+
```
lionel-: Suggestion: Maybe remove `detect_r_setup()` and inline at call site? I think reading the full call is a bit clearer.

crates/ark/src/version.rs
```diff
@@ -31,9 +35,15 @@ pub struct RVersion {
...
 
-pub fn detect_r() -> anyhow::Result<RVersion> {
-    let r_home: String = r_home_setup()?.to_string_lossy().to_string();
+impl RVersion {
+    pub fn is_supported(&self) -> bool {
+        self.major > MIN_R_MAJOR || (self.major == MIN_R_MAJOR && self.minor >= MIN_R_MINOR)
+    }
+}
 
+// It is assumed that r_home matches the R_HOME env var. Usually the input will
```
lionel-: That's a bit confusing. If we're passing `r_home` through, we should probably run that R?

lionel-: Right now we're only passing it to return it along with the version. That seems misleading, since it suggests the function needs it but not really.

thomasp85: yeah, I agree... this was the minimally invasive path but it does lead to some weirdness

thomasp85: How would you feel about `r_home_setup()` returning an `RVersion` instead of a `PathBuf`? That is the nicest way to fold all of this together I think?

lionel-: Good idea. How about a new `RHome` struct with `version` and `path` field?

thomasp85: I'm fine with that - the use of RVersion is pretty much confined to this so it's easy to generalise it

DavisVaughan: Do you mind if i merge https://github.com/posit-dev/ark/pull/1172 first? It uses / moves around some things with `r_home_setup()`

update: merged

thomasp85: ok, with R_HOME always set we should probably just remove `r_home` from the RVersion struct and use that..?

lionel- (approved):
LG!

crates/ark/src/version.rs
lionel-: > ok, with R_HOME always set we should probably just remove r_home from the RVersion struct and use that..?

It's still useful to have it in a struct, looking up an envvar is not thread safe and should in principle only be done from the R thread (we've seen crashes with 0mq init because of this in the past). The less we look up envvars, the better.

thomasp85:
@DavisVaughan Can you give the latest iteration I quick glance? I changed the version check to read from the DESCRIPTION file of base to completely avoid any R process execution for it

crates/ark/src/version.rs
```diff
@@ -7,14 +7,17 @@
...
 use anyhow::Context;
-use harp::command::r_command;
-use harp::command::r_home_setup;
 use harp::object::RObject;
 use itertools::Itertools;
 use libr::SEXP;
+use oak_package_metadata::description::Description;
+
+pub const MIN_R_MAJOR: u32 = 4;
+pub const MIN_R_MINOR: u32 = 2;
```
DavisVaughan, line 20: You may as well throw in `MIN_R_PATCH: u32: = 0`

I think it makes the `Ark requires R >= {}.{}.0` message clearer and we may need it one day

Don't forget to update `is_supported`

DavisVaughan (approved):
Looks great just one minor comment

PR #1201, opened 2026-05-06
Reinstate handlers during Rprofile sourcing

### LINKED ISSUE #1101: `message()` output during R startup appears red in Positron console
I noticed this while working on https://github.com/posit-dev/ark/pull/1088 and playing around with `message()` in `.Rprofile`, both inside and outside of session init hooks.

`message()` calls from `.Rprofile` (and presumably the site file) render like an error, i.e. in red in the Positron console. The same `message()` call renders in the normal text color when executed interactively. All of those startup messages are emitted with `message()` too.                       

Image

My understanding of what's going on:

ark registers a `globalCallingHandlers()` message handler ([`.ps.errors.globalMessageHandler` in `errors.R`](https://github.com/posit-dev/ark/blob/90d41ea2521f502a44c1664b49ebf6d606760f1f/crates/ark/src/modules/positron/errors.R#L81-L107)) that intercepts `message()` conditions and redirects output to stdout via `cat()`. This makes "normal" `message()` output look "normal".

During startup, `.Rprofile` is sourced inside [`top_level_exec()`](https://github.com/posit-dev/ark/blob/90d41ea2521f502a44c1664b49ebf6d606760f1f/crates/ark/src/startup.rs#L71-L85) (in `startup.rs`), which resets the R handler stack. The global calling handlers are not active in that context, so `message()` falls through to R's default behavior, which is to write to stderr. And Positron styles all stderr output with `--vscode-positronConsole-errorForeground` (red).

I don't know what the right thing to do about this is.

thomasp85:
Fix #1101

This PR ensures that our error, warning, and message handlers are all in place during startup-sourcing of .Rprofile etc.

It does this by calling `initialize_errors()` prior to sourcing which re-registers our handlers. It was necessary to modify the logic somewhat to avoid very noisy 

```
pushing duplicate `error` handler on top of the stack
pushing duplicate `warning` handler on top of the stack
pushing duplicate `message` handler on top of the stack
pushing duplicate `interrupt` handler on top of the stack
```

during startup, so we now filter the existing handlers before applying our own (again). Trying to surpress this by wrapping in `suppressWarnings()` etc results in system crash.

To test this PR add this to you .Rprofile

```
message("Hello")
globalCallingHandler(test = function(...) print("test"))
```

and check that "Hello" is printed in white, not red, and that a "test" handler has been correctly registered

### DIFF
--- a/crates/ark/src/modules/positron/errors.R
+++ b/crates/ark/src/modules/positron/errors.R
@@ -129,6 +129,17 @@ globalInterruptHandler <- function(cnd) {
     format_traceback(traceback)
 }
 
+# Sources `file` with ark's global condition handlers re-registered.
+# Intended to be called from a `top_level_exec()` context (which clears the
+# handler stack), so that `globalCallingHandlers()` succeeds. This restores
+# message/error/warning rendering to match interactive behavior during
+# `.Rprofile` sourcing without putting calling-handlers on the stack that
+# would block user `globalCallingHandlers()` calls.
+.ps.errors.source_with_handlers <- function(file) {
+    initialize_errors()
+    sys.source(file, envir = globalenv())
+}
+
 # If a sink is active (either on output or on messages) messages
 # are always streamed to `stderr`. This follows rlang behaviour
 # and ensures messages can be sinked from stderr consistently.
@@ -278,10 +289,21 @@ initialize_errors <- function() {
     # Unregister all handlers and hold onto them
     handlers <- globalCallingHandlers(NULL)
 
+    existing_ps_handler <- vapply(
+        handlers,
+        function(h) {
+            isTRUE(all.equal(h, .ps.errors.globalErrorHandler)) ||
+                isTRUE(all.equal(h, .ps.errors.globalWarningHandler)) ||
+                isTRUE(all.equal(h, .ps.errors.globalMessageHandler)) ||
+                isTRUE(all.equal(h, globalInterruptHandler))
+        },
+        logical(1)
+    )
+
     # Inject our global error handler at the end.
     # This allows other existing error handlers to run ahead of us.
     handlers <- c(
-        handlers,
+        handlers[!existing_ps_handler],
         list(
             error = .ps.errors.globalErrorHandler,
             warning = .ps.errors.globalWarningHandler,
--- a/crates/ark/src/startup.rs
+++ b/crates/ark/src/startup.rs
@@ -12,7 +12,6 @@ use std::str::FromStr;
 use amalthea::socket::iopub::IOPubMessage;
 use amalthea::wire::stream::Stream;
 use amalthea::wire::stream::StreamOutput;
-use harp::environment::R_ENVS;
 use harp::exec::RFunction;
 use harp::exec::RFunctionExt;
 use libr::Rf_eval;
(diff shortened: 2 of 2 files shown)


crates/ark/src/modules/positron/errors.R
```diff
@@ -278,10 +289,21 @@ initialize_errors <- function() {
     # Unregister all handlers and hold onto them
     handlers <- globalCallingHandlers(NULL)
 
+    existing_ps_handler <- vapply(
```
lionel-, line 292: _ark_ rather than _ps_

crates/ark/src/modules/positron/errors.R
```diff
@@ -278,10 +289,21 @@ initialize_errors <- function() {
...
+        handlers,
+        function(h) {
+            isTRUE(all.equal(h, .ps.errors.globalErrorHandler)) ||
```
lionel-, line 295: What is preventing usage of `identical()` here?

crates/ark/src/modules/positron/errors.R
lionel-: `registered_ark_handlers`

"registered" is clearer to me, and plural makes it clearer this is a vector

crates/ark/src/modules/positron/errors.R
```diff
@@ -129,6 +129,17 @@ globalInterruptHandler <- function(cnd) {
     format_traceback(traceback)
 }
 
+# Sources `file` with ark's global condition handlers re-registered.
+# Intended to be called from a `top_level_exec()` context (which clears the
+# handler stack), so that `globalCallingHandlers()` succeeds. This restores
+# message/error/warning rendering to match interactive behavior during
+# `.Rprofile` sourcing without putting calling-handlers on the stack that
+# would block user `globalCallingHandlers()` calls.
+.ps.errors.source_with_handlers <- function(file) {
```
lionel-, line 138: This is not exported so shouldn't use `.ps.` prefix.

crates/ark/src/modules/positron/errors.R
```diff
@@ -129,6 +129,17 @@ globalInterruptHandler <- function(cnd) {
...
+    initialize_errors()
```
lionel-, line 139: We'll need a comment in `initialize_errors()` to mention that particular calling context.

Are we certain that no ill side effect comes from reregistering the global handlers in that top-level context? Does R restore the handler stack to the previous state? (It might very well do that).

To make things clearer I think I'd prefer not using `initialize_errors()` which in my mind should only be called once per file. Probably best to split into `initialize_global_handlers()`.

lionel- (commented):
It would be nice to add some integration tests here (e.g. look at https://github.com/posit-dev/ark/blob/main/crates/ark/tests/kernel-r-profile.rs)

crates/ark/src/modules/positron/errors.R
thomasp85: I honestly don't know, except that it doesn't return `TRUE`. I couldn't find whatever was causing it so ended up being pragmatic about it instead

crates/ark/src/modules/positron/errors.R
thomasp85: I guess I'm still a bit uncertain about the meaning of the various prefixes. Will clean up

crates/ark/src/modules/positron/errors.R
thomasp85: The function now makes sure to be idempotent so reapplying it shouldn't change a thing. This is also why I think we need to reuse the function rather than split it because we would otherwise be dependent on the order of calling. I'd be fine with renaming `initialize_errors()` but don't see any upside to splitting it into two more or less identical functions

crates/ark/src/modules/positron/errors.R
lionel-: If you get a chance can you please try `identical(as.list(zap_srcref(h)))`? This should get rid of things like srcrefs / bytecode.

Maybe the environment is different too, but I can't see why that would be the case from the top of my head.

Not that big of a deal but `all.equal()` is certainly an eyebrow raiser ;)

lionel- (commented):
One of the CI failure is transient and the other should be fixed on main.

If you need help with the integration test please feel free to come chat with me.

PR #1204, opened 2026-05-06
Add support for source() directive in symbol resolution

### LINKED ISSUE #1148: Cross-file symbol resolution

lionel-:
Closes https://github.com/posit-dev/ark/issues/1148

I somehow closed https://github.com/posit-dev/ark/pull/1157 by mistake and couldn't reopen it. See PR description there.

I've addressed your first round of review @DavisVaughan. The main change I made follows your remark about ordering with:

```r
helper <- function() {}
source("helpers.R")  # Also defines `helper`
helper # Should resolve to definition in `helpers.R`
```

We really need the imported definitions to shadow any earlier ones. Cross-file resolution only kicks in for unbound symbols, so the cross-file directive approach couldn't work here. To fix this I made the `local = FALSE` case use the same approach as with `local = TRUE`. I.e. we inject actual definitions in the symbol table, so that they can shadow earlier ones.

The cross-file resolution mechanism would still have been useful in nested scopes like:

```r
fn <- function() {
  source("helpers.R")
  helper
}
```

Here the external resolution for unbound symbols _is_ correct because `local = FALSE`. However it didn't feel justified to keep this mechanism just for this case. So I've removed it and `source()` now always injects in the local environment, as if `local = TRUE`. I've added a TODO note about linting this usage of `source()` to guide users toward using `local = TRUE`.

The other change concerns how we represent these imported definitions. You were right to call out how they were represented in the first version of my PR, as definitions copied from the imported file. Although in ty definitions are self-contained and shareable, with a scope and file ID that refer to their original context, there is a major problem to importing a definition this way: it creates a dependency that's not trackable by Salsa in a targeted way. For instance if the sourced file changes in a way that doesn't affect imports (same exported names), we'd be forced to recompute the whole index even though it's not necessary. With a proper Salsa query for imported symbols, the computation will be short-circuited if the imports haven't changed.

So I've revised the approach so it can be resolved via queries. There is a new `DefinitionKind::Import` which references a definition in an external file, and a resolving procedure that looks up the external definition. All stored definitions are local.

One subtlety is that import definitions have an empty range (at the `source()` call offset), since there is no syntactic element in the file that represents them. Thanks to the empty range we can exclude these definitions in a natural way by using `def.contains(offset)`. Their empty range can't contain the offset. So for instance when we look for an actual definition at cursor, e.g. in goto-definition, these import definitions are excluded as expected via their empty range.

### DIFF
--- a/crates/ark/src/lsp/document.rs
+++ b/crates/ark/src/lsp/document.rs
@@ -212,9 +212,10 @@ impl Document {
         Ok(self.parse.syntax())
     }
 
-    /// Recomputed every time for now, we'll track this with Salsa soon.
-    pub fn semantic_index(&self) -> SemanticIndex {
-        oak_index::semantic_index(&self.parse.tree())
+    /// TODO(salsa) Recomputed every time for now, but we'll track this with
+    /// Salsa soon.
+    pub fn semantic_index(&self, file: &url::Url) -> SemanticIndex {
+        oak_index::semantic_index(&self.parse.tree(), file)
     }
 
     pub fn tree_sitter_point_from_lsp_position(
--- a/crates/ark/src/lsp/goto_definition.rs
+++ b/crates/ark/src/lsp/goto_definition.rs
@@ -550,7 +550,7 @@ mod tests {
     #[test]
     fn test_script_source_directive_resolves() {
         // script.R has `source("helpers.R")` then uses `helper`.
-        // WorldState::file_scope() should resolve the source() directive
+        // WorldState::file_analysis() should resolve the source() directive
         // and make helpers.R's exports visible via the search path.
         let script_dir = std::env::temp_dir().join("test_script_source");
 
--- a/crates/ark/src/lsp/goto_definition.rs
+++ b/crates/ark/src/lsp/goto_definition.rs
@@ -981,4 +981,53 @@ mod tests {
             }
         );
     }
+
+    #[test]
+    fn test_script_source_in_function_packages_scoped() {
+        // source() inside a function where the sourced file calls library(dplyr).
+        // The dplyr Attach directive should be scoped to the function: visible
+        // inside f (and nested scopes) but NOT at file scope.
+        //
+        // We verify by checking the scope chain at two offsets. Package symbols
+        // produce no NavigationTarget, so goto_definition can't distinguish
+        // "resolved to package" from "unresolved". The scope chain check is
+        // more direct.
+        let dir = tempfile::tempdir().unwrap();
+
+        std::fs::write(
+            dir.path().join("helpers.R"),
+            "library(dplyr)\nhelper <- function() 1\n",
+        )
+        .unwrap();
+
+        //  "f <- function() {\n"              offset 0
+        //  "  source(\"helpers.R\")\n"         offset 18
+        //  "  mutate\n"                        offset 39
+        //  "}\n"                               offset 48
+        //  "mutate\n"                          offset 50
+        let script_source = "f <- function() {\n  source(\"helpers.R\")\n  mutate\n}\nmutate\n";
+        let script_doc = Document::new(script_source, None);
+        let script_uri = lsp_types::Url::from_file_path(dir.path().join("script.R")).unwrap();
+
+        let mut state = WorldState::default();
+        state
+            .documents
+            .insert(script_uri.clone(), script_doc.clone());
+
+        let (index, file_scope) = state.file_analysis(&script_uri, &script_doc);
+
+        let has_dplyr = |layers: &[oak_index::external::ScopeLayer]| -> bool {
+            layers.iter().any(|l| matches!(l, oak_index::external::ScopeLayer::PackageExports(pkg) if pkg == "dplyr"))
+        };
+
+        // Inside f (offset 41, on "mutate"): dplyr should be in the scope chain
+        let inner_offset = biome_rowan::TextSize::from(41);
+        let inner_chain = file_scope.at(&index, inner_offset);
+        assert!(has_dplyr(&inner_chain));
+
+        // Outside f (offset 50, on "mutate"): dplyr should NOT be in the scope chain
+        let outer_offset = biome_rowan::TextSize::from(50);
+        let outer_chain = file_scope.at(&index, outer_offset);
+        assert!(!has_dplyr(&outer_chain));
+    }
 }
--- a/crates/ark/src/lsp/goto_definition.rs
+++ b/crates/ark/src/lsp/goto_definition.rs
@@ -26,7 +26,7 @@ pub(crate) fn goto_definition(
 
     let (index, scope) = state.file_analysis(&uri, document);
     let root = document.syntax()?;
-    let targets = oak_ide::goto_definition(offset, &uri, &root, &index, &scope, &state.library);
+    let targets = oak_ide::goto_definition(state, offset, &uri, &root, &index, &scope);
 
     if targets.is_empty() {
         return Ok(None);
@@ -1161,39 +1161,39 @@ mod tests {
     }
 
     #[test]
-    fn test_script_source_local_false_does_not_shadow_local_def() {
-        // `source()` (default `local = FALSE`) in a function scope does
-        // not shadow a prior local binding.
+    fn test_script_source_shadows_local_def() {
+        // A local definition followed by source() of a file that also
+        // defines the same name: the use after source() should resolve
+        // to the sourced file's definition (later shadows earlier).
         let dir = tempfile::tempdir().unwrap();
 
-        std::fs::write(dir.path().join("helpers.R"), "foo <- function() 1\n").unwrap();
+        std::fs::write(dir.path().join("helpers.R"), "helper <- function() 1\n").unwrap();
 
-        //  Line 0: "f <- function() {\n"
-        //  Line 1: "  foo <- \"local\"\n"
-        //  Line 2: "  source(\"helpers.R\")\n"
-        //  Line 3: "  foo\n"
-        //  Line 4: "}\n"
-        let script_source =
-            "f <- function() {\n  foo <- \"local\"\n  source(\"helpers.R\")\n  foo\n}\n";
+        //  Line 0: "helper <- 'local'\n"
+        //  Line 1: "source(\"helpers.R\")\n"
+        //  Line 2: "helper\n"
+        let script_source = "helper <- 'local'\nsource(\"helpers.R\")\nhelper\n";
         let script_doc = Document::new(script_source, None);
         let script_uri = lsp_types::Url::from_file_path(dir.path().join("script.R")).unwrap();
+
+        let helpers_uri = lsp_types::Url::from_file_path(dir.path().join("helpers.R")).unwrap();
 
         let mut state = WorldState::default();
         state
             .documents
             .insert(script_uri.clone(), script_doc.clone());
 
-        // `foo` on line 3 resolves to the local definition on line 1
-        let params = make_params(script_uri.clone(), 3, 2);
+        // `helper` on line 2 should resolve to helpers.R (source() shadows local)
+        let params = make_params(script_uri, 2, 0);
         assert_matches!(
             goto_definition(&script_doc, params, &state).unwrap(),
             Some(GotoDefinitionResponse::Link(ref links)) => {
-                assert_eq!(links[0].target_uri, script_uri);
+                assert_eq!(links[0].target_uri, helpers_uri);
                 assert_eq!(
                     links[0].target_range,
                     lsp_types::Range {
-                        start: lsp_types::Position::new(1, 2),
-                        end: lsp_types::Position::new(1, 5),
+                        start: lsp_types::Position::new(0, 0),
+                        end: lsp_types::Position::new(0, 6),
                     }
                 );
             }
--- a/crates/ark/src/lsp/goto_definition.rs
+++ b/crates/ark/src/lsp/goto_definition.rs
@@ -1030,4 +1030,33 @@ mod tests {
         let outer_chain = file_scope.at(&index, outer_offset);
         assert!(!has_dplyr(&outer_chain));
     }
+
+    #[test]
+    fn test_script_source_later_shadows_earlier() {
+        // When two sourced files define the same name, the later
+        // source() call shadows the earlier one.
+        let dir = tempfile::tempdir().unwrap();
+
+        std::fs::write(dir.path().join("a.R"), "foo <- 1\n").unwrap();
+        std::fs::write(dir.path().join("b.R"), "foo <- 2\n").unwrap();
+
+        let script_doc = Document::new("source(\"a.R\")\nsource(\"b.R\")\nfoo\n", None);
+        let script_uri = lsp_types::Url::from_file_path(dir.path().join("script.R")).unwrap();
+
+        let mut state = WorldState::default();
+        state
+            .documents
+            .insert(script_uri.clone(), script_doc.clone());
+
+        let b_uri = lsp_types::Url::from_file_path(dir.path().join("b.R")).unwrap();
+
+        // `foo` (line 2) should resolve to b.R (later source shadows earlier)
+        let params = make_params(script_uri, 2, 0);
+        assert_matches!(
+            goto_definition(&script_doc, params, &state).unwrap(),
+            Some(GotoDefinitionResponse::Link(ref links)) => {
+                assert_eq!(links[0].target_uri, b_uri);
+            }
+        );
+    }
 }
--- a/crates/ark/src/lsp/goto_definition.rs
+++ b/crates/ark/src/lsp/goto_definition.rs
@@ -1040,8 +1040,8 @@ mod tests {
 
         let (index, file_scope) = state.file_analysis(&script_uri, &script_doc);
 
-        let has_dplyr = |layers: &[oak_index::external::ScopeLayer]| -> bool {
-            layers.iter().any(|l| matches!(l, oak_index::external::ScopeLayer::PackageExports(pkg) if pkg == "dplyr"))
+        let has_dplyr = |layers: &[oak_index::scope_layer::ScopeLayer]| -> bool {
+            layers.iter().any(|l| matches!(l, oak_index::scope_layer::ScopeLayer::PackageExports(pkg) if pkg == "dplyr"))
         };
 
         // Before f (offset 0, on "mutate"): dplyr is NOT visible because the
--- a/crates/ark/src/lsp/goto_definition.rs
+++ b/crates/ark/src/lsp/goto_definition.rs
@@ -793,11 +793,8 @@ mod tests {
 
     #[test]
     fn test_script_source_in_function_scoping() {
-        // source() inside a function body: definitions are visible inside
-        // the function and its nested scopes, but NOT outside.
-        // This is independent of `local = TRUE/FALSE`. Even if sourced in the
-        // global environment, only code following the `source()` call in the
-        // current scope can assume these symbols are in scope.
+        // `source(local = FALSE)` inside a function scopes directives to the
+        // function scope, so definitions are NOT visible at file scope.
         let dir = tempfile::tempdir().unwrap();
 
         std::fs::write(dir.path().join("helpers.R"), "helper <- function() 1\n").unwrap();
@@ -838,7 +835,7 @@ mod tests {
             }
         );
 
-        // `helper` on line 5 (outside f) — should NOT resolve
+        // `helper` on line 5 (outside f) — NOT visible
         let params = make_params(script_uri, 5, 0);
         let result = goto_definition(&script_doc, params, &state).unwrap();
         assert_eq!(result, None);
@@ -984,14 +981,8 @@ mod tests {
 
     #[test]
     fn test_script_source_in_function_packages_scoped() {
-        // source() inside a function where the sourced file calls library(dplyr).
-        // The dplyr Attach directive should be scoped to the function: visible
-        // inside f (and nested scopes) but NOT at file scope.
-        //
-        // We verify by checking the scope chain at two offsets. Package symbols
-        // produce no NavigationTarget, so goto_definition can't distinguish
-        // "resolved to package" from "unresolved". The scope chain check is
-        // more direct.
+        // `source(local = FALSE)` inside a function scopes package directives
+        // to the function scope, so they are NOT visible at file scope.
         let dir = tempfile::tempdir().unwrap();
 
         std::fs::write(
@@ -1000,12 +991,14 @@ mod tests {
         )
         .unwrap();
 
-        //  "f <- function() {\n"              offset 0
-        //  "  source(\"helpers.R\")\n"         offset 18
-        //  "  mutate\n"                        offset 39
-        //  "}\n"                               offset 48
-        //  "mutate\n"                          offset 50
-        let script_source = "f <- function() {\n  source(\"helpers.R\")\n  mutate\n}\nmutate\n";
+        //  "mutate\n"                          offset 0
+        //  "f <- function() {\n"              offset 7
+        //  "  source(\"helpers.R\")\n"         offset 25
+        //  "  mutate\n"                        offset 46
+        //  "}\n"                               offset 55
+        //  "mutate\n"                          offset 57
+        let script_source =
+            "mutate\nf <- function() {\n  source(\"helpers.R\")\n  mutate\n}\nmutate\n";
         let script_doc = Document::new(script_source, None);
         let script_uri = lsp_types::Url::from_file_path(dir.path().join("script.R")).unwrap();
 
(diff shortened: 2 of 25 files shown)


crates/oak_index/src/semantic_index.rs
```diff
@@ -461,8 +498,13 @@ impl SymbolFlags {
 // declarations.
 #[derive(Debug)]
 pub struct Definition {
-    pub(crate) symbol: SymbolId,
     pub(crate) kind: DefinitionKind,
+    /// The file that owns this definition's index.
+    pub(crate) file: Url,
```
DavisVaughan, line 497: I believe you need to update all the docs here

```
// Our definitions don't carry file or scope because they live inside the
// `SemanticIndex` at a known position (`definitions[scope_id][def_id]`).
// In ty, `Definition<'db>` is a self-contained salsa tracked struct that
// carries file + scope + place, because it gets passed around independently
// to type inference queries and cross-file lookups. We'll add those fields
// when salsa is introduced.
```

Also, I assume `file` will eventually become a cheaper salsa tracked FileId?

Having a URL per Definition is "relatively" expensive, I think

crates/oak_index/src/semantic_index.rs
```diff
@@ -526,9 +583,9 @@ impl Ranged for Use {
 pub struct Directive {
     pub(crate) kind: DirectiveKind,
     pub(crate) offset: TextSize,
+    pub(crate) scope: ScopeId,
```
DavisVaughan, line 580: IIUC, this comment is now wrong since you have a scope `/// A file-level directive`

crates/oak_index/src/semantic_index.rs
```diff
@@ -295,7 +332,7 @@ impl SymbolTable {
         self.by_name.get(name).copied()
     }
 
-    pub fn symbol(&self, id: SymbolId) -> &Symbol {
+    pub fn symbol_id(&self, id: SymbolId) -> &Symbol {
```
DavisVaughan: I'm fairly confident this is still just `symbol()`?

You're requesting the `Symbol` as output, not the `SymbolId`

crates/oak_index/src/external.rs
```diff
@@ -80,10 +80,10 @@ pub fn resolve_external_name(
 
 /// Resolve a name in a specific package's exported symbols.
 pub fn resolve_in_package(
-    name: &str,
+    library: &Library,
     package: &str,
+    name: &str,
     visibility: PackageDefinitionVisibility,
```
DavisVaughan, line 86: your AI and I are fighting because I moved `name` to be first because you're resolving `name`, that's the most specific argument, so to me it should be first

like, the returned `ExternalDefinition` is very specific to `name`, so having `name` first makes the most sense to me (here and elsewhere in this file)

crates/oak_index/src/semantic_index.rs
```diff
@@ -115,14 +116,27 @@ impl SemanticIndex {
     }
 
     /// Top-level definitions exported by this file (definitions in the file scope).
-    pub fn file_exports(&self) -> Vec<(&str, TextRange)> {
+    /// Includes `Import`-kind forwarding definitions from `source()` calls.
+    /// Last definition of each name wins (later assignments shadow earlier ones).
+    pub fn file_exports(&self) -> FxHashMap<&str, TextRange> {
```
DavisVaughan, line 121: I think previously the Vec was ordered by TextRange, i.e. definitions earlier in the Vec came earlier in the file

I guess that doesn't really matter in practice for what we use it for? It's kind of hard to maintain both that property and uniqueness

Like, now if you add

```
defA(range 0)
defB(range 1)
defA(range 3)
```

Then you probably end up with a hash map with an iterable order of `defA(range 3), defB(range 1)` (although you don't really have any ordering guarantees with this hash map anyways)

crates/oak_index/src/semantic_index.rs
```diff
@@ -147,6 +161,29 @@ impl SemanticIndex {
...
+
+    /// Find the use site at `offset`, if any.
+    pub fn use_at_offset(&self, offset: TextSize) -> Option<(ScopeId, UseId)> {
+        let (scope, _) = self.scope_at(offset);
+        let use_id = self
+            .uses(scope)
+            .iter()
+            .find_map(|(id, u)| u.range().contains(offset).then_some(id));
+        Some((scope, use_id?))
+    }
```
DavisVaughan: These aren't used at all

And if i remember right, these are what I requested we remove because the actual `Definition` and `Use` should be returned to avoid multiple lookups

crates/oak_index/src/scope_layer.rs
```diff
@@ -28,14 +28,17 @@ pub enum ScopeLayer {
 /// Compute the scope layers that a single file contributes to the
 /// scope chain: one `FileExports` layer from its top-level definitions, plus
 /// one `PackageExports` layer per `library()`/`require()` directive.
+///
+/// Offsets are discarded since all of a predecessor file's layers are
+/// unconditionally visible.
```
DavisVaughan: i dont see this happening anywhere

crates/oak_index/src/builder.rs
```diff
@@ -41,18 +43,51 @@ use crate::semantic_index::UseId;
 use crate::use_def_map::UseDefMapBuilder;
 
 /// Build a [`SemanticIndex`] from a parsed R file.
-pub fn semantic_index(root: &RRoot) -> SemanticIndex {
+pub fn semantic_index(root: &RRoot, file: &Url) -> SemanticIndex {
     let range = root.syntax().text_trimmed_range();
-    let mut builder = SemanticIndexBuilder::new(range);
+    let mut builder = SemanticIndexBuilder::new(range, file.clone(), None);
```
DavisVaughan, line 50: i think you should just take `file: Url` to avoid a double clone when the caller already owns `file` and can feed it through

eventually it will probably be `file: FileId` anyways, I presume

crates/oak_index/src/builder.rs
```diff
@@ -41,18 +43,51 @@ use crate::semantic_index::UseId;
...
 }
 
+/// Build a [`SemanticIndex`] with cross-file `source()` resolution.
+///
+/// The resolver callback is called when the builder encounters a
+/// `source("path")` call. It should return the sourced file's exported
+/// names and any `library()` package attachments.
+pub fn semantic_index_with_source_resolver<'a>(
+    root: &RRoot,
+    file: &Url,
```
DavisVaughan, line 63: same here about `file: Url`

crates/oak_index/src/builder.rs
```diff
@@ -663,18 +712,28 @@ impl SemanticIndexBuilder {
...
+    }
 
+    // ## `library()` / `require()` scoping
+    //
+    // In R, `library()` always modifies the global search path regardless
+    // of where it's called. Statically, we scope the directive to
+    // `self.current_scope`: at file scope it's visible everywhere (sequential
+    // execution is guaranteed), but inside a function it's only visible
+    // within that function and its children, since the function might never
+    // be called. Same reasoning as `source(local = FALSE)` directives.
```
DavisVaughan:
```suggestion
    // be called. Same reasoning as `source()` directives.
```

? i.e. is `local` relevant anymore?

crates/oak_index/src/builder.rs
```diff
@@ -703,12 +761,115 @@ impl SemanticIndexBuilder {
...
+            self.add_definition(
+                &name,
+                SymbolFlags::IS_BOUND,
+                DefinitionKind::Import {
+                    call: call.syntax().clone(),
+                    file: file.clone(),
+                    name: name.clone(),
+                },
+                range,
+            );
```
DavisVaughan, line 858: I actually consider this `range` behavior to be correct no matter what

For _this file's_ semantic index, the definition comes from the call site of `source()`. I mostly really like that `range` now refers to a location in _this file_, rather than in the sourced file (which was super confusing!)

And then goto definition can do the chasing as needed, which feels totally fine and correct to me (because not everyone needs to chase, probably)

crates/ark/src/lsp/document.rs
```diff
@@ -212,9 +212,10 @@ impl Document {
         Ok(self.parse.syntax())
     }
 
-    /// Recomputed every time for now, we'll track this with Salsa soon.
-    pub fn semantic_index(&self) -> SemanticIndex {
-        oak_index::semantic_index(&self.parse.tree())
+    /// TODO(salsa) Recomputed every time for now, but we'll track this with
+    /// Salsa soon.
+    pub fn semantic_index(&self, file: &url::Url) -> SemanticIndex {
```
DavisVaughan, line 217: `file: Url`

crates/ark/src/lsp/state.rs
```diff
@@ -98,16 +123,65 @@ impl WorldState {
...
+
+    /// Resolve from the workspace root, falling back to the file's directory.
+    /// This matches RStudio behaviour. Potential improvements: support
+    /// `setwd()` calls and find the workspace containing the file if there are
+    /// multiple.
+    fn source_root(&self, file: &Url) -> Option<PathBuf> {
+        self.workspace
+            .folders
+            .first()
+            .and_then(|url| url.to_file_path().ok())
```
DavisVaughan, line 149: Shouldn't you check that `file` is under the workspace folder?

i.e. `~/Desktop/foo.R` should not be under my workspace folder

And if you start doing that check, you can probably iterate through all workspace folders rather than just looking at the first

crates/ark/src/lsp/state.rs
```diff
@@ -98,16 +123,65 @@ impl WorldState {
...
+    /// For packages, cross-file visibility comes from NAMESPACE imports and
+    /// collation ordering.
+    pub(crate) fn file_analysis(
+        &self,
+        file: &Url,
+        doc: &Document,
+    ) -> (SemanticIndex, ExternalScope) {
+        match self.root {
+            Some(SourceRoot::Package(ref pkg)) => self.package_file_analysis(file, doc, pkg),
+            _ => self.script_file_analysis(file, doc),
```
DavisVaughan, line 170: Personally I'd prefer to see the `None` here so the compiler can catch when we add more cases to `SourceRoot`

crates/ark/src/lsp/state.rs
```diff
@@ -149,19 +223,11 @@ impl WorldState {
...
-                let Some(contents) = std::fs::read_to_string(path).log_err() else {
-                    continue;
-                };
-                &Document::new(&contents, None)
+            let Some(doc) = self.workspace_document(&uri) else {
+                continue;
             };
 
-            let layers = file_layers(uri, &doc.semantic_index());
+            let layers = file_layers(uri.clone(), &doc.semantic_index(&uri));
```
DavisVaughan, line 230: Since `SemanticIndex`'s `Definition`s now know about the `file` they were defined in, would it make sense for the `SemanticIndex` struct to hold the `file` it was was created on?

To me that makes sense, because then you can produce `file_layers()` without needing the extra `uri` input, as its already part of the `SemanticIndex`.

(Technically right now you could screw it up by providing a `uri` here that doesn't match the file the `semantic_index()` was built on, and this would also remove that admittedly rare footgun)

crates/ark/src/lsp/state.rs
```diff
@@ -98,16 +123,65 @@ impl WorldState {
...
+    ///
+    /// For scripts, the index is built with a source resolver so that
+    /// `source()` directives carry the sourced file's exports.
+    /// For packages, cross-file visibility comes from NAMESPACE imports and
+    /// collation ordering.
+    pub(crate) fn file_analysis(
+        &self,
+        file: &Url,
+        doc: &Document,
+    ) -> (SemanticIndex, ExternalScope) {
```
DavisVaughan, line 167: Would it make sense for there to be `file_semantic_index()` and `file_external_scope()` for salsa granularity purposes? i.e. if the external scope changes but the semantic index doesn't, is there potential to avoid recomputation?

DavisVaughan: Maybe that's hard because you need one to compute the other

crates/ark/src/lsp/state.rs
```diff
@@ -171,43 +237,83 @@ impl WorldState {
...
+    fn resolve_source(
+        &self,
+        base_dir: &Path,
+        path: &str,
+        stack: &mut HashSet<Url>,
+    ) -> Option<SourceResolution> {
+        let resolved = base_dir.join(path);
+        let url = Url::from_file_path(&resolved).log_err()?;
+
+        if !stack.insert(url.clone()) {
```
DavisVaughan, line 280: Part of me wonders if we should add a safety wrapper of like `with_stack()` or something that handles the push/pop of the stack

Like, after this `insert()` succeeds you MUST pop it from the stack before you exit this function

But it would be _so easy_ to do a `.log_err()?` call somewhere down below, which would completely mess you up. Anything to make that harder to do accidentally would be nice

crates/ark/src/lsp/goto_definition.rs
```diff
@@ -551,4 +544,698 @@ mod tests {
...
+
+        let script_doc = Document::new("source(\"helpers.R\")\nmutate\nhelper\n", None);
+        let script_uri = lsp_types::Url::from_file_path(script_dir.join("script.R")).unwrap();
+
+        let mut state = make_state(&script_uri, &script_doc);
+        state.library = library;
+        state.documents.insert(helpers_uri.clone(), helpers_doc);
+
+        // `mutate` (line 1) resolves via dplyr, attached by helpers.R's library() call.
+        // Package symbol, no NavigationTarget.
```
DavisVaughan, line 641: technically you can fake it now, but i guess that's not really the point of the test!

crates/oak_ide/src/identifier.rs
```diff
@@ -37,7 +37,9 @@ impl Identifier {
     pub fn classify(root: &RSyntaxNode, index: &SemanticIndex, offset: TextSize) -> Option<Self> {
         let (scope_id, _) = index.scope_at(offset);
 
-        if let Some((def_id, _)) = index.definitions(scope_id).contains(offset) {
+        // Definitions with empty ranges (e.g. imports) are naturally excluded
+        // here since they can't contain the offset
+        if let Some((def_id, _def)) = index.definitions(scope_id).contains(offset) {
             return Some(Identifier::Definition { scope_id, def_id });
```
DavisVaughan: So you can't goto definition on an import that comes from a `source()` call? Like if `foo <- function() {}` was in a file that I sourced via `source("helper.R")` into `file.R`, then I can't goto definition on a call to `foo()` in `file.R`?

DavisVaughan: Hmm it seems like `resolve_import()` over in goto_definition is supposed to make this work, so what am I missing?

crates/oak_ide/src/goto_definition.rs
```diff
@@ -143,24 +146,71 @@ fn resolve_use(
...
+
+    // Imports always target top-level definitions in the sourced file.
+    let file_scope = ScopeId::from(0);
+    let symbols = &target_index.symbols(file_scope);
+
+    let (_def_id, def) = target_index
+        .definitions(file_scope)
+        .iter()
+        .filter(|(_id, def)| symbols.symbol_id(def.symbol()).name() == name)
+        .last()?;
```
DavisVaughan, line 179: Should there only ever be exactly 1 definition now? Since the builder itself throws out all but the last one already?

DavisVaughan (approved):
I think I'm a little confused about how `Identifier::classify` works with Import definitions, because it seems like we are throwing them out? So how do we ever get to `resolve_use()`? This is probably a me problem, but I would appreciate an explanation about what I'm missing

crates/oak_index/src/external.rs
lionel-: hmm it's probably me rather than the AI.
library → package → name → visibility is a logical progression from general to specific

crates/oak_index/src/builder.rs
lionel-: I think the point here is that in python imports all have a declaring symbol binding, to which a text range can be attached. With `source()` they appear out of nowhere.

crates/oak_index/src/semantic_index.rs
lionel-: This will become a `File`, which is a whole interned Salsa struct rather than `FileId`, but as cheap to copy around yes.

crates/oak_index/src/semantic_index.rs
lionel-: yes! I was confused

crates/oak_index/src/external.rs
lionel-: The other way to look at this is that this could be a method on `Library` but not on `name: &str`.

crates/oak_index/src/semantic_index.rs
lionel-: Right. For file exports, the order (or age) of definitions doesn't matter, you see them in bulk.

crates/oak_index/src/semantic_index.rs
lionel-: I'm using `use_at_offset()` in a later PR.

I'm not worried about the array lookups, these are basically free. But we could return `&Use` and `&Definition` alongside the ids for ergonomics.

Looking at the API again, it was a bit inconsistent though:

- `scope_at()` but these two have `_offset` suffixes. There shouldn't be any suffix, I'm introducing `resolve_at()` and `exports_at()` in a future PR.
- `scope_at()` returns `&Scope` alongside the `ScopeId`, which reinforces that we should return refs from the other two.

I've made both changes.

crates/oak_index/src/scope_layer.rs
lionel-: I think this is referring to `PackageExports` not carrying ranges/offsets like `FileExports`.

Agree this comment is unclear, and not very useful, I removed it.

crates/oak_index/src/builder.rs
lionel-: I take note of your comment (good practice to clone at call site) but I'll avoid any diff churn here because these functions will go away. We should avoid Salsa dependencies on coarse objects like semantic indexes, in favour of narrower Salsa queries (file exports etc). I've added a `TODO(salsa)` to make sure we don't forget.

crates/ark/src/lsp/state.rs
lionel-: This whole path is replaced in a future PR by `File::workspace_root()` with multi-workspace support. Files outside a workspace fall back to their parent dir.

crates/ark/src/lsp/state.rs
lionel-: Good point. But this code is going away soon and I believe we mostly use `if let Some(pkg) = file.package(db)` rather than `match`.

crates/ark/src/lsp/state.rs
lionel-: This works completely differently in subsequent branches, where you now access the layers via the `File::imports()` method. We no longer need to pass URLs around in most cases.

crates/ark/src/lsp/state.rs
lionel-: We don't want to expose the whole semantic index publicly via `File`, precisely for granularity.

But we will provide narrower / more granular methods like `File::imports()` ("imports" is the new name for "external scope").

crates/ark/src/lsp/state.rs
lionel-: Good point, but cycle detection is entirely handled via Salsa in subsequent PRs, so this code goes away.

crates/oak_ide/src/identifier.rs
lionel-: Yep you can goto-def on imported symbols.

The mechanism here is a bit subtle. Two things:

- This is about the case where the cursor is on a definition, not a use.

- And this is about the forwarding definition with kind `DefinitionKind::Import` that are created in the scope. Unlike in python where all imports have a syntactic binding that's physically present in the file, these imports are created out of nowhere. So we set the ranges of these import definitions at the `source()` call site (which is useful to order effects) with empty ranges (so they can't be matched with `contains()`).

So the consideration here is what happens when cursor in on `source()` and there's a bunch of import definitions registered there.

crates/oak_ide/src/goto_definition.rs
lionel-: The builder doesn't throw away these definitions, only the `file_exports()` view filters them out.

crates/oak_ide/src/identifier.rs
lionel-: I've updated the comment:

```rs
        // `Import` definitions have empty ranges (no physical text position,
        // since `source()` injects them) so `contains()` skips them. If the
        // cursor is on the `source` symbol, the offset classifies instead as a
        // use of `source` via the check below.
```

lionel-:
> I think I'm a little confused about how Identifier::classify works with Import definitions, because it seems like we are throwing them out? So how do we ever get to resolve_use()? This is probably a me problem, but I would appreciate an explanation about what I'm missing

Classification only throws out import definitions, not their uses. See https://github.com/posit-dev/ark/pull/1204#discussion_r3225936591 for more context.

So if cursor is on `source()` (the anchor of "synthetic" import definitions, as encoded by empty ranges at the start of `source()`), we filter them out. But if cursor is on a use-site of an imported symbol after the `source()`, we are able to look up the import definition from the use and follow the chain.

Does that make more sense?

crates/oak_index/src/semantic_index.rs
lionel-: Not really wrong (directives are accessed at the file level for resolution via imports), but rather unclear. I made it:

```
/// A directive that affects the file's imports (e.g. `library()` calls).
```

This disappears in a later PR though, we no longer need a discriminated directive type since `source()` no longer produces directives. It actually should have been removed in this PR.

crates/oak_index/src/external.rs
DavisVaughan: I view it more about trying to tightly connect inputs and outputs

- First input is `name`
- Tightly connects to output of `ExternalDefinition` for that `name`

When `name` is the 3rd or 4th argument down, I kind of lose the plot a bit.

I can live with whatever you want though

crates/ark/src/lsp/state.rs
DavisVaughan: Okay good to hear

crates/ark/src/lsp/state.rs
DavisVaughan: Yea that works just as well for me

crates/ark/src/lsp/state.rs
DavisVaughan: That sounds right i think. Or maybe `File::index().imports()` so imports are owned by the semantic index still, idk

DavisVaughan: > We don't want to expose the whole semantic index publicly via File for granularity 

oh, from a later comment

crates/ark/src/lsp/state.rs
DavisVaughan: It's baked in for us?

crates/oak_ide/src/identifier.rs
DavisVaughan: > So the consideration here is what happens when cursor in on source() and there's a bunch of import definitions registered there.

Oh! Yes, wasn't clear but the comment helps a lot

crates/ark/src/lsp/state.rs
lionel-: Almost, we just need to implement two methods that tell Salsa how to recover from a cycle (`recover_from_cycle`) and how to provide an initial fixpoint value (`cycle_initial`).

PR #1208, opened 2026-05-12
Fix namespace generation in `View()` hook

lionel-:
Follow up to https://github.com/posit-dev/ark/pull/1179

The positron-r extension test for `View()` is failing because the top-level check in the hook now incorrectly returns `FALSE`, due to the new hook wrapper. This prevented `View()` from generating virtual namespaces for package functions.


lionel-:
Merging to fix frontend CI and unblock other PRs

PR #1209, opened 2026-05-13
Allow MSRV to drift from rust-toolchain.toml

lionel-:
This walks back the Rust version bump in our Cargo.toml that I did in https://github.com/posit-dev/ark/pull/1207

For Rust version updates, dependabot only bumps rust-toolchain.toml, which is the specified toolchain for developers and CI. It doesn't bump the Rust requirement in Cargo.toml, which is user oriented. I thought we should keep them in lockstep and added in #1206 a script and CI job that checks that the versions match.

I now realise this was a bad idea, because the two files serve different audiences and having these versions drift apart is a feature, not a bug:

- `rust-toolchain.toml` is developer-facing, just like `Cargo.lock`. It pins devs and CI to the same version, and keeping it fresh lets us absorb new compiler and clippy requirements as they roll in, rather than all at once in one big update.

- `Cargo.toml`'s `rust-version` is user-facing. It's the minimum Rust version we promise to users. Bumping it forces a new Rust requirement on them, even when not needed.

MSRV stability isn't as relevant for applications as for libraries, but I think it's a good practice to take in our Rust repos. And Ark does include some potentially user-facing libraries like Amalthea and libR.

Relatedly, although somewhat orthogonally, having bumped the Cargo.toml version broke the frontend CI in https://github.com/posit-dev/positron/pull/13513. I think that's a bug in our CI images that I've fixed in https://github.com/posit-dev/positron-ci-images/pull/30. In the mean time though, walking back the Cargo.toml bump will fix my Positron PR.

Changes in this PR:

- Wack back user-facing MSRV to 1.94
- Change the check script to only fail in the Rust version in Cargo.toml is ahread of the rust-toolchain.toml one.
- Add a build in our CI to check that the project compiles with the version of Rust encoded in Cargo.toml.

The new CI job ensures that we don't inadvertently use newer features that won't compile with our declared minimal version. That has happen in the past and is confusing for developers. We should still feel free to bump `rust-version` deliberately when we need newer features. This CI job just ensures we do it on purpose rather than by accident.


lionel-:
Merging now so I can fix my frontend-side bump, but happy to discuss more!

PR #1211, opened 2026-05-13
Use `AstPtr` in `DefinitionKind`

lionel-:
Branched from #1210

Replaces the syntax node in `DefinitionSync` by an AST pointer to gain Send/Sync/PartialEq, which are needed for Salsa caching.

The pointer is just an offset + an AST type so we can recover the node from the original parse tree, or even a fresh parse tree. For context, it's common in Salsa-based LSPs to throw away entire parse trees until they are needed again, to spare memory usage.

### DIFF
--- a/crates/oak_semantic/src/semantic_index.rs
+++ b/crates/oak_semantic/src/semantic_index.rs
@@ -1,6 +1,10 @@
 use std::ops::Range;
 
-use aether_syntax::RSyntaxNode;
+use aether_syntax::RBinaryExpression;
+use aether_syntax::RCall;
+use aether_syntax::RForStatement;
+use aether_syntax::RParameter;
+use biome_rowan::AstPtr;
 use biome_rowan::TextRange;
 use biome_rowan::TextSize;
 use oak_core::range::Ranged;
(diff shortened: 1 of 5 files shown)


crates/oak_semantic/src/semantic_index.rs
```diff
@@ -36,7 +40,7 @@ define_index!(EnclosingSnapshotId);
 // (all indexed by `ScopeId`) rather than bundled into a single struct, so
 // that each can be cached and invalidated independently (when salsa is
 // introduced).
-#[derive(Debug)]
+#[derive(Debug, PartialEq, Eq)]
 pub struct SemanticIndex {
```
DavisVaughan: I understand the Send/Sync bits, but not the Eq/PartialEq bits

I don't see that ty's `SemanticIndex` is `Eq/PartialEq`

And I can't imagine that comparing two `SemanticIndex`s would be a cheap operation. When do you need this and what am I missing here?

In my head this smells quite suspicious right now, so I'd like to talk about it or at least fully understand what makes us different from ty here

DavisVaughan (approved):
A fairly important question below, but otherwise fine

crates/oak_semantic/src/semantic_index.rs
lionel-: I thought this was needed transitionally while we have coarse deps on the index, though I guess not.

Issue #1212, opened 2026-05-13
Caching and incremental computation with Salsa


lionel-:
This issue tracks the transition to a Salsa database and associated queries in the Oak LSP.

As we're adding semantic analysis for symbols (origin, type, and NSE behaviour), we need to compute increasingly more derived properties. And these properties transitively depend on properties in other files. For instance:

- The LSP sees a `source("foo.R", local = T)` call, which injects top-level exports into the symbol table of the calling environment. What are the top-level exports in that file? Requires parsing, building a semantic index to figure out scope nesting, and detect all top-level assignments.

- I'm looking at `test_that(...)`, is that an NSE function? Where is it defined? From a package? Does it have registered NSE properties? (In the future this will be: is it a function? does it have declarations?). Every single function call potentially changes the symbol tables, and whether they do depend on properties potentially defined outside the current file.

- Library paths and the contents of installed packages will also be LSP inputs. We'll track when a package is reinstalled, when renv is added to the repo, etc. These events all change the state of the world.

The LSP will produce a ton of these queries, and they all potentially get invalidated at _each keystroke_ when the user types in the frontend. These changes can produce a big blast across the world state (e.g. is it the ggplot or the ggplot2 package in that `library()` call), a small blast (the user just added a local variable), or no blast at all (a string or comment is being edited, the file was reformatted with Air, etc).

With Salsa, we're able to finely specify a dependency graph of computation (queries). Salsa caches each result and only recompute ones that are invalidated. That's the framework that makes Rust-Analyzer and ty perform well in an IDE environment.

lionel-:
- [x] Rename `UrlId` to `CanonicalUrl`, `as_url()` to `inner()`, `from_canonical()` to `from_url_unchecked()`

- [x] Take canonical URL in semantic index

- [ ] Split `Root` into enum:
  - `WorkspaceRoot { packages, scripts }`
  - `OrphanRoot { scripts }`
  - `LibraryRoot { packages }`


- [x] Make `File::contents` a query (https://github.com/posit-dev/ark/pull/1271)

- [ ] `name_range()` versus `Definition::range()`

- [ ] Align naming in semantic index: https://github.com/posit-dev/ark/pull/1231#discussion_r3312181531

- [x] Remove `encoding` from `ArkFile` if we can't delete `ArkFile` entirely

- [ ] Possibly: https://github.com/posit-dev/ark/pull/1257#discussion_r3419357290

lionel-:
[...]

Issue #1183, opened 2026-04-30
Moving towards a source graph


lionel-:
Rust-Analyzer has a notion of crate graph that is the main source of inputs for the whole LSP:

Building the crate graph
https://github.com/rust-lang/rust-analyzer/blob/210340894905b8f6e94aaba2084b03cbbbde4cd3/crates/base-db/src/input.rs#L128

Registering a workspace to the crate graph (workspace crates merge into this graph)
https://github.com/rust-lang/rust-analyzer/blob/b1f58e7f05642269c54e49dd114afd4cf1fcbc92/crates/project-model/src/workspace.rs#L944

I think a similar notion for packages is very relevant for Oak. Some thoughts:

- The LSP protocol allows multiple workspaces. This is becoming increasingly important as LLMs make it much easier to work on multiple projects of a vertically integrated stack. Devtools does not support the multi-package workflow very well but LLMs are perfectly able to work around this workflow gap.

- Workspaces do not have a clear form. They might or might not contain packages. Packages might be in a subfolder (e.g. multi-lingual monorepos). Scripts might live at the root, or in subfolders.

- A workspace-package might have a soft dependency (e.g. vctrs in tests of rlang) that has a hard-dependency on it (e.g. vctrs imports rlang)

I think this implies:

- We create a unique package graph. Scripts and package files in R/ folders consume the graph via imports (`library()` calls or `NAMESPACE` directives).

- Packages are detected anywhere in a workspace, no matter which workspace if we have several. These workspace-packages become source of authority in the graph. We never look at installed/cached files for those. For dependencies not found in any workspace, we fall back to installed packages via `.libPaths()` (via our source discovery mechanism).

This is a longer term and more general way of solving https://github.com/posit-dev/ark/issues/1168.

PR #1213, opened 2026-05-13
Introduce `oak_db` and `File`

### LINKED ISSUE #1141: Implement the semantic index

lionel-:
Branched from #1211 

Closes https://github.com/posit-dev/ark/issues/1141

This PR introduces

- The `oak_db` crate which hosts a Salsa database containing all LSP inputs.
- The `File` data structure that represents an R file (similar to `Document` in the legacy LSP). It holds the parse tree and semantic index for a file.
- Dependency on `salsa` (latest version).
- The parse tree is now a Salsa query on file contents, and the semantic index is a query on the parse tree.

The cache holds at most 128 parse trees at any time (following Rust-Analyser) to spare memory.

Higher level consumers should depend on narrow Salsa queries, to reduce the risk of being invalidated by unrelated edits. For this reason, although `File` contains the parse tree and semantic index, they are kept `pub(crate)` so they are not reachable from outside `oak_db` (e.g. from `oak_ide`).

For context we're working our way up towards:

- A file tree starting from roots (workspace roots and lib path folders) and indexing all R files wrapped up in `File`
- A Source Graph (#1183) relating scripts, workspace packages, and installed packages. The nodes point to `File` handles from the file tree.

### DIFF
--- a/crates/oak_db/src/db.rs
+++ b/crates/oak_db/src/db.rs
@@ -1,0 +1,6 @@
+/// Salsa Database trait.
+///
+/// Queries take a `dyn Db` rather than the concrete database owned by
+/// the LSP layer.
+#[salsa::db]
+pub trait Db: salsa::Database {}
--- a/crates/oak_db/src/file.rs
+++ b/crates/oak_db/src/file.rs
@@ -51,7 +51,7 @@ impl File {
     /// The full index re-runs typically on every edit (e.g. the `AstPtr` ranges
     /// inside `Definition`s shift, so it rarely backdates).
     #[cfg(not(test))]
-    #[salsa::tracked(returns(ref))]
+    #[salsa::tracked(returns(ref), no_eq)]
     fn semantic_index(self, db: &dyn Db) -> SemanticIndex {
         build_semantic_index(self, db)
     }
@@ -59,7 +59,7 @@ impl File {
     /// Tests use the `pub(crate)` variant gated behind `cfg(test)` so they can
     /// call into `semantic_index` directly to verify salsa caching behaviour.
     #[cfg(test)]
-    #[salsa::tracked(returns(ref))]
+    #[salsa::tracked(returns(ref), no_eq)]
     pub(crate) fn semantic_index(self, db: &dyn Db) -> SemanticIndex {
         build_semantic_index(self, db)
     }
--- a/crates/oak_db/src/file.rs
+++ b/crates/oak_db/src/file.rs
@@ -1,0 +1,45 @@
+use oak_semantic::semantic_index::SemanticIndex;
+use url::Url;
+
+use crate::parse::OakParse;
+use crate::Db;
+
+/// A source file tracked by Salsa.
+///
+/// Content is pushed into Salsa by the LSP layer, the database never does I/O.
+/// This matches rust-analyzer's push model and avoids tying parsing to
+/// disk/network I/O inside a Salsa query.
+#[salsa::input]
+pub struct File {
+    #[returns(ref)]
+    pub url: Url,
+    #[returns(ref)]
+    pub contents: String,
+}
+
+#[salsa::tracked]
+impl File {
+    /// Parse this file's contents into an R syntax tree.
+    ///
+    /// Crate-internal: kept `pub(crate)` so downstream crates can't take a
+    /// dependency on a full parse tree. They reach the tree only through
+    /// narrower public queries.
+    #[salsa::tracked(returns(ref))]
+    pub(crate) fn parse(self, db: &dyn Db) -> OakParse {
+        OakParse::new(aether_parser::parse(
+            self.contents(db),
+            aether_parser::RParserOptions::default(),
+        ))
+    }
+
+    /// Build this file's `SemanticIndex` from the parse tree.
+    ///
+    /// Crate-internal: cross-file consumers must use narrower public queries
+    /// (`exports`, `external_scope`) to avoid depending on the full index,
+    /// which would invalidate on every internal edit.
+    #[salsa::tracked(returns(ref))]
+    pub(crate) fn semantic_index(self, db: &dyn Db) -> SemanticIndex {
+        let parsed = self.parse(db);
+        oak_semantic::semantic_index(&parsed.tree(), self.url(db))
+    }
+}
--- a/crates/oak_db/src/file.rs
+++ b/crates/oak_db/src/file.rs
@@ -1,4 +1,9 @@
+use std::sync::Arc;
+
+use oak_semantic::semantic_index::ScopeId;
 use oak_semantic::semantic_index::SemanticIndex;
+use oak_semantic::semantic_index::SymbolTable;
+use oak_semantic::use_def_map::UseDefMap;
 use url::Url;
 
 use crate::parse::OakParse;
@@ -40,12 +45,39 @@ impl File {
 
     /// Build this file's `SemanticIndex` from the parse tree.
     ///
-    /// Crate-internal: cross-file consumers must use narrower public queries
-    /// (`exports`, `external_scope`) to avoid depending on the full index,
-    /// which would invalidate on every internal edit.
+    /// Private to this file to prevent coarse Salsa queries. Consumers should
+    /// go through the narrow tracked queries below.
+    ///
+    /// The full index re-runs typically on every edit (e.g. the `AstPtr` ranges
+    /// inside `Definition`s shift, so it rarely backdates).
+    #[cfg(not(test))]
+    #[salsa::tracked(returns(ref))]
+    fn semantic_index(self, db: &dyn Db) -> SemanticIndex {
+        build_semantic_index(self, db)
+    }
+
+    /// Tests use the `pub(crate)` variant gated behind `cfg(test)` so they can
+    /// call into `semantic_index` directly to verify salsa caching behaviour.
+    #[cfg(test)]
     #[salsa::tracked(returns(ref))]
     pub(crate) fn semantic_index(self, db: &dyn Db) -> SemanticIndex {
-        let parsed = self.parse(db);
-        oak_semantic::semantic_index(&parsed.tree(), self.url(db))
+        build_semantic_index(self, db)
     }
+
+    /// The symbol table for one scope of this file.
+    #[salsa::tracked]
+    pub fn symbol_table(self, db: &dyn Db, scope: ScopeId) -> Arc<SymbolTable> {
+        Arc::clone(self.semantic_index(db).symbols(scope))
+    }
+
+    /// The use-def map for one scope of this file.
+    #[salsa::tracked]
+    pub fn use_def_map(self, db: &dyn Db, scope: ScopeId) -> Arc<UseDefMap> {
+        Arc::clone(self.semantic_index(db).use_def_map(scope))
+    }
+}
+
+fn build_semantic_index(file: File, db: &dyn Db) -> SemanticIndex {
+    let parsed = file.parse(db);
+    oak_semantic::semantic_index(&parsed.tree(), file.url(db))
 }
--- a/crates/oak_db/src/file.rs
+++ b/crates/oak_db/src/file.rs
@@ -24,7 +24,13 @@ impl File {
     /// Crate-internal: kept `pub(crate)` so downstream crates can't take a
     /// dependency on a full parse tree. They reach the tree only through
     /// narrower public queries.
-    #[salsa::tracked(returns(ref))]
+    ///
+    /// `lru = 128` caps the number of live parse trees to 128. Matches
+    /// rust-analyzer's default for its analogous `parse(file_id)` query.
+    /// Rowan's green tree shares structure across edits, so eviction frees
+    /// memory cleanly. Derived queries (e.g. `semantic_index`) store
+    /// `AstPtr`s rather than tree nodes, so they don't pin an evicted tree.
+    #[salsa::tracked(returns(ref), lru = 128)]
     pub(crate) fn parse(self, db: &dyn Db) -> OakParse {
         OakParse::new(aether_parser::parse(
             self.contents(db),
--- a/crates/oak_db/src/tests.rs
+++ b/crates/oak_db/src/tests.rs
@@ -1,0 +1,1 @@
+mod file;
--- a/crates/oak_db/src/tests/file.rs
+++ b/crates/oak_db/src/tests/file.rs
@@ -1,0 +1,183 @@
+use std::sync::Arc;
+use std::sync::Mutex;
+
+use url::Url;
+
+use crate::Db;
+use crate::File;
+
+type Events = Arc<Mutex<Vec<salsa::Event>>>;
+
+#[salsa::db]
+#[derive(Clone)]
+struct TestDb {
+    storage: salsa::Storage<Self>,
+    events: Events,
+}
+
+impl TestDb {
+    fn new() -> Self {
+        let events = Events::default();
+        let storage = salsa::Storage::new(Some(Box::new({
+            let events = events.clone();
+            move |event| {
+                events.lock().unwrap().push(event);
+            }
+        })));
+        Self { storage, events }
+    }
+
+    /// Count `WillExecute` events whose `database_key`'s Debug form
+    /// contains `name`. Salsa's `DatabaseKeyIndex::fmt` resolves the
+    /// underlying function name only when a database is attached to the
+    /// current thread, so we wrap the scan in `salsa::attach`.
+    fn executions(&self, name: &str) -> usize {
+        salsa::attach(self, || {
+            self.events
+                .lock()
+                .unwrap()
+                .iter()
+                .filter(|event| match &event.kind {
+                    salsa::EventKind::WillExecute { database_key } => {
+                        format!("{database_key:?}").contains(name)
+                    },
+                    _ => false,
+                })
+                .count()
+        })
+    }
+}
+
+#[salsa::db]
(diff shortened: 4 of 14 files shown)


crates/oak_db/src/db.rs
```diff
@@ -0,0 +1,6 @@
+/// Salsa Database trait.
+///
+/// Queries take a `dyn Db` rather than the concrete database owned by
+/// the LSP layer.
+#[salsa::db]
+pub trait Db: salsa::Database {}
```
DavisVaughan, line 6: So this is the "subtrait" idea and is where we get to add some of our own methods onto a database?

DavisVaughan: I guess eventually we'll probably bring over some of these ty ideas, in particular `files()`?

```rust
/// Most basic database that gives access to files, the host system, source code, and parsed AST.
#[salsa::db]
pub trait Db: salsa::Database {
    fn vendored(&self) -> &VendoredFileSystem;
    fn system(&self) -> &dyn System;
    fn files(&self) -> &Files;
    fn python_version(&self) -> PythonVersion;
}
```

crates/oak_db/src/tests.rs
```diff
@@ -0,0 +1 @@
+mod file;
```
DavisVaughan, line 1: whats up with this file?

crates/oak_db/src/tests/file.rs
```diff
@@ -0,0 +1,183 @@
...
+
+    let _ = file.parse(&db);
+    assert_eq!(db.executions("parse"), 1);
+
+    let _ = file.parse(&db);
+    assert_eq!(db.executions("parse"), 1);
+}
+
+#[test]
+fn semantic_index_is_cached_across_calls() {
```
DavisVaughan, line 74: neat!

crates/oak_db/src/tests/file.rs
```diff
@@ -0,0 +1,183 @@
...
+    let source = "x <- 1\nx\n";
+    let url = file_url("a.R");
+    let file = File::new(&db, url.clone(), source.to_string());
+
+    let via_salsa = file.semantic_index(&db);
+
+    let parse = aether_parser::parse(source, aether_parser::RParserOptions::default());
+    let direct = oak_semantic::semantic_index(&parse.tree(), &url);
+
+    assert_eq!(*via_salsa, direct);
```
DavisVaughan, line 114:
```suggestion
    assert_eq!(via_salsa, &direct);
```

i got a bit confused by this because i thought `*via_salsa` was making a Copy of a SemanticIndex here, but Copy isn't even a method on SemanticIndex (which is good)

i had Claude help me understand that this is just creating an abstract "place" of a SemanticIndex which is then reborrowed with `&` later on by assert_eq!, so no ownership shifting actually ever occurs

it suggests making this clearer with this rewrite

crates/oak_db/src/file.rs
```diff
@@ -0,0 +1,51 @@
...
+        ))
+    }
+
+    /// Build this file's `SemanticIndex` from the parse tree.
+    ///
+    /// Crate-internal: cross-file consumers must use narrower public queries
+    /// (`exports`, `external_scope`) to avoid depending on the full index,
+    /// which would invalidate on every internal edit.
+    #[salsa::tracked(returns(ref))]
+    pub(crate) fn semantic_index(self, db: &dyn Db) -> SemanticIndex {
```
DavisVaughan, line 63: Claude has helped me understand that this currently unused but obviously useful helper is the reason for the Eq/PartialEq on SemanticIndex

(that and this one test `crates/oak_db/src/tests/file.rs:114` which could be tweaked)

It still felt like a footgun to me that we have a big SemanticIndex as Eq/PartialEq, so I asked claude to tell me what in the world ty is doing differently.

And look at this!

```rust
  // ruff/.../semantic_index.rs:56
  #[salsa::tracked(returns(ref), no_eq, heap_size=...)]
  pub(crate) fn semantic_index(db: &dyn Db, file: File) -> SemanticIndex<'_> { ... }
```

> no_eq tells salsa: don't bother comparing new vs old, just always invalidate downstream. So
> SemanticIndex needs Update (for safe revision-to-revision storage) and get_size2::GetSize (for the
> heap-size accounting), but PartialEq / Eq would be dead weight — they'd never get called.


And then it reiterated everything else we've been discussing regarding narrowly scoped queries

```rust
  // 2. ty exposes narrow per-scope tracked queries on top.

  // ruff/.../semantic_index.rs:70, 93
  #[salsa::tracked(returns(deref), heap_size=...)]
  pub(crate) fn place_table<'db>(db: &'db dyn Db, scope: ScopeId<'db>) -> Arc<PlaceTable> {
      let index = semantic_index(db, file);
      Arc::clone(&index.place_tables[scope.file_scope_id(db)])
  }
```

Where each of those does proper caching, but if Salsa determines this function needs to be rerun then hitting that call `semantic_index()` unconditionally regenerates the semantic index, which actually does feel right to me

And here's the rest of Claude's comments if you're interested

```
  Why this is a better design

  Backdating only pays off where you frequently produce structurally-equal output. At the aggregate
  level, that almost never happens — any meaningful edit touches something in the index, so a deep == on
  the whole SemanticIndex walks the entire file's worth of scopes/symbols/defs/uses and almost always
  returns false. Pure cost, no win.

  At the per-scope level, equality holds often: editing a function body leaves every sibling scope's
  PlaceTable and UseDefMap structurally identical. Backdating on those small pieces lets type inference
  for unaffected scopes skip entirely. That's the actual cache-granularity win.

  ty also gets a small parse-level safety net for free: a true no-op edit (type then delete) gets caught
  by OakParse-style structural equality at the parse query, so semantic_index never even re-runs.
  PartialEq on SemanticIndex itself isn't what catches that case.

  Where this leaves ark

  The PartialEq, Eq on oak_semantic::SemanticIndex is a placeholder, not a load-bearing design choice.
  The header comment at semantic_index.rs:47-51 and 46-65 actually anticipates exactly ty's strategy:

  ▎ "ty wraps per-scope tables in Arc so they can be returned from individual salsa tracked queries (e.g.
  ▎  place_table(db, scope) -> Arc<PlaceTable>). Salsa compares the returned Arc by Eq to skip re-running
  ▎  downstream queries when a scope's table hasn't changed. We defer that until salsa is introduced."

  Salsa is now introduced at the File/parse/semantic_index level, but the per-scope sub-queries aren't
  built yet. Until they are, the PartialEq derive on the aggregate buys you essentially nothing (and
  costs an expensive deep walk on every reparse where the index changed at all — i.e. most of them).

  Concrete next steps if you want to match ty:

  1. Change File::semantic_index to #[salsa::tracked(returns(ref), no_eq)] in
  crates/oak_db/src/file.rs:46.
  2. Drop PartialEq, Eq from SemanticIndex. You'll likely need salsa::Update (auto-derivable) instead;
  the inner field types may need it too.
  3. Wrap the per-scope fields in Arc: IndexVec<ScopeId, Arc<SymbolTable>>, IndexVec<ScopeId, 
  Arc<UseDefMap>>, etc.
  4. Add narrow tracked queries (symbol_table(db, scope), use_def_map(db, scope), file_exports(db, file),
   file_attached_packages(db, file)) that Arc::clone the relevant field. These keep their default eq
  behavior so they can backdate.
  5. Rewrite the test at crates/oak_db/src/tests/file.rs:114 (semantic_index_matches_oak_semantic) to
  compare via accessors (file_exports, scopes, etc.) instead of one big assert_eq!.

  The PartialEq derive isn't wrong today — it's just a forward-compatible placeholder that will get
  replaced when the per-scope query layer lands. If you want, you can do the no_eq flip now as a cheap,
  isolated win (skip the wasted deep comparison) and refactor the per-scope wrapping later.
```

crates/oak_db/src/file.rs
```diff
@@ -0,0 +1,51 @@
...
+        let parsed = self.parse(db);
+        oak_semantic::semantic_index(&parsed.tree(), self.url(db))
+    }
+}
```
DavisVaughan, line 83: Random comment, I noticed this in ty

```rust
/// Returns the place table for a specific `scope`.
///
/// Using [`place_table`] over [`semantic_index`] has the advantage that
/// Salsa can avoid invalidating dependent queries if this scope's place table
/// is unchanged.
#[salsa::tracked(returns(deref), heap_size=ruff_memory_usage::heap_size)]
pub(crate) fn place_table<'db>(db: &'db dyn Db, scope: ScopeId<'db>) -> Arc<PlaceTable> {
    let file = scope.file(db);
    let _span = tracing::trace_span!("place_table", scope=?scope.as_id(), ?file).entered();
    let index = semantic_index(db, file);
    Arc::clone(&index.place_tables[scope.file_scope_id(db)])
}
```

That's cool, they have `FileScopeId` which I believe is most similar to our `ScopeId` right now - i.e. it starts at `0` every time for any given file

And then their `ScopeId` is abstracted on top of that to generate a unique id of `File+FileScopeId` via

```rust
/// A cross-module identifier of a scope that can be used as a salsa query parameter.
#[salsa::tracked(debug, heap_size=ruff_memory_usage::heap_size)]
pub struct ScopeId<'db> {
    pub file: File,

    pub file_scope_id: FileScopeId,
}
```

So then their granular queries can query via a single `ScopeId`, which seems quite nice

crates/oak_db/src/file.rs
```diff
@@ -0,0 +1,51 @@
...
+    /// dependency on a full parse tree. They reach the tree only through
+    /// narrower public queries.
+    ///
+    /// `lru = 128` caps the number of live parse trees to 128. Matches
+    /// rust-analyzer's default for its analogous `parse(file_id)` query.
+    /// Rowan's green tree shares structure across edits, so eviction frees
+    /// memory cleanly. Derived queries (e.g. `semantic_index`) store
+    /// `AstPtr`s rather than tree nodes, so they don't pin an evicted tree.
+    #[salsa::tracked(returns(ref), lru = 128)]
+    pub(crate) fn parse(self, db: &dyn Db) -> OakParse {
```
DavisVaughan, line 39: Ok ok, so this is going to sound weird, but bear with me

I think all salsa tracked functions should be parameterized as

```rust
pub(crate) fn parse(db: &dyn Db, <one salsa tracked item>) -> Value {
```

i.e. you'll notice that in ty they have

```rust
#[salsa::tracked(returns(ref), no_eq, heap_size=ruff_memory_usage::heap_size, lru=200)]
pub fn parsed_module(db: &dyn Db, file: File) -> ParsedModule {
```

and in rust analyzer they have

```rust
impl EditionedFileId {
    pub fn parse(self, db: &dyn SourceDatabase) -> syntax::Parse<ast::SourceFile> {
        #[salsa::tracked(lru = 128)]
        pub fn parse(
            db: &dyn SourceDatabase,
            file_id: EditionedFileId,
        ) -> syntax::Parse<ast::SourceFile> {
            let _p = tracing::info_span!("parse", ?file_id).entered();
            let (file_id, edition) = file_id.unpack(db);
            let text = db.file_text(file_id).text(db);
            ast::SourceFile::parse(text, edition)
        }
        parse(db, self)
    }
```

where the inner tracked `parse()` is of this form

I think this is a rather important footgun, but I don't quite know why

Carefully read this part of the book
https://salsa-rs.netlify.app/tutorial/parser#parameters-to-a-tracked-function

```
The first parameter to a tracked function is always the database, db: &dyn crate::Db.

The second parameter to a tracked function is always some kind of Salsa struct.

Tracked functions may take other arguments as well, though our examples here do not. Functions that take additional arguments are less efficient and flexible. It's generally better to structure tracked functions as functions of a single Salsa struct if possible.
```

So I have a strong gut feeling we should rewrite these as free functions of this format!

DavisVaughan (approved):
A few particularly important comments about our `salsa::tracked` methods. It feels important to get these right so if you disagree with my comments it would be nice to get clarification before merging

crates/oak_db/src/file.rs
lionel-: hmm I think methods and free functions expand to the same thing but I'll double check

crates/oak_db/src/tests.rs
lionel-: Sorry I don't understand

crates/oak_db/src/file.rs
lionel-: yep I think I was confused on the need for coarse dep on the index while we transition towards narrow queries.
After all we don't care if the index is transitionally inefficient...

One subtlelty I now understand: even once we have only narrow tracked queries, we still need to track the index with `no_eq`. Otherwise each narrow query would rebuild it every time they are invalidated.

crates/oak_db/src/file.rs
lionel-: yep they expand and behave exactly the same way.

We can keep in mind the advice regarding additional arguments. (Although I imagine the performance cost is small)

crates/oak_db/src/db.rs
lionel-: yep Files is coming up!

crates/oak_db/src/file.rs
lionel-: Now `no_eq`, but we still need Eq/PartialEq.

I'm seeing a caveat about cycles in the doc, so we might have to remove the `no_eq` later on, hopefully not (but probably no big deal if we have to).

crates/oak_db/src/file.rs
lionel-: Nice, it would make sense to have a similar tracked oak_db type

crates/oak_db/src/tests.rs
DavisVaughan: Like, this file literally just has `mod file;` in it but never looks to be used. What is it for?

crates/oak_db/src/file.rs
DavisVaughan: yea i also double checked today and methods of `(self, db)` and free functions of `(db, tracked_struct)` expand the same way as long as you mark the impl block with `[salsa::tracked]` as well, which you do

Book does not reflect this 😛

crates/oak_db/src/tests.rs
lionel-: It loads the test module `tests/file.rs`

DavisVaughan: Oh you have a _unit test_ `tests/` folder, I totally just assumed that was integration tests

Interesting model!

crates/oak_db/src/file.rs
lionel-: oh I think salsa is under quite active development, so the book might not always be up to date

crates/oak_db/src/tests.rs
lionel-: yep this way we keep implementation files clean while also avoiding exposing stuff for integration tests

lionel-:
I made `File::semantic_index()` private to ensure no coarse dependency can be made by other modules.

(It'll temporarily go back to `pub(crate)` in another PR so I can better stage things)

PR #1214, opened 2026-05-13
Add `SourceGraph` to Oak database

lionel-:
Branched from #1213 

Progress towards #1212 
Progress towards #1183

Just simple data structures for now.

The source graph contains all workspace scripts, workspace packages, and installed packages. The edges are encoded differently for `Script` and `Package` nodes. In scripts they are populated from `library()` or `::` occurrences, stored in the file's semantic index. For packages, the package's metadata in the `namespace` field defines edges.


crates/oak_db/src/source_graph.rs
```diff
@@ -0,0 +1,92 @@
...
+    // every tracked query that read it. Future fix derives `Vec<File>`
+    // from a basename spec via `Root.revision` and a `Files` registry.
+    #[returns(ref)]
+    pub collation: Vec<File>,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub enum PackageOrigin {
+    Workspace { root: PathBuf },
+    Installed { version: String, libpath: PathBuf },
```
DavisVaughan: Is the idea that `version` is required to make a unique identifier in case the package is reinstalled?

crates/ark/src/lsp/state.rs
```diff
@@ -31,7 +31,9 @@ pub struct OakDatabase {
 
 impl OakDatabase {
     pub fn new() -> Self {
-        Self::default()
+        let db = Self::default();
+        oak_db::SourceGraph::new(&db, vec![], vec![], vec![]);
```
DavisVaughan: For an example of this odd pattern, see ty's `Program::from_settings()` in `ProjectDatabase::new()`, where `Program` is a singleton as well

crates/ark/src/lsp/state.rs
```diff
@@ -31,7 +31,9 @@ pub struct OakDatabase {
...
+        db
```
DavisVaughan: I would also have been happy with `SourceGraph::empty()` as a small helper method that calls the Salsa generated `new()` method with these empty vec![]s

crates/ark/src/lsp/state.rs
```diff
@@ -31,7 +31,9 @@ pub struct OakDatabase {
...
     }
```
DavisVaughan, line 39: I see that ty sets `.durability(Durability::HIGH)` on their `Program` singleton, but that's presumably because it is rare that it ever changes

Our `SourceGraph` on the other hand is likely to change quite a bit, I think

crates/oak_db/src/db.rs
```diff
@@ -1,6 +1,12 @@
...
 /// Salsa Database trait.
 ///
 /// Queries take a `dyn Db` rather than the concrete database owned by
 /// the LSP layer.
 #[salsa::db]
-pub trait Db: salsa::Database {}
+pub trait Db: salsa::Database {
+    fn source_graph(&self) -> SourceGraph {
+        SourceGraph::get(self)
+    }
```
DavisVaughan: This might be splitting hairs a bit, but, just to confirm, you want to inflict a `SourceGraph` on all salsa Databases we might ever want to create?

Like, IIUC, ty has an inheritance-like approach where they have this base db

```rust
/// Most basic database that gives access to files, the host system, source code, and parsed AST.
pub trait Db: salsa::Database {
    fn vendored(&self) -> &VendoredFileSystem;
    fn system(&self) -> &dyn System;
    fn files(&self) -> &Files;
    fn python_version(&self) -> PythonVersion;
}
```

they reference this as `SourceDb` throughout the project, i.e. all databases should be able to provide file sources and ASTs

then further up they have something they refer to as a `SemanticDb`, which inherits from `SemanticDb -> ModuleResolverDb -> SourceDb`, but is noted as

```rust
/// Database giving access to semantic information about a Python program.
```

it looks like this with these extra methods

```rust
// SemanticDb
pub trait Db: ModuleResolverDb {
    /// Returns `true` if the file should be checked.
    fn should_check_file(&self, file: File) -> bool;

    /// Resolves the rule selection for a given file.
    fn rule_selection(&self, file: File) -> &RuleSelection;

    fn lint_registry(&self) -> &LintRegistry;

    fn analysis_settings(&self, file: File) -> &AnalysisSettings;

    /// Whether ty is running with logging verbosity INFO or higher (`-v` or more).
    fn verbose(&self) -> bool;
}

// ModuleResolverDb
pub trait Db: SourceDb {
    /// Returns the search paths for module resolution.
    fn search_paths(&self) -> &SearchPaths;
}
```

And, for example, this `search_paths()` method is where they use their singleton `Program::get()`

```rust
#[salsa::db]
impl ty_module_resolver::Db for ProjectDatabase {
    fn search_paths(&self) -> &SearchPaths {
        Program::get(self).search_paths(self)
    }
}
```

So I just wondered if we should at least have this base `Db` with no trait methods for now (it will probably have `files()` eventually) and then a `SemanticDb` that has `source_graph()`

crates/oak_db/src/name.rs
```diff
@@ -0,0 +1,8 @@
+/// Interned identifier.
+///
+/// Lets tracked queries cache symbols or packages by name cheaply.
+#[salsa::interned]
+pub struct Name<'db> {
+    #[returns(ref)]
+    pub text: String,
```
DavisVaughan: Might be interesting to consider `CompactString` used by ty at some point

crates/oak_db/src/source_graph.rs
```diff
@@ -0,0 +1,92 @@
...
+        self.scripts(db)
+            .iter()
+            .find(|script| script.file(db).url(db) == url)
+            .copied()
+    }
+
+    /// Look up a `Package` by name. Workspace packages take precedence over
+    /// installed packages of the same name.
+    #[salsa::tracked]
+    pub fn package_by_name(self, db: &dyn Db, name: Name<'_>) -> Option<Package> {
```
DavisVaughan, line 91: Like other PR, I think this needs to probably be 

```rust
pub fn package_by_name(db: &dyn Db, name: Name<'_>)
```

which internally does

```rust
let source_graph = db.source_graph()
```

crates/oak_db/src/db.rs
```diff
@@ -1,6 +1,12 @@
...
+}
```
DavisVaughan: Since the downstream struct that implements `Db` is in charge of calling `SourceGraph::new()`, I feel like it would be symmetric for that struct to also be forced to implement a `source_graph()` method, even if its just `SourceGraph::get()` as you have here.

I feel like that's not totally unreasonable given

```
#[salsa::db]
impl ty_module_resolver::Db for ProjectDatabase {
    fn search_paths(&self) -> &SearchPaths {
        Program::get(self).search_paths(self)
    }
}
```

DavisVaughan: Like in the `TestDb` it feels like it would be nice to see both `SourceGraph::new()` and `SourceGraph::get()` in the same file, kind of close together, to see how the init + get dance works with it, and this would force that to be the case

DavisVaughan (approved):

crates/oak_db/src/db.rs
lionel-: I'm not sure I see the need for thinking about abstraction now?

I also don't think we "inflict" anything with the source graph. Are you concerned about more verbose tests? The other side of the coin is that it makes them more realistic.

For context, I can see the semantic index being useful on its own, independent of oak_db. So we should be careful to keep it that way (e.g. the `ImportResolver` coming up is generic) But as soon as cross-file resolution is needed, we need the full DB and its graph.

crates/oak_db/src/source_graph.rs
lionel-: hmm I don't think so, we'll need to depend on all package metadata and R files.

Let me delete both version and libpath here, we don't use either of those in future PRs.

crates/ark/src/lsp/state.rs
lionel-: As a whole yes. Down the line we'll want to set high durability on installed packages, and low durability on scripts and workspace packages.

crates/oak_db/src/name.rs
lionel-: Good idea, that's a Salsa feature. I've just switched.

crates/oak_db/src/db.rs
lionel-: hmm... I don't know, let's discuss it in our call

PR #1215, opened 2026-05-13
Better htmlwidget support

### LINKED ISSUE #1034: In notebook mode, emit HTML widgets as static, embeddable HTML
Currently HTML widgets are displayed in Positron as follows:

- The widget is saved to a temporary directory, along with all its supporting files and libraries
- The path is sent to Positron
- Positron creates a lightweight proxy to serve files from the path
- The URL to the proxy is shown in the Viewer pane

This is great for interactive consoles; it is performant, can be cached, works similarly to deployed widgets, etc. But doesn't work at all for notebooks because the temporary files are ephemeral and don't get saved with the notebook. Consequently, we need to implement an alternate way to send HTML widgets to Positron when Ark is in notebook mode. 

There are a few ways we could address this:

- Send a JSON representation of the widget data along with metadata listing all its dependencies, to be rendered statically by a [notebook output renderer](https://code.visualstudio.com/api/extension-guides/notebook#_notebook-renderer) in Positron. The data is rendered into HTML just in time using the htmlwidgets JS framework. (this was originally the approach I tried for HTML widgets in Positron proper)
- Use pandoc or a similar tool to convert the entire widget into a single blob of HTML that can be sent to Positron. This is [the approach used by IRKernel](https://github.com/IRkernel/repr/blob/826281bc655f2cec4601a2d7aee3f63f76af97ec/R/repr_htmlwidget.r#L129)  (which _does_ support HTML widgets in notebooks). Less performant, probably, but more self-contained.
- Map [Jupyter's "ipywidgets" concept](https://ipywidgets.readthedocs.io/en/stable/) onto HTML widgets and write whatever bridge/glue code is needed, so HTML widgets appear as Jupyter/ipywidgets from Positron's perspective
- [Run away from home and live in the woods](https://www.wikihow.com/Run-Away-from-Home-and-Live-in-the-Woods)

This is required for HTML widgets to work correctly in both notebooks (https://github.com/posit-dev/positron/issues/4219) and Quarto inline outputs (https://github.com/posit-dev/positron/issues/5640).

thomasp85:
Fix #1034

This PR adds proper htmlwidget support for notebooks and inline quarto content. It follows suggestion 2 from @jmcphers and inlines the html directly, like the IRKernel. The reasons for this is mainly that supporting Jupyter Widgets would be a substantial addition with no user-visible upside since the one thing it provides (bidirectional communication) isn't possible with htmlwidgets anyway. If a compelling use case for bidirectional communication comes up we can revisit it.

We don't need pandoc or any other heavy dependency because the html is trivial to construct.

One wrinkle is that IRKernel defaults to deduplicating html dependencies from consecutive cells. This facility also exist in this PR but since Positrons notebook view renders each cell in isolation. Due to that the default for ark is to not deduplicate.

All changes only has an effect for notebooks. The standard path for viewer/plot pane remains unchanged


crates/ark/src/viewer.rs
```diff
@@ -124,3 +130,51 @@ pub unsafe extern "C-unwind" fn ps_html_viewer(
...
+/// context (notebook or background session), where rich output must be emitted
+/// inline via `display_data` rather than routed through the Positron UI comm.
+#[harp::register]
+pub unsafe extern "C-unwind" fn ps_is_notebook() -> anyhow::Result<SEXP> {
+    let is_notebook = matches!(
+        Console::get().session_mode(),
+        SessionMode::Notebook | SessionMode::Background
+    );
+    Ok(libr::Rf_ScalarLogical(is_notebook as i32))
+}
```
DavisVaughan: Can we call this `ps_session_mode()` and emit a length 1 character vector of `"console"`, `"notebook"` or `"background"`? That seems a bit more general.

And then put this at the bottom of `console_repl.rs`, which is the file where `SessionMode` is defined.

And then probably `utils.R` would hold the R wrapper of `session_mode()`

crates/ark/src/viewer.rs
```diff
@@ -124,3 +130,51 @@ pub unsafe extern "C-unwind" fn ps_html_viewer(
     // No return value
     Ok(R_NilValue)
 }
+
+/// Emit a pre-rendered, self-contained HTML widget as a Jupyter `display_data`
+/// message. Called from R via `.ps.Call("ps_html_widget_emit", html, label)`.
```
DavisVaughan:
```suggestion
/// message.
```

crates/ark/src/modules/positron/html_widgets.R
```diff
@@ -1,16 +1,43 @@
...
 #' @export
 .ps.view_html_widget <- function(x, ...) {
-    # Render the widget to a tag list.
-    rendered <- htmltools::as.tags(x, standalone = TRUE)
+    if (isTRUE(.ps.is_notebook())) {
+        view_html_widget_inline(x)
+        return(invisible(x))
+    }
+
+    view_html_widget_viewer(x)
```
DavisVaughan: Nit: Make it a true `if / else` so you don't need a return? I think it makes it more pleasant to read that way

crates/ark/src/viewer.rs
```diff
@@ -19,26 +19,21 @@ use libr::SEXP;
...
-/// Emit HTML output on IOPub for delivery to the client
+/// Emit pre-rendered HTML output on IOPub for delivery to the client.
 ///
 /// - `iopub_tx` - The IOPub channel to send the output on
-/// - `path` - The path to the HTML file to display
-/// - `kind` - The kind of the HTML widget
+/// - `contents` - The complete HTML document to emit as `text/html`
+/// - `kind` - A short label used in the `text/plain` fallback
 fn emit_html_output_jupyter(
     iopub_tx: Sender<IOPubMessage>,
```
DavisVaughan: Small thing I just noticed.

You can make this `&Sender<IOPubMessage>` and then at the call sites you can just provide `console.iopub_tx()` without having to `.clone()` it out first.

crates/ark/src/modules/positron/html_widgets.R
```diff
@@ -19,20 +46,197 @@
...
+# returned string can be saved into a Jupyter notebook and reopened without
+# any external file references.
+embed_tags <- function(rendered) {
+    deps <- filter_seen_deps(rendered$dependencies)
+    dep_html <- vapply(deps, render_dep_inline, character(1))
+
+    head_parts <- c(
+        '<meta charset="utf-8"/>',
+        dep_html,
+        if (nzchar(rendered$head %||% "")) rendered$head else NULL
```
DavisVaughan: I feel like the only thing this is guarding against is a `""` `head`? But it's hard to figure that out.

I feel like you could lift this up as

```r
if (identical(rendered$head, "")) {
    rendered$head <- NULL
}
```

and that would accomplish the same thing, since an existing `NULL` passes through fine.

crates/ark/src/modules/positron/html_widgets.R
```diff
@@ -19,20 +46,197 @@
...
+        if (!is.null(file_base)) {
+            parts <- c(
+                parts,
+                sprintf(
+                    '<link rel="stylesheet" href="%s"/>',
+                    file_to_data_uri(file.path(file_base, css), "text/css")
+                )
+            )
+        } else if (!is.null(href_base)) {
+            parts <- c(
```
DavisVaughan, line 139: If you have `css` but both `file_base` and `href_base` are `NULL`, what happens? We just lose it? A comment would be nice, like in an extra `else {` fallthrough branch

Same with the js loop below

crates/ark/src/modules/positron/html_widgets.R
```diff
@@ -19,20 +46,197 @@
...
+                    '<script src="%s/%s"></script>',
+                    href_base,
+                    js
+                )
+            )
+        }
+    }
+
+    # Inline <script>/<style> blocks the dep wants in <head>.
+    if (length(dep$head) && nzchar(dep$head)) {
```
DavisVaughan:
```suggestion
    if (length(dep$head) == 1L && nzchar(dep$head)) {
```

?

Otherwise a length > 1 would produce an `nzchar()` result >1 and the if would be wrong

crates/ark/src/modules/positron/html_widgets.R
```diff
@@ -19,20 +46,197 @@
...
+        parts <- c(parts, dep$head)
+    }
+
+    paste(parts, collapse = "\n")
+}
+
+# `htmlDependency()` allows `script`/`stylesheet` to be either a character
```
DavisVaughan, line 185: i see `htmlDependency` mentioned 3 times but have no idea what it is

crates/ark/src/modules/positron/html_widgets.R
```diff
@@ -19,20 +46,197 @@
...
+# registered the library globally.
+#
+# This only works on frontends that render `text/html` outputs into a shared
+# DOM (classic Jupyter, JupyterLab). Positron's notebook view isolates each
+# cell's output, so a deduped second widget would find an empty global scope
+# and render blank. Default off; opt in with
+# `options(ark.html_widget.deduplicate = TRUE)` if you know your frontend
+# shares scope across cells.
+filter_seen_deps <- function(deps) {
+    if (!isTRUE(getOption("ark.html_widget.deduplicate", FALSE))) {
```
DavisVaughan, line 238: I would also be ok with not having this for now

Regardless, if you are confident you want this then please document this new option in `doc/configuration.md`

crates/ark/src/modules/positron/html_widgets.R
```diff
@@ -19,20 +46,197 @@
...
+    if (is.null(the$html_dep_cache)) {
+        the$html_dep_cache <- new.env(parent = emptyenv())
+    }
+    the$html_dep_cache
+}
+
+# Test hook: clear the per-session dedup cache so tests can assert dedup
+# behavior independent of each other.
+#' @export
+.ps.html_widget_reset_deps <- function() {
```
DavisVaughan: Rather than making it an export, you can also access it in tests via the secret `.ps.internal()` export, i.e. `.ps.internal(html_widget_reset_deps())`

crates/ark/src/viewer.rs
```diff
@@ -124,3 +130,51 @@ pub unsafe extern "C-unwind" fn ps_html_viewer(
...
+
+/// Emit a pre-rendered, self-contained HTML widget as a Jupyter `display_data`
+/// message. Called from R via `.ps.Call("ps_html_widget_emit", html, label)`.
+///
+/// Used by the notebook-mode widget print path, which renders the widget to a
+/// single HTML string with its JS/CSS dependencies inlined as `data:` URIs
+/// (see `crates/ark/src/modules/positron/html_widgets.R`). Console mode keeps
+/// the temp-file flow via `ps_html_viewer`.
+#[harp::register]
+pub unsafe extern "C-unwind" fn ps_html_widget_emit(
```
DavisVaughan: My biggest comment is that maybe we should consider a more generic name for this

It isn't realllllly widget specific, all this does it emit HTML via `display_data`

So I've renamed it in https://github.com/posit-dev/ark/pull/1242 along with some other minor tweaks, take a look and feel free to merge as is or take parts and close

crates/ark/src/viewer.rs
DavisVaughan, line 1: Is this supposed to work?

```
library(leaflet)
m <- leaflet() %>% addTiles()
m
```

I tried it in a qmd with inline output turned on and got a blank output pane

DavisVaughan: another one that doesn't work for me

```r
library(billboarder)

# data
data("prod_par_filiere")

billboarder(data = prod_par_filiere) %>%
  bb_barchart(
    mapping = aes(x = annee, y = prod_hydraulique),
    color = "#102246"
  ) %>%
  bb_y_grid(show = TRUE) %>%
  bb_y_axis(
    tick = list(format = suffix("TWh")),
    label = list(
      text = "production (in terawatt-hours)",
      position = "outer-top"
    )
  ) %>%
  bb_legend(show = FALSE) %>%
  bb_labs(
    title = "French hydraulic production",
    caption = "Data source: RTE (https://opendata.reseaux-energies.fr/)"
  )
```

DavisVaughan (commented):
The code all seems good, but i have not actually been able to get this feature to work for me yet

Ping me on slack if the examples below (i.e. leaflet) work for you and we can try to figure out what I'm doing wrong before merging

crates/ark/src/viewer.rs
DavisVaughan: (I did this in https://github.com/posit-dev/ark/pull/1242)

crates/ark/src/viewer.rs
DavisVaughan: (I did this in https://github.com/posit-dev/ark/pull/1242)

thomasp85:
The related positron PR is https://github.com/posit-dev/positron/pull/14150

This PR now handles a range of separate issues related to inline rendering of html widgets in quarto, the positron notebook and VS code notebook. A summary of the changes are:

* Use fully qualified blocks for scripts and stylesheets instead of data: refs to avoid issues with execution order
* Refire the wiring of html widgets at the end of the document as they may be injected after the onLoad event is fired
* Guard against the AMD module builder when executing the widget scripts

crates/ark/src/console/console_repl.rs
```diff
@@ -3025,3 +3025,15 @@ pub unsafe extern "C-unwind" fn ps_get_active_request() -> anyhow::Result<SEXP>
     let r_obj = RObject::try_from(json)?;
```
DavisVaughan, line 3025: Still seeing the weird font issue in the pop out, probably want to at least open an issue for this.

> i have a gut feeling the font thing is because in `quartoOutputManager.ts` in `_openWebviewInViewer` (which is what the popout button calls) it also does `createNotebookOutputWebview`, which may need to transition to the "raw" html version too (i.e. to `createRawHtmlOutputWebview`)

https://github.com/user-attachments/assets/160b1e80-94ec-4eb8-bab2-43be5c5b7a75

crates/ark/src/modules/positron/html_widgets.R
```diff
@@ -19,20 +43,241 @@
...
+        AMD_GUARD_OPEN,
+        dep_html,
+        rendered$head
+    )
+
+    paste0(
+        "<!DOCTYPE html>\n",
+        "<html>\n",
+        "<head>\n",
+        paste(head_parts, collapse = "\n"),
```
DavisVaughan, line 79: the amd_guard_open goes in the `<head>` but the amd_guard_close is outside of it? i dont know anything about this, it just smells odd

crates/ark/tests/html_widgets.rs
```diff
@@ -0,0 +1,165 @@
...
+        .to_string()
+}
+
+#[test]
+fn test_html_widget_emits_self_contained_html() {
+    let frontend = DummyArkFrontendNotebook::lock();
+
+    let mut code = String::from(".ps.internal(html_widget_reset_deps())\n");
+    // Defensive reset in case a sibling test leaked the option.
+    code.push_str("options(ark.html_widget.deduplicate = NULL)\n");
```
DavisVaughan: I think you should remove all of these "defensive resets"

With nextest, all Rust tests run in _their own individual process_, so this is a non issue and will only confuse future us

DavisVaughan (approved):
Combined with the Positron side PR, the two examples mentioned above work for me now in both a Quarto file with inline output enabled, and a ipynb file.

I can't comment much on the AMD bits or the forced `staticRender()` call for htmlwidgets, but if it seems required then sounds good.

crates/ark/src/modules/positron/html_widgets.R
thomasp85: Placing the opening guard in `<head>` ensures that it fires before the scripts are evaluated. There is no equivalent `<tail>` tag so the closing has to live in `<body>`

crates/ark/src/console/console_repl.rs
thomasp85: Yeah, I did not address this. I'm fairly certain this is a Positron side issue related to font resolution in different contexts.

PR #1216, opened 2026-05-13
Track `source()` calls as `SemanticCall` items on the index

lionel-:
Branched from #1214 
Progress towards #1212

This PR does two things:

- Renames `Directive` to `SemanticCall`. Better reflects what they are: calls that impact cross-file resolution.

- Stores `source()` calls as `SemanticCall::Source { path, resolved }`

I was originally going to use that for a downstream query, but I won't be needing it in the end.

I kept these changes because it seems useful to keep track of source calls and their resolution for reflective queries, e.g. for diagnostics (unresolved source path) or debugging.

### DIFF
--- a/crates/oak_semantic/src/builder.rs
+++ b/crates/oak_semantic/src/builder.rs
@@ -836,22 +836,28 @@ impl<'a> SemanticIndexBuilder<'a> {
         };
 
         let call_offset = call.syntax().text_trimmed_range().start();
+        let resolution = self.resolve_source(&path);
 
-        // Always record the source() call site. Downstream consumers
-        // (oak_db's `exports` query) translate the path to a `Script`
-        // and inject the target's exports into this file's exports.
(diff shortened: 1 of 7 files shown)


crates/oak_semantic/src/builder.rs
```diff
@@ -834,8 +834,28 @@ impl<'a> SemanticIndexBuilder<'a> {
         };
 
         let call_offset = call.syntax().text_trimmed_range().start();
+        let resolution = self.resolve_source(&path);
+
+        // Record every `source()` call site, independent of whether the resolver
+        // below resolves it. `resolved` pins the canonical URL when resolution
```
DavisVaughan:
```suggestion
        // above resolved it. `resolved` pins the canonical URL when resolution
```

?

DavisVaughan (approved):

PR #1218, opened 2026-05-14
Derive Equality on the semantic index after all

lionel-:
@DavisVaughan I thought we could use `no_eq` _instead_ of the Eq/PartialEq derivation, but that's not the case. You still need either Eq, or derive/implement `salsa::Update` for updating the value in the Salsa storage. The latter also does deep comparisons. 

I think Eq/PartialEq is better for oak_semantic because it feels useful to keep it completely independent of salsa.

One thing I've learned from ty is that they wrap symbol tables and use-def maps in `Arc` so that the comparisons are cheap. We now do the same, hopefully that addresses your concern about performance.

Also note that ty uses a `no_eq` attribute on their AST pointer in `Definition`. This needs `Definition` to be `#[salsa::tracked]` (the attribute only applies inside salsa-managed structs) and an Update derive. In our case we don't need all this complexity since our AST pointer type is very cheap to compare (a kind and a range).

### DIFF
--- a/.github/workflows/test-macos.yml
+++ b/.github/workflows/test-macos.yml
@@ -28,6 +28,8 @@ jobs:
 
       - name: Rust Cache
         uses: Swatinem/rust-cache@v2
+        with:
+          cache-bin: 'false'
 
       - name: Install nextest
         uses: taiki-e/install-action@v2
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -274,6 +274,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "allocator-api2"
+version = "0.2.21"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
+
+[[package]]
 name = "amalthea"
 version = "0.1.0"
 dependencies = [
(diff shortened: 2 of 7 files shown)


lionel-:
@DavisVaughan Following our discussion yesterday, `SemanticIndex` now conditionally implements `Update`, this way `oak_semantic` remains free of the Salsa dependency without having to implement PartialEq/Eq.

Digging a little more into this, I found out that your intuition was correct: Implementing `Update` does enable optimisations.

`Update` is the mechanism used to detect that a new computation is out of date, and replace the old value in the cache by copying the new value to the cache location.

There are two paths, depending on whether you implement `Update`:

- If you don't, the fallback simply compares the whole value with `==`, and moves/copy the whole value if the old is outdated.

- If you do, Update takes a different path where Salsa manually walks and compares a bunch of types (vectors, hashmaps, enums, etc). This is a deep walk similar to equality. The nice thing about that path is that if it finds a leave that's different, it's able to update it in place. E.g. if a hashmap value is different, only that value is updated in memory.

That's why it's important for entry points at least to use the `Update` mechanism. We should also strive to only have Salsa-compatible fields in our Salsa-tracked structs. If not compatible, we should try and provide an `Update` implementation. The main thing to watch out for are `struct` types, which `Update` does not recurse into by default.

That said, regarding the semantic index, this `Update` optimisation will not apply because the `SemanticIndex` will not become a tracked struct, we'll only track it via a tracked function marked as `no_eq`. Consequences:

- Output of tracked functions do not use `Update`, they always update wholesale.

- Because of `no_eq`, backdating check always returns `false`: https://github.com/salsa-rs/salsa/blob/7e77c49f27210dc85b49ba28606542d72836b5ab/components/salsa-macros/src/tracked_fn.rs#L107-L114

So it seems like implementing `Update` does not actually do anything here, except fullfilling one of two required bounds: equality or update.

DavisVaughan:
Nice, that's what I was hoping for!

DavisVaughan:
Did you really need `PartialEq, Eq` on the other fields in the end? They are a bit heavy for us to want to expose Eq comparison for (like `SymbolTables` gaining `Eq` feels off to me)

PR #1219, opened 2026-05-18
Resolve cross-file imports through Salsa

lionel-:
Progress towards #1212 

Reimplements the imports resolver (used to figure out symbols created by `source()` calls) on top of Salsa (incremental computation and caching mechanism). Coarse dependency on the sourced file's whole semantic index for now, but a future PR adds a `File::exports()` narrow query that fixes this.

Introduces an `ImportsResolver` trait with a method for resolving `source()` calls. When NSE lands, this trait will gain a `resolve_name()` method to resolve a symbol to an external definition. The point of this trait is keep `oak_semantic` totally free of Salsa details.

The concrete implementation is called `SalsaImportsResolver`, which evokes how that resolver leans on the salsa infrastructure. We could imagine a semantic index backed by a simpler non-Salsa resolver in the future.

When `source()` calls source each other, the cycle is resolved via the Salsa cycle resolution mechanism. The resolution is to rerun the semantic index computation with a noop-resolver (`NoopImportsResolver`). All local properties of the index remain, just the imports are missing, which allows the LSP to gracefully recover. For now we only emit a warning message, in the future we'll properly lint that case in a user-visible way.


crates/oak_db/src/imports.rs
```diff
@@ -0,0 +1,90 @@
...
+        })
+    }
+}
+
+/// Resolve `path` (the literal `source("path")` argument) against
+/// `calling_url`'s parent directory. Returns `None` if `calling_url` has
+/// no parent (root URL) or the joined path can't be turned back into a
+/// file URL.
+///
+/// TODO(salsa): Workspace anchoring + URL canonicalisation
```
DavisVaughan, line 81: yea im not exactly sure what this should statically anchor to

like if you see `source("b.R")` in `a.R`, does that resolve relative to the directory that `a.R` lives in, or the Workspace root that `a.R` is in?

DavisVaughan: thinking about the Positron console - it is typically started in the workspace root, so R's concept of the working directory is typically defaulting to the workspace root, I think

crates/oak_semantic/src/resolver.rs
```diff
@@ -0,0 +1,77 @@
+use url::Url;
+
+/// The result of resolving a `source()` call. Returned by
+/// [`ImportsResolver::resolve_source`].
+#[derive(Clone)]
+pub struct SourceResolution {
+    /// The resolved URL of the sourced file.
+    pub file: Url,
```
DavisVaughan: `url`? 

it makes sense to me when looking at `imports.rs` for this field to be url, especially since in that file you also have `File`s flying around

crates/oak_db/src/imports.rs
```diff
@@ -0,0 +1,90 @@
...
+    pub(crate) fn new(db: &'db dyn Db, calling_file: File) -> Self {
+        Self { db, calling_file }
+    }
+}
+
+impl<'db> ImportsResolver for SalsaImportsResolver<'db> {
+    fn resolve_source(&mut self, path: &str) -> Option<SourceResolution> {
+        let target_url = resolve_relative_to(self.calling_file.url(self.db), path)?;
+        let script = self.db.source_graph().script_by_url(self.db, &target_url)?;
+        let target = script.file(self.db);
```
DavisVaughan:
```suggestion
        let file = script.file(self.db);
```

?

crates/oak_db/src/imports.rs
```diff
@@ -0,0 +1,90 @@
...
+
+impl<'db> SalsaImportsResolver<'db> {
+    pub(crate) fn new(db: &'db dyn Db, calling_file: File) -> Self {
+        Self { db, calling_file }
+    }
+}
+
+impl<'db> ImportsResolver for SalsaImportsResolver<'db> {
+    fn resolve_source(&mut self, path: &str) -> Option<SourceResolution> {
+        let target_url = resolve_relative_to(self.calling_file.url(self.db), path)?;
```
DavisVaughan:
```suggestion
        let url = resolve_relative_to(self.calling_file.url(self.db), path)?;
```

i just feel like we could simplify the names a little bit

DavisVaughan: and if you make the change to SourceResolution, then `file: target_url` below just becomes `url`

crates/ark/src/lsp/document.rs
```diff
@@ -215,7 +215,11 @@ impl Document {
     /// TODO(salsa) Recomputed every time for now, but we'll track this with
     /// Salsa soon.
     pub fn semantic_index(&self, file: &url::Url) -> SemanticIndex {
-        oak_semantic::semantic_index(&self.parse.tree(), file)
+        oak_semantic::build_index(
+            &self.parse.tree(),
+            file,
+            &mut oak_semantic::NoopImportsResolver,
```
DavisVaughan: I suppose the noop resolver will also go away, because youll have access to the salsa database once we start tracking these with salsa

crates/oak_semantic/src/use_def_map.rs
```diff
@@ -211,6 +211,20 @@ pub struct UseDefMap {
...
+    /// TODO(salsa): delete with `SemanticIndex::empty` once cycle
+    /// handling moves from `File::semantic_index` to `File::exports`.
+    /// Sole caller today is `SemanticIndex::empty`.
+    pub fn empty() -> Self {
+        Self {
+            bindings_by_use: IndexVec::new(),
+            enclosing_snapshots: IndexVec::new(),
+        }
+    }
+
```
DavisVaughan, line 220: Can't you just do `UseDefMapBuilder::new().finish()`?

crates/oak_semantic/src/semantic_index.rs
```diff
@@ -103,6 +103,47 @@ impl SemanticIndex {
...
+    /// symbols, definitions, uses, or semantic calls. Used as the cycle
+    /// recovery value for `source()` chains where one file (transitively)
+    /// sources itself, an illegal user construct.
+    ///
+    /// TODO(salsa): delete once cycle handling moves from
+    /// `File::semantic_index` to `File::exports`. With the cycle path
+    /// running through exports, semantic_index no longer needs an
+    /// empty-recovery value.
+    pub fn empty() -> Self {
+        let mut scopes = IndexVec::new();
```
DavisVaughan, line 116: Same question with `SemanticIndexBuilder::new().finish()`?

crates/oak_semantic/src/builder.rs
```diff
@@ -45,50 +46,17 @@ use crate::semantic_index::Use;
...
-/// names and any `library()` package attachments.
-pub fn semantic_index_with_source_resolver<'a>(
-    root: &RRoot,
-    file: &Url,
-    resolver: impl FnMut(&str) -> Option<SourceResolution> + 'a,
-) -> SemanticIndex {
+/// Build a [`SemanticIndex`] from a parsed R file with cross-file
+/// information supplied by `resolver`. See [`ImportsResolver`] for the
+/// available impls.
+pub fn build_index(root: &RRoot, file: &Url, resolver: &mut dyn ImportsResolver) -> SemanticIndex {
```
DavisVaughan: I would advocate for an "own the `resolver`" approach, i.e. https://github.com/posit-dev/ark/pull/1223

I think this makes call sites simpler to reason about

And removes the need for lifetimes in `SemanticIndexBuilder`, which I find quite nice

DavisVaughan (approved):
Biggest comment is that I think `SemanticIndexBuilder` should own the `resolver`, i.e. https://github.com/posit-dev/ark/pull/1223

crates/oak_db/src/imports.rs
DavisVaughan: ah i see this

> Taking advantage of the new infrastructure, SalsaImportsResolver::resolve_source now anchors relative source() paths against the workspace root when the calling file is under one, falling back to the calling file's parent directory otherwise.

that sounds good

crates/ark/src/lsp/document.rs
lionel-: Right (just to be precise it'll go away here, but we still need the noop-resolver for cycle resolution)

crates/oak_semantic/src/use_def_map.rs
lionel-: I think I prefer explicitness of `empty()`? no strong opinions really. The method didn't go away btw, the TODO comment gets removed.

DavisVaughan: It can probably stay, just didn't want it to get out of sync with what the builder makes in the no-op case

`empty()` could even call `UseDefMapBuilder::new().finish()` to avoid any potential for out-of-sync ness, but that might be overkill

crates/oak_semantic/src/builder.rs
lionel-: No longer merges cleanly so I reimplemented here (feel free to use comments for this sort of small improvements)

PR #1220, opened 2026-05-18
Extract `aether_url` crate and canonicalise all paths

lionel-:
Branched from #1219 

Progress towards #1212 

Extract our canonical `Urld` struct used in Ark to get the legacy LSP and the DAP to agree on file URL identity. Now lives in `aether_url`, a dependency of both Oak and Ark. `File` now stores a `UrlId` so that all URLs get canonicalised at the LSP / Salsa DB boundary.

As agreed in our last call @DavisVaughan this will be renamed to `CanonicalUrl` once the current batch of PRs has been merged.


crates/oak_db/src/file.rs
```diff
@@ -15,10 +15,13 @@ use crate::Db;
 /// Content is pushed into Salsa by the LSP layer, the database never does I/O.
 /// This matches rust-analyzer's push model and avoids tying parsing to
 /// disk/network I/O inside a Salsa query.
+///
+/// The `url` field is a [`UrlId`], so the type system enforces "everything
+/// inside Salsa is a canonical URL".
```
DavisVaughan, line 20: 👍

crates/oak_db/src/file.rs
```diff
@@ -98,7 +101,7 @@ fn semantic_index_cycle_result(db: &dyn Db, _id: salsa::Id, file: File) -> Seman
     let parsed = file.parse(db);
     oak_semantic::build_index(
         &parsed.tree(),
-        file.url(db),
+        file.url(db).as_url(),
```
DavisVaughan, line 104: Do you think in the rename the field will be `canonical_url` so this is `file.canonical_url().as_url()`?

Seeing `url().as_url()` is fairly awkward IMO

I would also understand if you have a preference for the short `url: CanonicalUrl` field naming, but in that case idk what we could do better.

---

Oh, maybe `into_inner()`? So it just returns the "inner" `Url`?

So, `file.url(db).into_inner()` or maybe just `file.url(db).inner()` for a reference to it

crates/aether_url/src/lib.rs
```diff
@@ -0,0 +1,307 @@
...
+            .unwrap_or(uri);
+
+        #[cfg(windows)]
+        let uri = uppercase_windows_drive_in_uri(uri);
+
+        Self(uri)
+    }
+
+    /// Wrap a [`Url`] that the caller asserts is already canonical.
+    pub fn from_canonical(uri: Url) -> Self {
```
DavisVaughan, line 114: I would call this `from_url_unchecked()` to match the general vibe of Rust apis

```
# a bit odd
CanonicalUrl::from_canonical()

# seems right
CanonicalUrl::from_url_unchecked()
```

DavisVaughan (approved):

crates/oak_db/src/file.rs
lionel-: Added as bullet point for the planned refactor

crates/aether_url/src/lib.rs
lionel-: _unchecked has an unsafe flavour but I don't have a better alternative

PR #1222, opened 2026-05-19
Remove interrupt tasks entirely

### LINKED ISSUE #1187: On Windows, call some equivalent to `R_PolledEvents`
See https://github.com/posit-dev/ark/pull/1180 for more

On Unix we override `R_PolledEvents()` with

```
libr::set(R_PolledEvents, Some(r_polled_events));
```

We can't do this on Windows, this concept does not exist there.

There are 2 places on a Unix system where `R_PolledEvents()` is really called from that we are missing right now on Windows:

- `R_ProcessEvents() -> `R_PolledEvents()`
    - Used in many places, but the main one is from `R_CheckUserInterrupt()`
    - So on Windows I think our polled events aren't ever running at `R_CheckUserInterrupt()` time

- `R_runHandlers() -> R_PolledEvents()`
    - `R_runHandlers()` is a unix concept as well
    - `Sys.sleep() -> Rsleep() -> R_runHandlers() -> R_PolledEvents()`
        - But `Rsleep()` on Unix also calls `R_CheckUserInterrupt()`, so using `R_ProcessEvents()` would work here
        - And `Rsleep()` on Windows directly calls `R_ProcessEvents()`

I think for the most part this isn't the end of the world. For tasks that typically get run in `polled_events()` at interrupt time, they instead get run in the main `run_event_loop()` at idle time.

But we do do this in `polled_events()` which isn't replicated elsewhere

```rust
        if let Some(text) = self.debug_filter.check_timeout() {
            self.emit_stdout(text);
        }
```

---

We could consider removing our `polled_events()` hook altogether. If it isn't being invoked on Windows right now, how useful is it really? That would force us to remove interrupt tasks as a concept altogether, something we have been interested in. It would be worth finding a case where interrupt tasks are _very useful_ on Mac that we are currently lacking on Windows before discarding this idea.

If we want to keep interrupt tasks, on both OSes, we have access to hook into `R_ProcessEvents()`:
- Unix: `ptr_R_ProcessEvents`, settable at any moment
- Windows: `Rp->Callback` which sets a static `ptr_ProcessEvents`, settable only at startup

We'd have to reconcile the fact that we can't set `Rp->Callback` at any moment on windows, meaning `RLocalPolledEventsSuspended` wouldn't be precisely reproducible for process events, but maybe we could have a global atomic boolean that we flip instead, and our callback would just bail if that is set to `false` (i.e. suspended).

We also would need to reconcile with the fact that we currently manually call `R_ProcessEvents()`, and typically we seem to be expecting that tasks dont run at that time (i.e. the comment says we are typically suspended), so that might need some thinking as well

```
    fn process_idle_events() {
        // Process regular R events. We're normally running with polled
        // events disabled so that won't run here. We also run with
        // interrupts disabled, so on Windows those won't get run here
        // either (i.e. if `UserBreak` is set), but it will reset `UserBreak`
        // so we need to ensure we handle interrupts right before calling
        // this.
        unsafe { R_ProcessEvents() };
```

DavisVaughan:
Closes https://github.com/posit-dev/ark/issues/1187

In #1187 we realized that our "interrupt time" tasks have never been running on Windows, and...no one has complained.

Practically, this means that LSP features like completions or hover have not been running on Windows when R is busy, i.e. when R is running some long running computation, even if that code is checking for interrupts periodically via `R_CheckUserInterrupt()`. Unix machines, on the other hand, do return completions or hover results at interrupt time, but it is fairly unreliable and highly subject to the R code being run, i.e. this video shows how when you run a long running `lm()`, you can basically hang the LSP on a Mac because it almost never calls `R_CheckUserInterrupt()`:

https://github.com/user-attachments/assets/52d54ca3-47f2-4ae5-a693-81b1b8d9bc02

We've long through that doing extensive work at R interrupt time is rather dangerous (you could accidentally load a namespace mid-R execution, which would be wild), so we've wanted to remove these. And with oak, we are going to rely on interrupt time less and less anyways, because more things can be done statically without talking to the main R session.

So this PR removes interrupt tasks entirely. I'll leave more notes inline.


crates/ark/src/console/console_repl.rs
```diff
@@ -506,9 +505,12 @@ impl Console {
...
         // https://github.com/posit-dev/ark/blob/bd827e73/crates/ark/src/r_task.rs#L261.
-        r_task::spawn(RTask::interrupt({
+        //
+        // Use `idle_any_prompt` so DAP breakpoint invalidation still fires when
+        // the LSP signals a document change while R is paused at `browser()`.
+        r_task::spawn(RTask::idle_any_prompt({
             let dap_clone = console.debug_dap.clone();
-            async move || {
+            async move |_| {
                 Console::process_console_notifications(console_notification_rx, dap_clone).await
```
DavisVaughan, line 547: Added a test for this as well in `dap_breakpoints.rs`

crates/ark/src/console/console_repl.rs
```diff
@@ -1948,36 +1953,30 @@ impl Console {
...
-                        "{} milliseconds wait before running task.",
-                        start_info.start_time.elapsed().as_millis()
-                    )
-                });
-            }
+    /// The log message helps monitor excessively long-running tasks.
+    fn handle_task(&mut self, mut task: QueuedRTask) {
+        // For Sync tasks (i.e. only `r_task()`s), we want to log excessive waiting,
+        // because we are blocking the calling thread
+        if matches!(task, QueuedRTask::Sync(_)) {
```
DavisVaughan, line 2016: We used to log "excessive waiting" for interrupt tasks

We don't have interrupt tasks anymore, but we do have `Sync` tasks, which are only used by `r_task()`.

I thought it was appropriate to change this to log excessive waiting for `r_task()`s, since those are blocking, and since they previously were interrupt tasks. So it felt like the spirit was the same.

crates/ark/src/console/console_repl.rs
```diff
@@ -2475,33 +2503,6 @@ impl Console {
         if let Some(text) = self.debug_filter.check_timeout() {
             self.emit_stdout(text);
         }
```
DavisVaughan: I have kept the `debug_filter` check as part of an `R_ProcessEvents()` hook, which is now also being run on Windows.

This is now the ONLY THING special that we do at interrupt time. And it doesn't talk to the R API in any way, so should be very safe.

It's important we run this regularly-ish because the whole point is to not hold onto user output that might (for whatever reason) look like `debug:` output too long during long running computation

We talked about moving this to another thread, which would let us remove interrupt time hooks altogether.  I still think that idea is kind of nice in theory, but I wasn't quite sure how that thread would work.
- Does it just busy loop with a 25ms sleep between checks?
- Also, it would force `DebugFilterState` to be wrapped in a `Mutex` so it could be shared between `Console` and this extra thread, and that felt gross

I think we should keep this as is for this PR and if we still feel like we want to move it to a separate thread, we should do a follow up PR and aim for the simplest strategy possible (or, again, explore removing this check entirely)

crates/ark/src/console/console_repl.rs
```diff
@@ -2475,33 +2503,6 @@ impl Console {
...
-        // so we need to ensure we handle interrupts right before calling
-        // this.
-        unsafe { R_ProcessEvents() };
-
-        crate::sys::console::run_activity_handlers();
-
-        // Run pending finalizers. We need to do this eagerly as otherwise finalizers
-        // might end up being executed on the LSP thread.
-        // https://github.com/rstudio/positron/issues/431
-        unsafe { R_RunPendingFinalizers() };
```
DavisVaughan, line 2554: No longer needed since we don't run R code off the main R thread

crates/ark/src/lsp/backend.rs
```diff
@@ -99,21 +100,35 @@ macro_rules! cast_response {
...
+        actions: None,
     });
+    match tokio::time::timeout(Duration::from_secs(5), request).await {
+        Ok(result) => {
+            result.log_err();
+        },
+        Err(_) => {
+            log::warn!("Timed out waiting for frontend to acknowledge LSP crash notification");
+        },
+    }
```
DavisVaughan, line 134: `report_crash()` moves off needing to use the ui comm!

crates/ark/src/r_task.rs
```diff
@@ -340,90 +320,85 @@ impl RTask {
...
-            fun(capture).await
-        })
+        RTask::SendIdle(Self::pin_with_capture(fun))
     }
 
-    pub(crate) fn send_interrupt<F, Fut>(fun: F) -> Self
+    /// For rare performance sensitive cases where you'd like to avoid the cost of
+    /// poking R options via [Self::pin_with_capture()] and you know you don't need
+    /// the safety of capturing output because you aren't running R code
+    pub(crate) fn send_idle_without_capture<F, Fut>(fun: F) -> Self
```
DavisVaughan, line 440: I did end up wanting this for `Drop` in `RThreadSafe`

crates/harp/src/sys/unix/polled_events.rs
DavisVaughan, line 1: No longer relevant since `R_PolledEvents` is a no-op if no one sets it, and we don't set this anymore

crates/harp/src/exec.rs
```diff
@@ -509,6 +509,8 @@ where
     F: 'env,
     T: 'env,
 {
+    // TODO: These days this just means `RLocalInterruptsSuspended`.
+    // Should we simplify the name from `r_sandbox()` to `r_interrupts_suspended()`?
     let _scope = crate::raii::RLocalSandbox::new();
```
DavisVaughan: I was curious about your thoughts on this, since `r_sandbox()` does less now

crates/ark/src/console/console_repl.rs
```diff
@@ -853,7 +851,8 @@ impl Console {
         self.active_request.as_ref().map(|req| &req.request)
     }
 
-    // Async messages for the Console. Processed at interrupt time.
+    // Async messages for the Console. Polled at any idle prompt (top-level or
+    // browser) so DAP doc-change handling continues during debug sessions.
```
lionel-: doc-change handling?

crates/ark/src/console/console_repl.rs
```diff
@@ -1086,15 +1086,19 @@ impl Console {
...
-        // Process R's polled events regularly while waiting for console input.
-        // We used to poll every 200ms but that lead to visible delays for the
-        // processing of plot events, it also slowed down callbacks from the later
+        // Run activity handlers regularly while waiting for console input.
+        // We used to poll every 200ms but that slowed down callbacks from the later
         // package. 50ms seems to be more in line with RStudio (posit-dev/positron#7235).
-        let polled_events_rx = crossbeam::channel::tick(Duration::from_millis(50));
+        let activity_handlers_rx = crossbeam::channel::tick(Duration::from_millis(50));
+
+        // Run `R_ProcessEvents()` regularly out of good faith. We don't think this
```
lionel-: Here is what it does:

- Checks for interrupts on Windows (not needed at idle time)
- Checks time limits (not really supported by Ark)
- Run events callbacks: Cocoa/Quartz (graphics device), Windows GraphApp-based events (graphics device), TCL/TK (CRAN repo selection)

Native events can potentially be used in other contexts than Positron: CRAN repo selection, plot GUI, etc. To try this, set `options(device = "quartz")` in a Jupyter app and create plots. Note you'll need to try it outside of this branch because Quartz has been broken by setting `ptr_R_ProcessEvents`, see other comment.

Since it's involved in GUI processing, we should keep polling at idle time every 50ms.

crates/ark/src/console/console_repl.rs
lionel-: The main thing that gives me pause about this thread is using up 2mb stack space just for that. If a tokio task, that becomes much nicer, with the huge caveat that now Console depends on tokio :P

crates/ark/src/console/console_repl.rs
lionel-: That's a good idea

crates/ark/src/sys/unix/console.rs
```diff
@@ -77,6 +76,7 @@ pub fn setup_r(args: &Vec<String>) {
         libr::set(ptr_R_ShowMessage, Some(r_show_message));
         libr::set(ptr_R_Busy, Some(r_busy));
         libr::set(ptr_R_Suicide, Some(r_suicide));
+        libr::set(ptr_R_ProcessEvents, Some(r_process_events));
```
lionel-: This breaks the quartz device on macOS because it checks that this hook is still `NULL` before initialising: https://github.com/r-devel/r-svn/blob/714af4bf22bebd387437e31c397167996b4d5fbc/src/library/grDevices/src/qdCocoa.m#L712-L713

```r
options(device = "quartz")
plot(1)
```

While none of this is used in Positron, I think we should avoid unnecessary regressions on this front to make it easier to integrate Ark in other contexts, e.g. jupyter-based REPLs.

crates/ark/src/console/console_repl.rs
```diff
@@ -2873,11 +2872,10 @@ pub extern "C-unwind" fn r_suicide(buf: *const c_char) {
     panic!("Suicide: {}", msg.to_str().unwrap());
 }
 
-#[cfg(unix)]
 #[cfg_attr(not(test), no_mangle)]
-pub unsafe extern "C-unwind" fn r_polled_events() {
-    if let Err(err) = r_sandbox(|| Console::get_mut().polled_events()) {
-        panic!("Unexpected longjump while polling events: {err:?}");
+pub unsafe extern "C-unwind" fn r_process_events() {
+    if let Err(err) = r_sandbox(|| Console::get_mut().process_events()) {
```
lionel-: Should we remove the sandbox here, and document prominently that no R code should ever run in `process_events()`?

Or we keep it but mention here that is purely defensive against future changes.

crates/harp/src/exec.rs
lionel-: I think I'd rather keep the name distinct from implementation details?

lionel- (commented):
Looks like a step in the right direction!

The main thing I'd like us to be careful about is to not break native infrastructure that could potentially be used outside Positron:

- Native graphics devices: GraphApp-based on Windows, Cocoa/Quartz on macOS, X11 on Linux.
- Less important but comes up in practice: tcl/tk GUIs. R prompts user for a CRAN repo with a tcl/tk menu.

For some reason the tcltk package refuses to even load on my machine (R freezes) so I couldn't verify that it still works. But I could verify the Quartz graphics device no longer works, even though it does work on main. That's because Ark now sets `ptr_R_ProcessEvents`, see inline comment.

I did a deep dive in R's messy ecosystem of events loop (see below). I think we could simplify things and keep everything working with the following setup:

Windows:

```rs
pub fn pump_idle_events() {
    // R_ProcessEvents on Windows contains peekevent/doevent (GraphApp pump),
    // time limits, UserBreak, ptr_ProcessEvents (our hook), and R_Tcl_do.
    // No R_runHandlers / input-handler concept on Windows.
    unsafe { R_ProcessEvents() };
}

pub fn setup_r(args: &Vec<String>) {
    (*params).CallBack = Some(r_flush_debug_filter);
}
```

Unix:

```rs
pub fn pump_idle_events() {
    // ptr_R_ProcessEvents (Quartz), R_PolledEvents (debug_filter + tcltk),
    // R_CheckTimeLimits.
    unsafe { R_ProcessEvents() };

    // Drain ready input handlers (Quartz pipe, X11, later, help server).
    let mut fdset = unsafe { R_checkActivity(0, 1) };
    while !fdset.is_null() {
        unsafe { R_runHandlers(libr::get(R_InputHandlers), fdset) };
        fdset = unsafe { R_checkActivity(0, 1) };
    }

    // Match R's sys-std.c:1136 idiom: one final call with NULL fires
    // Rg_PolledEvents (Cairo X11 buffered display flush) and R_PolledEvents
    // (re-fires debug_filter + tcltk; both idempotent).
    unsafe { R_runHandlers(libr::get(R_InputHandlers), std::ptr::null_mut()) };
}

pub fn setup_r(args: &Vec<String>) {
    libr::set(R_PolledEvents, Some(r_flush_debug_filter));
    libr::set(R_wait_usec, 10000);
    // ptr_R_ProcessEvents intentionally not set as it's reserved for Quartz.
}
```

The other thing I'm wondering for a follow-up PR is if we should wait on the activity hanlder fds directly, on Unix. `R_InputHandlers` is a public linked list so it should be possible, I think Kevin had a comment about this (might have been deleted since): https://github.com/r-devel/r-svn/blob/714af4bf22bebd387437e31c397167996b4d5fbc/src/include/R_ext/eventloop.h#L93

The advantage would be that we'd run {later} callbacks and R's help server requests as soon as the events come in, without any timeout delays.

---

Notes from deep dive:

Interrupt-time: `R_CheckUserInterrupt()` calls `R_ProcessEvents()`

Unix `R_ProcessEvents()`
- `Ptr_R_ProcessEvents()`
- `R_PolledEvents()`
- `R_CheckTimeLimits()`

Windows `R_ProcessEvents()`
- GraphApp events
- `R_CheckTimeLimits()`
- Interrupt check
- `Ptr_R_ProcessEvents()`
- `R_Tcl_do()`

`R_runHandlers` (Unix only)
- Run activity/input handlers registered by packages (e.g. grDevices on mac, later on linux and mac)
- Runs only when R is idle, never when busy. Called by ReadConsole's Unix implementation.

`Ptr_R_ProcessEvents`
- macOS: Set by Cocoa, only if pointer still `NULL`. We can't set this without breaking Cocoa GUI.
- Windows: Mapped to `Rp->Callback`

`R_PolledEvents` (Unix only): Called from `ProcessEvents()`
-> Global pointer, only set by tcl/tk in stack-like fashion (calls old pointer)


`Rg_PolledEvents`: Separate hook for X11, called alongside `R_PolledEvents` when `R_runHandlers` is passed a `NULL` `readMask` (when `R_checkActivity()` returns `NULL`).


tcl/tk (Unix):
- Interrupt-time: Sets `R_PolledEvents` on Unix (in a stack-like pattern, calling the old handler if any is set), `R_Tcl_do` on Windows
- Idle: Does _not_ register an activity input handler
- Bidirectional: Sets a TCL event loop handler to call `R_runHandlers` when TCL is itself busy (e.g. running a modal dialog).

X11 (Unix): Runs on the R thread
- Interrupt-time: nothing. X11 does not register `Ptr_R_ProcessEvents`, so X11
  windows freeze during R computation. (Cairo-buffered X11 devices additionally
  set `Rg_PolledEvents = CairoHandler` for periodic buffer flushes, but
  `Rg_PolledEvents` only fires from the legacy `R_runHandlers(handlers, NULL)`
  branch, which ark never invokes.)
- Idle: Registers an activity input handler on the X11 connection fd
  (`addInputHandler(R_InputHandlers, ConnectionNumber(display),
   R_ProcessX11Events, XActivity)`). The X
  server pushes events to the connection fd, select() sees the fd ready,
  `R_runHandlers` dispatches to `R_ProcessX11Events`.

Quartz/Cocoa: Runs on the R thread
- Interrupt-time: Registers `Ptr_R_ProcessEvents` if NULL
- Idle-time: Registers an activity input handler for idle time events. Tickled by a background thread that sets a byte on the write end every 100ms.
- If `Ptr_R_ProcessEvents` is NULL, also initialises Cocoa GUI in the current process, which sets it as a full-blown foreground process visible in the Dock.

HTML help server (Unix): Server runs on the R thread
- Interrupt-time: Nothing, Help server is unresponsive when R is busy
- Idle-time: Registers an activity input handler on the listening TCP socket
 (`srv_input_handler`).
  When a connection arrives, that handler calls
 `accept()` and registers another input handler on the per-connection
 socket (`worker_input_handler`) to serve the request. So each in-flight
 request has its own input handler.

HTML help server (Windows): Server runs in a separate thread.

later (Unix):
- Interrupt-time: Nothing
- Idle-time: Registers an activity input handler (that checks call stack is empty in case of debugger read-prompt)

later (Windows):
- Spawns a message-only window that checks every 10ms

crates/ark/src/console/console_repl.rs
DavisVaughan: > document prominently that no R code should ever run in process_events()?

With your later PR that is going to poke `R_interrupts_pending` from inside here, this will no longer be true, so I'm going to keep the sandbox

crates/ark/src/console/console_filter.rs
```diff
@@ -263,7 +263,6 @@ impl ConsoleFilter {
     /// Check for timeout and handle state transitions.
     /// Timeout means we didn't reach ReadConsole to confirm debug output,
     /// so we emit the accumulated content back to the user.
-    #[cfg(any(unix, test))]
```
lionel-, line 266: 🎉

crates/ark/src/console/console_repl.rs
```diff
@@ -536,7 +534,13 @@ impl Console {
         // integration tests by spawning an async task. The deadlock is caused
         // by the `block_on()` behaviour in
         // https://github.com/posit-dev/ark/blob/bd827e73/crates/ark/src/r_task.rs#L261.
-        r_task::spawn(RTask::interrupt({
+        //
+        // Use idle-any so DAP breakpoint invalidation still fires when the LSP signals a
+        // document change while R is paused at `browser()`. Run without capture because
+        // this is a long running loop and we don't want capturing to permanently pin the
+        // `warn` global option to `1` for the lifetime of the loop. We don't emit any R
```
lionel-: I think we should toggle `warn` when the future is parked/reentered, the same way we toggle capture.

crates/ark/src/console/console_repl.rs
```diff
@@ -2498,24 +2500,72 @@ impl Console {
...
-        // This is 1/8th of the typical Windows stack space (1MB, whereas macOS
-        // and Linux have 8MB).
-        if harp::exec::r_check_stack(Some(128 * 1024)).is_err() {
-            return;
-        }
+    /// Invoke `R_ProcessEvents()`
+    ///
+    /// We call this out of good faith at regular intervals while idling in the event
+    /// loop. It is also one of the ways our [Self::interrupt_events()] hook is called,
+    /// which drains `debug_filter` during long computations, but that is a non-critical
```
lionel-: Should be less specific here, because that's also where we'll run the timeout check which I think is critical (prevents bricking Ark from DAP Evaluate requests)

crates/ark/src/console/console_repl.rs
```diff
@@ -2498,24 +2500,72 @@ impl Console {
...
+    /// - Calls `R_CheckTimeLimits()`
+    ///
+    /// Windows <https://github.com/wch/r-source/blob/bcc8ef90e50c65f143a54b2fde698bb16a135291/src/gnuwin32/system.c#L123-L158>:
+    /// - Calls graphapp's `doevent()` (we are unsure if you can use graphapp in Ark)
+    /// - Calls `R_CheckTimeLimits()`
+    /// - If `UserBreak=true`, sets it to `false` and calls `onintr()`. Never the case
+    ///   for us, since `run_event_loop()` always sets `set_interrupts_pending(false)`.
+    /// - Calls `ptr_R_ProcessEvents()`, bound to [Self::interrupt_events()] via
+    ///   `Rp->Callback`
+    /// - Calls `R_Tcl_do` (we are unsure if you can use tcktk in Ark)
```
lionel-: I've confirmed we can (not on your branch though, can you please try it?):

```r
library(tcltk)
tkmessageBox(title = "Info", message = "Hello!", icon = "info")
```

Image

crates/ark/src/console/console_repl.rs
```diff
@@ -2498,24 +2500,72 @@ impl Console {
...
+    /// Here is the platform specific call chain of `R_ProcessEvents()`, which shows how
+    /// it eventually calls our [Self::interrupt_events()] hook:
+    ///
+    /// Unix <https://github.com/wch/r-source/blob/bcc8ef90e50c65f143a54b2fde698bb16a135291/src/unix/sys-unix.c#L1168-L1181>:
+    /// - Calls `ptr_R_ProcessEvents()`, we don't set this, but Quartz is known to
+    /// - Calls `R_PolledEvents()`, bound to [Self::interrupt_events()]
+    /// - Calls `R_CheckTimeLimits()`
+    ///
+    /// Windows <https://github.com/wch/r-source/blob/bcc8ef90e50c65f143a54b2fde698bb16a135291/src/gnuwin32/system.c#L123-L158>:
+    /// - Calls graphapp's `doevent()` (we are unsure if you can use graphapp in Ark)
```
lionel-: graphapp is for Rgui, so no

crates/ark/src/console/console_repl.rs
```diff
@@ -2498,24 +2500,72 @@ impl Console {
...
+    /// [Self::run_process_events()], see that for details.
+    ///
+    /// It is also called at interrupt time via `R_CheckUserInterrupt()` calling
+    /// `R_ProcessEvents()`. This is the ONLY place we run our code at interrupt time
+    /// (i.e. in the middle of an R evaluation), so we must be very careful about what we
+    /// run here. Ideally we keep any R API access here to the absolute minimum, and it
+    /// should be very carefully analyzed. We currently use this to flush buffered debug
+    /// output from long running computations to avoid silently swallowing it for too
+    /// long.
+    /// https://github.com/wch/r-source/blob/fe64b85fc32fc9c3f2ef788f42566d6d2a182a0a/src/main/errors.c#L141-L159
```
lionel-: Mild pref for short links, here and elsewhere

```suggestion
    /// https://github.com/wch/r-source/blob/fe64b85f/src/main/errors.c#L141-L159
```

lionel- (approved):
That's great, thanks for getting this over the line!

crates/ark/src/console/console_repl.rs
DavisVaughan: Confirmed on my end with this branch as well

crates/ark/src/console/console_repl.rs
DavisVaughan: Opened https://github.com/posit-dev/ark/issues/1322

PR #1225, opened 2026-05-20
Add narrow tracked queries on `File`

lionel-:
Branched from #1221
Progress towards #1212

Adds narrow Salsa queries that shield downstream computations from irrelevant edits (e.g. a definition moves up or down) to a file:

- `File::exports()`: The top-level names this file exports, including those defined transitively through sourced files.

- `File::attached_packages()`: file attaches via `library()` / `require()`, including those propagated transitively through sourced files.

- `File::imports()`: The file's import chain in priority order: package imports from NAMESPACE for package files, attached packages for scripts, plus the default R search path. Offset-independent but order-dependent, so swapping two `library()` calls invalidates consumers of this query.

Each query carries its own `cycle_result` to handle `source()` cycles with empty/None results.

The resolver in `imports.rs` now reads `target.exports(db)` and `target.attached_packages(db)` instead of `target.semantic_index().file_exports()`. That's the one site where invalidation behaviour actually changes in this PR: editing a function body in `b.R` used to invalidate `a.R`'s semantic_index (because `a.R` sources `b.R`). This is no longer the case because `a.R`'s resolver depends on `b.R`'s exports, which is stable across body edits. In the future, this firewall query will be used in other places, providing similar benefits.

#### Resolving a definition

`File::resolve()` returns a new salsa-tracked `Definition` defined in `oak_db`. I've renamed the one in `oak_semantic` to `IndexDefinition` to disambiguate. The `Definition`  includes a `DefinitionKind` field that carries a pointer to the AST node for the binding expression, which includes the identifier. Handlers like goto-def or find-refs can reach for the identifier range through that `kind` field.

That `kind` field is by construction highly volatile: every edits that changes the text range of the definition (and remember our ranges are byte offsets in the file, not line/column coordinates) invalidate the kind.

ty has the right pattern for this. They explicitly mark `DefinitionKind` as `#[tracked]`, leaving other fields untracked. Untracked fields are the entity's identity. Once you've created a `Definition` for some `(file, scope, name)`, that salsa id is permanent. When a field like `kind` is `#[tracked]`, it's allowed to update across revisions without changing the id. This way, if a consumer only ever reads `file` / `scope` / `name` without reading `kind`, salsa keeps it cached across edits.

```rust
#[salsa::tracked(debug)]
pub struct Definition<'db> {
    pub file: File,         // identity
    pub scope: ScopeId,     // identity
    pub name: Name<'db>,    // identity
    #[tracked]
    #[no_eq]
    pub kind: DefinitionKind, // volatile
}
```

The alternative is to keep a lighterweight untracked `Resolution` structs for transient lookups. For now I think I like having a single entity for both cross-edit identity and transient lookups. Consumers only need to know about one.


crates/oak_db/src/definition.rs
```diff
@@ -0,0 +1,88 @@
...
+/// the AST update `kind` in place while the salsa id (keyed on `file`, `scope`,
+/// `name`) stays stable. Without `#[tracked]`, every AST change would produce a
+/// new entity id and invalidate all consumers, including those that only care
+/// about identity.
+///
+/// Also `no_eq` because the embedded `AstPtr`s shift on any edit moving the
+/// binding's line, so backdating would never fire. The point isn't backdating,
+/// it's per-field invalidation.
+#[salsa::tracked(debug)]
+pub struct Definition<'db> {
```
DavisVaughan, line 30: I'm not totally sure I understand, something feels wrong here. We have this:

```rust
#[derive(Debug, PartialEq, Eq)]
pub struct IndexDefinition {
    pub(crate) kind: DefinitionKind,
    /// The file that owns this definition's index.
    // TODO(salsa): Should become a File.
    pub(crate) file: Url,
    /// The scope within the file's index where this definition lives.
    pub(crate) scope: ScopeId,
    // TODO(salsa): Should become a PlaceId (like ty's `ScopedPlaceId`).
    pub(crate) symbol: SymbolId,
    pub(crate) range: TextRange,
}
```

This new `Definition` is a mirror of that type, but in the `SemanticIndex` itself, with lots of TODOs about how it should move to Salsa

I was expecting those TODOs to be removed, and for the `Definition` in the `SemanticIndex` to become "salsa aware"

ty only has 1 copy of this type, the `Definition` that lives in the `SemanticIndex` itself, which is salsa aware. ty's `SemanticIndex` seems to have fairly deep connections with Salsa, rather than trying to avoid taking it as a dependency, like we've been doing.

crates/oak_semantic/src/use_def_map.rs
```diff
@@ -213,11 +213,7 @@ pub struct UseDefMap {
 impl UseDefMap {
     /// An empty use-def map, suitable for a scope with no uses and no
     /// nested-scope snapshots. Used to construct an empty index on
-    /// `source()` cycle recovery.
-    ///
-    /// TODO(salsa): delete with `SemanticIndex::empty` once cycle
-    /// handling moves from `File::semantic_index` to `File::exports`.
-    /// Sole caller today is `SemanticIndex::empty`.
+    /// `source()` cycle recovery. Sole caller is `SemanticIndex::empty`.
```
DavisVaughan: I'm becoming more convinced that you should remove this in favor of just letting `SemanticIndex::empty()` do `SemanticIndexBuilder::new().finish()`.

Especially if you go look at `SemanticIndex::empty()`. That's a lot of boilerplate that also exists in `SemanticIndexBuilder::new()` and I would hate for that to get out of sync.

And then you don't need these extra `empty()` methods on the subtypes.

To me it makes sense 🤷

crates/oak_db/src/file.rs
```diff
@@ -75,22 +76,37 @@ impl File {
 
     /// Build this file's `SemanticIndex` from the parse tree.
     ///
-    /// `pub(crate)` so [`SalsaImportsResolver`] and tests can reach it. External
-    /// consumers should go through the narrow tracked queries below.
+    /// This is a coarse query that invalidates downstream on every edit
+    /// (`AstPtr` ranges inside `Definition`s shift). External consumers should
```
DavisVaughan, line 80: Every mention of `Definition` here should actually be `IndexDefinition`

But as mentioned elsewhere the fact that we have 2 overlapping types feels very wrong

crates/oak_db/src/file.rs
```diff
@@ -75,22 +76,37 @@ impl File {
...
+    ///   side is rebuilt with `NoopImportsResolver`. Cross-file
+    ///   injection drops, but local analysis (scopes, use-def maps,
+    ///   function bodies) is preserved.
+    ///
+    /// - `exports` / `imports` / `resolve` (FallbackImmediate, empty):
+    ///   the cycling side gets an empty fallback for that query.
+    ///
+    /// Which handler fires depends on which query salsa first re-enters.
+    /// R doesn't allow cyclic `source()`, so the asymmetric recovery is
+    /// acceptable. TODO(diagnostics): Lint `source()` cycles.
```
DavisVaughan, line 109: I think this is really the most important part of all this cycling handling

Like, ideally we wouldn't have to handle this edge case at all, but we have to defend against invalid partial states a user might be in, so we do

And ideally we will get to lint this invalid partial state for the user as they are shifting things around

crates/oak_db/src/file.rs
```diff
@@ -112,6 +128,20 @@ impl File {
         Arc::clone(self.semantic_index(db).use_def_map(scope))
     }
 
+    /// Package names from `library()` / `require()` calls in this file,
+    /// including those propagated transitively through `source()` chains.
+    /// Ordered by call-site position, which preserves R's search-path
+    /// semantics: a later `library(b)` shadows an earlier `library(a)`
+    /// when both export the same name.
+    #[salsa::tracked]
+    pub fn attached_packages(self, db: &dyn Db) -> Vec<Name<'_>> {
```
DavisVaughan, line 137: This feels like it should be a `returns(ref)`?

i.e. I don't think you really want a fresh `Clone` of the `Vec` at every call site, I have a feeling you rarely need full ownership (and if you do you probably iterate and convert to `String` anyways, but that should be up to the caller to decide)

crates/oak_db/src/file.rs
```diff
@@ -112,6 +128,20 @@ impl File {
...
+        self.semantic_index(db)
+            .file_attached_packages()
```
DavisVaughan: Any reason not to just call this `attached_packages()` too?

crates/oak_semantic/src/semantic_index.rs
```diff
@@ -541,7 +539,7 @@ impl SymbolFlags {
 // signature). Future `declare()` annotations will also produce pure
 // declarations.
 #[derive(Debug, PartialEq, Eq)]
-pub struct Definition {
+pub struct IndexDefinition {
     pub(crate) kind: DefinitionKind,
```
DavisVaughan: In addition to not loving having the double Definition types, I also think it would be a mistake to not rename DefinitionKind and DefinitionId to match the new name too.

crates/oak_db/src/file_exports.rs
```diff
@@ -0,0 +1,101 @@
...
+    /// `DefinitionKind::Import { file, name }` maps to `ExportEntry::Import {
+    /// file, name }` via [`Db::file_by_url`]. If the target file isn't interned
+    /// yet, the Import is dropped silently. Expected, since
+    /// [`SalsaImportsResolver`] only injects Imports when `file_by_url`
+    /// resolves the target.
+    ///
+    /// `cycle_result` (FallbackImmediate) recovers from cyclic `source()`
+    /// chains by returning empty exports for every cycle participant.
+    #[salsa::tracked(returns(ref), cycle_result = exports_cycle_result)]
+    pub fn exports(self, db: &dyn Db) -> FileExports {
```
DavisVaughan, line 62: Hmm, but we have `file_exports()` on `SemanticIndex` too?

So I'm expecting either that method to be used here and directly translated into some salsa compatible type like `FileExports`, or I'm expecting to see some TODO over on `file_exports()` that we should be trying to move away from it

I don't love that current have both with different impls

DavisVaughan: I feel like one of two things is bound to happen if you want to keep `SemanticIndex` unaware of Salsa:

- We move all the "smarts" like `exports()` into `SemanticIndex` itself and just reexpose `SemanticIndex::exports()` as `File::exports()`, possibly with a different return type that _is_ salsa aware so it can be tracked
- We trim `SemanticIndex` back to the _minimal_ amount of exposed things, like maybe it only exposes accessors to fields, and all the "smarts" live in `oak_db` as `File` methods, like `exports()` here

But having impls in both places feels not right

I feel like this is why ty ended up letting their `SemanticIndex` have deep integration with Salsa. In their world if they had a method like `exports()` then it would be directly on their salsa-aware `SemanticIndex`, and it would directly return a salsa-aware return type, so they wouldn't need a translation layer between salsa-aware and salsa-unaware worlds

crates/oak_db/src/file_exports.rs
```diff
@@ -0,0 +1,101 @@
...
+        let mut entries: HashMap<String, ExportEntry> = HashMap::new();
+        for (_id, def) in index.definitions(file_scope).iter() {
+            let name = symbols.symbol(def.symbol()).name();
+            let entry = match def.kind() {
+                DefinitionKind::Import {
+                    file: target_url,
+                    name: target_name,
+                    ..
+                } => {
+                    let target_url_id = UrlId::from_canonical(target_url.clone());
```
DavisVaughan, line 71: Should that already be canonical in `DefinitionKind`?

crates/oak_db/src/file_imports.rs
```diff
@@ -0,0 +1,130 @@
...
+    /// file. Names are resolved through `file.exports(db)`.
+    File(File),
+    /// NAMESPACE `importFrom(pkg, name)` entries. Maps each imported
+    /// symbol to its source package by name. Translation to `Package`
+    /// happens at resolution time.
+    PackageImports(HashMap<String, String>),
+    /// An attached package's exports. Missing packages are filtered
+    /// out by `imports`.
+    PackageExports(Package),
+}
```
DavisVaughan, line 25: I'm looking for some kind of indication of the fact that this replaces something

Like, it replaces `ScopeLayer` and the methods like `package_root_layers()`, right?

It would be really nice to have a TODO or something added over there to make it clear that we are trying to remove those now

crates/oak_db/src/file_imports.rs
```diff
@@ -0,0 +1,130 @@
...
+    let index = file.semantic_index(db);
+
+    // `library()` / `require()` calls in source order. We keep all of
+    // them; callers that need offset-based narrowing filter at query
+    // time.
+    for call in index.semantic_calls() {
+        let SemanticCallKind::Attach { package: name } = call.kind() else {
+            continue;
+        };
+        if let Some(package) = db.package_by_name(name) {
```
DavisVaughan: The other `file_layers()` function had this very nice comment from exhaustive mapping

```
            SemanticCallKind::Source { .. } => {
                // `source()` injects into local scope, not the search path;
                // not a scope-chain layer.
            },
```

without this i was confused about what is currently here

crates/oak_db/src/file_imports.rs
```diff
@@ -0,0 +1,130 @@
...
+    #[salsa::tracked(returns(ref))]
+    pub fn imports(self, db: &dyn Db) -> Vec<ImportLayer> {
+        match self.package(db) {
+            Some(package) => package_layers(self, db, package),
+            None => script_layers(self, db),
+        }
+    }
+}
+
+fn script_layers(file: File, db: &dyn Db) -> Vec<ImportLayer> {
```
DavisVaughan: not tracked?

DavisVaughan: also, not `self`???

crates/oak_db/src/file_imports.rs
```diff
@@ -0,0 +1,130 @@
...
+        }
+    }
+
+    // Default search path (R always attaches these at startup).
+    extend_with_default_search_path(db, &mut layers);
+
+    layers
+}
+
+fn package_layers(_file: File, db: &dyn Db, package: Package) -> Vec<ImportLayer> {
```
DavisVaughan: not tracked?

why is `_file` here?

why not a method on `Package`?

crates/oak_db/src/file_imports.rs
```diff
@@ -0,0 +1,130 @@
...
+            .iter()
+            .map(|imp| (imp.name.clone(), imp.package.clone()))
+            .collect();
+        layers.push(ImportLayer::PackageImports(map));
+    }
+
+    // NAMESPACE `import(pkg)` directives (bulk package imports).
+    for pkg_name in &namespace.package_imports {
+        if let Some(pkg) = db.package_by_name(pkg_name) {
+            layers.push(ImportLayer::PackageExports(pkg));
```
DavisVaughan, line 105: And IIUC this is currently almost always `None` right?

Because `LibraryRoots` isn't really hooked up right now and has no `Root`s

crates/oak_db/src/file_resolve.rs
```diff
@@ -0,0 +1,74 @@
...
+            if !visited.insert((current_file, current_name.clone())) {
+                log::error!(
+                    "Internal error: Cycle through `Import` forwards while resolving \
+                    `{current_name}` in {url}.",
+                    url = current_file.url(db),
+                );
+                return None;
+            }
+
+            let entry = current_file.exports(db).get(current_name.as_ref())?.clone();
```
DavisVaughan: The `clone()` here caught me off guard a bit

I think you can remove it and just use `*file` on the `Copy`-implementing `File` further below to be explicit about the small bit that actually needs a copy

crates/oak_db/src/file_resolve.rs
```diff
@@ -0,0 +1,74 @@
...
+            match entry {
+                ExportEntry::Local => {
+                    // Fetch exports again, this time through the semantic index
+                    // to get the volatile `kind` field that the firewall query
+                    // `File::exports()` doesn't expose.
+                    let kind = current_file
+                        .semantic_index(db)
+                        .file_exports()
```
DavisVaughan: I am uncomfortable with the fact that we have `exports()` and `file_exports()` with different impls being used here

DavisVaughan: Like the `?` is doing a lot of heavy lifting here by saying "if `exports()` and `file_exports()` disagree, we're cool with that, just return `None`" which feels wrong to me

crates/oak_db/src/file_resolve.rs
```diff
@@ -0,0 +1,74 @@
...
+                        .get(current_name.as_ref())
+                        .map(|def| def.kind().clone())?;
+
+                    let file_scope = ScopeId::from(0);
```
DavisVaughan, line 56: It feels wrong to me that you are specifying the file scope manually here instead of pulling from `def.scope()`, even if its the same value, i want to see that 1:1 mapping

crates/oak_db/src/definition.rs
```diff
@@ -0,0 +1,88 @@
...
+                };
+                target.into_syntax()
+            },
+            DefinitionKind::Parameter(param_ptr) => {
+                let param = param_ptr.to_node(&root);
+                param.name().ok()?.into_syntax()
+            },
+            DefinitionKind::ForVariable(for_ptr) => {
+                let for_stmt = for_ptr.to_node(&root);
+                for_stmt.variable().ok()?.into_syntax()
```
DavisVaughan: Can you make all of these just `ptr`? And all of the nodes just `node`? I think some naming cleanup would make this read cleaner

crates/oak_db/src/definition.rs
```diff
@@ -0,0 +1,88 @@
...
+            },
+            DefinitionKind::Import { .. } => return None,
+        };
+        Some(name_node.text_trimmed_range())
+    }
+}
+
+fn is_right_assignment(bin: &RBinaryExpression) -> bool {
+    bin.operator().is_ok_and(|op| {
```
DavisVaughan:
```suggestion
fn is_right_assignment(node: &RBinaryExpression) -> bool {
    node.operator().is_ok_and(|op| {
```

DavisVaughan (approved):
I have many questions and concerns that mainly stem from:

- `Definition` and `IndexDefinition` feel like they should be 1 type, as it is leading to some weird ergonomics and some duplication that feels wrong, and is also unlike ty

- If some of these new things are replacing old things, I would have appreciated some TODO comments on the old things that let me know what they are replacing. Otherwise it very much feels like we have 2 impls of everything right now. A notable weird one is `File::exports()` and `SemanticIndex::file_exports()` which currently does not look like it is being replaced, so we have 2 impls of how to get exports, and that feels wrong.

These concerns are called out in more detail below

crates/oak_db/src/definition.rs
lionel-: For now I'd like keep `oak_semantic` free of Salsa. I think the type duplication is a small cost.

Either way is fine really, but I've been working to keep the index to be independent of Salsa for a while now, so it can be used by independent code (e.g. jarl if they want to do their own imports resolution). It doesn't seem like now is the time to change course. I don't know if that separation will pay off, but I don't think the cost of maintaining it is prohibitive.

crates/oak_semantic/src/use_def_map.rs
lionel-: I've delegated for use-def-map but `SemanticIndex::empty()` is different because `SemanticIndexBuilder::new(range, file, resolver)` requires three arguments. The `empty()` method makes more sense to me.

There are no longer any callers of `SemanticIndex::empty()` anymore though so I've removed it for now.

crates/oak_db/src/file.rs
lionel-: I think originally I thought this would better reflect that this does not represent any global state, but I now agree this should be obvious from the semantic index context.

lionel-: Although most index methods take a scope, and these ones are "global" to the index, which the `file_` prefix makes clearer.

I think both names are fine. I already made the change locally, so keeping it.

crates/oak_db/src/file_exports.rs
lionel-: Good catch I missed that duplication. Now calls `exports()` on the index so it doesn't reimplement the walk.

Duplication on either side should be kept minimal, and I think we can achieve that even with the Salsa boundary.

crates/oak_db/src/file_exports.rs
lionel-: Tracked in https://github.com/posit-dev/ark/issues/1212#issuecomment-4451934067

crates/oak_db/src/file_imports.rs
lionel-: The caller is tracked.

I haven't included private methods in impl blocks so far.

crates/oak_db/src/file_imports.rs
lionel-: The caller is tracked.

_file comes from the dispatching-like pattern.

This is a `File` method

crates/oak_db/src/file_imports.rs
lionel-: Right, I've added the missing TODO

crates/oak_db/src/file_imports.rs
lionel-: None of the roots are hooked up right now, but that's coming in the next PRs.

The library scan will have access to NAMESPACE files and will be able to populate exports.

crates/oak_db/src/file_resolve.rs
lionel-: `File::exports()` uses `SemanticIndex::exports()` too. They use the same underlying implementation. Having to call twice is inherent to the problem at hand: the first call is the firewall, the second gets at the volatile `kind`.

crates/oak_db/src/file_resolve.rs
lionel-: I think pulling `def.scope()` would wrongly suggest the scope can be anything other than 0.

lionel-:
@DavisVaughan https://github.com/posit-dev/ark/pull/1225/commits/e052eb5a42c8405ec79396de2c019da227e9c413 should go a long way to make the duplicate structs more palatable. With this cleanup, `kind` is the only field the two types share.

PR #1226, opened 2026-05-20
Add offset-based queries

lionel-:
Branched from https://github.com/posit-dev/ark/pull/1225

Adds `File::resolve_at()` and `File::imports_at()`. To be used mainly to handle LSP requests that need information at a particular cursor position. These methods are not tracked since they take an offset, to avoid growing the cache unnecessarily.


crates/oak_semantic/src/semantic_index.rs
```diff
@@ -247,17 +247,36 @@ impl SemanticIndex {
     }
 
     /// Resolve a name starting from `scope`, walking up the scope chain.
-    pub fn resolve_symbol(&self, name: &str, scope: ScopeId) -> Option<(ScopeId, SymbolId)> {
+    /// Returns the scope that owns the binding and the `DefinitionId` of
+    /// the first matching [`IndexDefinition`] in that scope (source-order
+    /// first), so callers can fetch kind/range without a second lookup.
+    pub fn resolve_symbol(&self, name: &str, scope: ScopeId) -> Option<(ScopeId, DefinitionId)> {
```
DavisVaughan: This is starting to feel like a companion to `definition_at()`, but for looking up by `name` not by `offset`. So

- Maybe it gets a `definition_` prefix too
- Maybe it returns `(ScopeId, DefinitionId, &Definition)` like `definition_at()` does, because it is just as likely you may want immediate definition access too

crates/oak_db/src/file_resolve.rs
```diff
@@ -1,29 +1,160 @@
...
+    ///
+    /// For the offset-aware, sequential-semantics variant, see
+    /// [`File::resolve_at`].
     #[salsa::tracked]
     pub fn resolve(self, db: &'db dyn Db, name: Name<'db>) -> Option<Definition<'db>> {
+        if let Some(def) = self.resolve_export(db, name) {
+            return Some(def);
+        }
+
+        // For each sibling `ImportLayer::File`, check the target's exports
```
DavisVaughan, line 51:
```r
across <- function() {}

library(dplyr)

across<gotodefn>
```

When `PackageExports` are introduced, if they are part of the `imports() walk` then I'd expect this to give incorrect results with the current setup.

It seems like this would resolve to the local `across <-` because `self.resolve_export()` comes first, ignoring the positioning of `library(dplyr)` later on

DavisVaughan: I see that my example above assumes you put an `offset` somewhere, and technically you only are supposed to supply a `name` here, but IIUC you are supposed to assume that the implied offset here is at EOF, in which case I still think the answer is wrong.

crates/oak_db/src/file_resolve.rs
```diff
@@ -1,29 +1,160 @@
...
+
+        // 3. Top-level: exports chain (here) offset-narrowed imports (below).
+        if let Some(def) = self.resolve_export(db, interned) {
+            return Some(def);
+        }
+
+        // Same exports-only chase as `resolve`'s imports walk: avoids the
+        // sibling cycle, matches R's namespace semantics. TODO: Package-level
+        // layers.
+        for layer in self.imports_at(db, offset) {
```
DavisVaughan, line 140: I think I have the same concern here about the `across` example

crates/oak_db/src/file_resolve.rs
```diff
@@ -1,29 +1,160 @@
...
+    ///    uses the EOF state (full imports, source forwards). The function
+    ///    will run after the whole script/package is loaded, so we
+    ///    over-approximate accordingly.
+    /// 3. **At top-level (sequential context)**: walk the exports chain
+    ///    (handles `<-` and `source()` shadowing via `exports`'s
+    ///    last-wins ordering), then walk `imports_at(offset)` for the
+    ///    visible subset of attaches / collation predecessors.
+    ///
+    /// Not `#[salsa::tracked]` because keying on `(self, offset)` would
+    /// balloon the cache. `Definition` creation delegates to the tracked
```
DavisVaughan, line 96: That feels right to me. This feels like the start of one of the places where we are _heavily using_ the cache, but we aren't trying to cache this particular result. We're finally starting to answer some questions here, so it's reasonable to assume we are going to "do some work"

crates/oak_db/src/file_resolve.rs
```diff
@@ -1,29 +1,160 @@
...
+    /// [`File::resolve_export()`] and [`File::local_definition()`].
+    pub fn resolve_at(self, db: &'db dyn Db, offset: TextSize) -> Option<Definition<'db>> {
+        let index = self.semantic_index(db);
+        let (use_scope, use_id, _) = index.use_at(offset)?;
+        let use_symbol = index.uses(use_scope)[use_id].symbol();
```
DavisVaughan: You've got the `use` right there in the `_` and it knows its `.symbol()`, so no need to look up again

crates/oak_db/src/file_resolve.rs
```diff
@@ -1,29 +1,160 @@
...
+                return Some(self.local_definition(
+                    db,
+                    binding_scope,
+                    def_id,
+                    Name::new(db, name.as_str()),
+                ));
+            }
+        }
+
+        let (cursor_scope, _) = index.scope_at(offset);
```
DavisVaughan: You've already got this, its `use_scope`

crates/oak_db/src/file_resolve.rs
```diff
@@ -1,29 +1,160 @@
...
+            .symbol(use_symbol)
+            .name()
+            .to_string();
+
+        // Step 1, lexical. Function-scope hits return directly. File-scope
+        // hits fall through. `resolve_symbol()` only tracks IS_BOUND, but
+        // `exports()` orders bindings in source order so steps 2/3 pick the
+        // last winner between `<-` and a same-name `source()`.
+        let file_scope = ScopeId::from(0);
+        if let Some((binding_scope, def_id)) = index.resolve_symbol(&name, use_scope) {
```
DavisVaughan: yea, feels like maybe `resolve_symbol()` could be `definition_in()` or something?

crates/oak_db/src/file_resolve.rs
```diff
@@ -1,29 +1,160 @@
...
+            if binding_scope != file_scope {
```
DavisVaughan, line 113: And if `resolve_symbol()` returns `&IndexDefinition` as suggested elsewhere, then you don't even need `local_definition()`. You'll have the `def.kind()` right there available to you, no need to look up again.

But as mentioned in the other PR I'd really enjoy if we just had 1 `Definition` type you could just return from here directly 😢

crates/oak_db/src/file_imports.rs
```diff
@@ -42,43 +45,147 @@ const DEFAULT_SEARCH_PATH: [&str; 7] = [
...
+    /// (LIFO) priority order. Symbols that don't have local bindings (are
+    /// unbound in the file's semantic index) can be resolved against these
+    /// imports.
     ///
-    /// Offset-independent and stable across cursor moves. Recomputed
-    /// only when the file's package membership, NAMESPACE, or this
-    /// file's semantic calls actually change.
+    /// The last `library()` call comes before the first, and the latest
+    /// collation file before the earliest. The first hit in a forward search
+    /// then matches R's runtime semantics (last attached / latest sourced wins).
```
DavisVaughan: "`library()` calls further down the file come earlier in the returned `Vec`. Similarly, files collated last come earlier in the returned `Vec`. The first hit in a forward search then..."

?

crates/oak_db/src/file_imports.rs
```diff
@@ -42,43 +45,147 @@ const DEFAULT_SEARCH_PATH: [&str; 7] = [
...
+    ///
+    /// - **Cursor in lazy context**: returns the full lazy view. Lazy
+    ///   contexts like functions are treated as if they run after the
+    ///   file is fully sourced (over-approximation). Any `library()` /
+    ///   collation entry is potentially visible regardless of where it
+    ///   appears relative to the cursor.
+    ///
+    /// - **Top-level cursor (script)**: only `library()` calls that
+    ///   have run by `offset` (file-scope calls before the cursor).
+    ///   LIFO order, latest-attached comes first.
```
DavisVaughan:
```suggestion
    /// - **Top-level cursor (script)**: only `library()` calls that
    ///   have occurred before `offset`. Most recently attached comes
    ///   first.
```

crates/oak_db/src/file_imports.rs
```diff
@@ -42,43 +45,147 @@ const DEFAULT_SEARCH_PATH: [&str; 7] = [
...
+    ///
+    /// - **Top-level cursor (package)**: only collation predecessors
+    ///   of this file. LIFO order, latest-sourced predecessor comes
+    ///   first. The package imports and base namespace come last.
```
DavisVaughan, line 86:
```suggestion
    /// - **Top-level cursor (package)**: only collation predecessors
    ///   of this file. Most recently sourced predecessor comes
    ///   first. The package imports and base namespace come last.
```

crates/oak_db/src/file_imports.rs
```diff
@@ -42,43 +45,147 @@ const DEFAULT_SEARCH_PATH: [&str; 7] = [
...
+        // No File layers in `imports()` for a file that claims package
+        // membership. Either the package has only this file (and self
+        // is excluded from imports, so the File block is empty) or the
+        // populator is inconsistent. Conservative fallback: full
+        // imports.
+        return imports.clone();
+    };
+
+    // `imports()` emits the `files.len() - 1` non-self package entries
+    // in a contiguous File block, in *reverse* order (latest-sourced
```
DavisVaughan: Am I expecting to see a `rev()` call in this chain of functions?

crates/oak_db/src/file_imports.rs
```diff
@@ -42,43 +45,147 @@ const DEFAULT_SEARCH_PATH: [&str; 7] = [
...
+    // first). Successors of this file occupy the first
+    // `(files.len() - 1 - file_pos)` slots of that block; predecessors
+    // occupy the remainder, in LIFO order. For a top-level cursor we
+    // drop the successor prefix and keep predecessors + the surrounding
+    // namespace / base layers.
+    let drop_count = files.len() - 1 - file_pos;
+    imports[..start]
+        .iter()
+        .chain(&imports[start + drop_count..])
+        .cloned()
```
DavisVaughan: I'm fairly confused by this, I think an actual example in the comments would help

The combination of indices originating from `imports` (which may have package files, package imports, package exports, and base package exports) and `files` (just the R files of a package) is weirding me out, because i would not think they are compatible and yet i see `start + drop_count` being done

crates/oak_db/src/file_imports.rs
```diff
@@ -101,15 +208,19 @@ fn package_layers(_file: File, db: &dyn Db, package: Package) -> Vec<ImportLayer
...
     // TODO(scan): once `oak_scan` orders `Package.files` by the
     // DESCRIPTION `Collate` field, this iteration becomes
     // collation-ordered directly. Today `Package.files` is whatever
     // order the scanner produces.
-    for &pkg_file in package.files(db) {
+    for &pkg_file in package.files(db).iter().rev() {
+        if pkg_file == file {
+            continue;
+        }
         layers.push(ImportLayer::File(pkg_file));
```
DavisVaughan: Shouldn't `File` import layers be pushed _before_ the `PackageImports` and `PackageExports` above?

That would mean in the resulting `Vec` you end up with the `File` layers at the front of the vec, which I think is correct (local package definitions take precedence over some import from another package)

crates/oak_db/src/file_imports.rs
```diff
@@ -42,43 +45,147 @@ const DEFAULT_SEARCH_PATH: [&str; 7] = [
...
+        .rev()
+        .filter_map(|call| attach_layer(db, call))
+        .collect();
     extend_with_default_search_path(db, &mut layers);
 
     layers
 }
 
-fn package_layers(_file: File, db: &dyn Db, package: Package) -> Vec<ImportLayer> {
+fn package_layers(file: File, db: &dyn Db, package: Package) -> Vec<ImportLayer> {
```
DavisVaughan: I can't put this comment anywhere meaningful, but the naming of `ImportLayer::PackageExports` confuses me

We have:

- `ImportLayer::PackageImports` for `importFrom(pkg, name)`
- `ImportLayer::PackageExports` for `import(pkg)`

The mix of import and export in the same type name is hellllla confusing here. What about

- `ImportLayer::PackageImport` for `importFrom(pkg, name)`
- `ImportLayer::PackageBulkImport` for `import(pkg)`

Or something similar, the point being that we get `Export` out of the naming scheme

crates/oak_db/src/file_imports.rs
DavisVaughan: sounds like this is worth a test!

DavisVaughan (approved):
As mentioned for the other PR, if these directly replace utilities we are getting rid of, then I think I would like to see some TODOs about the removal of those things!

crates/oak_semantic/src/semantic_index.rs
lionel-: hmm `definition_at()` is a simple getter so is quite different despite the same return value. Keeping the "resolve" term is important here, since it's the standard term for this operation.

I've renamed to `resolve()` for symmetry with `File::resolve()`, since that's the direction we took in the other PR.

crates/oak_db/src/file_resolve.rs
lionel-: `library(dplyr)` resolves from the search path so it's really only hit for symbols unbound locally.

I know you know that no worries ;)

crates/oak_db/src/file_resolve.rs
lionel-: `Definition::new()` creates a salsa tracked struct and can only be called from inside a tracked function. `local_definition()` is that wrapper. `resolve_at` is intentionally not tracked so we have to go through `local_definition()`.

I don't like `local_definition()` too much so I've renamed to `intern_definition()`.

crates/oak_db/src/file_imports.rs
lionel-: I refactored for clarity.

crates/oak_db/src/file_imports.rs
lionel-: Good catch! Now fixed (and existing test fixed too)

crates/oak_db/src/file_imports.rs
lionel-: Agreed, I've struggled with these names. The current names come from the fact that with `import()` it's the imported package that decides what goes in via its exports, and with `importFrom()` it's the importing package that decides what goes in.

Here's what I managed to come up with:

```rs
pub enum ImportLayer {
    File(File),                          // Collation file
    From(HashMap<String, String>),       // importFrom(pkg, name)
    Package(Package),                    // import(pkg) / library(pkg)
}
```

Issue #1149, opened 2026-04-14
Implement goto-definition, find-references, and rename with oak's symbol resolution


lionel-:
Requires #1141 and #1148. Augmented by #1146.

- Goto definition: The semantic index provides file-local targets (#1141). For unbound symbols, we resort to cross-file resolution (#1148). These could land in a project file, in which case we provide that file's definition. Or these could be symbols from installed packages.

  In the latter case we can use either virtual files created from srcrefs, downloaded sources (https://github.com/posit-dev/positron/issues/2286), or our current namespace deparser fallback we use for the debugger (though we're considering removing it in favour of the new fallbacks just mentioned). See #1146. This will be a big UX improvement for advanced users. It allows jumping to any function's source in context, with comments and roxygen documentation available.

- Find references: ty and rust-analyzer both do a textual search across files, then resolve symbols to see if they match the definition for the symbol at cursor.

- Rename: Easy once we have the other two handlers. Find definitions and references, then rename.

No need for Salsa here, we can just recompute on the fly for the time being.

Main user-visible behaviour change to discuss: Symbols are resolved from the scope. Currently we just piggyback on the workspace indexer for all files without any scope consideration. For scripts in particular, this means symbols will be isolated, unless you're explicit about the provenance, e.g. by calling `source()`. We'd also allow declarations like `declare(source("foo.R"))` in case the side effect of sourcing is not desirable for some reason.

klmr:
Hi Lionel,

Apologies for hijacking this ticket to ask if there is any documentation/guidance for how `declare()` is (going to be) used by Ark. Due to time constraints I’m not following all R-devel discussions on this topic, and as far as I know the syntax or exact usage of `declare()` hasn’t been finalised yet. However, as soon as it’s being used I’d like to implement `declare()` hints in ‘[box](https://github.com/klmr/box)’ so that they can be picked up by static analysis tools. But for this I’d need to know how these hints are supposed to look like (and whether I *can* even implement suitable hints for dynamically-loaded code).

lionel-:
@klmr There are no norms or conventions as of now, and everything needs to be invented.

There were some discussions about NSE annotations at an R dev day a few years ago, but nothing concrete came out of this yet: https://bugs.r-project.org/show_bug.cgi?id=14354#c10

Our goal is to start experimenting with annotations to get a better idea of how they could work. And to start with we'll just hardcode a registry of semantic properties in the LSP.

[...]

PR #1230, opened 2026-05-25
Implement goto-definition for local symbols

### LINKED ISSUE #8631: Support for local definitions in "Go to definition"
Our "go to definition" handling is currently on par with RStudio, but in the landscape of LSPs we have incomplete support. Here is where we stand right now with static definitions:

- [x] Top-level function
- [x] Top-level variable (not supported in RStudio)
- [ ] Local variable
- [ ] Local variable inside an NSE function with local scope (`local()`, `with()`, etc)

While RStudio doesn't support jumping to variable definitions, this is a common LSP feature that I think we should support. In general, the "go to definition" feature should be consistent with "unknown symbol" diagnostics. If the LSP knows that a variable is defined then it should jump to the most relevant (closest) definition.

 We already made it possible to jump to top-level variables, which was mainly a byproduct of making top-level variables exported as workspace symbols (see https://github.com/posit-dev/positron/issues/5274). We still need to support local variables. Here are some examples of how this should all work:

```r
foo <- 1
foo # Jump to top-level variable 1

foo <- function(foo = 2) {
    foo # Jump to parameter 2

    foo <- 3
    foo # Jump to local variable 3

    if (cond) {
        foo <- 4
    }

    foo # Ambiguous, but I guess we want to jump to `foo <- 4` here

    if (cond) {
        foo <- 5
    } else {
        foo <- 6
    }
    foo # In this case jump to `foo <- 5`? I.e. first occurrence in the diverging paths?
        # Also fine to jump to 6 for consistency (last definition seen)
}

foo # Jump to top-level function (not variable)


local({
    foo <- 7
    foo # Jump to local variable 7
})

# Requires NSE handling: Ark should understand local scope inside `local()`
foo # Jump to top-level function

foo <- 8
foo <- foo + 1 # If cursor is on LHS, stay there. If cursor is on RHS, jump to 8
```

Currently Ark only supports top-level definitions because the "goto definition" handler is built around our workspace indexer.

To support local definitions, we'd ideally build a semantic index with symbol resolution information that would be used by multiple LSP features, including jump to definition. This sort of integrated data structure can wait for the independent LSP rewrite though, so we can design it holistically.

In the mean time, we could switch to the same sort of algorithm we use for "unknown symbol" diagnostics to add support for local variables. This first implementation would ignore any complications from NSE.

lionel-:
Part of https://github.com/posit-dev/positron/issues/13749
Closes https://github.com/posit-dev/positron/issues/8631 (should extract NSE bullet in its own issue, also affecting find-refs and rename).

Adds precise within-file goto-definition driven by Oak's semantic index.

The Salsa rewrite is taking more time and merging right before release seems a bit risky, given the added complexity. So it looks like full cross-file support for goto-def, find-refs, and rename is not going to make it to the June release. However we already have everything we need for within-file analysis.

This PR adds support for jumping to definition of locally-defined symbols (`def <-` and function parameters):

- Removes the `ARK_OAK_GOTO_DEF`-gated goto-definition handler, and all its non-Salsa baggage. That's a regression in our dev builds (we can no longer exercise the oak_sources stuff to visit installed package definitions), but that will be restored later this week with the proper Salsa implementation.

- Adds within-file support in Ark's goto-definition handler, based on Oak's semantic index. If the symbol is not bound in the file, we use the legacy indexer-based handler. This legacy handler includes all symbols of the same name in other files, including unrelated files. This will be fixed once we fully switch to Oak.

  Note that for conditionally unbound (via e.g. `if (cond) x <- 1`), we don't use the legacy handlers to populate external symbols, because the complexity wasn't worth the slight and temporary inconsistency.


crates/oak_semantic/src/semantic_index.rs
```diff
@@ -247,6 +247,43 @@ impl SemanticIndex {
...
+    /// also counts when `may_be_unbound` is true. That happens when the local
+    /// binding doesn't cover every control-flow path, so execution can fall
+    /// through to the outer scope.
+    ///
+    /// When `may_be_unbound` is false we deliberately skip the enclosing scope.
+    /// Otherwise a shadowed inner use would also bind to the outer def of the
+    /// same name.
+    pub fn reaching_definitions(
+        &self,
+        scope: ScopeId,
```
DavisVaughan, line 262: nit: since we have `Scope` and `ScopeId`, i have the mildest of consistency preference for `scope_id: ScopeId` and `scope: Scope` as a naming scheme (similarly for `Use` and `UseId`, etc)

crates/oak_ide/src/goto_definition.rs
```diff
@@ -1,57 +1,24 @@
...
+/// salsa-backed path and will be wired in later.
 pub fn goto_definition(
-    db: &dyn LegacyDb,
-    offset: TextSize,
-    file: &Url,
-    root: &RSyntaxNode,
     index: &SemanticIndex,
-    scope: &ExternalScope,
+    root: &RSyntaxNode,
+    pos: &FileOffset,
```
DavisVaughan:
```suggestion
    file_offset: &FileOffset,
```

?

When in doubt of a good name, I like explicit ones

DavisVaughan: Actually, this struct doesn't currently feel all that useful to me.

Looking at its uses, you kind of just combine into a `FileOffset` at the last moment only to destructure it. So I also would be fine with just `file: Url` and `offset: TextSize` as args (oh, I guess what I'm saying is that I was fine with what was there before)

crates/oak_ide/src/goto_definition.rs
```diff
@@ -61,163 +28,43 @@ pub fn goto_definition(
...
-    db: &dyn LegacyDb,
-    symbol: &str,
-    scope_chain: &[ScopeLayer],
-) -> Vec<NavigationTarget> {
-    let Some(external) = resolve_external_name(db.library(), scope_chain, symbol) else {
-        return Vec::new();
-    };
-    vec![external.into()]
+        })
+        .collect()
```
DavisVaughan, line 71: Nice!

crates/ark/src/lsp/main_loop.rs
```diff
@@ -189,7 +186,6 @@ impl GlobalState {
     ///   and auxiliary loop.
     pub(crate) fn new(
         client: Client,
-        r_home: PathBuf,
         console_notification_tx: TokioUnboundedSender<ConsoleNotification>,
```
DavisVaughan, line 190: I'm fairly confident we are going to need `r_home` in here and it did take some work to get it wired in right, can we just `_` it for now?

crates/ark/src/lsp/handlers.rs
```diff
@@ -279,13 +278,7 @@ pub(crate) fn handle_goto_definition(
...
 
-    if std::env::var("ARK_OAK_GOTO_DEF").is_ok() {
-        Ok(goto_definition(document, params, state).log_err().flatten())
-    } else {
-        Ok(goto_definition_legacy::goto_definition(document, params)
-            .log_err()
-            .flatten())
-    }
+    Ok(goto_definition(document, params).log_err().flatten())
 }
```
DavisVaughan, line 282: Probably very little risk of anything bad, right?

No salsa involved yet, so the main issue would be some slowness from computing a single file SemanticIndex on every goto definition call?

crates/ark/src/lsp/handler.rs
DavisVaughan, line 1: yea like id love to keep all this if possible

DavisVaughan (approved):

crates/oak_semantic/src/semantic_index.rs
DavisVaughan: it looks like future you agrees with me and made this change in https://github.com/posit-dev/ark/pull/1231

crates/oak_ide/src/goto_definition.rs
DavisVaughan: i do see how `FileRange` is a useful return value from `find_references()` in a future pr, so keeping it is fine

but maybe lets still change the `pos` name to `file_offset`?

lionel-: The offset is only meaningful in the context of the file, and the arguments are meant to be passed together, so gathering them in a struct reflects that ("keeps it honest" in claude-linguo).

That's what R-A does too. They call the type `FilePosition` even though the position field is `offset: TextSize`. I thought I'd go ahead and name it `FileOffset`

`file_offset` seems like an awfully long name for this parameter. How about just `offset`? In R-A the arguments take a `FilePosition` are called `position`. But unlike `position`, `offset` really does not evoke the file-anchored aspect. Now I'm tempted to just rename to `FilePosition` and use `position`.

I'm not sure discussing this sort of stuff asynchronously is useful, how about we take a few minutes to talk it through in our next sync?

crates/oak_ide/src/goto_definition.rs
lionel-: 2 definitions because there's a conditional path? yep nice!

crates/ark/src/lsp/handlers.rs
lionel-: yep but for a single file I'm not worried at all about performance

crates/oak_ide/src/goto_definition.rs
DavisVaughan: sure

PR #1231, opened 2026-05-25
Add support for local variables in `find-references`

lionel-:
Branched from https://github.com/posit-dev/ark/pull/1230

Part of https://github.com/posit-dev/positron/issues/13749

Adds a within-file Oak path to the existing find-references handler of the Ark LSP.

Same deal as for goto-definition: within-file behaviour is based on the semantic index and precise, cross-file is handled by the legacy handler and is approximate (matches any symbol of the same name without filtering based on expected definition).

### DIFF
--- a/crates/ark/src/lsp/find_references.rs
+++ b/crates/ark/src/lsp/find_references.rs
@@ -48,7 +48,7 @@ pub(crate) fn find_references(
         &document.line_index,
         document.position_encoding,
     )?;
-    let pos = oak_ide::FileOffset {
+    let pos = oak_ide::FilePosition {
         file: uri.clone(),
         offset,
     };
--- a/crates/ark/src/lsp/find_references.rs
+++ b/crates/ark/src/lsp/find_references.rs
@@ -1,16 +1,11 @@
-//
-// references.rs
-//
-// Copyright (C) 2022-2026 Posit Software, PBC. All rights reserved.
-//
-//
-
 use std::path::Path;
 
+use aether_lsp_utils::proto::from_proto;
+use aether_lsp_utils::proto::to_proto;
 use anyhow::anyhow;
 use stdext::result::ResultExt;
+use stdext::unwrap;
 use stdext::unwrap::IntoResult;
-use stdext::*;
 use tower_lsp::lsp_types::Location;
 use tower_lsp::lsp_types::Position;
 use tower_lsp::lsp_types::Range;
@@ -31,6 +26,68 @@ use crate::lsp::traits::url::UrlExt;
 use crate::treesitter::ExtractOperatorType;
 use crate::treesitter::NodeType;
 use crate::treesitter::NodeTypeExt;
+
+pub(crate) fn find_references(
+    params: ReferenceParams,
+    state: &WorldState,
+) -> anyhow::Result<Vec<Location>> {
+    let uri = params.text_document_position.text_document.uri;
+    let position = params.text_document_position.position;
+    let include_declaration = params.context.include_declaration;
+
+    let document = state.get_document(&uri)?;
+
+    let mut locations: Vec<Location> = Vec::new();
+
+    let index = document.semantic_index();
+    let root = document.syntax()?;
+
+    // Intra-file resolution is precise via the semantic index
+    let offset = from_proto::offset_from_position(
+        position,
+        &document.line_index,
+        document.position_encoding,
+    )?;
+    let pos = oak_ide::FileOffset {
+        file: uri.clone(),
+        offset,
+    };
+    let intra = oak_ide::find_references(&index, &root, &pos, include_declaration);
+
+    for file_range in intra {
+        let Some(range) = to_proto::range(
+            file_range.range,
+            &document.line_index,
+            document.position_encoding,
+        )
+        .log_err() else {
+            continue;
+        };
+        locations.push(Location::new(file_range.file, range));
+    }
+
+    if !locations.is_empty() {
+        // If the intra-file pass resolved cleanly, we have the precise answer.
+        // The textual cross-file walk can only add unrefined external matches
+        // which, for a locally-bound symbol, would just be unrelated noise.
+        return Ok(locations);
+    }
+
+    // Truly free variable: no within-file binding. Fall back to a textual walk
+    // over workspace folders (including the current file: intra-file gave
+    // nothing, so no dedup needed). No semantic refinement yet, so unrelated
+    // same-name symbols in other files leak through.
+    if let Ok(context) = build_context(&uri, position, state) {
+        for folder in state.workspace.folders.iter() {
+            if let Ok(path) = folder.to_file_path() {
+                lsp::log_info!("searching references in folder {}", path.display());
+                find_references_in_folder(&context, &path, &mut locations, state);
+            }
+        }
+    }
+
+    Ok(locations)
+}
 
 #[derive(Debug, PartialEq)]
 enum ReferenceKind {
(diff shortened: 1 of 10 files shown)


crates/oak_semantic/src/semantic_index.rs
```diff
@@ -189,21 +189,26 @@ impl SemanticIndex {
     }
 
     /// Iterate direct child scopes of `scope`.
-    pub fn child_scopes(&self, scope: ScopeId) -> ChildScopesIter<'_> {
-        let descendants = &self.scopes[scope].descendants;
-        ChildScopesIter {
+    pub fn child_scope_ids(&self, scope_id: ScopeId) -> ChildScopeIdsIter<'_> {
```
DavisVaughan, line 192: i approve of this naming!

crates/oak_semantic/src/semantic_index.rs
```diff
@@ -189,21 +189,26 @@ impl SemanticIndex {
...
             index: self,
             current: descendants.start,
             end: descendants.end,
         }
     }
 
+    /// Iterate over every scope in the file (source order, file scope first).
+    pub fn scope_ids(&self) -> impl Iterator<Item = ScopeId> + '_ {
+        self.scopes.iter().map(|(id, _)| id)
+    }
```
DavisVaughan, line 204: Do you think it would be more useful and general to just expose this?

```rust
pub fn scopes(&self) -> &IndexVec<ScopeId, Scope>
```

Or do you purposefully try not to directly expose the IndexVec?

crates/oak_semantic/src/semantic_index.rs
```diff
@@ -216,7 +221,7 @@ impl SemanticIndex {
         name: &str,
         scope: ScopeId,
     ) -> Option<(ScopeId, DefinitionId, &Definition)> {
-        for ancestor in self.ancestor_scopes(scope) {
+        for ancestor in self.ancestor_scope_ids(scope) {
```
DavisVaughan, line 224:
```suggestion
        scope_id: ScopeId,
    ) -> Option<(ScopeId, DefinitionId, &Definition)> {
        for ancestor_id in self.ancestor_scope_ids(scope) {
```

probably would be useful to do a Claude pass once all of this is merged that does a mass renaming for consistency

crates/oak_ide/src/identifier.rs
```diff
@@ -7,21 +7,28 @@ use biome_rowan::TextRange;
...
         def_id: DefinitionId,
+        def: &'a Definition,
+        name: &'a str,
     },
 
     Use {
         scope_id: ScopeId,
         use_id: UseId,
+        use_site: &'a Use,
+        name: &'a str,
```
DavisVaughan:
```suggestion
pub enum Identifier<'index> {
    Definition {
        scope_id: ScopeId,
        def_id: DefinitionId,
        def: &'index Definition,
        name: &'index str,
    },

    Use {
        scope_id: ScopeId,
        use_id: UseId,
        use_site: &'index Use,
        name: &'index str,
```

I often find myself going "but _what_ is the lifetime bound to?"

In this case it's likely always the semantic index, so this informative-lifetime-name pattern seems useful here. I've also seen this done in other rust projects (tree sitter's rust bindings) and i find it makes things a bit easier to read

crates/oak_semantic/src/semantic_index.rs
DavisVaughan: hmm no we do expose IndexVec elsewhere, so i do feel like this might be more useful to expose?

```
pub fn uses(&self, scope: ScopeId) -> &IndexVec<UseId, Use> {
```

crates/oak_ide/src/goto_definition.rs
DavisVaughan, line 1: the changes in `identifier.rs` make this feeeeel more smooth!

crates/ark/src/lsp/find_references.rs
```diff
@@ -118,38 +173,43 @@ fn build_context(uri: &Url, position: Position, state: &WorldState) -> anyhow::R
...
-                .into_result()?;
+                .descendant_for_point_range(back, back)
+                .filter(|n| n.is_identifier())
+            {
+                node = retry;
+            }
+        }
+        if !node.is_identifier() {
+            let fwd = Point::new(point.row, point.column + 1);
+            if let Some(retry) = ast
```
DavisVaughan, line 193: i dont remember the details right now but id be pretty hesitant to change this right now

im not sure we really have a big suite of tests for this, but it seems like this was from some hard fought debugging based on the comment

can we just leave it as is till we remove it for good?

crates/ark/src/lsp/tests/find_references.rs
```diff
@@ -0,0 +1,264 @@
...
+
+#[test]
+fn test_cursor_past_trailing_edge_resolves_via_fallback() {
+    // Some LSP clients send the cursor one past the last character of a
+    // selected identifier (typical for double-click then "find references").
+    // We exercise this through the cross-file fallback because `mutate` is
+    // unbound: the intra-file pass returns nothing, and `build_context`'s
+    // boundary retry pulls the cursor back one column onto the identifier
+    // before the textual walk runs.
+    let dir = tempfile::tempdir().unwrap();
```
DavisVaughan, line 118: can we also exercise this in the intra file case?

it is an important use case, because its probably how i do find references 99% of the time (double click first, which puts my flashing cursor at the end of `mutate`, then right click and "find references")

crates/oak_ide/src/find_references.rs
```diff
@@ -0,0 +1,113 @@
...
+use aether_syntax::RSyntaxNode;
+use oak_semantic::semantic_index::SemanticIndex;
+use oak_semantic::DefinitionId;
+use oak_semantic::ScopeId;
+
+use crate::FileOffset;
+use crate::FileRange;
+use crate::Identifier;
+
+/// Find all in-file references to the symbol at `pos`.
```
DavisVaughan:
```suggestion
/// Find all in-file references to the symbol at `file_offset`.
```

yea that just reads better to me 😛

crates/oak_ide/src/find_references.rs
```diff
@@ -0,0 +1,113 @@
...
+                .any(|d| target_defs.contains(&d));
+            if !intersects {
+                continue;
+            }
+            results.push(FileRange {
+                file: pos.file.clone(),
+                range: use_site.range(),
+            });
+        }
+    }
```
DavisVaughan, line 105: Whew, `find_references()` is complicated! So IIUC, with:

```r
foo <- function() {}

foo() # <- find references from here

bar()

foo()
```

We first `classify()` `foo` to figure out that it is a `Identifier::Use` and collect its `reaching_definitions()` of `foo <- function() {}` as `target_defs`

But that's really only the first part of `find_references()`. The real part is figuring out _where else_ this thing is used.

And I guess we don't directly store that info in the SemanticIndex on an individual Definition, so we have to hunt for it.

So we scan the file top to bottom looking for uses of the `foo` symbol, and we see if its set of reaching definitions match ours, and those become part of the final result.

---

Now it _totally_ makes sense why other groups do a textual scan as part of this feature.

Scanning through every single file in the workspace from top to bottom would nuke performance. So the textual scan can probably at least do the `symbol` matching loop for you across all files very quickly, and then we'd just be in charge of matching up the uses of all of the hits with our `target_defs`!

crates/oak_ide/src/find_references.rs
```diff
@@ -0,0 +1,113 @@
...
+
+    // Defs are emitted in `target_defs` (HashSet) iteration order, which
+    // is non-deterministic. Sort by start offset so callers see source
+    // order regardless of how we collected results.
+    results.sort_by_key(|r| r.range.start());
```
DavisVaughan, line 115: When we do cross-file resolution this will have to be tweaked, probably prioritizing the current file, and then showing all other files in arbitrary (alphabetical?) order. And within each file we'd probably still sort in source order

Might be worth a TODO(salsa)?

DavisVaughan (approved):
Did some basic testing and it seems to work

crates/oak_semantic/src/semantic_index.rs
lionel-: I'll add it to the salsa post-pr bullet list

crates/ark/src/lsp/find_references.rs
lionel-: I'm just expanding the workaround, with a test. I've added a test for the original workaround too. It doesn't seem like a very risky path, so I think I'd prefer to go ahead with the change. Otherwise we'll need to (a) keep a hole in the impl, and (b) keep a hole in the test coverage, both with fixmes.

crates/oak_ide/src/find_references.rs
lionel-: Without the textual search I guess we'd maintain a map of some sort.

crates/ark/src/lsp/tests/find_references.rs
lionel-: Added a test, and a fix because the test was failing.

We now have a rudimentary version of r-a's `pick_best_token()` that snaps to a neighbouring identifier token when the offset sits in between tokens.

PR #1232, opened 2026-05-26
Add support for renaming local variables

lionel-:
Branched from https://github.com/posit-dev/ark/pull/1231

Addresses https://github.com/posit-dev/positron/issues/13749

Implements rename for local definitions. If a symbol is unbound locally, an error is emitted. Cross-file renaming will come along with the Salsa infrastructure.

See usage in screencast. The most surprising behaviour is with nested lazy contexts where the enclosing snapshot reaches definitions and uses below the context. That's a consequence of the over-approximation we inherit from ty. The alternative would be to perform full call analysis to determine exactly what a nested symbol can reach (intractable in general).

https://github.com/user-attachments/assets/b0f589f3-b28f-47e7-a09f-c3643840c387

See user-facing errors in second screencast, when trying to rename something not renamable, or to an invalid identifier. There should be no backtraces in user visible popups.

https://github.com/user-attachments/assets/82fa38f8-fbfc-4d63-bfe7-4b692479b11b

### DIFF
--- a/crates/ark/src/lsp/handlers.rs
+++ b/crates/ark/src/lsp/handlers.rs
@@ -26,13 +26,16 @@ use tower_lsp::lsp_types::HoverContents;
 use tower_lsp::lsp_types::HoverParams;
 use tower_lsp::lsp_types::Location;
 use tower_lsp::lsp_types::MessageType;
+use tower_lsp::lsp_types::PrepareRenameResponse;
 use tower_lsp::lsp_types::ReferenceParams;
 use tower_lsp::lsp_types::Registration;
+use tower_lsp::lsp_types::RenameParams;
 use tower_lsp::lsp_types::SelectionRange;
 use tower_lsp::lsp_types::SelectionRangeParams;
 use tower_lsp::lsp_types::SignatureHelp;
 use tower_lsp::lsp_types::SignatureHelpParams;
 use tower_lsp::lsp_types::SymbolInformation;
+use tower_lsp::lsp_types::TextDocumentPositionParams;
 use tower_lsp::lsp_types::TextEdit;
 use tower_lsp::lsp_types::WorkspaceEdit;
 use tower_lsp::lsp_types::WorkspaceSymbolParams;
@@ -58,6 +61,7 @@ use crate::lsp::indent::indent_edit;
 use crate::lsp::input_boundaries::InputBoundariesParams;
 use crate::lsp::input_boundaries::InputBoundariesResponse;
 use crate::lsp::main_loop::LspState;
+use crate::lsp::rename;
 use crate::lsp::selection_range::convert_selection_range_from_tree_sitter_to_lsp;
 use crate::lsp::selection_range::selection_range;
 use crate::lsp::signature_help::r_signature_help;
@@ -325,6 +329,22 @@ pub(crate) fn handle_references(
     } else {
         Ok(Some(locations))
     }
+}
+
+#[tracing::instrument(level = "info", skip_all)]
+pub(crate) fn handle_prepare_rename(
+    params: TextDocumentPositionParams,
+    state: &WorldState,
+) -> LspResult<Option<PrepareRenameResponse>> {
+    Ok(rename::prepare_rename(params, state).log_err().flatten())
+}
+
+#[tracing::instrument(level = "info", skip_all)]
+pub(crate) fn handle_rename(
+    params: RenameParams,
+    state: &WorldState,
+) -> LspResult<Option<WorkspaceEdit>> {
+    Ok(rename::rename(params, state).log_err().flatten())
 }
 
 #[tracing::instrument(level = "info", skip_all)]
--- a/crates/ark/src/lsp/handlers.rs
+++ b/crates/ark/src/lsp/handlers.rs
@@ -345,8 +345,8 @@ pub(crate) fn handle_rename(
     params: RenameParams,
     state: &WorldState,
 ) -> LspResult<Option<WorkspaceEdit>> {
-    // Propagate error to the frontend (unlike `prepare_rename()`) to give
-    // actionable feedback to the user
+    // Propagate error to the frontend to give actionable feedback to the user.
+    // All errors thrown by `rename()` must be informative for users.
     Ok(rename::rename(params, state)?)
 }
 
--- a/crates/ark/src/lsp/handlers.rs
+++ b/crates/ark/src/lsp/handlers.rs
@@ -336,8 +336,7 @@ pub(crate) fn handle_prepare_rename(
     params: TextDocumentPositionParams,
     state: &WorldState,
 ) -> LspResult<Option<PrepareRenameResponse>> {
-    // Don't propagate errors to the frontend
-    Ok(rename::prepare_rename(params, state).log_err().flatten())
+    Ok(rename::prepare_rename(params, state)?)
 }
 
 #[tracing::instrument(level = "info", skip_all)]
--- a/crates/ark/src/lsp/handlers.rs
+++ b/crates/ark/src/lsp/handlers.rs
@@ -336,6 +336,7 @@ pub(crate) fn handle_prepare_rename(
     params: TextDocumentPositionParams,
     state: &WorldState,
 ) -> LspResult<Option<PrepareRenameResponse>> {
+    // Don't propagate errors to the frontend
     Ok(rename::prepare_rename(params, state).log_err().flatten())
 }
 
@@ -344,7 +345,9 @@ pub(crate) fn handle_rename(
     params: RenameParams,
     state: &WorldState,
 ) -> LspResult<Option<WorkspaceEdit>> {
-    Ok(rename::rename(params, state).log_err().flatten())
+    // Propagate error to the frontend (unlike `prepare_rename()`) to give
+    // actionable feedback to the user
+    Ok(rename::rename(params, state)?)
 }
 
 #[tracing::instrument(level = "info", skip_all)]
--- a/crates/ark/src/lsp/rename.rs
+++ b/crates/ark/src/lsp/rename.rs
@@ -26,7 +26,7 @@ pub(crate) fn prepare_rename(
     )?;
     let index = document.semantic_index();
     let tree = document.syntax()?;
-    let pos = oak_ide::FileOffset { file: uri, offset };
+    let pos = oak_ide::FilePosition { file: uri, offset };
 
     let Some((range, placeholder)) = oak_ide::prepare_rename(&index, &tree, &pos) else {
         return Ok(None);
@@ -55,7 +55,7 @@ pub(crate) fn rename(
     )?;
     let index = document.semantic_index();
     let root = document.syntax()?;
-    let pos = oak_ide::FileOffset {
+    let pos = oak_ide::FilePosition {
         file: uri.clone(),
         offset,
     };
--- a/crates/ark/src/lsp/rename.rs
+++ b/crates/ark/src/lsp/rename.rs
@@ -1,0 +1,83 @@
+use std::collections::HashMap;
+
+use aether_lsp_utils::proto::from_proto;
+use aether_lsp_utils::proto::to_proto;
+use tower_lsp::lsp_types;
+use tower_lsp::lsp_types::PrepareRenameResponse;
+use tower_lsp::lsp_types::RenameParams;
+use tower_lsp::lsp_types::TextDocumentPositionParams;
+use tower_lsp::lsp_types::TextEdit;
+use tower_lsp::lsp_types::WorkspaceEdit;
+
+use crate::lsp::state::WorldState;
+
+pub(crate) fn prepare_rename(
+    params: TextDocumentPositionParams,
+    state: &WorldState,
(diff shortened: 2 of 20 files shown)


crates/oak_core/src/identifier.rs
```diff
@@ -0,0 +1,169 @@
...
+
+use anyhow::anyhow;
+
+/// Convert a semantic name into its canonical R source form.
+///
+/// A name that parses as a plain R identifier (and isn't a reserved
+/// word) returns as-is. Anything else gets wrapped in backticks. Empty
+/// names, reserved words, and names containing a literal backtick return
+/// `Err` (a backtick can't appear inside a backtick-quoted identifier).
+pub fn to_identifier_text(name: &str) -> anyhow::Result<String> {
```
DavisVaughan, line 17: I slightly question the decision to use anyhow in our lowest level crate

Typically anyhow isn't used in a "library" style crate, but our use case is a bit blurry because our only consumers won't care much about what type of error this is.

If you care about this, I would not be against seeing a structured `IdentifierError` or something

crates/oak_core/src/identifier.rs
```diff
@@ -0,0 +1,169 @@
...
+//! emit or recognise identifier text.
+
+use anyhow::anyhow;
+
+/// Convert a semantic name into its canonical R source form.
+///
+/// A name that parses as a plain R identifier (and isn't a reserved
+/// word) returns as-is. Anything else gets wrapped in backticks. Empty
+/// names, reserved words, and names containing a literal backtick return
+/// `Err` (a backtick can't appear inside a backtick-quoted identifier).
```
DavisVaughan: Can we return a `Cow<str>`? I feel like 95% of the time we will have a valid identifier

DavisVaughan: (I see it won't matter much for our current usage, but it still seems useful in case we use this a lot)

crates/oak_core/src/identifier.rs
```diff
@@ -0,0 +1,169 @@
...
+        }
+    }
+    chars.all(|c| c.is_ascii_alphanumeric() || c == '.' || c == '_')
+}
+
+/// R reserved words that cannot be used as identifier names. Source:
+/// `?Reserved` in R. Note that `return` is a function, not a reserved
+/// word, so it's missing from this list (`return <- 1` is valid R).
+/// `_` became reserved in R 4.2 for use as the `|>` pipe placeholder.
+pub fn is_reserved(name: &str) -> bool {
```
DavisVaughan, line 79: are you sure you want these to be pub?

crates/oak_core/src/identifier.rs
```diff
@@ -0,0 +1,169 @@
...
+///
+/// R's rule: starts with a letter or `.`, then letters, digits, `.`, or
+/// `_`. A leading `.` followed by a digit is a number literal, not an
+/// identifier (e.g. `.5`).
+pub fn is_valid_identifier(name: &str) -> bool {
+    let mut chars = name.chars();
+    let Some(first) = chars.next() else {
+        return false;
+    };
+    if !(first.is_ascii_alphabetic() || first == '.') {
```
DavisVaughan: i think this will fail for valid non ascii identifiers, like `μ <- 1`

DavisVaughan: getting it fully correct might be Extremely Hard™

crates/oak_core/src/identifier.rs
```diff
@@ -0,0 +1,169 @@
...
+            if second.is_ascii_digit() {
+                return false;
+            }
+        }
+    }
+    chars.all(|c| c.is_ascii_alphanumeric() || c == '.' || c == '_')
+}
+
+/// R reserved words that cannot be used as identifier names. Source:
+/// `?Reserved` in R. Note that `return` is a function, not a reserved
```
DavisVaughan, line 76: funny how we both touched this this week https://github.com/r-lib/tree-sitter-r/pull/197

crates/oak_core/src/identifier.rs
```diff
@@ -0,0 +1,169 @@
...
+        return Err(anyhow!("Identifier cannot contain a backtick"));
+    }
+    Ok(format!("`{name}`"))
+}
+
+/// Whether `name` is a valid bare R identifier (no backticks needed).
+///
+/// R's rule: starts with a letter or `.`, then letters, digits, `.`, or
+/// `_`. A leading `.` followed by a digit is a number literal, not an
+/// identifier (e.g. `.5`).
```
DavisVaughan: and this 😆 https://github.com/r-lib/tree-sitter-r/pull/198

crates/ark_test/src/lsp_client.rs
```diff
@@ -198,6 +198,27 @@ impl LspClient {
...
+        let mut message = json!({
+            "jsonrpc": "2.0",
+            "id": id,
+            "method": method,
+        });
+        if !params.is_null() {
+            message["params"] = params;
+        }
+
+        self.send_raw(&message);
```
DavisVaughan, line 213: seems like there is room for a little helper used by this and `send_request()` above. looks to be the same in all but the last line

crates/ark_test/src/lsp_client.rs
```diff
@@ -278,6 +299,46 @@ impl LspClient {
         }
     }
 
+    /// Receive an expected error response for `id`. Returns the error
+    /// `message` field. Panics on success responses or malformed errors.
+    #[track_caller]
+    fn recv_error_response(&mut self, method: &str, id: i64) -> String {
```
DavisVaughan, line 274: share implementation with `recv_response()`, but use an `error: bool` flag to determine what the expected behavior is? seems like lots of duplication

crates/oak_ide/src/rename.rs
DavisVaughan, line 1: Random note that it would be nice to one day support glue like strings

crates/ark/src/lsp/rename.rs
```diff
@@ -0,0 +1,83 @@
...
+    let offset = from_proto::offset_from_position(
+        position,
+        &document.line_index,
+        document.position_encoding,
+    )?;
+    let index = document.semantic_index();
+    let tree = document.syntax()?;
+    let pos = oak_ide::FileOffset { file: uri, offset };
+
+    let Some((range, placeholder)) = oak_ide::prepare_rename(&index, &tree, &pos) else {
```
DavisVaughan, line 31: i think sending the original name back as `placeholder` is nice because you often only want to make a small tweak to an existing name!

crates/oak_ide/src/rename.rs
```diff
@@ -0,0 +1,79 @@
...
+pub fn prepare_rename(
+    index: &SemanticIndex,
+    root: &RSyntaxNode,
+    pos: &FileOffset,
+) -> Option<(TextRange, String)> {
+    let ident = Identifier::classify(index, root, pos.offset)?;
+    match ident {
+        Identifier::Definition { def, name, .. } => Some((def.range(), name.to_string())),
+        Identifier::Use { use_site, name, .. } => Some((use_site.range(), name.to_string())),
+        Identifier::NamespaceAccess { .. } => None,
```
DavisVaughan, line 38: I'm not entirely sure what this should even do with NamespaceAccesses

crates/oak_core/src/identifier.rs
DavisVaughan: ACTUALLY I now see that these errors actually make their way to the USER when they type in an invalid name!

In that case I now feel even stronger that they should be a structured error type, with a note that they display to the user as is, and maybe even some tests to validate their behavior if you dont have any yet

(GREAT user experience though!)

crates/ark/src/lsp/handlers.rs
```diff
@@ -327,6 +331,25 @@ pub(crate) fn handle_references(
     }
 }
 
+#[tracing::instrument(level = "info", skip_all)]
+pub(crate) fn handle_prepare_rename(
+    params: TextDocumentPositionParams,
+    state: &WorldState,
+) -> LspResult<Option<PrepareRenameResponse>> {
+    // Don't propagate errors to the frontend
```
DavisVaughan: Would you mention why? Because the protocol definitely encourages it

```
error: code and message set in case the element can’t be renamed. Clients should show the information in their user interface.
```

IIUC, it is because `prepare_rename()` isn't set up to return actionable errors to the user right now

crates/ark/src/lsp/handlers.rs
```diff
@@ -327,6 +331,25 @@ pub(crate) fn handle_references(
...
+    Ok(rename::prepare_rename(params, state).log_err().flatten())
+}
+
+#[tracing::instrument(level = "info", skip_all)]
+pub(crate) fn handle_rename(
+    params: RenameParams,
+    state: &WorldState,
+) -> LspResult<Option<WorkspaceEdit>> {
+    // Propagate error to the frontend (unlike `prepare_rename()`) to give
+    // actionable feedback to the user
```
```suggestion
    // Propagate error to the frontend to give actionable feedback to the user.
    // All errors thrown by `rename()` must be informative for users.
```

crates/ark/src/lsp/rename.rs
```diff
@@ -0,0 +1,83 @@
...
+    let document = state.get_document(&uri)?;
+
+    let offset = from_proto::offset_from_position(
+        position,
+        &document.line_index,
+        document.position_encoding,
+    )?;
+    let index = document.semantic_index();
+    let tree = document.syntax()?;
+    let pos = oak_ide::FileOffset { file: uri, offset };
```
DavisVaughan: another pos->file_offset rename if you end up doing that in the other PRs

crates/oak_core/src/identifier.rs
```diff
@@ -0,0 +1,169 @@
...
+    if name.is_empty() {
+        return Err(anyhow!("Identifier cannot be empty"));
+    }
+    if is_reserved(name) {
+        return Err(anyhow!("`{name}` is a reserved word in R"));
+    }
+    if is_valid_identifier(name) {
+        return Ok(name.to_string());
+    }
+    if name.contains('`') {
```
DavisVaughan, line 42: I think you should add a special case before this of "starts with and ends with a backtick"

I tried renaming a symbol to

```
`i know what im doing`
```

and it refused to allow me to do so, but i think this should be allowed. i know i can drop the backticks here, but i feel like i shouldnt have to because im an "expert" and gave you the right thing to begin with

DavisVaughan (approved):

crates/oak_core/src/identifier.rs
lionel-: I'm agnostic about it since it's an internal crate.

crates/oak_core/src/identifier.rs
lionel-: oops of course

crates/oak_ide/src/rename.rs
lionel-: yes! The one thing that's not nice is how we'll have to hardcode all of these r-lib/tidyverse idiosyncracies, but I don't see a practical alternative.

crates/oak_ide/src/rename.rs
lionel-: Would only work when the package is in the workspace. Sometimes you've got to qualify your own package for reason x or y (I had to do it in rlang). Sometimes you're working on two packages at a time and one of them calls the other with qualified names.

crates/oak_core/src/identifier.rs
lionel-: Makes sense!

crates/oak_core/src/identifier.rs
lionel-: hmm I'm very skeptical about the benefits of proactively creating error types unless we need specific handling somewhere in the stack, and quite confident about the costs.

crates/oak_core/src/identifier.rs
lionel-: Returning `Cow` only helps if `RenameTargets::new_text` is also `Cow` (so the caller doesn't `.into_owned()` and re-allocate). But that would require adding a lifetime to `RenameTargets`, which doesn't feel great right now.

crates/ark/src/lsp/handlers.rs
lionel-: I copied ty's behaviour which doesn't propagate. And then I got super confused when I demo'd rename to Hadley because I was expecting an error right away but instead had to pick a name first! We now propagate errors as recommended in the spec, good catch.

crates/oak_core/src/identifier.rs
lionel-: surprisingly easy!

crates/oak_core/src/identifier.rs
DavisVaughan: My argument was that the correct thing for `to_identifier_text()` to return is a `Cow` regardless of who calls it. Like, if you look at just that one function in isolation, a `Cow` makes sense to me because most of the time the string won't need to be changed.

As I mentioned here https://github.com/posit-dev/ark/pull/1232#discussion_r3312612139, I understand the current usage in `RenameTargets` doesn't benefit from it and would need to call `to_owned()`, but if we end up using this function elsewhere a lot, it could be useful to avoid the allocation

PR #1233, opened 2026-05-27
Return latest version from `.ps.rpc.pkg_outdated`

bricestacey:
`utils::old.packages()` already returns `ReposVer` (the latest version
in the user's configured repositories), so surface it alongside the
package name.

The Positron R extension uses this to prefer ark's session-resolved
`latestVersion` over P3M's generic upstream mirror in the packages pane
-- mirroring the pip/uv approach in
[positron#13453](https://github.com/posit-dev/positron/issues/13453).

### Shape change

`pkg_outdated` now returns:

```json
[{"name": "dplyr", "latestVersion": "1.1.5"}, ...]
```

This will break the packages pane in Positron until this PR lands https://github.com/posit-dev/positron/pull/13696 so we should merge this PR, update #13696 to use the new commit hash, and then merge it maybe.

### DIFF
--- a/crates/ark/src/modules/positron/packages_pane.R
+++ b/crates/ark/src/modules/positron/packages_pane.R
@@ -98,13 +98,21 @@
     as.list(version)
 }
 
-# Return the list of outdated pacakages.
+# Return the list of outdated packages with their latest available versions.
+#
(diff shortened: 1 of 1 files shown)


jmcphers (approved):
Image

PR #1237, opened 2026-05-28
Fix regression in find-references handler

lionel-:
Fixes:

```
  [***-electron] › test/***/tests/references/references.test.ts:26:6 › References › R - Verify References Pane Lists All Function References Across Files @:references @:web @:win @:ark 
```

The compatibility mechanism for cross-file references wasn't quite right. If there were any intra-file results, cross-file lookup was skipped entirely. Instead cross-file results should be included for both unbound symbols _and_ symbols bound to a top-level definition, since it might be validly sourced by another file. For locally bound symbols though, we should keep excluding cross-file references since those may not be validly referenced from another file.

### DIFF
--- a/crates/ark/src/lsp/find_references.rs
+++ b/crates/ark/src/lsp/find_references.rs
@@ -53,6 +53,7 @@ pub(crate) fn find_references(
         offset,
     };
     let intra = oak_ide::find_references(&index, &root, &pos, include_declaration);
+    let intra_resolved = !intra.is_empty();
 
     for file_range in intra {
         let Some(range) = to_proto::range(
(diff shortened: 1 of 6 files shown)


crates/ark/src/lsp/find_references.rs
```diff
@@ -66,22 +68,37 @@ pub(crate) fn find_references(
...
 
-    if !locations.is_empty() {
-        // If the intra-file pass resolved cleanly, we have the precise answer.
-        // The textual cross-file walk can only add unrefined external matches
-        // which, for a locally-bound symbol, would just be unrelated noise.
+    // Skip the cross-file textual walk when the target is function-scoped.
+    // Local bindings aren't visible to other files, so any same-name match
+    // elsewhere is by definition a different binding. TODO(salsa): the
+    // short-circuit moves inside `oak_ide::find_references` once cross-file
+    // resolution lands; the `function_scoped` flag goes away.
```
DavisVaughan:
```suggestion
    // resolution lands; the `locally_scoped` flag goes away.
```

DavisVaughan (approved):

Issue #1095, opened 2026-03-10
Remove usage of `r_task()` in the LSP


lionel-:
Parent issue: Remove usage of `r_task()` https://github.com/posit-dev/ark/issues/691

We should consider two setups:

- Ark + Oak. The LSP runs alongside a user-operated Jupyter kernel. It has access to the user's R session.
- Oak. The LSP runs in standalone mode. It still has access to a sidecar R session, with or (preferably) without the Jupyter harness.

There are 3 categories of handlers:

- **Static**: can be rewritten with static analysis (package metadata, workspace index, semantic index).
- **Help**: depend on R's help server (see `RHtmlHelp`), and will remain as R queries.
- **Dynamic**: require runtime object state (data frames defined in the session, environment variables, R options, ...)

The most important category is the first one. Rewriting on top of static analysis:

- Makes the standalone LSP actually useful
- Makes the LSP in Positron more accurate, and sensitive to where cursor position is (e.g. package completions provided only when cursor is after corresponding `library()` call).

### Rewriting with static analysis

Main sources of static information:

- **Package metadata**: reading from installed package files like `NAMESPACE` and `DESCRIPTION`, building on the existing `Library` / `Package` infrastructure used for diagnostics.

- **Source code**:  Whereas information like exports is readily available from package metadata, a lot of context is missing from these files, such as whether an export is a function, what its formals are, what are the internal symbols in a namepsace, etc. This information requires analysing top-level assignments in R source files, which aren't readily available in installed packages. That's the main blocking point. Some ideas in https://github.com/posit-dev/positron/issues/2286.

- **Semantic index**: a new index modelled after ty internals that tracks scopes and usage/definitions. Needed for position-sensitive completions and symbol resolution.

Package info will need to be merged with local scope information from the semantic index for callee resolution, function lookup, etc.

### How can we set up dynamic queries

Help and dynamic queries need to be resolved by R, from a session whose `.libPaths()` matches the LSP. In packages, help for development documentation provided via pkgload needs some more thinking.

The simplest setup is to have dynamic features only available when R is idle. The queries would be sent from the LSP to the Console event loop and processed in between execute requests.

But this could be improved:

[...]

PR #1239, opened 2026-05-28
Pull roxygen2 location from `Library`

DavisVaughan:
Part of https://github.com/posit-dev/ark/issues/1095

I'm going to assume that `Library` holds a good enough `.libPaths()` for roxygen2 completions purposes, and that it will only get better and more accurate (i.e. more dynamic to users changing `.libPaths()`) as time goes on


lionel- (approved):
Nice

PR #1241, opened 2026-05-28
Bump version

jmcphers:
We're > 200 commits past 251, so feels like time to bump the minor version.


lionel- (approved):
Is this to make the git distance smaller? I was wondering whether we'd now use Ark releases to communicate something meaningful on the Jupyter side (documented in changelog).

By the way how about we use the `--first-parent` option to compute the number of PRs merged (plus commits pushed directly to main) rather than the number of commits merged?

```sh
git --no-pager rev-list --count 0.1.251..main
#> 242

git --no-pager rev-list --first-parent --count 0.1.251..main
#> 49
```

This should make the frontend-side release tags clearer.

PR #1243, opened 2026-05-29
Add `oak_scan` crate with initial library scanning

lionel-:
Branched from https://github.com/posit-dev/ark/pull/1226

Introduces `oak_scan` with a library scanner. The LSP layer calls (in the next PR) the scanner on startup with `.libPaths()`. Each libpath is set as a library root in `oak_db`.

`oak_scan` is the only writer of Salsa inputs in `oak_db`, and is in charge of maintaining invariants, such as making sure backpointers are up-to-date.

Library roots are static for now. Watching them for changes (package removed or installed) would require our own filesystem watcher. Since we don't need one for the workspace (we'll use LSP file events), the machinery would be a bit too heavy for what we would gain for it at this stage. For now the LSP can be restarted to register changes to libpaths.


crates/oak_scan/tests/integration/library.rs
```diff
@@ -0,0 +1,546 @@
...
+    write_package(&tmp1.path().join("pkg1"), "pkg1", &[]);
+    write_package(&tmp2.path().join("pkg2"), "pkg2", &[]);
+    let mut db = OakDatabase::new();
+
+    let paths: Vec<PathBuf> = vec![tmp1.path().to_path_buf(), tmp2.path().to_path_buf()];
+    db.set_library_paths(&paths);
+
+    let roots = db.library_roots().roots(&db).clone();
+    assert_eq!(roots.len(), 2);
+    assert_eq!(roots[0].packages(&db)[0].name(&db), "pkg1");
```
DavisVaughan, line 100: Random thought

We have quite a few things in oak that are ordering dependent, but I always forget which is and is not order dependent

It feels like it would be cool to have `SortedVec` be returned here, or something kind of like it, where its clear that the result is sorted (then, of course, the question becomes "sorted by what"?).

But maybe `LibraryRoots` suffices, since its docs state that the order invariant is upheld there

crates/oak_scan/tests/integration/library.rs
DavisVaughan, line 1: Lets teach claude to always do `tests/integration/` for all integration tests!

crates/oak_scan/src/inputs.rs
```diff
@@ -0,0 +1,244 @@
...
+/// initial content whenever the helper mints a new `File` entity, i.e.
+/// the first time a URL is seen, whether at the initial scan or on a
+/// later rescan that discovers a newly-created file.
+///
+/// If a `File` already exists at this URL (scanner-created from an
+/// earlier scan, or VFS-created via `didOpen`), the helpers reuse that
+/// entity and leave its content alone. `set_contents` (driven by the
+/// VFS) is the authoritative way to update content.
+#[derive(Clone, Debug)]
+pub struct FileEntry {
```
DavisVaughan, line 48: I had the thought "is it always an R file?"

It seems to be, so it might be nice if the name reflected this, maybe just `RFile`? Or `RFileEntry`.

But the generic `FileEntry` left me wondering if it was also used for other things besides just R files

crates/oak_scan/src/inputs.rs
```diff
@@ -0,0 +1,244 @@
...
+    pub url: UrlId,
+    pub contents: String,
+}
+
+/// Extension methods on the database for scanner orchestration and
+/// placement-aware updates that don't have a natural `Root` receiver.
+pub trait DbExt: Db + DbInputs {
```
DavisVaughan: I would like to advocate for a more informative trait name than `DbExt`

Perhaps `DbLibraryPaths` or `DbScan`

crates/oak_scan/src/library.rs
```diff
@@ -0,0 +1,100 @@
...
+            Some(&r) => r,
+            None => scan_new_library_path(db, &path, url),
+        };
+        new_roots.push(root);
+    }
+    db.library_roots().set_roots(db).to(new_roots);
+}
+
+/// Initial scan of a path that wasn't previously a library root. Walks depth-1,
+/// calls `set_package` per package directory, returns the freshly-built `Root`.
```
DavisVaughan:
```suggestion
/// Initial scan of a path that wasn't previously a library root. Walks only the package directories,
/// not the package directory contents. Calls `set_package` per package directory, returns the freshly-built `Root`.
```

crates/oak_scan/src/packages.rs
```diff
@@ -0,0 +1,110 @@
+//! Filesystem-level R package discovery. Pure I/O, no salsa access.
+//! Reused by both the library scanner (which walks depth-1 over a
+//! library directory) and the workspace scanner (which walks the
```
DavisVaughan:
```suggestion
//! Reused by both the library scanner (which walks the package folders of a
//! library directory) and the workspace scanner (which walks the
```

I really don't understand this depth-1 terminology, and feel that just saying "walks the package folders but not their contents" is much more intuitive

crates/oak_scan/src/packages.rs
```diff
@@ -0,0 +1,110 @@
...
+use crate::inputs::FileEntry;
+
+/// One package discovered on disk: its `DESCRIPTION`-derived metadata
+/// plus the R files under `R/`.
+#[derive(Debug)]
+pub(crate) struct PackageDescriptor {
+    /// URL of the `DESCRIPTION` file. Stable identity for the `Package`
+    /// entity across rescans (see `Package::description_url`). The `name`
+    /// is *not* stable identity because two packages can declare the
+    /// same `Package:` field, and dedup picks one of them per root.
```
DavisVaughan: How is the url stable? Or maybe, what does stable mean here?

Because if I install a new version of a package in the middle of a session, that has the same url, but will be a different package

crates/oak_scan/src/packages.rs
```diff
@@ -0,0 +1,110 @@
...
+/// Read every `*.R` / `*.r` file directly under `r_dir`, in alphabetical
+/// order by basename. Subdirectories are skipped (the standard R package
+/// layout is flat). R files that fail to read are logged at warn level
+/// and skipped. Symlinks resolving to non-files (the `is_r_file` check)
+/// are skipped quietly.
+fn scan_r_files(r_dir: &Path) -> Vec<FileEntry> {
+    let mut entries: Vec<(PathBuf, String)> = Vec::new();
+    let Ok(read_dir) = fs::read_dir(r_dir) else {
+        // Normal for packages without an `R/` directory (data-only,
+        // header-only). Don't log.
```
DavisVaughan, line 75: IIUC this is ANYTHING in your library paths, i.e. this is fairly misleading

IIUC, the only way `scan_r_files()` is ever used is if `read_package()` is called on a workspace package. A library package will never have `R/` sources.

DavisVaughan: And since this is only called from `scan_new_library_path()`, I don't see how this is ever invoked

crates/oak_scan/src/packages.rs
```diff
@@ -0,0 +1,110 @@
...
+    entries
+        .into_iter()
+        .filter_map(|(path, contents)| {
+            let url = UrlId::from_file_path(&path).ok()?;
+            Some(FileEntry { url, contents })
+        })
+        .collect()
+}
+
+fn is_r_file(path: &Path) -> bool {
```
DavisVaughan, line 109: I think we have a helper for this somewhere?

crates/oak_scan/src/packages.rs
DavisVaughan: Oh maybe stable means "the contents may change" (like version), but when the contents change "replace me directly rather than making a new Root"

crates/oak_scan/src/stale.rs
```diff
@@ -0,0 +1,113 @@
...
+    // home, their old `Package` entity is itself in stale and the backpointer
+    // would lie about live containment. Setting to `None` here keeps the
+    // placement invariant honest.
+    for &file in &all_files {
+        file.set_package(db).to(None);
+    }
+
+    let (to_orphan, to_stale): (Vec<File>, Vec<File>) = all_files
+        .into_iter()
+        .partition(|f| editor_owned.is_some_and(|owned| owned.contains(f.url(db))));
```
DavisVaughan: Could save some time if you branch on `editor_owned`'s Option-ness earlier on outside the `partition()`

crates/oak_scan/src/stale.rs
```diff
@@ -0,0 +1,113 @@
...
+
+    if !to_orphan.is_empty() {
+        let orphan = db.orphan_root();
+        let mut files = orphan.files(db).clone();
```
DavisVaughan, line 78: Should `orphan.files()` and `stale.files()` be a `HashSet`? Lots of `contains()` on a Vec

crates/oak_scan/src/stale.rs
```diff
@@ -0,0 +1,113 @@
...
+        stale.set_packages(db).to(stale_packages);
+    }
+
+    // Clear the dropped root's containers and each package's files vec.
+    // The packages themselves now live in `stale_root.packages`; keeping
+    // their `files` populated would leave stale references that
+    // `package_by_url` can resurrect with inconsistent contents.
+    root.set_scripts(db).to(Vec::new());
+    for &pkg in &packages {
+        pkg.set_files(db).to(Vec::new());
```
DavisVaughan, line 114:
```suggestion
    for &package in &packages {
        package.set_files(db).to(Vec::new());
```

The naming inconsistency hurts me 😆

crates/oak_db/src/db.rs
```diff
@@ -53,6 +59,57 @@ pub trait Db: DbInputs {
...
+
+    /// Look up a `Package` by its `DESCRIPTION` URL.
+    ///
+    /// Walks workspace packages, then library packages, then falls back
+    /// to [`StaleRoot`]. Stale matches are intentional: scanner upserts
+    /// use this to find a `Package` entity whose live container was
+    /// dropped on a previous `set_*_paths` call, so the entity gets
+    /// reused on re-add. Analysis paths should not call this — they use
+    /// [`Db::package_by_name`] which is stale-blind.
+    fn package_by_url(&self, url: &UrlId) -> Option<Package>;
```
DavisVaughan: I have no idea if I should use this vs package_by_name based on the names alone 😢

I wonder if it can be `package_by_url_including_stale()` or something to really discourage usage of it

Or separate into `package_by_url()` and `stale_package_by_url()` for use by the scanner

I wonder if we are going to have to expose some ugly internals like this by having the scanner and db in separate crates :/

crates/oak_db/src/db.rs
```diff
@@ -53,6 +59,57 @@ pub trait Db: DbInputs {
...
+    /// the ownership query here. The shorter root's vec still transiently
+    /// lists the package, but it self-heals on its next scan since
+    /// `set_packages` replaces the vec wholesale.
+    fn root_by_package(&self, pkg: Package) -> Option<Root>;
+
+    /// All live roots in lookup-precedence order: workspace folders first, then
+    /// library paths (mirroring R's `.libPaths()`), then the orphan bucket.
+    /// Stale roots are not included. Salsa-cached and invalidates when one of
+    /// `workspace_roots` / `library_roots` / `orphan_root` changes.
+    fn live_roots(&self) -> &[LiveRoot];
```
DavisVaughan, line 82: I would also accept `Dead root` instead of `Stale root` since we have terminology like:
- live roots
- "resurrect"ing of dead/stale roots

crates/oak_db/src/db.rs
```diff
@@ -62,33 +119,79 @@ pub trait Db: DbInputs {
...
+        if let LiveRoot::Workspace(r) | LiveRoot::Library(r) = root {
+            if let Some(&pkg) = root_package_url_index(db, r).get(url) {
+                return Some(pkg);
+            }
         }
     }
-    for root in db.library_roots().roots(db) {
-        if let Some(&pkg) = root_package_index(db, *root).get(name) {
-            return Some(pkg);
+    stale_package_url_index(db).get(url).copied()
```
DavisVaughan: yea i would be very happy to see a pair of functions like
- `package_by_url_query()` / `stale_package_by_url_query()`
- `live_package_by_url_query()` / `dead_package_by_url_query()`

that feels less confusing to me in the long run, i feel like its a bit hard to tell exactly which set of roots are checked just from the function name. like, is the stale package set checked or not?

crates/oak_db/src/db.rs
```diff
@@ -62,33 +119,79 @@ pub trait Db: DbInputs {
...
+}
+
+/// Implementation of [`Db::root_by_package`]. Walks all live roots looking for
+/// `pkg` in their `packages` vec, picking the longest-path root on ties.
+pub fn root_by_package_query(db: &dyn Db, pkg: Package) -> Option<Root> {
```
DavisVaughan: should all these be pub(crate)?

DavisVaughan (approved):

crates/oak_scan/tests/integration/library.rs
lionel-: The roots are not sorted lexicographically though, or against a well defined sort key that can be easily used from elsewhere, so `SortedVec` would be a bit weird. It's more like they are ordered than sorted. I agree `LibraryRoots` suffices as the upholder of the invariant.

crates/oak_scan/src/inputs.rs
lionel-: Good idea, I like `DbScan` after the crate name.

crates/oak_scan/src/inputs.rs
lionel-: On the other hand there is nothing constraining the type to an R file, it's just that we only use it for R ones.

crates/oak_scan/src/packages.rs
lionel-: Claude really like hyphened terms. I'm fighting it but there's always going to be comments that leak through.

I find myself deleting comments more and more because it's so much work to rephrase them appropriately.

crates/oak_scan/src/stale.rs
lionel-: yep I know, sorry. Using short names in small scopes and longer names in larger ones is my writing style (https://www.lysator.liu.se/c/pikestyle.html). Not sure how to resolve our difference on this :-)

crates/oak_scan/src/packages.rs
lionel-: yep exactly. I've updated the comment:

```rust
    /// URL of the `DESCRIPTION` file. This is the identity key for the
    /// `Package` entity: the same path produces the same entity across
    /// rescans, even when the package's version or files change. So a
    /// version bump updates the existing entity in place rather than
    /// minting a new one (see `Package::description_url`). The `name`
    /// can't serve as identity because two packages can declare the
    /// same `Package:` field, and dedup picks one of them per root.
```

crates/oak_scan/src/packages.rs
lionel-: `read_package()` is called for workspace packages in the next PR

crates/oak_scan/src/packages.rs
lionel-: yep in oak_core, good catch

crates/oak_scan/src/stale.rs
lionel-: Good idea, I'll do it in https://github.com/posit-dev/ark/pull/1244 to avoid conflicts

crates/oak_db/src/db.rs
lionel-: hmm good point.

I moved the by-url lookup to oak_scan, which is the only one that should be concerned with stale roots.

If an analysis crate has a URL for a package, they can look up the name on disk and use `package_by_name()`, which resolves the effective package for that name.

crates/oak_db/src/db.rs
lionel-: That'd be fine, just a bit grim to my taste 😄

crates/oak_db/src/db.rs
lionel-: I think moving the lookup to oak_scan mostly resolves the confusion.

crates/oak_scan/src/packages.rs
lionel-: 🤣  Screenshot 2026-06-10 at 11 46 32

PR #1244, opened 2026-05-29
Add workspace scanning to `oak_scan` and wire it to the LSP

lionel-:
Branched from #1243
Progress towards #1212

This PR adds workspace scanning to `oak_scan` and wires it into the LSP. The scanner walks a workspace folder, finds packages (any `DESCRIPTION` file in the tree with a `Package` field, honoring `.gitignore`) and top-level R scripts, and registers them under a `Root` in `oak_db`.

The LSP feeds three event types into the scanner:

- `initialize` and `didChangeWorkspaceFolders` update workspace roots in the DB.
- `didChangeWatchedFiles` applies surgical updates for single R files. When a change to a `DESCRIPTION` file is detected, this triggers a full rescan of the containing root (package files might be demoted to scripts).

Editor-owned URLs (files the editor has open) get special handling because we ignore disk state for those. The editor contents are the source of truth. When a workspace is removed, these files survive and get placed in `OrphanRoot`. This way analysis keeps working when a user closes a workspace folder while a buffer from it is still open.

Files that leave a live workspace (folder removed, file deleted, buffer closed) route to `StaleRoot`. We keep them there for as long as they remain deleted, and we bring them back if they are added again. This allows reusing the `File` inputs without creating new ones. Salsa inputs are never garbage collected so reusing them (based on their URL key) avoids unbounded memory growth when e.g. switching git branches, which can delete and bring back many files at a time.

Workspace scans run synchronously in the main loop of the LSP in this PR. Since these operations are too heavy to be run synchronously, and would stall requests from user interactions, the next PR moves
them off the main loop.

### DIFF
--- a/crates/ark/src/lsp/diagnostics.rs
+++ b/crates/ark/src/lsp/diagnostics.rs
@@ -1130,7 +1130,6 @@ fn check_symbol_in_scope(
 #[cfg(test)]
 mod tests {
     use std::path::PathBuf;
-
     use std::sync::Mutex;
     use std::sync::OnceLock;
 
--- a/crates/ark/src/lsp/diagnostics.rs
+++ b/crates/ark/src/lsp/diagnostics.rs
@@ -1131,13 +1131,15 @@ fn check_symbol_in_scope(
 mod tests {
     use std::path::PathBuf;
 
+    use std::sync::Mutex;
+    use std::sync::OnceLock;
+
     use harp::eval::RParseEvalOptions;
     use oak_package_metadata::dcf::Dcf;
     use oak_package_metadata::description::Description;
     use oak_package_metadata::namespace::Namespace;
     use oak_semantic::library::Library;
     use oak_semantic::package::Package;
-    use once_cell::sync::Lazy;
     use stdext::SortedVec;
     use tower_lsp::lsp_types;
     use tower_lsp::lsp_types::Position;
@@ -1147,8 +1149,18 @@ mod tests {
     use crate::lsp::state::WorldState;
     use crate::r_task;
 
-    // Default state that includes installed packages and default scopes.
-    static DEFAULT_STATE: Lazy<WorldState> = Lazy::new(current_state);
+    // `WorldState` contains `OakDatabase`, which is `Send` but not `Sync`
+    // (salsa stores thread-local query state). That rules out `Lazy<WorldState>`
+    // in a static, since `Lazy` requires `Sync`. Wrap in a mutex and hand out
+    // fresh clones so each caller gets its own copy.
+    fn default_state() -> WorldState {
+        static CACHE: OnceLock<Mutex<WorldState>> = OnceLock::new();
+        CACHE
+            .get_or_init(|| Mutex::new(current_state()))
+            .lock()
+            .unwrap()
+            .clone()
+    }
 
     fn generate_diagnostics(doc: Document, state: WorldState) -> Vec<lsp_types::Diagnostic> {
         super::generate_diagnostics(doc, state, false)
@@ -1174,7 +1186,7 @@ foo
 
 1 }";
             let document = Document::new(text, None);
-            let diagnostics = generate_diagnostics(document, DEFAULT_STATE.clone());
+            let diagnostics = generate_diagnostics(document, default_state());
             assert_eq!(diagnostics.len(), 2);
 
             let diagnostic = diagnostics.first().unwrap();
@@ -1198,7 +1210,7 @@ foo
                 2 # hi there
             )";
             let document = Document::new(text, None);
-            let diagnostics = generate_diagnostics(document, DEFAULT_STATE.clone());
+            let diagnostics = generate_diagnostics(document, default_state());
             assert!(diagnostics.is_empty());
         })
     }
@@ -1208,7 +1220,7 @@ foo
         r_task(|| {
             let text = "base::";
             let document = Document::new(text, None);
-            let diagnostics = generate_diagnostics(document, DEFAULT_STATE.clone());
+            let diagnostics = generate_diagnostics(document, default_state());
             assert_eq!(diagnostics.len(), 1);
             let diagnostic = diagnostics.first().unwrap();
             insta::assert_snapshot!(diagnostic.message);
@@ -1221,7 +1233,7 @@ foo
             let text = "..1 + ..2 + 3";
             let document = Document::new(text, None);
 
-            let diagnostics = generate_diagnostics(document, DEFAULT_STATE.clone());
+            let diagnostics = generate_diagnostics(document, default_state());
 
             assert!(diagnostics.is_empty());
         })
@@ -1264,7 +1276,7 @@ foo
                 y + x + z
             ";
             let document = Document::new(text, None);
-            let diagnostics = generate_diagnostics(document.clone(), DEFAULT_STATE.clone());
+            let diagnostics = generate_diagnostics(document.clone(), default_state());
             assert!(diagnostics.is_empty());
         })
     }
@@ -1278,7 +1290,7 @@ foo
                 y + x
             ";
             let document = Document::new(text, None);
-            let diagnostics = generate_diagnostics(document.clone(), DEFAULT_STATE.clone());
+            let diagnostics = generate_diagnostics(document.clone(), default_state());
             assert!(diagnostics.is_empty());
         })
     }
@@ -1293,7 +1305,7 @@ foo
             ";
             let document = Document::new(text, None);
 
-            let diagnostics = generate_diagnostics(document.clone(), DEFAULT_STATE.clone());
+            let diagnostics = generate_diagnostics(document.clone(), default_state());
             assert_eq!(diagnostics.len(), 1);
 
             // Only marks the `x` before the `x <- 1`
@@ -1312,7 +1324,7 @@ foo
                 identity(~foo)
             ";
             let document = Document::new(text, None);
-            let diagnostics = generate_diagnostics(document, DEFAULT_STATE.clone());
+            let diagnostics = generate_diagnostics(document, default_state());
             assert!(diagnostics.is_empty());
         })
     }
@@ -1329,7 +1341,7 @@ foo
 
             let document = Document::new(code, None);
 
-            let diagnostics = generate_diagnostics(document.clone(), DEFAULT_STATE.clone());
+            let diagnostics = generate_diagnostics(document.clone(), default_state());
             assert_eq!(diagnostics.len(), 1);
 
             let diagnostic = diagnostics.first().unwrap();
@@ -1349,7 +1361,7 @@ foo
 
             let document = Document::new(code, None);
 
-            let diagnostics = generate_diagnostics(document.clone(), DEFAULT_STATE.clone());
+            let diagnostics = generate_diagnostics(document.clone(), default_state());
             assert_eq!(diagnostics.len(), 1);
 
             let diagnostic = diagnostics.first().unwrap();
@@ -1370,7 +1382,7 @@ foo
 
             let document = Document::new(code, None);
 
-            let diagnostics = generate_diagnostics(document.clone(), DEFAULT_STATE.clone());
+            let diagnostics = generate_diagnostics(document.clone(), default_state());
             assert_eq!(diagnostics.len(), 1);
 
             let diagnostic = diagnostics.first().unwrap();
@@ -1390,7 +1402,7 @@ foo
 
             let document = Document::new(code, None);
 
-            let diagnostics = generate_diagnostics(document.clone(), DEFAULT_STATE.clone());
+            let diagnostics = generate_diagnostics(document.clone(), default_state());
             assert_eq!(diagnostics.len(), 1);
 
             let diagnostic = diagnostics.first().unwrap();
@@ -1408,7 +1420,7 @@ foo
 
             let document = Document::new(code, None);
 
-            let diagnostics = generate_diagnostics(document.clone(), DEFAULT_STATE.clone());
+            let diagnostics = generate_diagnostics(document.clone(), default_state());
             assert_eq!(diagnostics.len(), 0);
         })
     }
@@ -1423,7 +1435,7 @@ foo
 
             let document = Document::new(code, None);
 
-            let diagnostics = generate_diagnostics(document.clone(), DEFAULT_STATE.clone());
+            let diagnostics = generate_diagnostics(document.clone(), default_state());
             assert_eq!(diagnostics.len(), 0);
         })
     }
@@ -1444,7 +1456,7 @@ foo
             ";
             let document = Document::new(code, None);
             assert_eq!(
-                generate_diagnostics(document.clone(), DEFAULT_STATE.clone()).len(),
+                generate_diagnostics(document.clone(), default_state()).len(),
                 0
             );
 
@@ -1454,7 +1466,7 @@ foo
             ";
             let document = Document::new(code, None);
             assert_eq!(
-                generate_diagnostics(document.clone(), DEFAULT_STATE.clone()).len(),
+                generate_diagnostics(document.clone(), default_state()).len(),
                 0
             );
         });
@@ -1468,7 +1480,7 @@ foo
             ";
             let document = Document::new(code, None);
             assert_eq!(
-                generate_diagnostics(document.clone(), DEFAULT_STATE.clone()).len(),
+                generate_diagnostics(document.clone(), default_state()).len(),
                 0
             );
 
@@ -1479,7 +1491,7 @@ foo
             ";
             let document = Document::new(code, None);
             assert_eq!(
-                generate_diagnostics(document.clone(), DEFAULT_STATE.clone()).len(),
+                generate_diagnostics(document.clone(), default_state()).len(),
                 0
             );
         });
@@ -1493,7 +1505,7 @@ foo
             ";
             let document = Document::new(code, None);
             assert_eq!(
-                generate_diagnostics(document.clone(), DEFAULT_STATE.clone()).len(),
+                generate_diagnostics(document.clone(), default_state()).len(),
                 0
             );
 
@@ -1504,7 +1516,7 @@ foo
             ";
             let document = Document::new(code, None);
             assert_eq!(
-                generate_diagnostics(document.clone(), DEFAULT_STATE.clone()).len(),
+                generate_diagnostics(document.clone(), default_state()).len(),
                 0
             );
         });
@@ -1523,7 +1535,7 @@ foo
             ";
             let document = Document::new(code, None);
             assert_eq!(
-                generate_diagnostics(document.clone(), DEFAULT_STATE.clone()).len(),
+                generate_diagnostics(document.clone(), default_state()).len(),
                 0
             );
 
@@ -1535,7 +1547,7 @@ foo
             ";
             let document = Document::new(code, None);
             assert_eq!(
-                generate_diagnostics(document.clone(), DEFAULT_STATE.clone()).len(),
+                generate_diagnostics(document.clone(), default_state()).len(),
                 0
             );
 
@@ -1546,7 +1558,7 @@ foo
             ";
             let document = Document::new(code, None);
             assert_eq!(
-                generate_diagnostics(document.clone(), DEFAULT_STATE.clone()).len(),
+                generate_diagnostics(document.clone(), default_state()).len(),
                 1
             );
         });
@@ -1563,7 +1575,7 @@ foo
             ";
             let document = Document::new(code, None);
             assert_eq!(
-                generate_diagnostics(document.clone(), DEFAULT_STATE.clone()).len(),
+                generate_diagnostics(document.clone(), default_state()).len(),
                 0
             );
 
@@ -1576,7 +1588,7 @@ foo
             ";
             let document = Document::new(code, None);
             assert_eq!(
-                generate_diagnostics(document.clone(), DEFAULT_STATE.clone()).len(),
+                generate_diagnostics(document.clone(), default_state()).len(),
                 0
             );
         });
@@ -1589,7 +1601,7 @@ foo
             ";
             let document = Document::new(code, None);
             assert_eq!(
-                generate_diagnostics(document.clone(), DEFAULT_STATE.clone()).len(),
+                generate_diagnostics(document.clone(), default_state()).len(),
                 0
             );
         });
@@ -1605,7 +1617,7 @@ foo
             ";
             let document = Document::new(code, None);
             assert_eq!(
-                generate_diagnostics(document.clone(), DEFAULT_STATE.clone()).len(),
+                generate_diagnostics(document.clone(), default_state()).len(),
                 0
             );
 
@@ -1617,7 +1629,7 @@ foo
             ";
             let document = Document::new(code, None);
             assert_eq!(
-                generate_diagnostics(document.clone(), DEFAULT_STATE.clone()).len(),
+                generate_diagnostics(document.clone(), default_state()).len(),
                 0
             );
 
@@ -1628,7 +1640,7 @@ foo
             ";
             let document = Document::new(code, None);
             assert_eq!(
-                generate_diagnostics(document.clone(), DEFAULT_STATE.clone()).len(),
+                generate_diagnostics(document.clone(), default_state()).len(),
                 0
             );
         })
@@ -1659,7 +1671,8 @@ foo
             // Simulate a search path with `library` in scope
             let console_scopes = vec![vec!["library".to_string()]];
 
-            // Whereas `DEFAULT_STATE` contains base package attached, this world state
+            // Whereas `default_state()` returns a state with the base package
+            // attached, this world state
             // only contains `mockpkg` as installed package and `library()` on
             // the search path.
             let state = WorldState {
@@ -1861,7 +1874,7 @@ foo
             let library = Library::new(vec![]).insert("penguins", palmerpenguins_pkg);
 
             // Simulate a world state with the penguins package installed and attached
-            let mut state = DEFAULT_STATE.clone();
+            let mut state = default_state();
             state.library = library;
             state.console_scopes = vec![vec!["library".to_string()]];
 
--- a/crates/ark/src/lsp/diagnostics.rs
+++ b/crates/ark/src/lsp/diagnostics.rs
@@ -1130,8 +1130,6 @@ fn check_symbol_in_scope(
 #[cfg(test)]
 mod tests {
     use std::path::PathBuf;
-    use std::sync::Mutex;
-    use std::sync::OnceLock;
 
     use harp::eval::RParseEvalOptions;
     use oak_package_metadata::dcf::Dcf;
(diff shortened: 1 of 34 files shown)


crates/ark/src/lsp/state_handlers.rs
```diff
@@ -216,6 +237,21 @@ pub(crate) fn initialize(
...
+/// Falls back to the legacy `rootUri` for clients that don't send
+/// multi-root workspaces. The returned vector is empty when neither
+/// produces a folder, i.e. the server is running in single-file mode.
+pub(super) fn effective_workspace_uris(params: &InitializeParams) -> Vec<Url> {
+    if let Some(folders) = params.workspace_folders.as_ref() {
+        if !folders.is_empty() {
+            return folders.iter().map(|f| f.uri.clone()).collect();
+        }
+    }
+    params.root_uri.iter().cloned().collect()
```
DavisVaughan: Do you know of any clients? I don't

DavisVaughan: and really if you are going to do this, i think we should respect the docs on `workspace_folders`

```
    /// This property is only available if the client supports workspace folders.
    /// It can be `null` if the client supports workspace folders but none are
    /// configured.
```

i.e. maybe we check the client configuration to see if workspace folders are supported. and if so, we ONLY look at `workspace_folders` and dont do any fallback

and supposedly you're also supposed to look at `root_path` too if `root_uri` isn't there...

but again id be happy with just blindly taking `workspace_folders` as we've been doing

crates/ark/src/lsp/state_handlers.rs
```diff
@@ -86,49 +95,61 @@ pub(crate) fn initialize(
...
+    let mut workspace_paths: Vec<PathBuf> = Vec::new();
+
+    for uri in workspace_uris {
+        state.workspace.folders.push(uri.clone());
+        if let Ok(path) = uri.to_file_path() {
+            workspace_paths.push(path.clone());
+            // Try to load package from this workspace folder and set as
+            // root if found. This means we're dealing with a package
+            // source.
+            if state.root.is_none() {
```
DavisVaughan, line 112: Are we eventually going to remove this `root` field? Probably right? It's not really correct for multi root workspaces anyways, and I feel like we will have this stored elsewhere

crates/ark/src/lsp/state.rs
```diff
@@ -57,6 +58,9 @@ pub(crate) struct WorldState {
     /// libraries. Lazily populated.
     pub(crate) library: Library,
 
+    /// Salsa input tree for Oak queries.
+    pub(crate) oak: OakDatabase,
```
DavisVaughan: I would like to strongly advocate for `db` naming scheme like everywhere else!

crates/ark/src/lsp/state_handlers.rs
```diff
@@ -86,49 +95,61 @@ pub(crate) fn initialize(
...
+    // no `didOpen` has fired yet, but build the set through the same shape we
+    // use elsewhere so the call site reads consistently.
+    let editor_owned: HashSet<UrlId> = state
+        .documents
+        .keys()
+        .map(|url| UrlId::from_url(url.clone()))
+        .collect();
+    state
+        .oak
+        .set_workspace_paths(&workspace_paths, &editor_owned);
```
DavisVaughan: I think just seeing something like

```
// Start first round of indexing. We are initializing, so no documents have been opened yet.
set_workspace_paths(&workspace_paths, &HashSet::new())
```

is less confusing

crates/ark/src/lsp/state_handlers.rs
```diff
@@ -260,6 +299,10 @@ pub(crate) fn did_change(
 
     document.on_did_change(parser, &params);
 
+    let new_contents = document.contents.to_string();
+    let url_id = UrlId::from_url(uri.clone());
+    state.oak.upsert_editor(url_id, new_contents);
```
DavisVaughan: Maybe just inline the temporary variables? I thought for a second that they were being given names so they could be used elsewhere. That also avoids the need to come up with a new name for them.

DavisVaughan: (Here and in the other handlers if you make this change)

crates/oak_scan/src/inputs.rs
```diff
@@ -62,12 +68,144 @@ pub trait DbExt: Db + DbInputs {
...
+
+    fn add_watched_file(&mut self, url: UrlId, contents: String) {
+        watch::add_watched_file(self, url, contents);
+    }
+
+    fn remove_watched_file(&mut self, url: UrlId) {
+        watch::remove_watched_file(self, url);
+    }
+
+    fn apply_watcher_events(&mut self, events: Vec<FileEvent>, skip: &HashSet<UrlId>) {
```
DavisVaughan: Do you want to use consistent `editor_owned` parameter naming here?

DavisVaughan: I think that would be nice since `editor_owned` is also used elsewhere in this file and it means the same thing

crates/ark/src/lsp/state_handlers.rs
```diff
@@ -350,6 +396,76 @@ pub(crate) fn did_rename_files(
     Ok(())
 }
 
+#[tracing::instrument(level = "info", skip_all)]
+pub(crate) fn did_change_watched_files(
```
DavisVaughan, line 393: Hmm we now have
- did_create_files / did_delete_files / did_rename_files
- did_change_watched_files

Is there any room for confusion there? It seems like there is a lot of overlap there.

Maybe we get a did_create_files notif for EVERY kind of created file, but only a did_change_watched_files notif for explicit ones we are watching? I'm not sure.

crates/ark/src/lsp/handlers.rs
```diff
@@ -94,6 +97,27 @@ pub(crate) async fn handle_initialized(
     // Register capabilities to the client
     let mut regs: Vec<Registration> = vec![];
 
+    // Watch R files and DESCRIPTION across the workspace. The client
+    // sends `workspace/didChangeWatchedFiles` on disk changes that the
+    // editor itself didn't make (git checkouts, external edits, etc.).
```
DavisVaughan: Is it ONLY for edits the editor itself didn't make?

If so, how do you know that? I don't get that impression from reading https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_didChangeWatchedFiles

crates/ark/src/lsp/main_loop.rs
```diff
@@ -637,6 +642,28 @@ fn send_auxiliary(event: AuxiliaryEvent) {
...
+#[cfg(test)]
+pub(crate) fn init_aux_for_test() {
+    use std::sync::OnceLock;
+    static RX_GUARD: OnceLock<TokioUnboundedReceiver<AuxiliaryEvent>> = OnceLock::new();
+    RX_GUARD.get_or_init(|| {
+        let (tx, rx) = tokio_unbounded_channel::<AuxiliaryEvent>();
+        *AUXILIARY_EVENT_TX.write().unwrap() = Some(tx);
+        rx
+    });
+}
```
DavisVaughan, line 656: I think I am a bit uncomfortable with this 

I don't think we need the notes about parallel-test-safe, we run tests with nextest so this isn't an issue

I think I would appreciate if we create a sender/receiver pair here and return the receiver rather than storing it. Then in the test i think it would be nice if we could assert! that we got the expected message. in this case it looks like a `publish_diagnostics()` after the `did_close()` that clears the diagnostics.

I think that would be nice and would help me when writing tests. If I hit an auxiliary panic in my test, I can add this and get a `auxiliary_rx` that I can `read()` on, expecting the message that I previously panicked on. I think that sounds nicer than just throwing everything out.

crates/ark/src/lsp/diagnostics.rs
```diff
@@ -1147,8 +1148,18 @@ mod tests {
...
+        static CACHE: OnceLock<Mutex<WorldState>> = OnceLock::new();
+        CACHE
+            .get_or_init(|| Mutex::new(current_state()))
+            .lock()
+            .unwrap()
+            .clone()
+    }
 
     fn generate_diagnostics(doc: Document, state: WorldState) -> Vec<lsp_types::Diagnostic> {
         super::generate_diagnostics(doc, state, false)
```
DavisVaughan: This seems a bit silly in nextest world. Each test is in a separate process so there is no sharing between tests.

The only thing that would be shared is a cached `console_inputs()` anyways

I vote we just remove this and call `current_state()` everywhere instead.

crates/oak_db/src/inputs.rs
```diff
@@ -205,11 +205,23 @@ pub struct Package {
...
+    /// the loadable namespace: `tests/`, `inst/`, `vignettes/`,
+    /// `data-raw/`, etc. These get LSP analysis (parse, semantic index)
+    /// but aren't loaded with the package, so name resolution treats
+    /// them as standalone scripts that just happen to live next to the
+    /// package's code.
+    ///
+    /// **Placement invariant.** Same as [`Self::files`]: backpointer
+    /// stays `Some(self)`, file lives in one of the two containers.
+    #[returns(ref)]
+    pub scripts: Vec<File>,
```
DavisVaughan: One edge case is `testthat/helper-foo.R` which is loaded with testthat and should be goto-defn-able from `test-foo.R`

crates/oak_db/src/inputs.rs
```diff
@@ -205,11 +205,23 @@ pub struct Package {
...
-    /// `oak_scan`'s helpers, which keep the back-pointer and the
-    /// container in sync.
+    /// `package(db) == Some(self)`, and a file with
+    /// `package == Some(self)` must live here or in [`Self::scripts`].
+    /// Call this setter only through `oak_scan`'s helpers, which keep
+    /// the back-pointer and the container in sync.
     #[returns(ref)]
     pub files: Vec<File>,
+    /// Other R files inside the package directory that aren't part of
+    /// the loadable namespace: `tests/`, `inst/`, `vignettes/`,
```
DavisVaughan: I probably would not list vignettes/ here

crates/oak_scan/src/stale.rs
```diff
@@ -101,13 +102,14 @@ pub(crate) fn set_root_stale<DB: Db + DbInputs>(
...
-    // Clear the dropped root's containers and each package's files vec.
-    // The packages themselves now live in `stale_root.packages`; keeping
+    // Clear the dropped root's containers and each package's files / scripts
+    // vec. The packages themselves now live in `stale_root.packages`. Keeping
     // their `files` populated would leave stale references that
     // `package_by_url` can resurrect with inconsistent contents.
     root.set_scripts(db).to(Vec::new());
     for &pkg in &packages {
         pkg.set_files(db).to(Vec::new());
+        pkg.set_scripts(db).to(Vec::new());
```
DavisVaughan, line 111: Do we want to change `set_files()` to `set_sources()`? i.e. with a field of `sources` instead?

crates/oak_scan/src/packages.rs
```diff
@@ -1,21 +1,24 @@
...
 use stdext::result::ResultExt;
 
 use crate::inputs::FileEntry;
 
 /// One package discovered on disk: its `DESCRIPTION`-derived metadata
-/// plus the R files under `R/`.
+/// plus the R files under `R/`, plus any package-internal R files in
+/// `tests/`, `inst/`, etc. (populated only by the workspace scanner).
 #[derive(Debug)]
 pub(crate) struct PackageDescriptor {
```
DavisVaughan: Why is this not just `Package`? `PackageDescriptor` means nothing meaningful to me 😛

crates/oak_scan/src/packages.rs
```diff
@@ -108,3 +119,163 @@ fn is_r_file(path: &Path) -> bool {
     };
     ext.eq_ignore_ascii_case("r")
 }
+
+/// Read just the package name from `dir/DESCRIPTION`. Cheaper than
+/// [`read_package`] when the caller only needs to look up an existing
+/// `Package` by name.
+pub(crate) fn read_description_name(dir: &Path) -> Option<String> {
```
DavisVaughan: Maybe should take `description_url` as argument? Might be less ambiguous than `dir`?

crates/oak_scan/src/packages.rs
```diff
@@ -26,12 +29,19 @@ pub(crate) struct PackageDescriptor {
...
+    /// scanner doesn't recurse into pkg_dir for these).
+    pub scripts: Vec<FileEntry>,
     pub collation: Option<Vec<String>>,
 }
 
 /// Read a candidate package directory. Returns `None` if `DESCRIPTION`
-/// is missing or malformed.
+/// is missing or malformed. Populates `files` (R/*.R) only; `scripts`
+/// is filled by [`scan_workspace`] for workspace packages.
 pub(crate) fn read_package(dir: &Path) -> Option<PackageDescriptor> {
```
DavisVaughan: It seems quite wrong to me that this doesn't populate `scripts`

What is the reason for that? It only seems to be used:
- Here, where it immediately populates `scripts`
- In `scan_new_library_path()`, where there probably aren't any `scripts`

I am also fairly confident we are going to want two different types of packages.

- `enum Workspace(WorkspacePackage)` where `WorkspacePackage` is the current package you are working on
    - This would have `tests/` and all kinds of other script-y folders
- `InstalledPackage` which is an installed package in libPaths
    - This would probably not have scripts
    - This may have additional fields for accessing installed help
    - This probably has special behavior for getting at `sources` in a different way from a `WorkspacePackage`

crates/oak_scan/src/packages.rs
```diff
@@ -108,3 +119,163 @@ fn is_r_file(path: &Path) -> bool {
...
+                pkg.scripts = scan_package_scripts(dir);
+                (dir.clone(), pkg)
+            })
+        })
+        .collect();
+
+    let packages = dedup_packages_by_name(pairs);
+    let scripts = collect_scripts(root, &description_dirs);
+
+    (packages, scripts)
```
DavisVaughan: I feel like you could return just `Vec<PackageDescriptor>` from here as `scan_workspace_packages()`, and then put `collect_scripts()` into a separate helper like `scan_workspace_scripts()` with an argument called `exclude` which is a `&[Path]` of directories to exclude while scanning, in this case that would be anything `scan_workspace_packages()` already covered

DavisVaughan: That feeeeels like a nice separation of powers, especially when I go look at `rescan_into()` and see how it is used

crates/oak_scan/src/inputs.rs
```diff
@@ -62,12 +68,144 @@ pub trait DbExt: Db + DbInputs {
...
+    ///   churn. The file watcher handles in-folder changes.
+    ///
+    /// - New paths: scanned (`DESCRIPTION` files at any depth, honouring
+    ///   `.gitignore`, plus top-level R scripts) and added.
+    ///
+    /// - Removed paths: their `Root` is evicted. Files whose URLs are in
+    ///   `editor_owned` move to [`oak_db::OrphanRoot`] (analysis-visible: the
+    ///   buffer is still open). Everything else moves to [`oak_db::StaleRoot`]
+    ///   for entity reuse if the path comes back.
+    fn set_workspace_paths(&mut self, paths: &[PathBuf], editor_owned: &HashSet<UrlId>);
```
DavisVaughan, line 88: Random thought - should a `File` have a field for `editor_owned` so we don't have to keep passing it around? Would be updated on `didOpen` and `didClose`?

crates/oak_scan/src/inputs.rs
```diff
@@ -62,12 +68,144 @@ pub trait DbExt: Db + DbInputs {
...
+    /// `didOpen` / `didChange` content for any URL the editor touches.
+    ///
+    /// If a `File` already exists at this URL (in a live root or orphan),
+    /// only its contents are updated. Classification is left as-is: a file
+    /// the scanner had previously placed in a package stays in that package
+    /// (`didOpen` is a content event, not a reclassification).
+    ///
+    /// If no live `File` exists but one is in [`StaleRoot`] from a prior
+    /// [`Self::close_editor`], it gets resurrected into `orphan_root` with
+    /// the new content. This ways, reopening a previously-closed buffer reuses
```
DavisVaughan, line 104: Is it always resurrected into `orphan_root`? It feels like it could be adopted be a workspace root too? Not sure.

DavisVaughan: Or maybe it is that the next workspace scan has to reclassify it

crates/oak_scan/src/inputs.rs
```diff
@@ -62,12 +68,144 @@ pub trait DbExt: Db + DbInputs {
...
+    /// React to a Deleted watcher event. Unlinks the file from whichever
+    /// container holds it (package files, root scripts, or orphan).
+    fn remove_watched_file(&mut self, url: UrlId);
+
+    /// Apply a batch of file-watcher events. Routes DESCRIPTION events to a
+    /// coarse rescan of the containing workspace root (deduped within the
+    /// batch), and R-file events to per-file add / remove. URLs in `skip` are
+    /// left alone, so callers can defer to an in-memory source of truth (e.g.
+    /// the editor's open buffers).
+    fn apply_watcher_events(&mut self, events: Vec<FileEvent>, skip: &HashSet<UrlId>);
```
DavisVaughan: Hmmmmmmm do add_watched_file and remove_watched_file need to be exposed?

It seems like `apply_watcher_events` should handle all types of updates, and that these other 2 functions are "internal" implementations that `apply_watcher_events` hands off to, and should not be exposed directly

DavisVaughan: and then the internal versions of `add_watched_file` and `remove_watched_file` would no longer be pub(crate)

crates/oak_scan/src/inputs.rs
```diff
@@ -62,12 +68,144 @@ pub trait DbExt: Db + DbInputs {
...
+
+    /// React to a Deleted watcher event. Unlinks the file from whichever
+    /// container holds it (package files, root scripts, or orphan).
+    fn remove_watched_file(&mut self, url: UrlId);
+
+    /// Apply a batch of file-watcher events. Routes DESCRIPTION events to a
+    /// coarse rescan of the containing workspace root (deduped within the
+    /// batch), and R-file events to per-file add / remove. URLs in `skip` are
+    /// left alone, so callers can defer to an in-memory source of truth (e.g.
+    /// the editor's open buffers).
```
DavisVaughan, line 138: Especially with `upsert_editor()` and `close_editor()` being right there, it does feel like `skip` should instead be `editor_owned` to really call out the difference

crates/oak_scan/src/inputs.rs
```diff
@@ -223,6 +411,15 @@ fn remove_from_orphan<DB: Db + DbInputs>(db: &mut DB, file: File) {
     orphan.set_files(db).to(files);
 }
 
+fn add_to_orphan_files<DB: Db + DbInputs>(db: &mut DB, file: File) {
+    let orphan = db.orphan_root();
+    let mut files = orphan.files(db).clone();
+    if !files.contains(&file) {
+        files.push(file);
+        orphan.set_files(db).to(files);
+    }
```
DavisVaughan: Delay the `clone()` on `files` until you need it?

crates/oak_scan/src/watch.rs
```diff
@@ -0,0 +1,283 @@
...
+    for event in events {
+        let Ok(path) = event.url.to_file_path() else {
+            continue;
+        };
+
+        if path.file_name().is_some_and(|n| n == "DESCRIPTION") {
+            if let Some(root) = roots
+                .iter()
+                .find(|(p, _)| path.starts_with(p))
+                .map(|(_, r)| *r)
```
DavisVaughan: I find these 1 character names so hard to understand. Can we use `name`, `path`, and `root`? I feel like this is what I'd normally type for all of these if writing this myself. Can we ask claude to always prefer these kinds of name? I feel like they improve overall readability

crates/oak_scan/src/watch.rs
```diff
@@ -0,0 +1,283 @@
...
+            }
+            continue;
+        }
+
+        if skip.contains(&event.url) {
+            continue;
+        }
+
+        match event.kind {
+            FileEventKind::Created | FileEventKind::Changed => match std::fs::read_to_string(&path)
```
DavisVaughan, line 99: Hopefully `apply_watcher_events` doesn't run all that often - mostly only for git branch switches?

Otherwise we are reading from disk on every `Changed` event, which feels like it could be expensive

crates/oak_scan/src/watch.rs
```diff
@@ -0,0 +1,283 @@
...
+}
+
+fn workspace_root_paths<DB: Db + DbInputs>(db: &DB) -> Vec<(PathBuf, Root)> {
+    db.workspace_roots()
+        .roots(db)
+        .iter()
+        .filter_map(|r| match r.path(db).to_file_path() {
+            Ok(p) => Some((p, *r)),
+            Err(err) => {
+                log::warn!("Skipping workspace root: {err}");
```
DavisVaughan: It's more like "Can't retrieve workspace root file path"?

crates/oak_scan/src/watch.rs
```diff
@@ -0,0 +1,283 @@
...
+}
+
+/// React to a Created or Changed event on an R file. Idempotent: if a `File`
+/// already exists at this URL, its contents are updated and its placement is
+/// left alone. If not, the URL is classified against the current workspace
+/// roots and the new file lands in the right container: `pkg.files` for
+/// `<pkg>/R/*.R`, `pkg.scripts` for other R files under a package
+/// (tests/, inst/, vignettes/, ...), `root.scripts` for R files outside
+/// every package. Mirrors the placement the bulk scanner would pick.
+pub(crate) fn add_watched_file<DB: Db + DbInputs>(db: &mut DB, url: UrlId, contents: String) {
```
DavisVaughan, line 134: I'm mildly nervous about the fact that we have 2 implementations of file classification. One for bulk updates and one for incremental ish updates. It seems likely they could easily get out of sync :/

crates/oak_scan/src/watch.rs
```diff
@@ -0,0 +1,283 @@
...
+                files.push(file);
+                pkg.set_files(db).to(files);
+            }
+        },
+        Placement::PackageScript(pkg) => {
+            let mut scripts = pkg.scripts(db).clone();
+            if !scripts.contains(&file) {
+                scripts.push(file);
+                pkg.set_scripts(db).to(scripts);
+            }
```
DavisVaughan: for all 3, only clone if !contains

crates/oak_scan/src/watch.rs
```diff
@@ -0,0 +1,283 @@
...
+    }
+
+    for &root in &db.workspace_roots().roots(db).clone() {
+        let mut scripts = root.scripts(db).clone();
+        if scripts.contains(&file) {
+            scripts.retain(|f| *f != file);
+            root.set_scripts(db).to(scripts);
+            return;
+        }
+    }
```
DavisVaughan, line 199:
```suggestion
    for root in db.workspace_roots().roots(db) {
        let scripts = root.scripts(db);
        if scripts.contains(&file) {
            let mut scripts = scripts.clone();
            scripts.retain(|f| *f != file);
            root.set_scripts(db).to(scripts);
            return;
        }
    }
```

?

crates/oak_scan/src/watch.rs
```diff
@@ -0,0 +1,283 @@
...
+
+    let orphan = db.orphan_root();
+    let mut files = orphan.files(db).clone();
+    if files.contains(&file) {
+        files.retain(|f| *f != file);
+        orphan.set_files(db).to(files);
+    }
```
DavisVaughan:
```suggestion
    let orphan = db.orphan_root();
    let files = orphan.files(db);
    if files.contains(&file) {
        let mut files = files.clone();
        files.retain(|f| *f != file);
        orphan.set_files(db).to(files);
    }
```

DavisVaughan (approved):

crates/oak_scan/src/inputs.rs
DavisVaughan: Seems like this was all resolved in the next scheduler pr!

lionel-:
https://github.com/posit-dev/ark/pull/1243#discussion_r3374234073

crates/ark/src/lsp/state.rs
lionel-: I would like to strongly agree with you!

crates/ark/src/lsp/state_handlers.rs
lionel-: yes we're heading towards a single notion of workspace root that can contain any number of scripts and packages. This way mono-repos are properly supported.

crates/oak_scan/src/inputs.rs
lionel-: yep sounds good

crates/ark/src/lsp/handlers.rs
lionel-: That's the VS Code behaviour. But it works fine for clients that send events for both, since events for editor-owned files are skipped. I've updated the comment to make this clearer.

crates/oak_db/src/inputs.rs
lionel-: Fixed in an upcoming PR.

crates/oak_scan/src/stale.rs
lionel-: Hmm `files` and `set_files()` are the consistent names throughout `oak_db` inputs

crates/oak_scan/src/packages.rs
lionel-: `Package` is the Salsa input that is created from the descriptor later on.

I've renamed to `PackageEntry` for symmetry with `FileEntry`, which is the equivalent scanned structure produced by the scanner before creating `File` inputs.

crates/oak_scan/src/packages.rs
lionel-: I renamed to `package_dir`

crates/oak_scan/src/inputs.rs
lionel-: hmmm

crates/oak_scan/src/packages.rs
lionel-: We don't scan scripts here because scanning the folders would be wasteful in libraries.
I'm not that confident that we will need two different types of packages. It's certainly something that's been on my mind, for now I've chosen not to split.

For clarity I've splitted into `scan_workspace_packages()` and `scan_workspace_scripts()`.

crates/ark/src/lsp/state_handlers.rs
lionel-: Agreed that could be simpler, and that's how ty and r-a work: everything relies on `DidChangeWatchedFiles` events. I've changed that in an ulterior PR, after the indexer has been migrated to `File`.

crates/oak_scan/src/inputs.rs
lionel-: Yes a stale file is one that used to belong in a workspace but doesn't anymore. If the workspace gets added again, it gets reclassified from orphan at the next scan.

crates/oak_scan/src/watch.rs
lionel-: yep mostly for git branches / agent edits. No I/O for opened files.

crates/oak_scan/src/watch.rs
lionel-: I've extracted a placement classifier to prevent some of the potential drift

PR #1245, opened 2026-05-29
Scan workspaces in background tasks

lionel-:
Branched from #1244
Progress towards #1212

This PR moves workspace scanning off the LSP main loop, to avoid blocking frontend requests on initialize and on every `didChangeWorkspaceFolders`. This PR introduces the same setup `rust-analyzer` and `ty` use: the handler dispatches scanning work to an asynchronous task pool and replies immediately.

The new piece is `oak_scan::ScanScheduler`, which owns the policy (when to scan, how to handle events arriving mid-scan) but not the runtime. The LSP layer is free to use any appropriate mechanism for running off-thread tasks: The LSP takes a `ScanRequest` from the `oak_scan` scheduler, run it via `ScanRequest::run()` off-thread task, and hand the resulting `ScanCompleted` results back to `oak_scan` via `ScanScheduler::apply_scan_completed()`.

In Ark's case we spawn each request on `lsp::spawn_blocking()` (which handles things like crashes gracefully with frontend reports) and ships the result back via a new `Event::OakScanCompleted` for the main loop. Scheduler state and Salsa inputs are mutated only on the main loop, which reduces the surface for race handling issues. We just need to ensure proper event ordering in `oak_scan`, there are no issues of concurrent accesses at all in the LSP layer.

See module doc in `scheduler.rs` for how race conditions are resolved.


crates/ark/src/lsp/main_loop.rs
```diff
@@ -299,13 +313,15 @@ impl GlobalState {
...
+                            let pending = state_handlers::did_change_workspace_folders(params, &mut self.world, &mut self.lsp_state)?;
+                            dispatch_scan_requests(&self.events_tx, pending);
                         },
                         LspNotification::DidChangeConfiguration(params) => {
                             state_handlers::did_change_configuration(params, &self.client, &mut self.world).await?;
                         },
                         LspNotification::DidChangeWatchedFiles(params) => {
-                            state_handlers::did_change_watched_files(params, &mut self.world)?;
+                            let pending = state_handlers::did_change_watched_files(params, &mut self.world, &mut self.lsp_state)?;
+                            dispatch_scan_requests(&self.events_tx, pending);
```
DavisVaughan:
```suggestion
                            let requests = state_handlers::did_change_workspace_folders(params, &mut self.world, &mut self.lsp_state)?;
                            dispatch_scan_requests(&self.events_tx, requests);
                        },
                        LspNotification::DidChangeConfiguration(params) => {
                            state_handlers::did_change_configuration(params, &self.client, &mut self.world).await?;
                        },
                        LspNotification::DidChangeWatchedFiles(params) => {
                            let requests = state_handlers::did_change_watched_files(params, &mut self.world, &mut self.lsp_state)?;
                            dispatch_scan_requests(&self.events_tx, requests);
```

?

seems to match naming better

crates/ark/src/lsp/main_loop.rs
```diff
@@ -452,6 +488,24 @@ impl GlobalState {
...
+) {
+    for req in requests {
+        let tx = events_tx.clone();
+        spawn_blocking(move || {
+            let scan = req.run();
+            tx.send(Event::OakScanCompleted(scan)).log_err();
+            Ok(None)
+        });
+    }
+}
```
DavisVaughan, line 549: Rather than tweaking the value of each LSP Request to possibly return `ScanRequest`s, I think I would much rather allow `state_handlers::initialize()` and `state_handlers::did_change_workspace_folders()` to kick off the scan requests themselves.

i.e. if you just pass `events_tx` to each of those, they should have everything they need to kick off a `spawn_blocking()` call.

That way you can, for example, keep the dispatch close to the `lsp_state.oak_scheduler.set_workspace_paths()` call that requires it, which seems nice to me.

I also really like that that would mean that we would not need the `respond_with()` changes (I'm not super comfortable with how the `::default()` behavior there works) and tweaks to the return values _just_ for this one feature.

DavisVaughan: (This one feels important enough to me to talk about)

DavisVaughan: Seems like this idea has merit, with `dispatch_scan_requests()` calls moving right after `oak_scheduler` usage like this




and then this nicely simplifies back to where we started

crates/oak_scan/src/scheduler.rs
```diff
@@ -0,0 +1,414 @@
...
+        }
+    }
+}
+
+/// Output of [`ScanRequest::run`]. Opaque payload from the scan to
+/// [`ScanScheduler::apply_scan_completed`]; the only field callers inspect is
+/// [`ScanCompleted::root`] (e.g. for logging which scan just finished).
+#[derive(Debug)]
+pub struct ScanCompleted {
+    pub root: Root,
```
DavisVaughan: i would also support not letting this be `pub` and instead having some kind of minimal accessor for anything that might be needed for logging

(though i currently dont see any reason this needs to be `pub` at all)

crates/oak_scan/src/scheduler.rs
```diff
@@ -0,0 +1,414 @@
...
+impl ScanCompleted {
+    /// Push the scan's output into salsa inputs for `self.root`.
+    ///
+    /// Atomic full-replacement of the root's packages and scripts.
+    /// Existing `File` and `Package` entities are reused by URL where
+    /// possible (see [`RootExt::set_package`] /
+    /// [`RootExt::set_workspace_scripts`]), so a rescan that doesn't
+    /// actually change anything is a no-op as far as downstream salsa
+    /// caches are concerned.
+    pub(crate) fn apply<DB: Db + DbInputs>(self, db: &mut DB) {
```
DavisVaughan: probably doesnt need to be pub at all

crates/oak_scan/src/scheduler.rs
```diff
@@ -0,0 +1,414 @@
...
+                vec![ScanRequest { root, path }]
+            },
+            Some(ScanState::Scanning) | None => {
+                // We're now idle. Drain buffered events through the normal path.
+                let buffered = self.buffered.remove(&root).unwrap_or_default();
+                if buffered.is_empty() {
+                    Vec::new()
+                } else {
+                    self.apply_watcher_events(db, buffered, editor_owned)
+                }
```
DavisVaughan: Shouldn't `None` be handled separately? i.e. "we got a ScanCompleted, but we weren't ScanState::Scanning, this is weird"

crates/oak_scan/src/scheduler.rs
```diff
@@ -0,0 +1,414 @@
...
+                let Some(path) = root.path(db).to_file_path().warn_on_err() else {
+                    return Vec::new();
+                };
+                vec![ScanRequest { root, path }]
+            },
+            Some(ScanState::Scanning) | None => {
+                // We're now idle. Drain buffered events through the normal path.
+                let buffered = self.buffered.remove(&root).unwrap_or_default();
+                if buffered.is_empty() {
+                    Vec::new()
```
DavisVaughan: Probably a little cleaner to `match` on the `Option` returned from `remove()` rather than doing unwrap-or-default and checking is-empty

crates/oak_scan/src/scheduler.rs
```diff
@@ -0,0 +1,414 @@
...
+        let prior = self.state.remove(&root);
+        match prior {
+            Some(ScanState::ScanningWithRescanQueued) => {
+                // A rescan was queued mid-scan. Spawn it and keep the
+                // buffer; replay happens when this next scan finishes idle.
+                self.state.insert(root, ScanState::Scanning);
+                let Some(path) = root.path(db).to_file_path().warn_on_err() else {
+                    return Vec::new();
+                };
+                vec![ScanRequest { root, path }]
```
DavisVaughan: The early `return Vec::new()` seems like a problem. You set this root up as `Scanning` but if you don't return a `ScanRequest` for it, then it will never get scanned and will be infinitely left in `Scanning` mode

DavisVaughan: do you want to do this there?

```
            self.state.remove(&root);
            self.buffered.remove(&root);
```

crates/oak_scan/src/scheduler.rs
```diff
@@ -0,0 +1,414 @@
...
+        return;
+    };
+    match event.kind {
+        FileEventKind::Created | FileEventKind::Changed => match std::fs::read_to_string(&path) {
+            Ok(contents) => add_watched_file(db, event.url, contents),
+            Err(err) => log::warn!("Skipped watched file {}: {err:?}", path.display()),
+        },
+        FileEventKind::Deleted => remove_watched_file(db, event.url),
+    }
+}
```
DavisVaughan: i think you should just inline this!

DavisVaughan (approved):

crates/oak_scan/src/scheduler.rs
lionel-: Good catch!

crates/ark/src/lsp/main_loop.rs
lionel-: Good idea

PR #1246, opened 2026-05-29
Help proxy: use 127.0.0.1 and surface error source chain

bricestacey:
Helps debug posit-dev/positron#13806.

R hands the proxy URLs formed with `127.0.0.1`, but `proxy_request` rewrites them to `localhost` before re-issuing the request. On Windows this needlessly introduces dual-stack resolution (`::1` + `127.0.0.1`) even though R's `tools::startDynamicHelp()` only binds IPv4. Switch the rewrite to `127.0.0.1`.

Also walks `reqwest::Error::source()` when logging proxy failures and logs `content-length` / `transfer-encoding` alongside `content-type`. The top-level reqwest error often hides the underlying cause (truncated chunk, decompression failure, socket close), and we've been seeing intermittent body-decode failures on Windows CI without enough signal to diagnose them.

This is a partial fix - it may resolve the symptom on its own (by eliminating the IPv6 path), and if not, the extra diagnostics give us what we need to pick the next mitigation.

@:packages-pane @:help @:ark

### DIFF
--- a/crates/ark/src/help_proxy.rs
+++ b/crates/ark/src/help_proxy.rs
@@ -131,6 +131,11 @@ async fn proxy_request(req: HttpRequest, app_state: web::Data<AppState>) -> Http
     // Disable proxy since we're connecting to localhost; otherwise HTTP_PROXY
     // and other env vars can cause the request to get incorrectly routed to a
     // proxy.
+    //
+    // Note: `RetryTransientMiddleware` only retries the request *setup* (before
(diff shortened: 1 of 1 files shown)


bricestacey:
This PR doesn't contain a solution to the problem. We just needed a remote branch so we could update the git submodule in Positron.

This branch is primarily diagnostics and, like I said doesn't actually solve the problem.

Closing because we can still reference random branches from a submodule, even if there is no accompanying PR.

PR #1248, opened 2026-05-29
Update action-gh-release to v3

### LINKED ISSUE #1247: Bump action-gh-release to v3
The Ark builds have a warning that node.js v20 is deprecated and v24 will be forced June 2, 2026. The version will need to be bumped to v3.0.0 to support v24.

timtmok:
Fixes #1247

Maintenance to support node.js v24

> Node.js 20 actions are deprecated. The following actions are running on Node.js 20 and may not work as expected: 1password/load-secrets-action@v3, aws-actions/configure-aws-credentials@v4. Actions will be forced to run with Node.js 24 by default starting June 2nd, 2026. Node.js 20 will be removed from the runner on September 16th, 2026. Please check if updated versions of these actions are available that support Node.js 24. To opt into Node.js 24 now, set the FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true environment variable on the runner or in your workflow file. Once Node.js 24 becomes the default, you can temporarily opt out by setting ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/
 
We will need to update to continue using the runner. There's also setting `ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true` but that won't work on September 16, 2026.

This was addressed in Positron with https://github.com/posit-dev/positron-builds/pull/1018

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -88,7 +88,7 @@ jobs:
             GITHUB_TOKEN: ${{ github.token }}
         steps:
             - name: Create release
-              uses: softprops/action-gh-release@v2
+              uses: softprops/action-gh-release@v3.0.0
               id: create_release
               with:
                   draft: false
@@ -147,7 +147,7 @@ jobs:
                   name: ark-${{ matrix.flavor }}-linux-arm64-archive
 
             - name: Upload macOS release artifact (universal)
-              uses: softprops/action-gh-release@v2
+              uses: softprops/action-gh-release@v3.0.0
               env:
                   GITHUB_TOKEN: ${{ github.token }}
               with:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -147,55 +147,55 @@ jobs:
...
+              uses: softprops/action-gh-release@v3.0.0
               env:
                   GITHUB_TOKEN: ${{ github.token }}
               with:
                   tag_name: ${{ needs.get_version.outputs.ARK_VERSION }}
                   files: ark-${{ needs.get_version.outputs.ARK_VERSION }}${{ env.DEBUG_FLAG }}-darwin-universal.zip
 
             - name: Upload macOS release artifact (arm64)
-              uses: softprops/action-gh-release@v2
+              uses: softprops/action-gh-release@v3.0.0
```
DavisVaughan, line 158: i dont see any reason to lock ourselves to `v3.0.0`, can we just use the sliding `v3` tag instead?

DavisVaughan (approved):
LGTM other than one comment

.github/workflows/release.yml
timtmok: Positron had previously locked to `v2.0.5` since an update had regressed in some functionality (unable to update assets in a release). I think it'll be good to stick with a version we know works until we have a reason to update.

DavisVaughan: On the flip side any bugs that are fixed won't flow through to us (particularly anything security related).

But sure, I don't care too much

PR #1249, opened 2026-05-29
POC: `package_by_file_query()` rather than `package` field in `File`

DavisVaughan:
🔥 UNREVIEWED

**Core change (`oak_db`):**
- `crates/oak_db/src/file.rs` — removed the `package: Option<Package>` input field from `File`. Added a `#[salsa::tracked] File::package(db)` query that derives ownership. `File::root` is unchanged (still routes through `root_by_package` for correct nested-root selection).
- `crates/oak_db/src/db.rs` — added `root_file_package_index` (a per-root `File → Package` tracked index, parallel to the existing `root_url_index`) and `package_by_file_query`, which walks live roots' indexes in workspace-then-library order and returns the first hit. Same granularity and walk shape as `file_by_url`.
- `crates/oak_db/src/inputs.rs` — rewrote the placement-invariant doc blocks on `Package.files`, `Root.scripts`, and `OrphanRoot.files` to describe single-source-of-truth instead.

**Scanner (`oak_scan`):**
- `upsert_file` no longer writes a back-pointer. It still drops a file from a *different* old package (the rare nested-root case) using the derived `old.package(db)`, and still cleans up orphan placement.
- `set_root_stale` lost its `file.set_package(None)` loop entirely — eviction now falls out of containment automatically.

---

Also

>   One note for later: narrow_package_top_level's "file not in package files" warning branch
  is now logically unreachable (the package is derived from that membership) — I kept it as
  a defensive guard and reworded the message.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2623,6 +2623,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "oak_scan"
+version = "0.1.0"
+dependencies = [
+ "aether_url",
+ "log",
+ "oak_db",
+ "oak_package_metadata",
+ "salsa",
+ "stdext",
+ "tempfile",
+ "url",
+ "walkdir",
+]
+
+[[package]]
 name = "oak_semantic"
 version = "0.1.0"
 dependencies = [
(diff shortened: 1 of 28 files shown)


lionel-:
That's exactly how it worked in a previous iteration where I implemented a map query for packages, just like there is one for files. I made it a field because it seems nice as a discriminator in operations over files (natural `match file.package` syntax to dispatch package vs script behaviour). It also has fewer invalidation events (no dependencies on roots), which while not crucial, still seems nice for such a common lookup.

PR #1250, opened 2026-06-03
Don't canonicalise file keys

lionel-:
Branched from https://github.com/posit-dev/ark/pull/1245

While I was working on breakpoints I ran into difficulties matching files from different locations: frontend URIs via DAP requests and source references from R whose paths went through `normalizePath()`. The latter does file canonicalisation, which means that in addition to resolving `.` and `..`, it also follows symlinks. The frontend URIs are not canonicalised, which caused CI failures when running tests in projects storedc in temporary folders, with mismatches between `/tmp` and `/private/tmp`. The fix I implemented was to canonicalise everywhere and encode that canonicalisation happened with the `UrlId` file, which we stored as file keys in our maps tracking state (e.g. breakpoints) by file.

The `UrlId` normalisation also dealt with other issues that surfaced over time, such as windows letter drive casing because to fix issues with lowercase on one end and uppercase on the other.

The same sort of issue came up with the LSP: we get URIs from the frontend, but we also encounter `source()` calls that may be relative and we need to resolve them. So I reached for the same `UrlId` approach to make the different file paths agree with each other.

It turns out that doing any sort of normalisation that requires I/O is a red flag. If a file gets deleted on disk, and we get an event about that file, there is literally no way to match it back to what we know. For the LSP that's a real issue because we get file deletion events after the file is deleted.

The other thing to watch out for: Always report back the unormalised URI/path to the frontend, because that's what they use to map files on their end. Any changes might cause the frontend to treat the file as different.

With this PR:

- We no longer canonicalise files that we store as keys. We only do lexical normalisation: resolve `.` and `..`. That matches r-a and ty.
- We still do canonicalisation at comparison time, in a secondary lookup. This way we can still match file paths coming from different sources.
- We keep track of original file URI/paths for communication with the frontend. For now only in the DAP, this will be done for the LSP in another PR.

### DIFF
--- a/crates/aether_url/src/lib.rs
+++ b/crates/aether_url/src/lib.rs
@@ -10,99 +10,63 @@ use std::fmt;
 use stdext::result::ResultExt;
 use url::Url;
 
-/// Canonicalised file URL identity.
+/// Lexically normalised file URL identity.
 ///
-/// # The multi-source URI reconciliation problem
+/// Internal identity key for files received from any source (LSP, DAP,
+/// scanner, R runtime). Constructed via the same lexical normalisation
+/// at every entry point so that two paths the editor considers "the
+/// same file" produce the same [`UrlId`].
 ///
-/// File URIs for the same file can arrive from independent sources, each
-/// with its own representation:
+/// What we normalise: drive-letter casing on Windows; percent-encoding
+/// of `:` (decoded via `Url -> PathBuf -> Url` round-trip). No I/O:
+/// `std::fs::canonicalize()`, no symlink resolution. The same input URI
+/// produces the same [`UrlId`] whether or not the file exists on disk.
 ///
-/// - **DAP (SetBreakpoints)**: Receives raw file paths from the frontend,
-///   converted to URIs via `UrlId::from_file_path`. These are stored as
-///   HashMap keys for breakpoint lookup.
+/// # Bridging across symlinks
 ///
-/// - **LSP (didChange, etc.)**: Receives URIs directly from the editor
-///   client, which may use non-canonical forms (e.g. percent-encoded
-///   colons on Windows, or symlinked paths on macOS).
+/// R's `normalizePath()` resolves symlinks on its own. A srcref URI
+/// from the R runtime may name `/private/tmp/foo.R` while the editor
+/// sent us `/tmp/foo.R`. The two don't compare equal, so a `HashMap`
+/// keyed on `UrlId` treats them as separate files. Code that needs to
+/// match a srcref URI back to an open document or a breakpoint should
+/// maintain a secondary index of `fs::canonicalize`d paths and fall
+/// back to it on a primary miss.
+/// [`crate::dap::dap_state::BreakpointMap`] in `ark` does this for
+/// breakpoints.
 ///
-/// - **Execute requests**: A frontend may attach a `code_location` URI
-///   that comes straight from the editor's document model, again
-///   potentially non-canonical.
+/// # Important: don't leak normalised URIs back out
 ///
-/// - **R runtime**: When R evaluates `source()` or annotates code, it
-///   passes URIs that went through R's `normalizePath()`, which resolves
-///   symlinks to their canonical target (e.g. `/tmp` resolves to `/private/tmp`
-///   on macOS), producing a path the editor never sent. More generally,
-///   arbitrary R code can create source references that we may end up
-///   consuming for breakpoint or debug purposes, and the paths in those
-///   references may or may not be canonical.
-///
-/// All four sources must agree on file identity. For instance breakpoints set
-/// via DAP are looked up in a HashMap keyed by URI when code is executed or
-/// sourced, and invalidated when documents change via LSP.
-///
-/// # Design decision
-///
-/// We solve this by canonicalizing URIs into [`UrlId`] at every entry
-/// point, rather than interning paths into opaque IDs (as rust-analyzer
-/// does with its VFS `FileId` approach). Interning would be a larger
-/// architectural change and is not warranted here since we only need
-/// canonical keys at a handful of call sites.
-///
-/// Canonicalization uses `std::fs::canonicalize()` to resolve symlinks
-/// (e.g. `/tmp` to `/private/tmp` on macOS), round-trips through the
-/// filesystem path to normalize encoding variants (e.g. `%3A` to `:` on
-/// Windows), and uppercases drive letters on Windows. When the file does
-/// not exist on disk, we fall back to the original URI.
-///
-/// # Important: canonical URIs must not leak
-///
-/// [`UrlId`] is strictly for internal identity. When a URI flows back
-/// to R (e.g. in `#line` directives or injected breakpoint calls) or to
-/// the frontend (e.g. in DAP stack frames), always use the original raw
-/// URI. The frontend (and possibly R code) expects their own URI
-/// representation, and a canonical URI (e.g. `/private/tmp/...` instead of
-/// `/tmp/...`) could be treated as a different file (e.g. open a new editor in
-/// the frontend instead of an existing one).
-///
-/// A canonicalized file URI for use as a stable identity key.
-///
-/// Wraps a [`Url`] that has been canonicalized to resolve symlinks,
-/// normalize encoding variants, and uppercase drive letters on Windows.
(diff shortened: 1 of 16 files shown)


DavisVaughan:
> It turns out that doing any sort of normalisation that requires I/O is a red flag. If a file gets deleted on disk, and we get an event about that file, there is literally no way to match it back to what we know.

LOL OUCH

crates/ark_test/src/dummy_frontend.rs
```diff
@@ -1431,7 +1431,12 @@ impl DummyArkFrontend {
         // Use forward slashes for R compatibility on Windows (backslashes would be
         // interpreted as escape sequences in R strings)
         let path = file.path().to_string_lossy().replace('\\', "/");
-        let uri_id = UrlId::from_file_path(file.path()).unwrap().to_string();
+        // `uri_id` is what R-side code reports back. R's `path_to_file_uri()`
+        // applies `normalizePath()` which resolves symlinks (e.g. macOS
+        // `/var/...` -> `/private/var/...`), so canonicalize here so tests
+        // can compare against R-reported URIs byte for byte.
+        let canonical = file.path().canonicalize().unwrap();
+        let uri_id = UrlId::from_file_path(&canonical).unwrap().to_string();
```
DavisVaughan:
```suggestion
        // R side `path_to_file_uri()` applies `normalizePath()`, which resolves
        // symlinks (e.g. macOS `/var/...` -> `/private/var/...`). To match that
        // in tests, we also canonicalize here.
        let canonical = file.path().canonicalize().unwrap();
        let uri_id = UrlId::from_file_path(&canonical).unwrap().to_string();
```

crates/ark_test/src/dummy_frontend.rs
```diff
@@ -1822,7 +1827,11 @@ impl SourceFile {
         // Use forward slashes for R compatibility on Windows (backslashes would be
         // interpreted as escape sequences in R strings)
         let path = file.path().to_string_lossy().replace('\\', "/");
-        let url = UrlId::from_file_path(file.path()).unwrap();
+        // `uri_id` is what R reports via `path_to_file_uri()` (which goes
+        // through `normalizePath()`), so canonicalize here so tests can
+        // compare against R-reported URIs byte for byte.
+        let canonical = file.path().canonicalize().unwrap();
+        let url = UrlId::from_file_path(&canonical).unwrap();
```
DavisVaughan:
```suggestion
        // R side `path_to_file_uri()` applies `normalizePath()`, which resolves
        // symlinks (e.g. macOS `/var/...` -> `/private/var/...`). To match that
        // in tests, we also canonicalize here.
        let canonical = file.path().canonicalize().unwrap();
        let url = UrlId::from_file_path(&canonical).unwrap();
```

crates/ark_test/src/dummy_frontend.rs
DavisVaughan: just seemed like it could be written simpler

crates/aether_url/src/lib.rs
```diff
@@ -10,125 +10,83 @@ use std::fmt;
...
+/// Internal identity key for files received from any source (LSP, DAP,
+/// scanner, R runtime). Constructed via the same lexical normalisation
+/// at every entry point so that two paths the editor considers "the
+/// same file" produce the same [`UrlId`].
 ///
-/// File URIs for the same file can arrive from independent sources, each
-/// with its own representation:
+/// What we normalise: drive-letter casing on Windows; percent-encoding
+/// of `:` (decoded via `Url -> PathBuf -> Url` round-trip). No I/O:
+/// `std::fs::canonicalize()`, no symlink resolution. The same input URI
```
DavisVaughan: dont we also normalize `..`?

crates/aether_url/src/lib.rs
```diff
@@ -10,125 +10,83 @@ use std::fmt;
...
-/// prefixed with `\\?\` (e.g. `\\?\C:\Users\...`). Projects like Ruff
-/// use the `dunce` crate to strip this prefix, but we don't need it
-/// because `Url::from_file_path` already handles
-/// `Prefix::VerbatimDisk` and produces a clean `file:///C:/...` URI.
+/// Even though [`UrlId`] no longer fs-canonicalises, it still
+/// uppercases the Windows drive letter and decodes the percent-encoded
+/// colon. When sending URIs back to the editor or to R, prefer the
+/// original bytes the frontend sent. The frontend treats a URI as the
+/// editor's identity for the file; a normalised form may look like a
+/// different file to it (e.g. open a new editor pane).
```
DavisVaughan:
```suggestion
/// When sending URIs back to the editor or to R, prefer the original bytes the
/// frontend sent rather than the lexically normalized form. The frontend treats
/// a URI as the editor's identity for the file; a normalised form may look like
/// a different file to it (e.g. open a new editor pane).
```

crates/aether_url/src/lib.rs
DavisVaughan: you mention it in the pr, but i dont see it anywhere here, so thats why im asking

> We only do lexical normalisation: resolve . and ...

DavisVaughan: would be nice to see a test in this file either way!

crates/ark/src/dap/dap_state.rs
```diff
@@ -220,6 +221,92 @@ impl DapBackendEvent {
     }
 }
 
+/// Breakpoint storage keyed on the verbatim `UrlId` the frontend sent in
```
```suggestion
/// Breakpoint storage keyed on the lexically normalized `UrlId` the frontend sent in
```

its not quite verbatim IIUC

crates/ark/src/dap/dap_state.rs
```diff
@@ -220,6 +221,92 @@ impl DapBackendEvent {
...
+/// `setBreakpoints`. A secondary index maps `fs::canonicalize`d paths
+/// back to the primary key, bridging R-runtime srcref URIs (which go
+/// through `normalizePath()` and therefore arrive symlink-resolved)
+/// against the editor's pre-resolution form.
+///
+/// `fs::canonicalize` runs at insert time and at lookup-fallback time
+/// only. Lookups try the primary first (cheap, no fs touch), then
+/// canonicalise the incoming URI and consult the secondary. DAP wire
+/// output (`debugInfo` source paths, breakpoint events) round-trips the
+/// primary key unchanged, so the frontend always sees the URI it sent.
```
DavisVaughan: Yea IIUC this isn't actually what is happening

In `handle_set_breakpoints()` we do this

```
let uri = match UrlId::from_file_path(path)
```

and we store that `uri` in `by_url` here.

So I'm not entirely sure "the frontend sees the uri it sent". It might be lexically normalized?

DavisVaughan: Which is either fine, and needs a doc comment update, or isn't fine, and we need a _third_ map that gets us back to the original `path: String` the DAP provided us 😬

DavisVaughan: And either way, a test could be good

DavisVaughan (approved):
I simultaneously:
- am sad we dont have a canonical uri everywhere, but get it, especially with the deleted file example
- am happy we dont have to do IO for uri normalization

crates/aether_url/src/lib.rs
lionel-: I can't find it either! Must have made a rebase mistake along the way or other... I distinctly remember component parsing code.

Claude says "`Url::parse` already collapses dot segments for `file:` URLs, but `Url::from_file_path` doesn't". Maybe I missed that the agent decided to remove that code at some point because of the parse behaviour 😬

crates/ark/src/dap/dap_state.rs
lionel-: I think we should be consistent and send the verbatim path back here, thanks for catching this.

Only concerns the `debugInfo` request, which is only sent from notebooks. For the DAP this aspect is much less crucial. Breakpoints are keyed by ID, and source frame paths come from R srcrefs.

lionel-: Instead of a third map, I've just included the verbatim path in the first map. I wonder if I should store the verbatim in the path variant of `FilePath`, in the next PR.

crates/aether_url/src/lib.rs
DavisVaughan: I think it may come in in a future PR, I saw it somewhere else

PR #1251, opened 2026-06-03
Refactor `UrlId` into `FilePath` enum

lionel-:
Branched from https://github.com/posit-dev/ark/pull/1250 (see that PR first for context)

- Rename `UrlId` to `FilePath`
- Make it our own custom enum instead of piggybacking on the `Url` enum. Variants: `File` (lexically normalised) and `Virtual` (wrap an `Uri` e.g. for untitled files). We could add `Vendored` later for vendored base R sources, mirroring ty.

This refactor better reflects that the main representation for our files are on-disk paths.


crates/aether_path/src/file_path.rs
```diff
@@ -0,0 +1,266 @@
...
+            }
+            // Fall through: a `file:` URL we can't extract a path from
+            // stays as Virtual so the input isn't lost. Rare in practice.
+        }
+        Self::Virtual(VirtualUri::new(url.clone()))
+    }
+
+    /// Build a [`FilePath::File`] from a filesystem path. Errors if
+    /// the path can't be expressed as a UTF-8 absolute path.
+    pub fn from_file_path(path: impl AsRef<Path>) -> anyhow::Result<Self> {
```
DavisVaughan: I think this should just be `from_path()` now, because I saw `FilePath::from_file_path()` elsewhere and was pretty confused by that naming

Or actually, probably `from_path_buf()` to mimic the `AbsPathBuf` change

crates/aether_path/src/file_path.rs
```diff
@@ -0,0 +1,266 @@
...
+    pub fn from_path(path: &Path) -> Option<Self> {
+        let utf8 = Utf8PathBuf::from_path_buf(path.to_path_buf())
+            .map_err(|p| anyhow::anyhow!("Path is not valid UTF-8: {}", p.display()))
+            .warn_on_err()?;
+        Self::from_utf8(utf8)
+    }
+
+    /// Build from a UTF-8 path. Returns `None` if the path is not
+    /// absolute.
+    pub fn from_utf8(path: Utf8PathBuf) -> Option<Self> {
```
DavisVaughan: Like `Utf8PathBuf::from_path_buf` and `to_path_buf()`, I'd prefer the unambiguous `from_utf8_path_buf()`

crates/aether_path/src/file_path.rs
```diff
@@ -0,0 +1,266 @@
...
+        Self::from_path(&path)
+    }
+
+    /// Build from a filesystem path. Returns `None` if the path can't
+    /// be represented as UTF-8 or is not absolute.
+    pub fn from_path(path: &Path) -> Option<Self> {
+        let utf8 = Utf8PathBuf::from_path_buf(path.to_path_buf())
+            .map_err(|p| anyhow::anyhow!("Path is not valid UTF-8: {}", p.display()))
+            .warn_on_err()?;
+        Self::from_utf8(utf8)
```
DavisVaughan:
```suggestion
    pub fn from_path_buf(path: PathBuf) -> Option<Self> {
        let utf8 = Utf8PathBuf::from_path_buf(path)
            .map_err(|p| anyhow::anyhow!("Path is not valid UTF-8: {}", p.display()))
            .warn_on_err()?;
        Self::from_utf8(utf8)
```

I think this will make `from_url()` a bit simpler too

Also, I generally treat `from_` APIs as consuming the input

I also like the general idea of not hiding clones from the caller, i.e. this forces the `to_path_buf()` caller up onto the caller, making it obvious at the call site that the clone occurs. And if they happen to have a PathBuf already, they aren't forced into a second clone.

crates/ark/src/dap/dap_state.rs
```diff
@@ -299,11 +299,11 @@ impl BreakpointMap {
...
 /// when the file isn't on disk.
-fn canonical_path(url: &UrlId) -> Option<PathBuf> {
+fn canonical_path(url: &FilePath) -> Option<PathBuf> {
     if !url.is_file() {
         return None;
     }
-    let path = url.to_file_path().ok()?;
+    let path = url.to_path_buf()?;
     std::fs::canonicalize(path).ok()
 }
```
```suggestion
fn canonical_path(url: &FilePath) -> Option<PathBuf> {
    match url {
        FilePath::File(path) => std::fs::canonicalize(path.as_path()).ok(),
        FilePath::Virtual(_) => None,
    }
}
```

Hmm. I'd like to challenge the addition of the `to_path_buf()` method.

I understand the desire to have a way to go from `FilePath -> PathBuf` for compatibility, but I have a strong feeling that it would be better to stay in `UTF8PathBuf` form as long as possible.

Like, I think this rewrite of `canonical_path()` is more obvious and takes advantage of UTF8PathBuf features

Can we avoid adding `to_path_buf()` entirely in favor of rewrites likes this one? And should `canonical_path()` itself also return another `UTF8PathBuf`?

crates/aether_path/src/file_path.rs
```diff
@@ -0,0 +1,266 @@
...
+                }
+                return out;
+            }
+        }
+    }
+    prefix.to_string()
+}
+
+#[cfg(test)]
+mod tests;
```
DavisVaughan, line 260: Put this in lib.rs?

crates/aether_path/src/file_path.rs
```diff
@@ -0,0 +1,266 @@
...
+        let url = Url::parse(s)?;
+        Ok(Self::from_url(&url))
+    }
+
+    /// Filesystem path buffer for the `File` arm. Returns `None` for the
+    /// `Virtual` arm.
+    pub fn to_path_buf(&self) -> Option<std::path::PathBuf> {
+        self.as_file()
+            .map(|p| p.as_path().as_std_path().to_path_buf())
+    }
```
DavisVaughan: See comment in dap_state.rs, would be nice if we could avoid this one. Leaking a PathBuf feeeeeels wrong to me

crates/oak_scan/tests/integration/library.rs
```diff
@@ -255,7 +255,7 @@ fn test_set_library_paths_re_add_preserves_file_identity() {
     db.set_library_paths(&[]);
     // After removal, the file is not reachable through analysis.
     let r_path = tmp.path().join("pkg").join("R").join("a.R");
-    let url = UrlId::from_file_path(&r_path).unwrap();
+    let url = FilePath::from_file_path(&r_path).unwrap();
```
DavisVaughan: I'd like to propose a big pass to catch any `let url = FilePath::` and turn it into `let path = FilePath::` because these names don't make sense to me.

crates/oak_db/src/storage.rs
```diff
@@ -60,15 +60,15 @@ impl DbInputs for OakDatabase {
 
 #[salsa::db]
 impl Db for OakDatabase {
-    fn file_by_url(&self, url: &aether_url::UrlId) -> Option<crate::File> {
+    fn file_by_url(&self, url: &aether_path::FilePath) -> Option<crate::File> {
         crate::db::file_by_url_query(self, url)
     }
```
DavisVaughan:
```suggestion
    fn file_by_path(&self, path: &aether_path::FilePath) -> Option<crate::File> {
        crate::db::file_by_path_query(self, path)
    }
```

these names don't make sense to me either now

crates/oak_scan/src/watch.rs
```diff
@@ -32,7 +32,7 @@ use crate::packages::read_description_name;
 #[derive(Clone, Debug)]
 pub struct FileEvent {
     pub kind: FileEventKind,
-    pub url: UrlId,
+    pub url: FilePath,
```
DavisVaughan: Also a pass looking for `url: FilePath` to change to `path: FilePath`

crates/oak_scan/src/watch.rs
```diff
@@ -50,13 +50,13 @@ pub enum FileEventKind {
...
     if let Some(existing) = db.file_by_url(&url) {
         existing.set_contents(db).to(contents);
         return;
     }
 
-    let Ok(path) = url.to_file_path() else {
+    let Some(path) = url.to_path_buf() else {
         log::warn!("Skipping add_watched_file: URL is not a file path");
         return;
     };
```
DavisVaughan:
```suggestion
    let Some(path) = url.as_file() else {
        log::warn!("Skipping add_watched_file: URL is not a file path");
        return;
    };

    let Some(placement) = classify(db, path.as_path().as_std_path()) else {
        // Either the URL falls outside every workspace, or it lives
        // inside a package subdir we don't track (tests/, inst/, ...).
        return;
    };
```

i would have written as this to avoid the expensive `to_path_buf()`. no need to make a clone here.

more evidence to me that `to_path_buf()` is a possible anti pattern

DavisVaughan: and really we'd probably make `classify()` take `&Utf8PathBuf`

crates/oak_db/src/inputs.rs
```diff
@@ -186,7 +186,7 @@ pub struct Package {
     /// package belongs to whichever `Root.packages` currently holds it.
     /// Workspace-vs-library is then `root.kind(db)`.
     #[returns(ref)]
-    pub description_url: UrlId,
+    pub description_url: FilePath,
```
DavisVaughan:
```suggestion
    pub description_path: FilePath,
```

crates/oak_db/src/db.rs
```diff
@@ -52,7 +52,7 @@ pub trait Db: DbInputs {
     /// then falls back to the orphan bucket. The walk short-circuits
     /// on the first hit, so callers depend only on the index maps
     /// actually visited.
-    fn file_by_url(&self, url: &UrlId) -> Option<File>;
+    fn file_by_url(&self, url: &FilePath) -> Option<File>;
```
DavisVaughan:
```suggestion
    fn file_by_path(&self, url: &FilePath) -> Option<File>;
```

and argument name update

DavisVaughan: same with package_by_url

DavisVaughan: and many other functions in this file

crates/ark/src/url.rs
```diff
@@ -5,20 +5,20 @@
...
-use aether_url::UrlId;
+use aether_path::FilePath;
 use amalthea::wire::execute_request::CodeLocation;
 use url::Url;
 
-/// Extract a canonical [`UrlId`] from a [`CodeLocation`].
-pub fn url_id_from_code_location(loc: &CodeLocation) -> UrlId {
-    UrlId::from_url(loc.uri.clone())
+/// Extract a canonical [`FilePath`] from a [`CodeLocation`].
+pub fn url_id_from_code_location(loc: &CodeLocation) -> FilePath {
```
DavisVaughan:
```suggestion
pub fn file_path_from_code_location(loc: &CodeLocation) -> FilePath {
```

DavisVaughan (approved):
I really like the idea of `FilePath`, but we are currently in a state where A LOT of existing code is still using `url`, `url_id`, or something similar, and I think it results in a net negative in terms of code readability. I have marked a few cases, but have definitely not captured them all. I'd like to advocate for a close review of all uses of `FilePath` before merging to ensure that the corresponding binding is also called `path` or `file_path`.

crates/aether_path/src/file_path.rs
lionel-: I pondered these alternatives and went with `from_file_path()` for consistency with `UrlId`. Happy to go back on that though.

crates/ark/src/dap/dap_state.rs
lionel-: That makes sense, I've removed it

PR #1252, opened 2026-06-03
Map LSP documents by normalised `FilePath`

lionel-:
Branched from https://github.com/posit-dev/ark/pull/1251

See  https://github.com/posit-dev/ark/pull/1250 for more context.

- `WorldState` now keys documents by normalised `FilePath` instead of `Url`. See `state.rs`. Keying by normalised paths will prevent bugs when matching to paths that don't come from the frontend, e.g. `source()` paths in user code.

- The original `Url` sent by the LSP frontend is stored in `Document`. This is used in communications with the frontend so that we send back exact byte-for-byte Urls, preventing any file matching bugs that could be caused by our internal normalisation of file paths.

Unlike the DAP which has to deal with paths normalised by R, we never resolve symlinks on the LSP side. All matching is done on `FilePath` based on pure lexical normalisation of paths.

### DIFF
--- a/crates/ark/src/lsp/state.rs
+++ b/crates/ark/src/lsp/state.rs
@@ -1,6 +1,7 @@
 use std::collections::HashMap;
 use std::path::Path;
 
+use aether_path::FilePath;
 use anyhow::anyhow;
 use oak_db::OakDatabase;
 use oak_semantic::library::Library;
@@ -18,8 +19,10 @@ pub(crate) struct WorldState {
     /// Salsa input tree for Oak queries.
     pub(crate) db: OakDatabase,
 
-    /// Watched documents
-    pub(crate) documents: HashMap<Url, Document>,
+    /// Watched documents, keyed on the normalised [`FilePath`] form.
+    /// The verbatim editor URL is preserved on each [`Document::url`]
+    /// for wire output.
+    pub(crate) documents: HashMap<FilePath, Document>,
 
     /// Watched folders
     pub(crate) workspace: Workspace,
@@ -79,7 +82,8 @@ impl WorldState {
     }
 
     pub(crate) fn get_document(&self, uri: &Url) -> anyhow::Result<&Document> {
-        if let Some(doc) = self.documents.get(uri) {
+        let key = FilePath::from_url(uri);
+        if let Some(doc) = self.documents.get(&key) {
             Ok(doc)
         } else {
             Err(anyhow!("Can't find document for URI {uri}"))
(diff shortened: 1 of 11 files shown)


crates/ark/src/lsp/state.rs
```diff
@@ -79,15 +82,17 @@ impl WorldState {
...
-        if let Some(doc) = self.documents.get(uri) {
+        let key = FilePath::from_url(uri);
+        if let Some(doc) = self.documents.get(&key) {
             Ok(doc)
         } else {
             Err(anyhow!("Can't find document for URI {uri}"))
         }
     }
 
     pub(crate) fn get_document_mut(&mut self, uri: &Url) -> anyhow::Result<&mut Document> {
```
DavisVaughan: I feel like these should take `path: &FilePath`. It would encourage conversion to `FilePath` at the LSP boundary

crates/ark/src/lsp/state.rs
```diff
@@ -15,8 +16,10 @@ use crate::lsp::inputs::source_root::SourceRoot;
 /// code. This is a pure value. There is no interior mutability in this data
 /// structure. It can be cloned and safely sent to other threads.
 pub(crate) struct WorldState {
-    /// Watched documents
-    pub(crate) documents: HashMap<Url, Document>,
+    /// Watched documents, keyed on the normalised [`FilePath`] form.
+    /// The verbatim editor URL is preserved on each [`Document::url`]
+    /// for wire output.
+    pub(crate) documents: HashMap<FilePath, Document>,
```
DavisVaughan, line 25: I would like to advocate for the following strategy

```rs
pub(crate) documents: HashMap<FilePath, (Url, Document)>,
```

- `Document` no longer knows about `url`. I think this is correct, as you should be able to create a `Document` without a corresponding `url`.
- `Document` no longer has that weird `PLACEHOLDER_URL` hack, which i really do not like
- `insert_document()` would be removed since we only `.insert()` in 1 place in production code, and then in a few tests, so the abstraction doesn't feel worth it

`WorldState::get_document()` and `WorldState::get_document_mut()` would then:

- Take a `path: FilePath` as I mentioned elsewhere
- Abstract over the `(Url, Document)` enum and just pull out the `Document` and return it

We would need to go update 2 direct usages of `state.document.get()` to be `state.get_document()`, but I think we should do that anyways.

This feels much much better to me, and I looked at all call sites of `state.documents.` and think it would work quite well

I am quite strongly against `PLACEHOLDER_URL`, so am going to block this pr until that is resolved

DavisVaughan (changes requested):
I really like the idea of this PR! But one blocking comment about `Document` that I'd like to resolve first

crates/ark/src/lsp/state.rs
lionel-: oops sorry, my fault. I asked Claude to get rid of `Option` and didn't closely review that one. The placeholder is not only unsightly, it would leak to the user if we made a mistake somewhere.

I did consider hoisting url out of there but it conflicted with existing code IIRC, and I wanted to avoid making too many changes to legacy Ark handlers. Let me take another look.

lionel-: oh wait I forgot it's already gone in an ulterior PR!

PR #1253, opened 2026-06-03
Fix stability of Definition entities

lionel-:
Branched from https://github.com/posit-dev/ark/pull/1252

Like in ty, our `Definition`s are a tracked struct with some degree of stability across edits. Unlike in ty, the definitions are not created within the semantic index, they are created outside so that we keep the index salsa-free, which allows it to be easily reused to figure out data flow dependencies in non-salsa contexts.

Because the definitions are created outside the index we need to be a bit careful. This is a tracked struct, not an interned one, and that means that different Salsa queries create different instances of the definition that don't compare equal, even if they are the same definition. (See https://hackmd.io/5ddaZLKYSVC_YTTD3SzRDw?view#Tracked-Structs, thanks to @DavisVaughan for finding this doc.)

To ensure definitions have a single identity, we now funnel all definitions lookup through a single `File::definitions()` query that creates a map of `Definition` keyed by `Scope` and `DefId` (bundled in a new struct `DefinitionSite`).

New tests ensure that definitions are not invalidated when they shift in a file or when definitions are inserted before (which changes the lookup key but not the definition ID, so downstream queries remain valid even if their key is now outdated).

### DIFF
--- a/crates/oak_db/src/definition.rs
+++ b/crates/oak_db/src/definition.rs
@@ -135,10 +135,6 @@ impl<'db> File {
 }
 
 /// Every `Definition` in a file, keyed by its definition site.
-///
-/// A `salsa::Update` struct rather than a bare map so the salsa-tracked
-/// `File::definitions` query can store it: salsa needs the `Update` impl to
-/// handle the `Definition<'db>` lifetime that the map values carry.
 #[derive(Debug, PartialEq, Eq, salsa::Update)]
 struct FileDefinitions<'db> {
     by_site: FxHashMap<DefinitionSite, Definition<'db>>,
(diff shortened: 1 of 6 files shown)


crates/oak_db/src/definition.rs
```diff
@@ -86,3 +89,62 @@ fn is_right_assignment(node: &RBinaryExpression) -> bool {
...
+#[salsa::tracked]
+impl<'db> File {
+    /// Look up the `Definition` entity for the binding at `(scope, def_id)` in
+    /// this file. The entity has already been minted in [`File::definitions`]
+    /// and this is a plain lookup. The salsa id stays stable across edits that
+    /// renumber `def_id`.
+    pub(crate) fn definition(
+        self,
+        db: &'db dyn Db,
+        scope: ScopeId,
```
DavisVaughan:
```suggestion
        scope_id: ScopeId,
```

consistency with `def_id` and other usages

DavisVaughan: should pass through to `DefinitionSite` too i think

crates/oak_db/src/definition.rs
```diff
@@ -86,3 +89,62 @@ fn is_right_assignment(node: &RBinaryExpression) -> bool {
...
+            }
+        }
+
+        FileDefinitions { by_site }
+    }
+}
+
+/// Every `Definition` in a file, keyed by its definition site.
+#[derive(Debug, PartialEq, Eq, salsa::Update)]
+struct FileDefinitions<'db> {
```
DavisVaughan:
```suggestion
struct Definitions<'db> {
```

would also be happy with this since the method is `definitions()`

DavisVaughan (approved):
I like this!

PR #1254, opened 2026-06-03
Use Salsa infrastructure for cross-file goto-definition 

lionel-:
Branched from https://github.com/posit-dev/ark/pull/1253
Progress towards https://github.com/posit-dev/ark/issues/1149

- Use new Salsa infrastructure for cross-file queries (via `source()`d files in scripts, or file collation in packages)
- `File` gains a `line_index()` query.
- `PositionEncoding` is stored in `LspConfig`, mirroring r-a
- Non-redundant parked LSP-layer tests are restored (either in oak_db, oak_ide, or ark).

One regression with the Salsa infra: we no longer analyse files outside of workspace. In the future we could add these as Salsa inputs via oak_scan. Would also require a non-lsp file watcher for out-of-workspace updates.

I recommend testing with https://github.com/posit-dev/ark/pull/1259 to benefit from bug fixes and improvements in ulterior branches.


crates/oak_ide/src/goto_definition.rs
```diff
@@ -1,70 +1,28 @@
...
+/// (lexical scopes, function-body resolution, top-level reaching defs, file
+/// imports). Each binding that could reach the use becomes a
+/// [`NavigationTarget`] in its own file's coordinates, so an ambiguous name
+/// (e.g. defined on both arms of an `if`/`else`) yields several. Empty means
+/// the name isn't reachable, rather than guessing by name across the workspace.
+pub fn goto_definition(db: &dyn Db, file: File, offset: TextSize) -> Vec<NavigationTarget> {
+    file.resolve_at(db, offset)
+        .into_iter()
+        .filter_map(|def| {
+            let range = def.name_range(db)?;
```
DavisVaughan, line 19: I see name_range "returns None for `Import` kinds". That's correct here? Just checking

crates/oak_db/src/file_resolve.rs
```diff
@@ -75,72 +78,86 @@ impl<'db> File {
...
+    pub fn resolve_at(self, db: &'db dyn Db, offset: TextSize) -> Vec<Definition<'db>> {
         let index = self.semantic_index(db);
-        let (use_scope, _use_id, use_site) = index.use_at(offset)?;
+        let Some((use_scope, use_id, use_site)) = index.use_at(offset) else {
+            // Cursor on a binding's own name (a def site, not a use): jump to
+            // it, like rust-analyzer / ty.
+            return match index.definition_at(offset) {
+                Some((scope, def_id, _)) => {
+                    self.definition(db, scope, def_id).into_iter().collect()
+                },
```
DavisVaughan, line 101: Another peculiarity of 2 `Definition` types is that we can't just return the SemanticIndex's `Definition` captured in the `_`. I understand why and all that, just pointing out that we nearly had it "right there"

crates/oak_db/src/file_resolve.rs
```diff
@@ -75,72 +78,86 @@ impl<'db> File {
...
+    /// both arms of an `if`/`else` yields two. A cursor on a binding's own name
+    /// resolves to that binding. Empty means nothing reachable binds the name
+    /// here.
     ///
     /// Not `#[salsa::tracked]` because keying on `(self, offset)` would
     /// balloon the cache. The `Definition` entities it returns are minted by
-    /// the tracked [`File::definitions()`], so they stay cached and
-    /// identity-stable even though this lookup isn't.
-    pub fn resolve_at(self, db: &'db dyn Db, offset: TextSize) -> Option<Definition<'db>> {
+    /// the tracked [`File::definitions()`] and looked up via
```
DavisVaughan, line 90: So has the meaning of `resolve_at()` changed quite a bit? It looks like the impl is quite different now

crates/oak_db/src/file_resolve.rs
```diff
@@ -75,72 +78,86 @@ impl<'db> File {
...
+
+    fn resolve_definition(
+        self,
+        db: &'db dyn Db,
+        scope: ScopeId,
+        def_id: DefinitionId,
+    ) -> Option<Definition<'db>> {
+        let index = self.semantic_index(db);
+        if let DefinitionKind::Import {
+            file: target_url,
```
DavisVaughan, line 152: Should this be a `FilePath`?

crates/ark/src/lsp/config.rs
```diff
@@ -81,11 +83,27 @@ pub static DOCUMENT_SETTINGS: &[Setting<DocumentConfig>] = &[
...
+#[derive(Clone, Debug)]
 pub(crate) struct LspConfig {
     pub(crate) diagnostics: DiagnosticsConfig,
     pub(crate) symbols: SymbolsConfig,
     pub(crate) workspace_symbols: WorkspaceSymbolsConfig,
+
+    /// Session-wide position encoding for offset <-> LSP-position conversion.
+    /// One value for the whole session, not per document. Hard-coded to UTF-16,
+    /// the encoding we advertise at `initialize`.
+    pub(crate) position_encoding: PositionEncoding,
```
DavisVaughan, line 95: I think I generally agree with this, but can we actually implement what it says?

i.e. go ahead and remove `position_encoding` from `Document`. It does not look too hard, as every usage of `document.position_encoding` seems very close to an LSP boundary where you have the `WorldState` available.

crates/ark/src/lsp/goto_definition.rs
```diff
@@ -1,1172 +1,62 @@
...
+        .collect::<anyhow::Result<Vec<_>>>()?;
 
-    Ok(Some(LocationLink {
-        origin_selection_range: None,
-        target_uri: file_id.as_uri().clone(),
-        target_range: entry.range,
-        target_selection_range: entry.range,
-    }))
+    Ok(Some(GotoDefinitionResponse::Link(links)))
 }
```
DavisVaughan, line 42: I really like how cleanly this reads on the LSP impl side of things!

crates/oak_db/src/imports.rs
```diff
@@ -52,6 +52,18 @@ impl<'db> ImportsResolver for SalsaImportsResolver<'db> {
...
+        // invisible. Minting can't happen here, so the work belongs on the
+        // write side in `oak_scan`. We should carry the resolved path on the
+        // directive even when no `File` exists (today the miss returns `None`
+        // and drops it), then have `oak_scan` enumerate source directives after
+        // a scan, mint an `OrphanRoot` `File` from disk for each
+        // out-of-workspace target, and iterate for `source()` chains. A file
+        // watcher is only needed for freshness (re-reading after an external
+        // edit), plus GC to drop the orphan once the directive goes away.
+        // TODO(diagnostics): Until we support out-of-workspace sourced files,
+        // should we at least lint so user knows that we can't analyse the file?
```
DavisVaughan, line 66: Are we sure we _want_ to support out of workspace `source()` files?

i.e. it might actually be more practically useful to _only_ lint to discourage this

crates/oak_db/src/file_resolve.rs
```diff
@@ -75,72 +78,86 @@ impl<'db> File {
...
-        // last winner between `<-` and a same-name `source()`.
-        let file_scope = ScopeId::from(0);
-        if let Some((binding_scope, def_id, _def)) = index.resolve(&name, use_scope) {
-            if binding_scope != file_scope {
-                return self.definition(db, binding_scope, def_id);
-            }
-        }
-
-        let in_function = use_scope != file_scope;
         let interned = Name::new(db, name.as_str());
```
DavisVaughan:
```suggestion
        let name = Name::new(db, name.as_str());
```

?

seeing `interned` later on isn't very helpful

crates/oak_db/src/file_resolve.rs
```diff
@@ -75,72 +78,86 @@ impl<'db> File {
...
         }
 
-        None
+        Vec::new()
+    }
+
+    fn resolve_definition(
+        self,
+        db: &'db dyn Db,
+        scope: ScopeId,
```
DavisVaughan: `scope_id`?

crates/oak_db/src/file_resolve.rs
```diff
@@ -75,72 +78,86 @@ impl<'db> File {
...
+        def_id: DefinitionId,
+    ) -> Option<Definition<'db>> {
+        let index = self.semantic_index(db);
+        if let DefinitionKind::Import {
+            file: target_url,
+            name: forwarded,
+            ..
+        } = index.definitions(scope)[def_id].kind()
```
DavisVaughan: So a `reaching_definitions()` definition could still be an Import from another file? I had some intuition that a reaching definition would only be local to this file

DavisVaughan (approved):

crates/oak_ide/src/goto_definition.rs
lionel-: Yes and that should never happen since `resolve_at()` resolves imports

crates/ark/src/lsp/config.rs
lionel-: This is out of scope for this PR but eventually yes

crates/oak_db/src/file_resolve.rs
lionel-: Possibly, but likely not. The `Url` is just a vessel in oak_semantic and doesn't need to guarantee lexical normalisation or anything of that sort. On the other hand it wouldn't hurt much. Overall I'd say let's worry about this later.

crates/oak_db/src/file_resolve.rs
lionel-: If you `source()` a file you get local imports that you need to resolve

crates/oak_db/src/imports.rs
lionel-: It could be useful in monorepos but not high priority for sure

crates/oak_db/src/file_resolve.rs
lionel-: It returns Vec instead of Option, handles the cursor-on-def case, and is more flow-sensitive via `reaching_definitions()`.

PR #1257, opened 2026-06-05
Use Salsa infrastructure for find-references and rename

### LINKED ISSUE #1149: Implement goto-definition, find-references, and rename with oak's symbol resolution

lionel-:
Branched from https://github.com/posit-dev/ark/pull/1254
Closes https://github.com/posit-dev/ark/issues/1149

This PR moves find-references and rename onto the Salsa infra and deletes the old implementation (the textual workspace walk).

Candidate search: Unlike Rust-Analyzer that does a textual search across the workspace and resolves all occurrences, we're more aligned with ty: we filter files to inspect with a textual search, but then use a post-parse structure to filter occurrences. ty walks the AST, we walk the index instead with a new `uses_of()` method that is more to the point than the raw AST.

Candidates are then narrowed with `File::resolve_at()` to check that they match the definition of the symbol at point.

New supporting infra:

- Salsa-based `Identifier::classify()`
- `all_files()` (Salsa query) that returns all files known to the database
- `root_by_file()` that disambiguates to which workspace root a file belong when roots are nested. That is meant to be the one source of truth for root ownership of files. 

Removed infra: The `Document` copies of the Rowan parse tree and semantic index are now removed, in favour of the Salsa DB.

Other fix: Goto-definition now benefits from the new Salsa-based `Identifier::classify()` which snaps the cursor to the symbol at point. Fixes goto-def when cursor in on RHS boundary.

I recommend testing with https://github.com/posit-dev/ark/pull/1259 to benefit from bug fixes and improvements in ulterior branches.


crates/oak_semantic/src/semantic_index.rs
```diff
@@ -188,6 +188,22 @@ impl SemanticIndex {
...
+            let Some(symbol_id) = self.symbols(scope_id).id(name) else {
+                continue;
+            };
+            for (_use_id, use_site) in self.uses(scope_id).iter() {
+                if use_site.symbol() == symbol_id {
+                    ranges.push(use_site.range());
+                }
+            }
+        }
+        ranges
```
DavisVaughan: Looking at the definition of this function and the return value of `use_at()` being `Option<(ScopeId, UseId, &Use)>`, to me it feels like the right return value for this would be `Vec<(ScopeId, UseId, &Use)>`.

You've got all the pieces to build that return value already. It feeeels like it would make sense to return the maximal amount of looked up info here

crates/ark/src/lsp/tests/find_references.rs
```diff
@@ -96,225 +89,103 @@ fn test_cursor_past_trailing_edge_resolves_intra_file() {
...
-
-    let file2_path = dir.path().join("b.R");
-    std::fs::write(&file2_path, "x <- 99\nx\n").unwrap();
-
-    let doc1 = Document::new(file1_code, None);
-    let mut state = make_state(&file1_uri, &doc1);
-    state.workspace.folders = vec![lsp_types::Url::from_directory_path(dir.path()).unwrap()];
+fn test_namespace_access_returns_empty() {
+    // Cursor on `mutate` in `dplyr::mutate` is a namespace access, returns nothing.
+    let code = "dplyr::mutate\n";
```
DavisVaughan: i guess it could return every time you call dplyr's `mutate()`?

crates/ark/src/lsp/tests/find_references.rs
```diff
@@ -96,225 +89,103 @@ fn test_cursor_past_trailing_edge_resolves_intra_file() {
...
     let locs = find_references(params, &state).unwrap();
-
-    // Only file1's two `x` occurrences (the parameter and its use). No
-    // file2 hits.
-    let expected = vec![
-        Location::new(file1_uri.clone(), range((0, 14), (0, 15))),
-        Location::new(file1_uri, range((1, 2), (1, 3))),
-    ];
-    assert_eq!(locs, expected);
+    assert!(locs.is_empty());
```
DavisVaughan: did we lose `test_function_scope_target_skips_cross_file_walk` as a test?

crates/oak_db/src/identifier.rs
```diff
@@ -0,0 +1,192 @@
...
+    At,
+}
+
+impl<'db> Identifier<'db> {
+    /// Classify the identifier at `offset` in `file`, snapping to the
+    /// nearest name-token boundary first.
+    ///
+    /// Returns `None` when the cursor isn't on a variable binding/use or
+    /// a member name (e.g. cursor is on an operator, a keyword, or a
+    /// `pkg::sym` namespace access).
```
DavisVaughan: I noticed that `NamespaceAccess` went away as an Identifier type?

crates/oak_db/src/identifier.rs
```diff
@@ -0,0 +1,192 @@
...
+/// The semantic identity of the identifier the cursor is on, after snapping
+/// to the nearest name-token boundary.
+#[derive(Debug, Clone)]
+pub enum Identifier<'db> {
+    /// Cursor on a binding use or definition site tracked by the semantic
+    /// index. The range is the use or def site's text range.
+    Variable { name: Name<'db>, range: TextRange },
+    /// Cursor on the RHS name of a `$` or `@` extract expression. Member
+    /// names are not tracked by the semantic index.
+    Member {
```
DavisVaughan, line 28: does this all work with nested `$` calls? like `foo$bar$baz` regardless of where the cursor is?

crates/oak_db/src/identifier.rs
```diff
@@ -0,0 +1,192 @@
...
+            return Some(Identifier::Variable {
+                name: Name::new(db, name_str),
+                range: use_site.range(),
+            });
+        }
+
+        if let Some((scope_id, _def_id, def)) = index.definition_at(snapped) {
+            let name_str = index.symbols(scope_id).symbol(def.symbol()).name();
+            return Some(Identifier::Variable {
+                name: Name::new(db, name_str),
```
DavisVaughan:
```suggestion
            let name = index.symbols(scope_id).symbol(use_site.symbol()).name();
            return Some(Identifier::Variable {
                name: Name::new(db, name),
                range: use_site.range(),
            });
        }

        if let Some((scope_id, _def_id, def)) = index.definition_at(snapped) {
            let name = index.symbols(scope_id).symbol(def.symbol()).name();
            return Some(Identifier::Variable {
                name: Name::new(db, name),
```

crates/oak_db/src/identifier.rs
```diff
@@ -0,0 +1,192 @@
...
+    ///
+    /// Returns `None` when the cursor isn't on a variable binding/use or
+    /// a member name (e.g. cursor is on an operator, a keyword, or a
+    /// `pkg::sym` namespace access).
+    pub fn classify(db: &'db dyn Db, file: File, offset: TextSize) -> Option<Identifier<'db>> {
+        let parse = file.parse(db);
+        let root = parse.syntax();
+        let index = file.semantic_index(db);
+
+        let snapped = snap_to_name_at_boundary(&root, offset);
```
DavisVaughan:
```suggestion
        let offset = snap_to_name_at_boundary(&root, offset);
```

would be in favor of this

crates/oak_db/src/identifier.rs
```diff
@@ -0,0 +1,192 @@
...
+
+        if let Some((scope_id, _use_id, use_site)) = index.use_at(snapped) {
+            let name_str = index.symbols(scope_id).symbol(use_site.symbol()).name();
+            return Some(Identifier::Variable {
+                name: Name::new(db, name_str),
+                range: use_site.range(),
+            });
+        }
+
+        if let Some((scope_id, _def_id, def)) = index.definition_at(snapped) {
```
DavisVaughan: What about `DefinitionKind::Import`? That returns the right thing here? i.e. the returned `range` is relevant?

crates/oak_db/src/identifier.rs
```diff
@@ -0,0 +1,192 @@
...
+
+impl<'db> File {
+    /// All use-site ranges for `name` in this file, across every scope.
+    pub fn uses_of(self, db: &'db dyn Db, name: Name<'db>) -> Vec<TextRange> {
+        self.semantic_index(db).uses_of(name.text(db).as_str())
+    }
+
+    /// All ranges where `name` appears as the RHS of a `$` or `@` with the
+    /// given `kind` in this file. Structural scan of the parse tree.
+    pub fn member_uses(self, db: &'db dyn Db, name: &str, kind: MemberKind) -> Vec<TextRange> {
```
DavisVaughan: rename to `uses_of_member()`? to match `uses_of()`?

crates/oak_db/src/identifier.rs
```diff
@@ -0,0 +1,192 @@
...
+        let root = self.parse(db).syntax();
+        root.descendants()
```
DavisVaughan, line 149: Is `descendants()` only at top level? Or does it recurse though braces and parens?

crates/oak_db/src/identifier.rs
```diff
@@ -0,0 +1,192 @@
...
+    Some((name, kind, op.text_trimmed_range(), name_range))
+}
+
+fn member_name_and_range(selector: &AnyRSelector) -> Option<(String, TextRange)> {
+    match selector {
+        AnyRSelector::RIdentifier(ident) => {
+            Some((ident.name_text(), ident.syntax().text_trimmed_range()))
+        },
+        AnyRSelector::RStringValue(s) => Some((s.string_text()?, s.syntax().text_trimmed_range())),
+        _ => None,
```
DavisVaughan: Did you test `find_references` with ``` foo$`backticked var` ``` and `foo$"string var"` to ensure that the stripping of backticks and quotes here is consistent with whatever `name` is passed in?

crates/oak_db/src/identifier.rs
```diff
@@ -0,0 +1,192 @@
...
+    let (name, name_range) = member_name_and_range(&rhs)?;
+    Some((name, kind, op.text_trimmed_range(), name_range))
+}
+
+fn member_name_and_range(selector: &AnyRSelector) -> Option<(String, TextRange)> {
+    match selector {
+        AnyRSelector::RIdentifier(ident) => {
+            Some((ident.name_text(), ident.syntax().text_trimmed_range()))
+        },
+        AnyRSelector::RStringValue(s) => Some((s.string_text()?, s.syntax().text_trimmed_range())),
```
DavisVaughan, line 253: regarding the `None` fallthrough, i mean technically you can do `foo$...` and `foo$..1` sooooooo maybe we should include those and make this exhaustive

DavisVaughan: plus a test

crates/oak_db/src/identifier.rs
```diff
@@ -0,0 +1,192 @@
...
+            if is_name_token(&left) {
+                left.text_trimmed_range().start()
+            } else if is_name_token(&right) {
+                right.text_trimmed_range().start()
+            } else {
+                offset
+            }
+        },
+    }
+}
```
DavisVaughan: What if this was `snap_to_token()` and returned `Option<RSyntaxToken>`?

That would mean
- you can early exit from `classify()` if there is no token at cursor
- you can always look up the offset as `token.text_trimmed_range().start()`
- `classify_member()` doesn't have to re-call `token_at_offset().right_biased()` with a special note there. you'd already have the token of interest

crates/oak_db/src/identifier.rs
```diff
@@ -0,0 +1,192 @@
...
+
+fn is_name_token(token: &RSyntaxToken) -> bool {
+    // FIXME: Right now we're too liberal with strings. We should only match
+    // them when they stand for an identifier, e.g. in the LHS of `<-` or in
+    // function call position.
+    matches!(
+        token.kind(),
+        RSyntaxKind::IDENT | RSyntaxKind::R_STRING_LITERAL
```
DavisVaughan, line 297: Might have to include `DotDotDot` and `DotDotI` to make `foo$...` and `foo$..i` work

crates/oak_db/src/db.rs
```diff
@@ -155,8 +191,42 @@ pub(crate) fn root_by_package_query(db: &dyn Db, pkg: Package) -> Option<Root> {
...
+/// matching the package tiebreak. Ownership is keyed on reachability (the
+/// file is in the root's [`root_url_index`]), not bare path prefix, so the
+/// owner always actually contains the file. That's what makes it safe for
+/// [`all_files`] to emit each file exactly once: the owner is guaranteed to
+/// visit it. A bare path prefix would name a freshly-added but not-yet-scanned
+/// nested root and drop the file until its scan landed.
+///
+/// Returns `None` for orphan files (they live in no workspace or library
+/// root). [`File::root`] handles that case with a path-prefix fallback.
+pub(crate) fn root_by_file(db: &dyn Db, file: File) -> Option<Root> {
```
DavisVaughan, line 216: Why is `root_by_package()` a member function on `db`, but this is a free function?

DavisVaughan: Very much feels like this should be `root_by_file_query()`?

crates/oak_db/src/db.rs
```diff
@@ -102,6 +103,41 @@ pub(crate) fn live_roots_query(db: &dyn Db) -> Vec<LiveRoot> {
...
+/// Used as the workspace-wide candidate pool for find-references: callers
+/// apply a textual name filter before building indexes.
+///
+/// Nested roots overlap on disk, so the same `File` is reachable from
+/// several roots (open `/proj` and `/proj/sub-pkg` and the outer scan walks
+/// into `sub-pkg`). We emit each file once by attributing it to a single
+/// owner via `root_by_file`, the same deepest-root rule
+/// [`root_by_package`] uses.
+#[salsa::tracked(returns(ref))]
+pub fn all_files(db: &dyn Db) -> Vec<File> {
```
DavisVaughan, line 119: So this result would change anytime a new file is added or removed from the workspace right? And it seems like it could get pretty big too. When the result changes, is the old one cleaned up?

crates/oak_db/src/db.rs
```diff
@@ -102,6 +103,41 @@ pub(crate) fn live_roots_query(db: &dyn Db) -> Vec<LiveRoot> {
...
+    let mut files: Vec<File> = Vec::new();
+
+    for &root in db.live_roots() {
+        match root {
+            LiveRoot::Workspace(r) | LiveRoot::Library(r) => {
+                let owned = |f: File| root_by_file(db, f) == Some(r);
+                files.extend(r.scripts(db).iter().copied().filter(|&f| owned(f)));
+                for &pkg in r.packages(db) {
+                    let pkg_files = pkg.files(db).iter().chain(pkg.scripts(db));
```
DavisVaughan: Would it be easier push all files into a HashSet that is converted to a Vec at the end?

Or would that possibly not retain the right ordering properties of the resulting `Vec<File>` in the case of a File in both `/proj` and `/proj/sub-pkg`? I guess it depends on the order that we iterate over 2 workspaces via `live_roots()`

DavisVaughan: Mostly im wondering how expensive the `root_by_file()` check is on every File, because it looks fairly expensive. I wonder if there is a simpler way

crates/oak_ide/src/rename.rs
```diff
@@ -19,61 +22,75 @@ pub struct RenameTargets {
...
+    offset: TextSize,
+) -> anyhow::Result<Option<(TextRange, Name<'db>)>> {
+    let Some(Identifier::Variable { name, range }) = Identifier::classify(db, file, offset) else {
+        return Ok(None);
+    };
+
+    for def in file.resolve_at(db, range.start()) {
+        if is_library_def(db, def) {
+            return Err(anyhow!(
+                "Can't rename: symbol is defined in an installed package."
```
DavisVaughan: Small code smell about the fact that we have to precisely repeat this logic here and in `find_references()`

I'm not sure we can really do anything about that, but they will have to carefully stay in sync

I originally thought maybe we could just make `find_references()` smarter by giving it a `for_rename: true` type of argument (kind of like what `include_declarations` already it), but I don't think that would work because `prepare_rename()` needs to be fast, right? Like it does the much simpler "is this even renamable?" check before the expensive thing.

So some amount of duplication is probably unavoidable, I think I just wish this duplication was closer to the `find_references()` implementation so you can see them side by side in the same file to hopefully help keep them in sync?

crates/oak_ide/src/find_references.rs
```diff
@@ -1,159 +1,133 @@
...
+        Identifier::Member { name, kind, .. } => {
+            find_member_references(db, file, name.text(db).as_str(), kind)
+        },
+    }
+}
 
+fn find_variable_references(
+    db: &dyn Db,
+    file: File,
+    snapped: TextSize,
```
DavisVaughan:
```suggestion
    offset: TextSize,
```

idk, i'd prefer this? just for overall consistency of using `offset` as much as possible

crates/oak_ide/src/find_references.rs
```diff
@@ -1,159 +1,133 @@
...
-            name.to_string(),
-        ),
-        Identifier::NamespaceAccess { .. } => {
-            return References {
-                ranges: Vec::new(),
-                locally_scoped: false,
-            };
+    match ident {
+        Identifier::Variable { range, .. } => {
+            find_variable_references(db, file, range.start(), include_declaration)
```
DavisVaughan: What if you expanded the `..` and passed `name` through so you don't need to do `let name = target_defs[0].name(db);` later on? That currently feels a bit awkward to me. Like, is `target_defs[1].name()` potentially different??

crates/oak_ide/src/find_references.rs
```diff
@@ -1,159 +1,133 @@
...
-    References {
-        ranges: results,
-        locally_scoped,
+    sort_file_ranges(&mut results, db, file);
+    results
+}
+
+fn find_member_references(
+    db: &dyn Db,
+    primary: File,
```
DavisVaughan, line 170:
```suggestion
    db: &dyn Db,
    file: File,
```

For consistency with `find_variable_references()`. Only `sort_file_ranges()` should probably have the concept of `primary`?

crates/ark/src/lsp/state.rs
```diff
@@ -129,6 +129,7 @@ impl WorldState {
     }
 }
 
+#[allow(dead_code)]
```
DavisVaughan: Could remove it?

crates/oak_ide/src/rename.rs
```diff
@@ -19,61 +22,75 @@ pub struct RenameTargets {
...
 pub fn rename(
-    index: &SemanticIndex,
-    root: &RSyntaxNode,
-    position: &FilePosition,
+    db: &dyn Db,
+    file: File,
+    offset: TextSize,
     new_name: &str,
 ) -> anyhow::Result<RenameTargets> {
     let new_text = to_identifier_text(new_name)?;
```
DavisVaughan: I could see this call moving out of `rename()` up to the lsp level

So then you would not pass `new_name` through or return it. And `RenameTargets` could just be `Vec<FileRange>`

Mostly I was a bit surprised to see it flow through to here since this function is mostly about "find the sites to rename at", only to find that its passed through here _just_ for name normalization.

I guess its a question of who owns the normalization process, oak_ide or the lsp layer?

DavisVaughan (approved):
Very cool!

crates/ark/src/lsp/tests/find_references.rs
lionel-: yep good catch, now done. Part of the feature needs to be implemented in https://github.com/posit-dev/ark/pull/1258#issuecomment-4716567688

crates/oak_db/src/identifier.rs
lionel-: I thought I'd leave it off until later, but resurrected it as part of https://github.com/posit-dev/ark/pull/1257#discussion_r3396604946

crates/ark/src/lsp/tests/find_references.rs
lionel-: Renamed to `test_function_scope_target_stays_in_file` and there's a similar test in IDE layer

crates/oak_db/src/identifier.rs
lionel-: Yep it does, I added tests for this

crates/oak_db/src/identifier.rs
lionel-: Imports have an empty range so they never contain `offset` and never fall through here. I wonder if we should tighten the `Definition` type for places where imports have been resolved.

crates/oak_db/src/identifier.rs
lionel-: With AI, which drastically reduces the use of completions, we're slowly diverging from the prefix-class style of naming, and `uses_of_member()` reads weird. That said we can still make them more consistent, I've renamed `member_uses()` to `member_uses_of()` and did the same for the new `namespace_uses()`.

crates/oak_db/src/identifier.rs
lionel-: I've intentionally excluded those, as they are not regular identifiers.

crates/oak_db/src/identifier.rs
lionel-: Full recursive descent

crates/oak_db/src/identifier.rs
lionel-: I prefer to keep those separate because they do have very different semantics than other identifiers.

But making the handling explicit in the match arms is clearer, now done.

crates/oak_db/src/identifier.rs
lionel-: Good idea! I made it `snap_to_name_token()`

crates/oak_db/src/db.rs
lionel-: `root_by_package()` is public, and `root_by_file()` is private

crates/oak_db/src/db.rs
lionel-: Yes all structural workspace changes (and library changes once they have source files) invalidate this query.

It's a tracked query, so invalidated values are cleaned up.

crates/oak_db/src/db.rs
lionel-: I've replaced `root_by_file()` with a seen set. Which root owns the file doesn't matter here, so we can return the first occurrence.

crates/oak_ide/src/rename.rs
lionel-: I've extracted `File::resolve_variable_at()` in `identifier.rs`

crates/oak_ide/src/rename.rs
lionel-: I think you're right, that's a better separation of concerns

PR #1258, opened 2026-06-05
Resolve symbols from attached and imported packages

lionel-:
Branched from https://github.com/posit-dev/ark/pull/1257

-  `File::resolve()` and `resolve_at()` now handle the `Package` (NAMESPACE `import()`) and `From` (`importFrom()`) import layers (previously only `File` layers were walked; the rest were dropped).

- `Package::resolve()` answers "is this name exported by this package?" by walking the package's files and collecting their exports. It provides a lazy / post-load view of package exports. For instance it returns two definitions when a top-level stub is overwritten in the `.onLoad()` hook via `<<-`.

- `Package::files` are now stored in collation order, which simplifies downstream calls that need to respect R's load order.

`File::imports()` and `Package::resolve()` are both used in `File::resolve()` / `File::resolve_at()`, which themselves support goto-def, find-refs, and rename.

With these changes, when a workspace contains both scripts and a package, you can now use goto-def, find-refs, and rename with symbols imported via `library()`.

goto-def will also work with installed packages when oak_sources is wired up again.

### DIFF
--- a/crates/oak_db/src/file_resolve.rs
+++ b/crates/oak_db/src/file_resolve.rs
@@ -13,6 +13,7 @@ use crate::ExportEntry;
 use crate::File;
 use crate::ImportLayer;
 use crate::Name;
+use crate::PackageVisibility;
 
 #[salsa::tracked]
 impl<'db> File {
@@ -29,12 +30,11 @@ impl<'db> File {
     ///    `source()`-forwarded entries. `ExportEntry::Import` is chased
     ///    through `exports(target)` until it lands on a `Local`. Cycles in
     ///    `source()` resolve to empty exports via `exports`'s `cycle_fn`.
-    /// 2. **`imports()` walk**: each `ImportLayer::File` sibling is checked
-    ///    via its own exports chain only (not its full `resolve`). Sibling
-    ///    package files would otherwise cycle through each other's
-    ///    `imports`, and R's namespace semantics don't transitively include
-    ///    siblings' imports anyway. Package-level layers (`From`,
-    ///    `Package`) are deferred to PR 4.
+    /// 2. **`imports()` walk**: each layer is checked in priority order.
+    ///    `File` siblings are checked via their exports chain only (not their
+    ///    full `resolve`), to avoid the cycle that recursing would create.
+    ///    `Package` and `From` layers call [`Package::resolve`] with
+    ///    `Exported` visibility.
     ///
     /// The returned `Definition` is keyed by `(file, scope, name)`, so
     /// downstream queries that only depend on identity stay cached across
@@ -62,16 +62,36 @@ impl<'db> File {
         // and the installed-package search path appear in this file's own
         // `imports()` directly, as `From` / `Package` layers, so finding
         // them does not require walking through siblings.
-        //
-        // TODO(sources): consume `From` and `Package` layers here. Today
-        // resolve only walks `ImportLayer::File`.
-        // Requires materializing installed-package files as `oak_db::File`
-        // entities first.
         for layer in self.imports(db) {
-            if let ImportLayer::File(target) = layer {
-                if let Some(def) = target.resolve_export(db, name) {
-                    return Some(def);
-                }
+            match layer {
+                ImportLayer::File(target) => {
+                    if let Some(def) = target.resolve_export(db, name) {
+                        return Some(def);
+                    }
+                },
+                ImportLayer::Package(pkg) => {
+                    if let Some(def) = pkg
+                        .resolve(db, name, PackageVisibility::Exported)
+                        .into_iter()
+                        .next()
+                    {
+                        return Some(def);
+                    }
+                },
+                ImportLayer::From(map) => {
+                    let name_str = name.text(db).as_str();
+                    if let Some(pkg_name) = map.get(name_str) {
+                        if let Some(pkg) = db.package_by_name(pkg_name) {
+                            if let Some(def) = pkg
+                                .resolve(db, name, PackageVisibility::Exported)
+                                .into_iter()
+                                .next()
+                            {
+                                return Some(def);
+                            }
+                        }
+                    }
+                },
             }
         }
 
@@ -129,12 +149,31 @@ impl<'db> File {
 
         // Top level: collation predecessors / other visible files (exports-only
         // chase, same as `resolve`'s imports walk). Avoids the sibling cycle and
-        // matches R's namespace semantics. TODO: Package-level layers.
+        // matches R's namespace semantics.
         for layer in self.imports_at(db, offset) {
-            if let ImportLayer::File(target) = layer {
-                if let Some(def) = target.resolve_export(db, name) {
-                    return vec![def];
-                }
+            match layer {
+                ImportLayer::File(target) => {
+                    if let Some(def) = target.resolve_export(db, name) {
+                        return vec![def];
+                    }
+                },
+                ImportLayer::Package(pkg) => {
+                    let defs = pkg.resolve(db, name, PackageVisibility::Exported);
+                    if !defs.is_empty() {
+                        return defs;
+                    }
+                },
+                ImportLayer::From(map) => {
+                    let name_str = name.text(db).as_str();
+                    if let Some(pkg_name) = map.get(name_str) {
+                        if let Some(pkg) = db.package_by_name(pkg_name) {
+                            let defs = pkg.resolve(db, name, PackageVisibility::Exported);
+                            if !defs.is_empty() {
+                                return defs;
+                            }
+                        }
+                    }
+                },
             }
         }
 
@@ -164,7 +203,11 @@ impl<'db> File {
     /// `Import` entries through target exports until a `Local` is found. Cycles
     /// resolve to `None` via `exports`'s `cycle_fn`.
     #[salsa::tracked]
-    fn resolve_export(self, db: &'db dyn Db, name: Name<'db>) -> Option<Definition<'db>> {
+    pub(crate) fn resolve_export(
+        self,
+        db: &'db dyn Db,
+        name: Name<'db>,
+    ) -> Option<Definition<'db>> {
         let mut current_file = self;
         let mut current_name: Rc<str> = Rc::from(name.text(db).as_str());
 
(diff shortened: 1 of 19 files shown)


crates/oak_scan/src/packages.rs
```diff
@@ -185,25 +185,88 @@ fn read_workspace_package(package_dir: &Path) -> Option<PackageEntry> {
...
 }
 
+/// Split the package's `R/*.R` files into `(loadable, leftover)` using the
+/// `Collate:` directive: `loadable` is the listed files in that order, and
+/// `leftover` is the R/ files not listed. Logs mismatches in either direction:
+/// `Collate:` entries with no file on disk, and R/ files absent from `Collate:`
+/// (which become standalone scripts, see [`read_workspace_package`]).
+///
+/// TODO(diagnostics): surface these as LSP diagnostics on `DESCRIPTION`
+/// instead of just log lines.
```
DavisVaughan, line 212: I question whether we should include them at all as standalone scripts vs just logging a diagnostic!

crates/oak_ide/tests/integration/rename.rs
```diff
@@ -268,6 +242,40 @@ fn test_rename_refuses_library_package_symbol() {
...
+    // refuses with the installed-package guard. Before package-layer resolution
+    // this use was unbound and errored with "no binding" instead.
+    let mut db = OakDatabase::new();
+    let _pkg_file =
+        install_library_package(&mut db, "mypkg", &["foo"], "a.R", "foo <- function() 42\n");
+    let script = upsert(&mut db, "script.R", "library(mypkg)\nfoo\n");
+
+    let use_start = "library(mypkg)\n".len() as u32;
+    let err = rename(&db, script, offset(use_start), "bar").unwrap_err();
+    assert!(err.to_string().contains("installed package"));
```
DavisVaughan: cool!

maybe a snapshot test for the error message?

crates/oak_ide/tests/integration/rename.rs
```diff
@@ -268,6 +242,40 @@ fn test_rename_refuses_library_package_symbol() {
...
+    let pkg_file =
+        install_workspace_package(&mut db, "mypkg", &["foo"], "a.R", "foo <- function() 42\n");
+    let script = upsert(&mut db, "script.R", "library(mypkg)\nfoo\n");
+
+    let use_start = "library(mypkg)\n".len() as u32;
+    let result = rename(&db, script, offset(use_start), "bar").unwrap();
+    assert_eq!(pairs(&result.ranges), vec![
+        (script, range(use_start, use_start + 3)),
+        (pkg_file, range(0, 3)),
+    ]);
```
DavisVaughan, line 229: ehhhh maybe?

any time i see `library(mypkg)` in a script, I assume that `mypkg` is installed and read only, even if im working on `mypkg` itself

i think the reason i would disagree with this is that a simple `load_all()` would not get things back in working order after the rename. with a `library()` call at the top, you really need to build and install and restart R first, then youll be back in working order.

crates/oak_ide/tests/integration/goto_definition.rs
```diff
@@ -80,3 +61,27 @@ fn test_navigates_across_source_directive() {
...
+
+    // A script attaches `mypkg`, then uses the exported `foo`.
+    let script = upsert(&mut db, "script.R", "library(mypkg)\nfoo\n");
+    let offset = TextSize::from("library(mypkg)\n".len() as u32);
+
+    let targets = goto_definition(&db, script, offset);
+    assert_eq!(targets.len(), 1);
+    let target = &targets[0];
+
+    // Jumps into the package file, at the `foo` binding's coordinates.
```
DavisVaughan, line 83: nice!

crates/oak_ide/tests/integration/find_references.rs
```diff
@@ -319,3 +288,23 @@ fn test_locally_scoped_stays_in_file() {
...
+        install_library_package(&mut db, "mypkg", &["foo"], "a.R", "foo <- function() 42\n");
+    let script = upsert(&mut db, "script.R", "library(mypkg)\nfoo\n");
+
+    let use_start = "library(mypkg)\n".len() as u32;
+    let refs = find_references(&db, script, offset(use_start), true);
+    assert_eq!(pairs(&refs), vec![
+        (script, range(use_start, use_start + 3)),
+        (pkg_file, range(0, 3)),
+    ]);
+}
```
DavisVaughan: maybe also include `include_declarations: false` to prove that the `pkg_file` reference isn't found in that case?

crates/oak_db/src/package_resolve.rs
```diff
@@ -0,0 +1,74 @@
...
+    ///   directives: an internal binding is invisible even if it exists.
+    /// - `Internal` (R's `pkg:::name`) returns any top-level binding regardless
+    ///   of NAMESPACE.
+    ///
+    /// Iterates [`Package::files`] and aggregates each file's
+    /// [`File::resolve_export`] for `name`.
+    ///
+    /// Returns a `Vec` because a package's namespace can carry more than one
+    /// binding per name. A common pattern is a top-level stub overridden from
+    /// an `.onLoad` hook via `<<-`:
```
DavisVaughan, line 32: But overwriting via `<<-` would never result in a 2nd top level binding, so how can it come back as a candidate?

DavisVaughan: Oh, or is it that `<<-` itself is special in the semantic index and registers `foo` as a top level binding even though its found in the `onLoad` scope?

crates/oak_db/src/package_resolve.rs
```diff
@@ -0,0 +1,74 @@
...
+                .namespace(db)
+                .exports
+                .contains_str(name.text(db).as_str())
+        {
+            return Vec::new();
+        }
+
+        let mut results = Vec::new();
+        for &file in self.files(db) {
+            results.extend(file.resolve_export(db, name));
```
DavisVaughan, line 69: Do re-exports work? i.e. does `export(tibble)` work when you just have `tibble::tibble` in the R file?
https://github.com/tidyverse/dplyr/blob/d5e94e7fa8fd4a5f79c1a707d1842216bb4c691f/R/reexport-tibble.R#L23

crates/oak_db/src/file_resolve.rs
```diff
@@ -62,16 +62,36 @@ impl<'db> File {
...
+                },
+                ImportLayer::From(map) => {
+                    let name_str = name.text(db).as_str();
+                    if let Some(pkg_name) = map.get(name_str) {
+                        if let Some(pkg) = db.package_by_name(pkg_name) {
+                            if let Some(def) = pkg
+                                .resolve(db, name, PackageVisibility::Exported)
+                                .into_iter()
+                                .next()
+                            {
```
DavisVaughan, line 89: These use `next()` and only take the first hit, should this function return `Vec<Definition>`?

DavisVaughan: And if so, should there be corresponding tests somewhere that ensure that multiple definitions work?

DavisVaughan (approved):

crates/oak_db/src/file_resolve.rs
DavisVaughan: ha, resolved by https://github.com/posit-dev/ark/pull/1259

crates/oak_ide/tests/integration/rename.rs
lionel-: hmm I'm not convinced by the argument that you need to reinstall. The LSP has a strictly static view of the world with the workspace package taking precedence over what's installed, that's its normal way of operating.

That said, it does feel a little safer to require first jumping to the definition of the symbol at point to rename, just as an extra safety measure. But also there's a convenience to just being able to rename functions from your scratch files, when you're working on your utility package (less so for a public CRAN package).

I'm on the fence too, just leaning towards that extra flexibility which is kind of cool.

crates/oak_scan/src/packages.rs
lionel-: I'm not a fan of rust-analyzer's behaviour where you get absolutely no LSP features when a file is not attached to a crate (generally due to a bug or latency when editing crate/module relationships). Classifying as standalone scripts strands the file, but at least you can still use internal navigation features etc.

lionel-:
Fix `TODO(namespace-refs)` notes from https://github.com/posit-dev/ark/pull/1257#discussion_r3396604946

crates/oak_ide/tests/integration/rename.rs
lionel-: We're asserting strings rather than snapshot them

crates/oak_db/src/package_resolve.rs
lionel-: Right `<<-` is additive, not overwriting, like when two arms of an if/else condition define the same name.

crates/oak_db/src/package_resolve.rs
lionel-: They did not! Now done, good catch

crates/oak_ide/tests/integration/rename.rs
DavisVaughan: I think my main point comes down to the fact that if I see `library(mypkg)`, then I was expecting all LSP related features for `mypkg` to come from a version of it that lives in my `.libPaths()`, not one that comes from what is in my workspace.

(That would include go-to definition, which would jump to cached sources even though you are in the mypkg workspace. And rename, which would not be renamable on `fn()` of `mypkg::fn()`)

I think my brain wants to trade "this is a cool feature if you're also in this package's workspace" for "i can always predict what this does really easily"

lionel-: I think the current approach is really nice for the development of monorepos.

Another viewpoint is that the LSP should be consistent with what you can observe in the Console, but that becomes hard quickly (load_all, outdated package loaded, etc).

In comparison I submit that the "workspace entity has precedence" is a very predictable rule, even if surprising at first if it doesn't match your initial expectation.

A toml/yaml dependency file could clear it up, but it brings heavier machinery that is not guaranteed to pay off for the R audience.

DavisVaughan: Okay, as-is is fine with me

PR #1259, opened 2026-06-05
Resolve names to all their candidate definitions

lionel-:
Branched from https://github.com/posit-dev/ark/pull/1258

Symbol resolvers that don't take cursor offsets (`File::resolve()`, `resolve_at()`, etc) now return all potential candidates, rather than the last one. The goto-def, find-refs, and rename handlers now offer all of them.

- Candidates come back in definition order, with R's runtime winner (last assignment wins) as the last element, and that order is preserved across `source()` imports.

- A top-level `<<-` now records a single file-scope definition carrying both `IS_SUPER_BOUND` and `IS_BOUND,` rather than a marker plus a same-scope target. At file scope there's no enclosing frame to walk to, so the two coincide. This avoids a spurious duplicate candidate.

### DIFF
--- a/crates/ark/src/lsp/tests/goto_definition.rs
+++ b/crates/ark/src/lsp/tests/goto_definition.rs
@@ -147,9 +147,35 @@ fn test_local_def_shadows_sourced() {
 
 #[test]
 fn test_sourced_file_with_repeated_def_offers_both() {
-    // When the sourced file binds the same name twice, goto-def offers both
-    // candidate definitions, in definition order. The last link is the binding
-    // R picks at runtime. Ranges are in the target file's coordinates.
+    // When the sourced file binds the same name on both arms of a top-level
+    // `if`/`else`, either arm could run, so goto-def offers both candidate
+    // definitions, in definition order. Ranges are in the target file's
+    // coordinates.
+    let script_uri = test_path("script.R");
+    let helpers_uri = test_path("helpers.R");
+    let state = make_state_with(&[
+        (&script_uri, "source(\"helpers.R\")\nfn\n"),
+        (&helpers_uri, "if (cond) fn <- 1 else fn <- 2\n"),
+    ]);
+
+    let params = make_params(script_uri, 1, 0);
+    assert_matches!(
+        goto_definition(params, &state).unwrap(),
+        Some(GotoDefinitionResponse::Link(ref links)) => {
+            assert_eq!(links.len(), 2);
+            assert_eq!(links[0].target_uri, helpers_uri);
+            assert_eq!(links[0].target_range, range((0, 10), (0, 12)));
+            assert_eq!(links[1].target_uri, helpers_uri);
+            assert_eq!(links[1].target_range, range((0, 23), (0, 25)));
+        }
+    );
+}
+
+#[test]
+fn test_sourced_file_with_sequential_redef_offers_runtime_winner() {
+    // When the sourced file binds the same name twice in sequence, the second
+    // overwrites the first, so only the last binding is in effect when
+    // `source()` finishes. goto-def offers just that one.
     let script_uri = test_path("script.R");
     let helpers_uri = test_path("helpers.R");
     let state = make_state_with(&[
@@ -164,11 +190,9 @@ fn test_sourced_file_with_repeated_def_offers_both() {
     assert_matches!(
         goto_definition(params, &state).unwrap(),
         Some(GotoDefinitionResponse::Link(ref links)) => {
-            assert_eq!(links.len(), 2);
+            assert_eq!(links.len(), 1);
             assert_eq!(links[0].target_uri, helpers_uri);
-            assert_eq!(links[0].target_range, range((0, 0), (0, 2)));
-            assert_eq!(links[1].target_uri, helpers_uri);
-            assert_eq!(links[1].target_range, range((1, 0), (1, 2)));
+            assert_eq!(links[0].target_range, range((1, 0), (1, 2)));
         }
     );
 }
--- a/crates/ark/src/lsp/tests/goto_definition.rs
+++ b/crates/ark/src/lsp/tests/goto_definition.rs
@@ -146,10 +146,10 @@ fn test_local_def_shadows_sourced() {
 }
 
 #[test]
-fn test_last_def_in_sourced_file_wins() {
-    // When the sourced file binds the same name twice, the use navigates to
-    // the last definition. The link range must point at that second def, in
-    // the target file's coordinates.
+fn test_sourced_file_with_repeated_def_offers_both() {
+    // When the sourced file binds the same name twice, goto-def offers both
+    // candidate definitions, in definition order. The last link is the binding
+    // R picks at runtime. Ranges are in the target file's coordinates.
     let script_uri = test_path("script.R");
     let helpers_uri = test_path("helpers.R");
     let state = make_state_with(&[
@@ -164,8 +164,11 @@ fn test_last_def_in_sourced_file_wins() {
     assert_matches!(
         goto_definition(params, &state).unwrap(),
         Some(GotoDefinitionResponse::Link(ref links)) => {
+            assert_eq!(links.len(), 2);
             assert_eq!(links[0].target_uri, helpers_uri);
-            assert_eq!(links[0].target_range, range((1, 0), (1, 2)));
+            assert_eq!(links[0].target_range, range((0, 0), (0, 2)));
+            assert_eq!(links[1].target_uri, helpers_uri);
+            assert_eq!(links[1].target_range, range((1, 0), (1, 2)));
         }
     );
 }
--- a/crates/oak_db/src/file_exports.rs
+++ b/crates/oak_db/src/file_exports.rs
@@ -15,7 +15,7 @@ use crate::File;
 /// internal edits.
 #[derive(Debug, Default, Clone, PartialEq, Eq)]
 pub struct FileExports {
-    entries: HashMap<String, ExportEntry>,
+    entries: HashMap<String, Vec<ExportEntry>>,
 }
 
 #[derive(Debug, Clone, PartialEq, Eq)]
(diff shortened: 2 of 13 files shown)


crates/ark/src/lsp/tests/goto_definition.rs
```diff
@@ -164,8 +164,11 @@ fn test_last_def_in_sourced_file_wins() {
     assert_matches!(
         goto_definition(params, &state).unwrap(),
         Some(GotoDefinitionResponse::Link(ref links)) => {
+            assert_eq!(links.len(), 2);
             assert_eq!(links[0].target_uri, helpers_uri);
-            assert_eq!(links[0].target_range, range((1, 0), (1, 2)));
+            assert_eq!(links[0].target_range, range((0, 0), (0, 2)));
+            assert_eq!(links[1].target_uri, helpers_uri);
+            assert_eq!(links[1].target_range, range((1, 0), (1, 2)));
```
DavisVaughan: Is the idea that you aren't really going for technical accuracy here?

Like, from a technical perspective only the 2nd one is relevant.

But from a user experience perspective, I do think I would want to be made aware of both. Likely I have made some kind of mistake resulting in two `fn` of the same name.

Is that what you are thinking?

crates/oak_semantic/src/builder.rs
```diff
@@ -175,6 +175,27 @@ impl<R: ImportsResolver> SemanticIndexBuilder<R> {
     // the first scope where the symbol is already bound. If no binding is
     // found, it assigns in the global (file) scope.
     fn add_super_definition(&mut self, name: &str, kind: DefinitionKind, range: TextRange) {
+        let target_scope = self.resolve_super_target(name);
+
+        // A top-level `<<-` resolves its target to the file scope it already
+        // sits in (no enclosing frame to walk to), so the marker scope and the
+        // binding scope coincide. Record one definition carrying both flags,
+        // rather than pushing duplicate definition entries.
+        if target_scope == self.current_scope {
```
DavisVaughan: I was really hoping to see some comparison against `file_scope` here, for clarity.

Like, would it work to do this?

```
if self.current_scope == file_scope {
    // go ahead and record that 1 def and then early exit
}

// now you do the target resolution
let target_scope = self.resolve_super_target(name);
```

or maybe even

```
match self.scopes[self.current_scope].parent {
    Some(parent) => {
        let target_scope = self.resolve_super_target(name, parent);
        // remainder of this branch
        // resolve_super_target would simplify to just the `loop`
    }
    None => {
        // current scope is file scope, use that path
    }
}
```

that actually feels quite nice to me

crates/oak_semantic/tests/integration/builder.rs
```diff
@@ -573,17 +574,12 @@ fn test_super_assignment_at_file_scope() {
...
+    assert_eq!(index.definitions(file).len(), 1);
     assert!(matches!(
         index.definitions(file)[DefinitionId::from(0)].kind(),
         DefinitionKind::SuperAssignment(_)
     ));
-    assert!(matches!(
-        index.definitions(file)[DefinitionId::from(1)].kind(),
-        DefinitionKind::SuperAssignment(_)
-    ));
     assert_eq!(index.uses(file).len(), 0);
```
DavisVaughan, line 583: that feels right!

crates/oak_semantic/src/semantic_index.rs
```diff
@@ -125,15 +125,18 @@ impl SemanticIndex {
...
 
-        let mut exports = FxHashMap::default();
+        let mut exports: FxHashMap<&str, Vec<(DefinitionId, &Definition)>> = FxHashMap::default();
         for (id, def) in self.definitions[file_scope].iter() {
             let name = symbols.symbol(def.symbol()).name();
-            exports.insert(name, (id, def));
+            exports.entry(name).or_default().push((id, def));
         }
 
         exports
```
DavisVaughan, line 161: Here's a thought

What if this was

```rust
pub fn definitions_by_name(&self, scope: ScopeId) -> FxHashMap<&str, Vec<(DefinitionId, &Definition)>>
```

that would go quite nicely alongside

```rust
pub fn definitions(&self, scope: ScopeId) -> &IndexVec<DefinitionId, Definition>
```

which is the "flat" view of the exact same information

and would make it more useful outside just the file-scope

---

I personally also think this is "more accurate" now that we return EVERY definition and not just the last one.

Like, I think of the "exports" as being a set of unique names. And I can derive the "exports" from this set of definitions, but this set of definitions doesn't correspond to the exports to me.

It makes sense to me for SemanticIndex to provide the maximal amount of information, i.e. every definition. And callers can filter that down as required for their semantic use case. I just think the name doesn't quite line up now.

crates/oak_db/src/file_exports.rs
```diff
@@ -60,27 +64,42 @@ impl File {
...
+                // entry is the binding R picks at runtime (statements run in
+                // order, last write wins). `Local` carries no `def_id` (that
+                // contentlessness is what keeps `exports()` stable across body
+                // edits), so a name's several top-level definitions collapse to
+                // one `Local`, identical `source()` forwards to one `Import`,
+                // and `resolve_export` re-mints from the marker either way.
+                if let Some(pos) = list.iter().position(|e| e == &entry) {
+                    list.remove(pos);
+                }
+                list.push(entry);
```
DavisVaughan: So IIUC `Vec<ExportEntry>` will have:
- At most 1 `ExportEntry::Local` entry
- Possibly multiple `ExportEntry::Import` entries if the same `name` came from multiple files

But `ExportEntry`s themselves are deduplicated based on equality

crates/oak_db/src/file_exports.rs
```diff
@@ -60,27 +64,42 @@ impl File {
     /// chains by returning empty exports for every cycle participant.
     #[salsa::tracked(returns(ref), cycle_result = exports_cycle_result)]
     pub fn exports(self, db: &dyn Db) -> FileExports {
```
DavisVaughan, line 66: Is `resolve_source()`'s usage of `exports()` altered in any way due to these changes?

crates/oak_db/src/file_resolve.rs
```diff
@@ -185,61 +187,100 @@ impl<'db> File {
...
-                    // to recover its `def_id`, then fetch the interned
-                    // definition. `exports()` returns the last-wins
-                    // definitions, so this is the right binding for the name.
-                    let index = current_file.semantic_index(db);
-                    let def_id = index.exports().get(current_name.as_ref())?.0;
-                    return current_file.definition(db, ScopeId::from(0), def_id);
+                    // The `Local` marker doesn't carry a `def_id`, so recover
+                    // every file-scope `def_id` for the name from the semantic
+                    // index and mint each through the single site. A name bound
+                    // more than once at top level fans out here.
```
DavisVaughan: I admittedly do feel a bit weird about the dance we are doing with `Local` here:
- Find all exports
- Collapse multiple into 1 `Local`
- Rediscover all exports for that `Local` here

I saw the note about "keeps `exports()` stable across body edits, so a name's several top-level definitions collapse", but how common is that really?

I would trade a little performance for not doing the collapse and rediscover dance.

Plus, surely rediscovering here also takes some time too, right? And maybe we'd gain back what we lost?

DavisVaughan (approved):
A few medium importance level comments, but looks nice and matches my gut feeling about not using `next()` in `File::resolve()`

crates/ark/src/lsp/tests/goto_definition.rs
lionel-: Actually there's an issue with the approach initially taken in this PR.

From the start I've taken the perspective that `<-` is an entirely new binding. All `<-` are both a declaration and an assignment. I.e. in the following:

```r
function() {
  foo <- 1
  foo <- 2
  foo
}
```

Goto-def on the use only shows you the 2nd definition, and renaming that variable only affects that second definition. Another approach would be to treat all homonyms in a given scope to be the same declared variable. That would technically mean only the first assignment is a declaration, all the subsequent are pure assignments. (Yet another is to make a semantic distinction between `<-` and `=`, but that idea has a socialisation challenge to overcome first).

Oak consistently takes the position that `<-` creates a new binding. The initial version of this PR contradicted this principle. I think my mistake was to think about non-offset exports as providing a lazy view, like in lazy contexts in the semantic index where we report extra definitions. The thing is, that's only true for definitions that occur _after_ the lazy context, not before. Before the lazy context, everything is eagerly evaluated, and only the last live bindings count. And this is the exact situation we're in with file imports / exports.

I've now changed the PR to take this approach. The exports are still a vector because an if/else expression may produce competing definitions, same as in the local index. But a series of `<-` assignments collapses to the last one.

If you're worried about multiple-definition mistakes in packages, we can think about linting them. But for now I'd like to implement the consistent view now offered in this PR. The bottom line is that before this revision, `foo <- 1; foo <- 2` resolved to one definition when used in the same file but two across `source()`. Now both agree.

crates/oak_semantic/src/builder.rs
lionel-: Yep that's nicer

crates/oak_semantic/src/semantic_index.rs
lionel-: It's now really more like exports, following changes described in https://github.com/posit-dev/ark/pull/1259#discussion_r3421227160

crates/oak_db/src/file_resolve.rs
lionel-: It will be increasingly common with type checking and NSE resolution. At the moment, I'm fairly confident these firewalls will pay off.

crates/oak_db/src/file_exports.rs
lionel-: I don't think so because it only depends on the names of exports

crates/ark/src/lsp/tests/goto_definition.rs
DavisVaughan: Soo IIUC it would now goto `foo <- 2` directly since it is a new binding?

lionel-: Yes both in the `source()` case and locally

PR #1260, opened 2026-06-08
Add import layers for testthat files

lionel-:
Branched from https://github.com/posit-dev/ark/pull/1259

Whereas scripts in `tests/` folders are standalone, scripts in `tests/testthat/` are sourced in an environment where:

- Setup and helper files are run.
- The package namespace is injected
- The `testthat` exports are attached

This PR detects testthat files and adjusts the import layers accordingly, so that symbols injected by testthat are visible in these files.

### DIFF
--- a/crates/aether_path/src/file_path.rs
+++ b/crates/aether_path/src/file_path.rs
@@ -118,8 +118,11 @@ impl FilePath {
                     .as_url()
                     .path_segments()?
                     .next_back()
+                    // A trailing slash (`.../foo/`) yields a final empty segment,
+                    // meaning the URL points at a directory. Drop it so we return
+                    // `None` rather than `""`.
                     .filter(|seg| !seg.is_empty())?;
-                Some(percent_decode_str(last).decode_utf8_lossy())
+                percent_decode_str(last).decode_utf8().warn_on_err()
             },
         }
     }
--- a/crates/aether_path/src/file_path.rs
+++ b/crates/aether_path/src/file_path.rs
@@ -12,11 +12,13 @@
 //! is the job of secondary canonical-path indexes at the specific call
 //! sites that need it, never of this type.
 
+use std::borrow::Cow;
 use std::path::PathBuf;
 
 use camino::Utf8Component;
 use camino::Utf8Path;
 use camino::Utf8PathBuf;
+use percent_encoding::percent_decode_str;
 use stdext::result::ResultExt;
 use url::Url;
 
@@ -99,6 +101,26 @@ impl FilePath {
         match self {
             Self::Virtual(u) => Some(u),
             Self::File(_) => None,
+        }
+    }
+
+    /// File name (final path component) of this identity, percent-decoded.
+    ///
+    /// For the `File` arm it's the path basename. For the `Virtual` arm it's
+    /// the last path segment of the URI, dropping any query and fragment.
+    /// `None` for URIs like `untitled:Untitled-1`, which have no path structure
+    /// to take a segment from.
+    pub fn file_name(&self) -> Option<Cow<'_, str>> {
+        match self {
+            Self::File(p) => p.as_path().file_name().map(Cow::Borrowed),
+            Self::Virtual(u) => {
+                let last = u
+                    .as_url()
+                    .path_segments()?
+                    .next_back()
+                    .filter(|seg| !seg.is_empty())?;
+                Some(percent_decode_str(last).decode_utf8_lossy())
+            },
         }
     }
 }
--- a/crates/oak_db/src/file_imports.rs
+++ b/crates/oak_db/src/file_imports.rs
@@ -66,7 +66,9 @@ impl File {
     #[salsa::tracked(returns(ref))]
     pub fn imports(self, db: &dyn Db) -> Vec<ImportLayer> {
         match self.package(db) {
-            Some(package) if is_testthat_file(self, db) => testthat_imports(self, db, package),
+            Some(package) if is_testthat_file(self, db) => {
+                testthat_imports(self, db, package, None)
+            },
             Some(package) => package_imports(self, db, package),
             None => script_imports(self, db),
         }
@@ -104,10 +106,13 @@ impl File {
 
         // Top-level cursor: sequential narrowing.
         match self.package(db) {
-            // testthat sources every helper/setup file before any test
-            // runs, so there's nothing to narrow by offset. The EOF view
-            // already reflects what a test file sees at any point.
-            Some(_) if is_testthat_file(self, db) => self.imports(db).clone(),
+            // Helpers, setup, the package, and testthat are all sourced or
+            // attached before a test file's body runs, so they stay visible
+            // at any offset. Only the file's own top-level `library()` calls
+            // narrow.
+            Some(package) if is_testthat_file(self, db) => {
+                testthat_imports(self, db, package, Some(offset))
+            },
             Some(package) => narrow_package_top_level(self, db, package),
             None => narrow_script_top_level(self, db, offset),
         }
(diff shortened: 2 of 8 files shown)


crates/aether_path/src/file_path.rs
```diff
@@ -101,6 +103,26 @@ impl FilePath {
...
+    /// to take a segment from.
+    pub fn file_name(&self) -> Option<Cow<'_, str>> {
+        match self {
+            Self::File(p) => p.as_path().file_name().map(Cow::Borrowed),
+            Self::Virtual(u) => {
+                let last = u
+                    .as_url()
+                    .path_segments()?
+                    .next_back()
+                    .filter(|seg| !seg.is_empty())?;
```
DavisVaughan, line 124: is this because `https://example.com/foo/` sees that last `/` and returns an empty string for the (not really there) segment after the `/`?

crates/aether_path/src/file_path.rs
```diff
@@ -101,6 +103,26 @@ impl FilePath {
...
+                Some(percent_decode_str(last).decode_utf8_lossy())
```
DavisVaughan: are you sure you dont want `decode_utf8().log_err()?` given that this already returns an option?

crates/oak_db/src/file_imports.rs
```diff
@@ -101,26 +106,21 @@ impl File {
 
         // Top-level cursor: sequential narrowing.
         match self.package(db) {
+            // Helpers, setup, the package, and testthat are all sourced or
+            // attached before a test file's body runs, so they stay visible
+            // at any offset. Only the file's own top-level `library()` calls
+            // narrow.
```
DavisVaughan, line 112: "Only the file's own top-level `library()` calls"

GROSSSSSSSS, THE HORROR

DavisVaughan: LINT THIS?

crates/oak_db/src/file_imports.rs
```diff
@@ -157,30 +157,41 @@ fn narrow_package_top_level(file: File, db: &dyn Db, package: Package) -> Vec<Im
     layers
 }
 
+/// The file's `library()` / `require()` attaches as `Package` layers, in LIFO
+/// order (latest-attached first). `before` narrows to a top-level cursor: with
+/// `Some(offset)`, keep only file-scope calls that have run by `offset`; with
+/// `None`, keep every attach (the end-of-file view, used for lazy contexts).
```
DavisVaughan: is the cursor itself top level? not always right? like you can be in a `test_that({<cursor>})` block and i wouldnt consider that top level?

i would also appreciate if this comment were rewritten as like a bulleted list of the 2 cases. it was very hard for me to mentally parse with the colon and semicolon

crates/oak_db/src/file_imports.rs
```diff
@@ -214,6 +225,117 @@ fn package_imports(file: File, db: &dyn Db, package: Package) -> Vec<ImportLayer
...
+///
+/// 1. helper/setup files (sourced into the test env, shadow everything),
+/// 2. the whole package's `R/` code,
+/// 3. the package's NAMESPACE imports,
+/// 4. the file's own top-level `library()` calls,
+/// 5. `testthat` on the search path,
+/// 6. base.
+///
+/// `offset` narrows the components of layer 4. `None` produces the end-of-file
+/// view and keeps every `library()` call. `Some(offset)` keeps only the calls
```
DavisVaughan:
```suggestion
/// view and utilizes every `library()` call. `Some(offset)` utilizes only the calls
```

"keep" felt like a weird verb here

crates/oak_db/src/file_imports.rs
```diff
@@ -214,6 +225,117 @@ fn package_imports(file: File, db: &dyn Db, package: Package) -> Vec<ImportLayer
...
+    }
+}
+
+fn in_testthat_dir(path: &Utf8Path) -> bool {
+    let Some(parent) = path.parent() else {
+        return false;
+    };
+    parent.file_name() == Some("testthat") &&
+        parent.parent().and_then(Utf8Path::file_name) == Some("tests")
+}
```
DavisVaughan, line 314: we should upgrade to 2024 edition rust at some point for let chains so you can do things like this

```rust
fn in_testthat_dir(path: &Utf8Path) -> bool {
    if let Some(parent) = path.parent() &&
        parent.file_name() == Some("testthat") &&
        let Some(parent) = parent.parent() &&
        parent.file_name() == Some("tests")
    {
        return true
    } else {
        return false
    }
}
```

crates/oak_db/src/file_imports.rs
```diff
@@ -214,6 +225,117 @@ fn package_imports(file: File, db: &dyn Db, package: Package) -> Vec<ImportLayer
...
+    }
+}
+
+/// Sort key for support files, matching testthat's `sort(dir(...))` order.
+/// We sort by raw basename (byte order = C locale for ASCII): case-sensitive
+/// like testthat, and platform-stable. This is a bit different to testthat
+/// which currently sorts based on locale, but arguably this should be fixed on
+/// the testthat side.
+fn testthat_support_key(file: File, db: &dyn Db) -> Cow<'_, str> {
+    file.path(db).file_name().unwrap_or_default()
```
DavisVaughan, line 339: What is the default? An empty Owned string i guess? can you just leave it as `Option<Cow>`? I have no idea if that would sort right

crates/oak_db/src/tests/file_imports.rs
```diff
@@ -187,6 +187,191 @@ fn test_package_file_emits_namespace_and_collation_layers() {
...
+    let test_bar = File::new(
+        &db,
+        file_path("w/pkg/tests/testthat/test-bar.R"),
+        "test_that('y', expect_true(TRUE))\n".to_string(),
+        Some(pkg),
+    );
+
+    pkg.set_files(&mut db).to(vec![r_file]);
+    pkg.set_scripts(&mut db)
+        .to(vec![helper, setup, test_foo, test_bar]);
```
DavisVaughan, line 244: i wonder if it will be useful to have a specific field for `testthat_scripts` or something, so we dont have to tease them apart from regular scripts every time

DavisVaughan (approved):
Cool

crates/aether_path/src/file_path.rs
lionel-: yep I've added this comment:

```
                    // A trailing slash (`.../foo/`) yields a final empty segment,
                    // meaning the URL points at a directory. Drop it so we return
                    // `None` rather than `""`.
```

crates/oak_db/src/file_imports.rs
lionel-: This is now:

```
/// The file's `library()` / `require()` attaches as `Package` layers, in LIFO
/// order (latest-attached first). `before` selects which calls to include:
///
/// - `None`: every attach. The end-of-file view, used for lazy contexts.
/// - `Some(offset)`: only top-level (file-scope) calls that have run by
///   `offset`. Calls nested in a block (e.g. inside `test_that({})`) are
///   dropped, as are calls after the offset.
```

crates/oak_db/src/file_imports.rs
lionel-: Default but it's unreachable.

crates/oak_db/src/tests/file_imports.rs
lionel-: Potentially, but I'd rather have some kind of collation annotation to branch on (via a tracked query).

crates/oak_db/src/file_imports.rs
lionel-: Possibly, but it feels to me more like Jarl's territory than Oak's.

PR #1261, opened 2026-06-09
Return package URL from .ps.rpc.pkg_list

bricestacey:
`.ps.rpc.pkg_list` now returns a `url` field for each installed package,
taken from the first entry of its DESCRIPTION `URL` field (conventionally
the package's canonical website). The field is omitted when a package
advertises no URL.

This feeds the new external-link button in Positron's Packages pane, which
validates the scheme before opening. Multiple / labeled links (repository,
issues, etc.) are left for a future change.

See posit-dev/positron#13890

### DIFF
--- a/crates/ark/src/modules/positron/packages_pane.R
+++ b/crates/ark/src/modules/positron/packages_pane.R
@@ -54,9 +54,7 @@
                 attached = attached,
                 description = description
             )
-            if (nzchar(url)) {
-                entry$url <- url
-            }
+            entry$url <- if (!is.na(url)) url
             entry
         },
         id,
--- a/crates/ark/src/modules/positron/packages_pane.R
+++ b/crates/ark/src/modules/positron/packages_pane.R
@@ -38,11 +38,7 @@
     # whitespace; the first is conventionally the package's canonical website.
     # Surface it as the single best URL (Positron validates the scheme before
     # opening). Drop everything from the first separator on; "" when absent.
-    url <- sub(
-        "[[:space:],].*$",
-        "",
-        trimws(ifelse(is.na(ip[, "URL"]), "", ip[, "URL"]))
-    )
+    url <- sub("[[:space:],].*$", "", trimws(ip[, "URL"]))
     # Build each package as a list so `url` can be omitted entirely when a
     # package advertises none -- a vectorized `Map(list, url = url)` forces the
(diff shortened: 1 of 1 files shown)


crates/ark/src/modules/positron/packages_pane.R
```diff
@@ -34,18 +34,41 @@
...
-    # expects. Strip the names with unname().
+    # DESCRIPTION's URL field can list several URLs separated by commas and/or
+    # whitespace; the first is conventionally the package's canonical website.
+    # Surface it as the single best URL (Positron validates the scheme before
+    # opening). Drop everything from the first separator on; "" when absent.
+    url <- sub(
+        "[[:space:],].*$",
+        "",
+        trimws(ifelse(is.na(ip[, "URL"]), "", ip[, "URL"]))
+    )
```
jennybc:
```suggestion
    url <- sub("[[:space:],].*$", "", trimws(ip[, "URL"]))
```

You could just let missing data be `NA`.

crates/ark/src/modules/positron/packages_pane.R
```diff
@@ -34,18 +34,41 @@
...
+                id = id,
+                name = name,
+                displayName = name,
+                version = version,
+                attached = attached,
+                description = description
+            )
+            if (nzchar(url)) {
+                entry$url <- url
+            }
```
jennybc:
```suggestion
            entry$url <- if (!is.na(url)) url
```

Then this becomes conditional on being non-missing instead of not-the-empty-string.

jennybc (approved):
LGTM!

I made a minor stylistic suggestion that leans more into letting `NA` do its job (indicating missing data), but this isn't a big deal.

I would have thought sending back the same shape all the time might be useful (so always sending `url` even if it's empty), but having an optional `url` must be better on the other side.

For any package that is on CRAN, if DESCRIPTION doesn't list a URL, you could fallback to `https://cran.r-project.org/package={package}`, but then you'd have to determine if it's on CRAN or not, which is extra work. But if you already have easy access to this info, this is an idea.

crates/ark/src/modules/positron/packages_pane.R
bricestacey: This feel more canonical so let's roll with it.

PR #1263, opened 2026-06-11
Add `ArkDb`, `ArkFile`, and `tree_sitter()` query for legacy handlers

lionel-:
Branched from https://github.com/posit-dev/ark/pull/1260
Progress towards https://github.com/posit-dev/ark/issues/1212

This PR starts migrating legacy Ark handlers to the Oak database so that we get one source of truth and remove redundancy of representations (source, syntax trees) in the LSP to save on memory.

Introduces `ArkFile` in the process. Note this is meant as a convenient temporary vessel until we fully migrate handlers to Oak.

- Add `ArkFile` which bundles an Oak `File` with some metadata used by Ark handlers (`version`, wire `url`, `encoding`). This is meant to replace `Document` in a subsequent PR. The LSP <> TS position converters are now methods on `ArkFile` (previously on `Document`).

- Add `ArkDb` and `FileExt` as extension traits for Oak.
  The main addition is a tree-sitter query, capped at 128 trees (like the Rowan trees).

- `spawn_blocking()` now catches Salsa cancellations when reads on backaground tasks (e.g. diagnostics) are cancelled by a concurrent write.

- Start migrating a bunch of handlers to that new Oak infra.

PR #1264, opened 2026-06-11
Migrate hover, signature help, and completions to `ArkFile`

lionel-:
Branched from https://github.com/posit-dev/ark/pull/1263

The handlers migrated use a slightly different approach because they use `r_task()` heavily and the Oak DB shouldn't be touched from an `r_task()` because of the risk of cancellation panics, which would be UB over the R stack. We could catch panics in `r_task()` and unwind for cancellation but I didn't want to think through the various implications.

- `DocumentContext` drops `Document` and stores the pieces it needs from the `ArkDb` and LSP state (`tree`, `contents`, `line_index`, `encoding`). This way the DB is not sent over an `r_task()`.

- Add LSP <> TS position/range converters as methods on `DocumentContext` (parallel to the ones on `ArkFile`, also replacements for the `Document` methods).

- Migrate hover, signature help, and completion to the refactored `DocumentContext`.

### DIFF
--- a/crates/ark/src/fixtures/utils.rs
+++ b/crates/ark/src/fixtures/utils.rs
@@ -29,17 +29,6 @@ pub fn r_test_init() {
     Console::test_init();
 }
 
-pub const TEST_ENCODING: aether_lsp_utils::proto::PositionEncoding =
-    aether_lsp_utils::proto::PositionEncoding::Wide(biome_line_index::WideEncoding::Utf16);
-
-pub fn tree_sitter_parse(code: &str) -> tree_sitter::Tree {
-    let mut parser = tree_sitter::Parser::new();
-    parser
-        .set_language(&tree_sitter_r::LANGUAGE.into())
-        .unwrap();
-    parser.parse(code, None).unwrap()
-}
-
 pub fn point_from_cursor(x: &str) -> (String, Point) {
     // i.e. looking for `@` in something like `fn(x = @1, y = 2)`, and it treats the
     // `@` as the cursor position
--- a/crates/ark/src/fixtures/utils.rs
+++ b/crates/ark/src/fixtures/utils.rs
@@ -29,6 +29,17 @@ pub fn r_test_init() {
     Console::test_init();
 }
 
+pub const TEST_ENCODING: aether_lsp_utils::proto::PositionEncoding =
+    aether_lsp_utils::proto::PositionEncoding::Wide(biome_line_index::WideEncoding::Utf16);
+
+pub fn tree_sitter_parse(code: &str) -> tree_sitter::Tree {
+    let mut parser = tree_sitter::Parser::new();
+    parser
+        .set_language(&tree_sitter_r::LANGUAGE.into())
+        .unwrap();
+    parser.parse(code, None).unwrap()
+}
+
 pub fn point_from_cursor(x: &str) -> (String, Point) {
     // i.e. looking for `@` in something like `fn(x = @1, y = 2)`, and it treats the
     // `@` as the cursor position
--- a/crates/ark/src/lsp/document_context.rs
+++ b/crates/ark/src/lsp/document_context.rs
@@ -5,9 +5,7 @@
 //
 //
 
-use aether_lsp_utils::proto::to_proto;
 use aether_lsp_utils::proto::PositionEncoding;
-use tower_lsp::lsp_types;
 use tree_sitter::Node;
 use tree_sitter::Point;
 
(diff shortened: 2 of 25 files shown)


crates/ark/src/fixtures/utils.rs
```diff
@@ -29,6 +29,17 @@ pub fn r_test_init() {
...
+pub const TEST_ENCODING: aether_lsp_utils::proto::PositionEncoding =
+    aether_lsp_utils::proto::PositionEncoding::Wide(biome_line_index::WideEncoding::Utf16);
+
+pub fn tree_sitter_parse(code: &str) -> tree_sitter::Tree {
+    let mut parser = tree_sitter::Parser::new();
+    parser
+        .set_language(&tree_sitter_r::LANGUAGE.into())
+        .unwrap();
+    parser.parse(code, None).unwrap()
+}
```
DavisVaughan: This is now the 3rd use of the same helper. See my request in the other PR where we should use exactly 1 parse helper.

crates/ark/src/lsp/document_context.rs
```diff
@@ -28,22 +35,25 @@ pub struct DocumentContext<'a> {
...
-        // get reference to AST
-        let ast = &document.ast;
+    pub fn new(
+        tree: &'a tree_sitter::Tree,
+        contents: &'a str,
+        encoding: PositionEncoding,
+        point: Point,
+        trigger: Option<String>,
+    ) -> Self {
+        let line_index = biome_line_index::LineIndex::new(contents);
```
DavisVaughan: We have an `ArkFile` query for the `line_index()` now too, right? i.e. it can be cached across multiple `DocumentContext`s being created for the same `ArkFile`? It feels worth it to me to try and use that instead of recreating it here, even if that adds to test noise

crates/ark/src/lsp/document_context.rs
```diff
@@ -28,22 +35,25 @@ pub struct DocumentContext<'a> {
 }
 
 impl<'a> DocumentContext<'a> {
-    pub fn new(document: &'a Document, point: Point, trigger: Option<String>) -> Self {
-        // get reference to AST
-        let ast = &document.ast;
+    pub fn new(
```
DavisVaughan, line 40: Would it help to add `DocumentContext::new_test(contents: &'a str, point: Point)` that internally:
- parses the contents
- makes the line index
- sets trigger to None
- sets encoding to UTF16

and use that in most or all of the tests?

it seems like that might clean things up and avoid the need for `TEST_ENCODING`?

crates/ark/src/lsp/document_context.rs
```diff
@@ -85,13 +95,36 @@ impl<'a> DocumentContext<'a> {
...
+    }
+
+    pub fn lsp_range_from_tree_sitter_range(
+        &self,
+        range: tree_sitter::Range,
+    ) -> anyhow::Result<lsp_types::Range> {
+        let start = self.lsp_position_from_tree_sitter_point(range.start_point)?;
+        let end = self.lsp_position_from_tree_sitter_point(range.end_point)?;
+        Ok(lsp_types::Range::new(start, end))
+    }
```
DavisVaughan: I don't like having implementations for these in two places (here and ArkFile)

Can you extract out free functions of the same name (these would probably live in ArkFile) that directly take the LineIndex and PositionEncoding as arguments and use those in both ArkFile and here?

Or, actually, if they were free functions then you could probably expose them as methods on ArkFile (to abstract over the db stuff) but for DocumentContext it is probably better to just use the free functions directly (because supplying `self.line_index` and `self.encoding` isn't a big deal) rather than having these methods at all.

crates/ark/src/lsp/document_context.rs
```diff
@@ -5,17 +5,24 @@
...
 use tree_sitter::Node;
 use tree_sitter::Point;
 
-use crate::lsp::document::Document;
 use crate::lsp::traits::node::NodeExt;
 use crate::treesitter::NodeType;
 use crate::treesitter::NodeTypeExt;
 
 #[derive(Debug)]
 pub struct DocumentContext<'a> {
```
DavisVaughan, line 21: Should it be renamed to `ArkFileContext` at some point?

DavisVaughan (approved):
Looks good, my main comments are about reducing duplication and noise

crates/ark/src/lsp/document_context.rs
```diff
@@ -5,17 +5,24 @@
...
-    pub document: &'a Document,
+    /// We store extracted components of `&ArkFile` + `&dyn ArkDb` here because
+    /// the latter can't be sent over an `r_task()`.
+    pub tree: &'a tree_sitter::Tree,
```
DavisVaughan, line 24: I would like to point out that we have

```
    // TODO(oak/completions): Clone so the closure captures by value. `r_task()`
    // sends the closure across threads, and `&WorldState` isn't `Send` because
    // `OakDatabase`'s salsa storage keeps thread-local query state.
    let state = state.clone();
    let completions = r_task(move || provide_completions(&context, &state))?;
```

So is something wrong here?

It's pretty odd to me that we:
- Are creating `DocumentContext` to make `context` here because of `ArkDb` issues
- Then we send over `WorldState`, which has those exact issues

Something feels off

DavisVaughan: Like, isn't this us using a salsa db _inside an r_task_? Which we definitely aren't supposed to do in case it panics?

lionel-: The completion handler runs on the main loop and thus can't be cancelled.

That said I agree this state of things is not great, but it's actually "fixed" in this PR: https://github.com/posit-dev/ark/pull/1269

DavisVaughan: > completion handler runs on the main loop and thus can't be cancelled

ah because it blocks the next request/notification handler from running at all, so we can't process a `did_change()` that might result in us being cancelled. gotcha.

PR #1266, opened 2026-06-12
Migrate the workspace indexer to `ArkFile` and make it a query

lionel-:
Branched from https://github.com/posit-dev/ark/pull/1264

With this PR the Ark indexer that currently powers legacy workspace symbols, completions, and diagnostics is migrated to sources from the Oak DB, instead of walking the disk and watching.

Benefits:

- Single source of truth
- Less memory consumption
- Reactivity to changes for workspace files that are not opened in the editor

This last property is quite nice when changing git branches or editing files with an agent. The changes used to be ignored until the file was opened, but now the IDE reacts instantly.

https://github.com/user-attachments/assets/0b99be22-54a8-4522-b233-164b02be979f

- Building a file index is now a Salsa query over `oak_db::File`. Making it a query per file is especially nice during initial workspace scans where Salsa cancellations don't invalidate finished work per file, they remain cached across edits without having to finish building the entire index. This replaces the old global mutable index.

- New `workspace_files()` helper that collects, as the name says, all workspace files. Unlike `all_files()`, library roots are excluded. This is what `indexer::find()` and `map()` use to walk over files.

- Dropped the combined indexer + diagnostics queue. Now diagnostics query the index, so the ordering we used to enforce by hand (finish indexing, then run diagnostics) just falls out of the pull model. I've kept diagnostics batching and deduping as these are still useful.

- Now that the indexer is a query, it's fully lazy. That's a good property in principle but can lead to user-visible delays when requesting workspace symbols in an empty workspace on startup. If a file is opened, diagnostics will query the index and warm it up, but if none are opened, nothing drives that.

  To fix this issue, we now warm up the index manually after the initial workspace scan has settled. This mirrors how Rust-Analyser warms up its own general-purpose queries after the language server reaches "quiescence" (after the initial scan, after a workspace refresh caused by changing Cargo.toml, etc). There is no such mechanism in ty, it's fully on-demand. I opted to go the r-a way as it's closer to what we had before.

- Diagnostics triggering is now simplified by looking at `salsa::plumbing::current_revision()` from the main loop. Rust-Analyser uses it too, ty doesn't (they have a different pull model for diagnostics that I didn't investigate much yet, our push model is closer to R-A's).

- The indexer was the only consumer of the rename events. We now watch files purely through `DidChangeWatchedFiles`, which is consistent which is also what ty and r-a do.
 
  The one event we might want to subscribe later on is `WillRename`. This would be useful to e.g. update `source()` paths when a file is renamed.

### DIFF
--- a/crates/ark/src/lsp/indexer.rs
+++ b/crates/ark/src/lsp/indexer.rs
@@ -24,7 +24,6 @@ use crate::treesitter::BinaryOperatorType;
 use crate::treesitter::NodeType;
 use crate::treesitter::NodeTypeExt;
 use crate::treesitter::TsQuery;
-use crate::url::ExtUrl;
 
 #[derive(Clone, Debug, PartialEq, Eq, salsa::Update)]
 pub enum IndexEntryData {
@@ -91,8 +90,7 @@ pub(crate) struct FileIndex {
 /// foreign code the user can't edit.
 pub(crate) fn find(db: &dyn ArkDb, symbol: &str) -> Option<IndexEntry> {
     for &file in oak_db::workspace_files(db) {
-        let url = file.path(db).to_url();
-        if !ExtUrl::is_indexable(&url) {
+        if !is_indexable(db, file) {
             continue;
         }
         if let Some(entry) = file_index(db, file).symbols.get(symbol) {
@@ -129,10 +127,10 @@ fn file_index(db: &dyn ArkDb, file: File) -> FileIndex {
 /// Visit every workspace symbol across all indexable files.
 pub(crate) fn map(db: &dyn ArkDb, mut callback: impl FnMut(&Url, &str, &IndexEntry)) {
     for &file in oak_db::workspace_files(db) {
-        let url = file.path(db).to_url();
-        if !ExtUrl::is_indexable(&url) {
+        if !is_indexable(db, file) {
             continue;
         }
+        let url = file.path(db).to_url();
         for (symbol, entry) in file_index(db, file).symbols.iter() {
             callback(&url, symbol, entry);
         }
@@ -145,6 +143,13 @@ pub(crate) fn map(db: &dyn ArkDb, mut callback: impl FnMut(&Url, &str, &IndexEnt
 pub(crate) fn warm(db: &dyn ArkDb) {
     for &file in oak_db::workspace_files(db) {
         file_index(db, file);
+    }
+}
+
+fn is_indexable(db: &dyn ArkDb, file: File) -> bool {
+    match file.path(db) {
+        aether_path::FilePath::File(_) => true,
+        aether_path::FilePath::Virtual(uri) => uri.as_url().scheme() != "ark",
     }
 }
 
--- a/crates/ark/src/lsp/indexer.rs
+++ b/crates/ark/src/lsp/indexer.rs
@@ -18,7 +18,7 @@ use url::Url;
 
 use crate::lsp;
 use crate::lsp::db::ArkDb;
-use crate::lsp::db::FileExt;
+use crate::lsp::db::FileArkExt;
 use crate::lsp::traits::node::NodeExt;
 use crate::treesitter::BinaryOperatorType;
 use crate::treesitter::NodeType;
--- a/crates/ark/src/lsp/indexer.rs
+++ b/crates/ark/src/lsp/indexer.rs
@@ -11,6 +11,7 @@ use std::sync::Arc;
 use std::sync::LazyLock;
 use std::sync::Mutex;
 
+use aether_lsp_utils::proto::PositionEncoding;
 use regex::Regex;
 use stdext::unwrap;
 use stdext::unwrap::IntoResult;
@@ -18,11 +19,10 @@ use tower_lsp::lsp_types::Range;
 use tree_sitter::Node;
 use tree_sitter::Query;
 use url::Url;
-use walkdir::DirEntry;
-use walkdir::WalkDir;
 
 use crate::lsp;
-use crate::lsp::document::Document;
+use crate::lsp::ark_file::ArkFile;
+use crate::lsp::db::ArkDb;
 use crate::lsp::traits::node::NodeExt;
 use crate::treesitter::BinaryOperatorType;
 use crate::treesitter::NodeType;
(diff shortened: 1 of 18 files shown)


crates/oak_db/src/db.rs
```diff
@@ -119,25 +119,44 @@ pub fn all_files(db: &dyn Db) -> Vec<File> {
...
+pub fn workspace_files(db: &dyn Db) -> Vec<File> {
+    let mut files: Vec<File> = Vec::new();
+
+    for &root in db.live_roots() {
+        match root {
+            LiveRoot::Workspace(r) => collect_root_files(db, &mut files, r),
+            LiveRoot::Library(_) => {},
+            LiveRoot::Orphan(orphan) => files.extend(orphan.files(db).iter().copied()),
         }
     }
```
DavisVaughan, line 163: Hmm, orphans?

If I open `~/Desktop/foo.R` and `~/Desktop/bar.R` then I'm not sure I want symbols in `foo.R` to be seen by `bar.R`.

DavisVaughan: And that does seem like what happens now

https://github.com/user-attachments/assets/a6a85625-efe5-4305-b81c-60506a21b5e4

crates/ark/src/lsp/main_loop.rs
```diff
@@ -496,6 +495,10 @@ impl GlobalState {
             lsp::log_info!("Handler took {}ms", loop_tick.elapsed().as_millis());
         }
 
+        if salsa::plumbing::current_revision(&self.world.db) != old_revision {
+            diagnostics_refresh_all(self.world.clone());
+        }
```
DavisVaughan, line 500: Stinks that we have to use an internal method for this

I will note that rust-analyzer doesn't use it like this for _diagnostics_. It uses it for some gc related things. Diagnostics are updated based on 

```rust
            let project_or_mem_docs_changed =
                became_quiescent || state_changed || memdocs_added_or_removed;
            if project_or_mem_docs_changed
                && !self.config.text_document_diagnostic()
                && self.config.publish_diagnostics(None)
            {
                self.update_diagnostics();
            }
```

so they do track state changes manually for this purpose

crates/ark/src/lsp/main_loop.rs
```diff
@@ -1067,116 +989,7 @@ fn refresh_diagnostics(task: RefreshDiagnosticsTask) -> RefreshDiagnosticsResult
 
 /// Run `f`, swallowing a salsa cancellation as `None`. Any other panic propagates.
 fn catch_cancellation<T>(f: impl FnOnce() -> T) -> Option<T> {
```
DavisVaughan, line 989: `log::trace!("Salsa query {cancelled}");` was removed. we dont want to know when it cancels? i know it happens naturally, but is it like super noisy or something?

crates/ark/src/lsp/symbols.rs
```diff
@@ -131,6 +141,37 @@ pub(crate) fn symbols(
...
+    let to_position = |point: indexer::IndexPoint| {
+        to_proto::position_from_line_col(
+            biome_line_index::LineCol {
+                line: point.row,
+                col: point.column,
+            },
+            line_index,
+            encoding,
+        )
+        .ok()
```
DavisVaughan: Swallowing the error feels wrong to me

crates/ark/src/lsp/symbols.rs
```diff
@@ -131,6 +141,37 @@ pub(crate) fn symbols(
     Ok(info)
 }
 
+/// Convert an index entry's tree-sitter point range to an LSP range, resolving
+/// the file's line index from the db. Returns `None` if the file is no longer
+/// in the db or the points fall outside it, in which case the symbol is
+/// dropped from the results.
+fn index_range_to_lsp_range(
```
DavisVaughan: Should this live alongside `IndexRange` instead?

crates/ark/src/lsp/indexer.rs
```diff
@@ -70,112 +44,116 @@ pub enum IndexEntryData {
...
 }
 
-/// Search a specific workspace file for a symbol
-pub fn find_in_file(symbol: &str, uri: &Url) -> Option<(FileId, IndexEntry)> {
-    let index = WORKSPACE_INDEX.lock().unwrap();
+/// Extract a file's workspace symbols.
+#[salsa::tracked(returns(ref))]
+fn file_index(db: &dyn ArkDb, file: File) -> FileIndex {
+    let tree = file.tree_sitter(db);
+    let contents = file.contents(db).as_str();
```
DavisVaughan, line 143: Is this kind of expensive to force up front on every workspace file?

Like, we have to read and store the contents and parse tree of EVERY workspace file up front in the initial round of indexing, that sounds very expensive memory wise

DavisVaughan (approved):

crates/oak_db/src/db.rs
lionel-: Yes diagnostics and the workspace indexer are legacy Ark handlers, they don't use Oak semantics.

The plan is to first expose users to stricter semantics with goto-def, find-refs, and rename, and then tighten diagnostics.

crates/ark/src/lsp/main_loop.rs
lionel-: I'm not too worried about it, and indeed we can work around if it's removed.

crates/ark/src/lsp/main_loop.rs
lionel-: It's noisy

crates/ark/src/lsp/indexer.rs
lionel-: It's bounded by our LRU eviction settings

lionel-: Also worth pointing out that opening a file would (likely) warm up the whole workspace anyway for diagnostics. So it's really about the empty workspace case, so that the user can use cmd + t to navigate to a symbol without delay.

crates/oak_db/src/db.rs
DavisVaughan: Okay

crates/ark/src/lsp/indexer.rs
DavisVaughan: > So it's really about the empty workspace case, so that the user can use cmd + t to navigate to a symbol without delay.

I do this with rust fairly often too, so makes sense

PR #1267, opened 2026-06-12
Finish migrating off `Document` to `ArkFile`

lionel-:
Branched from https://github.com/posit-dev/ark/pull/1266
Progress towards https://github.com/posit-dev/ark/issues/1212

With this PR everything now goes through Oak's DB.

- Remove Ark's `Document` structure, along with its copy of the tree-sitter tree, source code, and line index. These are now all accessed via Oak queries on `File`.

- `WorldState`'s `documents` field becomes `open_files: HashMap<FilePath, ArkFile>`. (Maybe we should rename to `ArkOpenFile`?)

PR #1268, opened 2026-06-12
Consistently send verbatim URLs back to frontend

lionel-:
Branched from https://github.com/posit-dev/ark/pull/1267

This PR straightens out the handing of verbatim wire URLs back to the frontend, so that our lexically normalised paths don't leak out (see #1250).

- New `WorldState::wire_url()` method that resolves verbatim URLs for open files. The LSP layer calls it when communicating URLs back to the frontend, instead of the lexically-normalised paths that travel through the analysis layers.

- The goto-definition, find-references, and rename handlers now use that method to create wire URLs.

- `indexer::map()` now hands `File` instead of URLs, and callers that need a URL call `wire_url()`.

Note that `ArkFile` still contains a copy of the verbatim URLs because Ark's diagnostics and roxygen handlers still use them. This is a temporary situation until these handlers are updated to the modern layering.

### DIFF
--- a/crates/ark/src/lsp/goto_definition.rs
+++ b/crates/ark/src/lsp/goto_definition.rs
@@ -35,7 +35,7 @@ pub(crate) fn goto_definition(
     // to several bindings; the client offers all of them.
     let links = targets
         .iter()
-        .map(|target| nav_target_to_link(db, encoding, target))
+        .map(|target| nav_target_to_link(state, encoding, target))
         .collect::<anyhow::Result<Vec<_>>>()?;
 
     Ok(Some(GotoDefinitionResponse::Link(links)))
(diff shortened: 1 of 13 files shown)


crates/ark/src/lsp/goto_definition.rs
```diff
@@ -45,17 +45,18 @@ pub(crate) fn goto_definition(
...
 ) -> anyhow::Result<LocationLink> {
+    let db = &state.db;
     let line_index = target.file.line_index(db);
     let target_range = to_proto::range(target.full_range, line_index, encoding)?;
     let target_selection_range = to_proto::range(target.focus_range, line_index, encoding)?;
 
     Ok(LocationLink {
         origin_selection_range: None,
-        target_uri: target.file.path(db).to_url(),
+        target_uri: state.wire_url(target.file),
```
DavisVaughan, line 59: We had a discussion on Friday about how it was nice that goto-definition was completely separate from ArkFile

But this ties it back together because `wire_url()` queries `ArkFile.url`

What is the end goal?

crates/ark/src/lsp/symbols.rs
```diff
@@ -68,15 +67,17 @@ pub(crate) fn symbols(
...
         if !symbol.fuzzy_matches(query) {
             return;
         }
 
-        let Some(range) = index_range_to_lsp_range(db, uri, entry.range, encoding) else {
+        let Some(range) = index_range_to_lsp_range(db, file, entry.range, encoding) else {
             return;
         };
 
+        let uri = state.wire_url(file);
```
DavisVaughan, line 76: You can drop the `uri.clone()` below now

DavisVaughan (approved):

crates/ark/src/lsp/goto_definition.rs
lionel-: Nope as we discussed ArkFile is temporary state. But I'll pull the wire URL into an OpenFile type as we discussed, this will look more like the final shape

PR #1271, opened 2026-06-12
Make `File` contents lazy

lionel-:
Branched from https://github.com/posit-dev/ark/pull/1269

This PR implements ty's lazy model for storing sources. This should help keep memory under control, especially when we integrate library packages with the `oak_sources` cache.

- `File` no longer stores the source string as a Salsa input, instead it includes a `revision` field, an optional override for when you do want to store a source (e.g. virtual files and open editors), and a lazy `source_text()` tracked query that reads from disk if needed. Failures return empty sources with an error logged, so the model degrades to empty contents. The query stores at most 128 sources in the Salsa cache, following our storage retention for parse trees.

- `oak_scan` reads the file's `mtime` instead of the contents and store that as revision. The watcher's Changed/Created events bump the revision.

This makes the Salsa model a bit impure. If we miss a FS event, or it's lagging a bit, and re-read from disk at that point, we'll get invalid contents. It becomes possible for different queries to be inconsistent. Should be self-healing as long as we do get the events eventually. Since ty does this, I'm not too worried about this approach. The alternative is that we'll have to store all the sources of the entire dependency tree, which is likely a bit much especially if there are many Ark sessions live in the workspace (e.g. multiple notebooks, shiny apps, etc).


DavisVaughan:
I have not fully read this yet but my brain is also thinking about https://github.com/posit-dev/ark/pull/1266#discussion_r3405685367 and the fact that we are going to be doing a full workspace index (which means reading contents from disk for every file) on startup. So....is it even lazy anymore?

Is it mostly for non workspace `File`s?

DavisVaughan:
Ok I also see it is not just about being lazy from a performance POV

It is probably mostly about gaining that `lru = 128` bound on the number of `File` contents we hold in memory at a time!

crates/oak_db/src/file.rs
```diff
@@ -48,15 +54,59 @@ use crate::Root;
...
+
+        let Some(path) = self.path(db).as_path() else {
+            // Our virtual documents (e.g. untitled://) are push-based, the
+            // editor writes them to the source override field via
+            // `upsert_editor`. If we ever do lazy virtual documents, that is
+            // where we'd hook them up. Until then it would be unexpected to
+            // reach here.
+            log::warn!("Can't read virtual document `{}`", self.path(db));
+            return String::new();
+        };
```
DavisVaughan: This actually would have been a bit more obvious to me via a `match self.path(db)` where you actually see the `Virtual` branch

crates/ark/src/lsp/ark_file.rs
```diff
@@ -55,7 +55,7 @@ impl ArkFile {
     }
 
     pub(crate) fn contents<'db>(&self, db: &'db dyn ArkDb) -> &'db str {
-        self.file.contents(db).as_str()
+        self.file.source_text(db).as_str()
     }
```
DavisVaughan, line 59: Also make this `source_text()` for consistency?

DavisVaughan:
Should `file_index()` over in the indexer also get `lru = 128` since it runs on all workspace files?

crates/oak_scan/src/inputs.rs
```diff
@@ -147,6 +147,11 @@ impl<DB: Db + DbInputs> DbScan for DB {
             return;
         };
 
+        // Clear the editor override so the disk contents becomes the source of
+        // truth again. This immediately invalidates queries that depend on
+        // `source_text`, because the latter depends on the override.
+        file.set_source_text_override(self).to(None);
```
DavisVaughan, line 153: 👍 i was explicitly looking for this, nice

DavisVaughan (approved):
Cool!

crates/ark/src/lsp/ark_file.rs
lionel-: Good idea, I'll do it in the `OpenFile` PR

lionel-:
> Should file_index() over in the indexer also get lru = 128 since it runs on all workspace files?

I don't think so, this needs to stay warm to timely answer workspace queries from the user.

In general I think of LRU mostly as a safeguard for data about _installed_ packages, once they're hooked up to oak_sources. We can revisit if we see issues in large workspaces though.

PR #533, opened 2024-09-18
Add LSP request for input boundaries

lionel-:
For https://github.com/posit-dev/positron/issues/1326

PR #1272, opened 2026-06-13
UI comm: handle inputBoundaries RPC for Positron Quarto

metehankaygsz:
## Summary
- handle `callMethod("inputBoundaries", code)` directly in the UI comm backend
- reuse the existing Rust-side `analysis::input_boundaries` parser logic
- return the same `{ boundaries: [...] }` payload shape expected by Positron
- add a unit test covering the new RPC response

## Why
Positron Quarto now asks the R backend for all statement boundaries in a single request so it can split multi-expression inline chunks without paying for one completeness probe per line.

## Testing
- `cargo test -p ark test_input_boundaries_call_method -- --nocapture`

## Notes
- pairs with the Positron Quarto execution change on [`posit-dev/positron#14185`](https://github.com/posit-dev/positron/pull/14185)
- this is intentionally limited to the UI comm RPC path; the boundary analysis implementation already existed

### DIFF
--- a/crates/ark/src/ui/ui_comm.rs
+++ b/crates/ark/src/ui/ui_comm.rs
@@ -26,12 +26,14 @@ use harp::object::RObject;
 use serde_json::Value;
 use stdext::result::ResultExt;
 
+use crate::analysis::input_boundaries::input_boundaries;
 use crate::comm_handler::handle_comm_message;
 use crate::comm_handler::CommHandler;
 use crate::comm_handler::CommHandlerContext;
 use crate::comm_handler::EnvironmentChanged;
 use crate::console::Console;
 use crate::console::ConsoleOutputCapture;
+use crate::lsp::input_boundaries::InputBoundariesResponse;
 use crate::modules::ARK_ENVS;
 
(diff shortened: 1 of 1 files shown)


lionel-:
Input boundaries are meant to be provided by the LSP, not by the UI comm (we plan to split the LSP and kernel in the future). But I see that you do call the LSP in the Positron-side PR to get input boundaries, which seems right.

Your Positron-side PR already works without this backend-side one, right?

For context: https://github.com/posit-dev/ark/pull/522 and https://github.com/posit-dev/ark/pull/533

metehankaygsz:
Yes, you’re right. I verified that the Positron-side PR already works without this change. `RSession.callMethod("inputBoundaries")` routes the request through the LSP using the existing `positron/inputBoundaries` endpoint in Ark.

I opened this PR because I misunderstood an earlier review comment asking for the accompanying Ark implementation and assumed the UI comm RPC still needed a backend handler. I now understand that the required Ark implementation already exists in the LSP and that adding it to UI comm would duplicate the functionality in the wrong architectural layer.

Thanks for clarifying and for the references to #522 and #533. I’ll close this PR as unnecessary.

PR #1273, opened 2026-06-15
Read metadata of installed packages lazily

### LINKED ISSUE #1265: Improve on laziness of package metadata reading during LSP startup
i noticed that `test_rename_to_empty_name_returns_clean_error` runs quite slow for me (like 5 seconds or so most of the time when the test is run in debug mode, the default)

it looks like `set_library_paths -> scan_new_library_path` can take a loooooooong time when you have quite a few packages, like 600 of them :grimacing:

it does run faster in release mode, i.e. `just test test_rename_to_empty_name_returns_clean_error --release` runs in more like 1 second.

the problem seems to mostly be Air's parsing of 600 package NAMESPACE files, which I think we can probably do lazily and delay until required.

Image
Image

lionel-:
Branched from https://github.com/posit-dev/ark/pull/1271
Closes #1265 

Uses same approach as file contents in #1271 to read `NAMESPACE` and `DESCRIPTION` lazily. This way the only thing that we need to initialise a library of installed packages are the folder names and the presence of a DESCRIPTION file. Unlike workspace packages, we don't read the `Package` field and just assume that a folder with a description file is indeed a package named after the folder name.

This avoids I/O and parsing costs for installed packages that are not imported or mentioned in the workspace files.


crates/oak_db/src/inputs.rs
```diff
@@ -200,12 +201,28 @@ pub struct Package {
     // `Package` directly because salsa inputs are lifetime-free.
     #[returns(ref)]
     pub name: String,
-    /// Installed-package version (from `DESCRIPTION`). `None` for
-    /// workspace packages.
+    /// Mtime of the package's `DESCRIPTION` file, or [`FileRevision::zero`]
+    /// when it can't be stat'd. Drives the lazy [`Package::description`]
```
DavisVaughan: In the "cant be stat'd" case of returning zero, we could end up stuck with an outdated version of the file right? Like we read it once and never again because we never move off 0?

Is it better to force it to read it every time in the case where we cant determine a FileRevision?

(I have no idea if you can even do that with salsa's caching model)

crates/oak_db/src/inputs.rs
```diff
@@ -200,12 +201,28 @@ pub struct Package {
...
+    /// touches disk. Mirrors [`File::source_text_override`]: `None` means
+    /// "read from disk". The scanners always leave it `None`. It's the
+    /// injection point unit tests use to supply a namespace for a synthetic
+    /// path with no file behind it, and the natural hook for an editor
+    /// editing a package's `NAMESPACE` live.
     #[returns(ref)]
-    pub version: Option<String>,
-    #[returns(ref)]
-    pub namespace: Namespace,
+    pub namespace_override: Option<Namespace>,
```
DavisVaughan: But not for description?

crates/oak_db/src/package.rs
```diff
@@ -0,0 +1,112 @@
...
+
+use oak_package_metadata::description::Description;
+use oak_package_metadata::namespace::Namespace;
+use stdext::result::ResultExt;
+
+use crate::Db;
+use crate::Package;
+
+#[salsa::tracked]
+impl Package {
```
DavisVaughan, line 88: Maybe move all of `pub struct Package {` here?

DavisVaughan: And merge with `package_resolve.rs`?

crates/oak_db/src/package.rs
```diff
@@ -0,0 +1,112 @@
...
+    ///
+    /// Returns the in-memory override if one is set
+    /// (`namespace_override`), otherwise reads `NAMESPACE` from disk and
+    /// parses it. Lazy and tracked so we only pay the read and the R-parse
+    /// for packages whose imports actually get resolved. Parsing every
+    /// installed package's `NAMESPACE` eagerly would dominate the cost of
+    /// scanning a library with hundreds of packages, so we defer it.
+    ///
+    /// A missing or unparseable `NAMESPACE` yields an empty `Namespace`.
+    #[salsa::tracked(returns(ref))]
```
DavisVaughan, line 99: lru bound for this and `description()`?

DavisVaughan: i have 600 packages installed fwiw 😬

crates/oak_db/src/package.rs
```diff
@@ -0,0 +1,112 @@
...
+#[salsa::tracked]
+impl Package {
+    /// The package's parsed `NAMESPACE`.
+    ///
+    /// Returns the in-memory override if one is set
+    /// (`namespace_override`), otherwise reads `NAMESPACE` from disk and
+    /// parses it. Lazy and tracked so we only pay the read and the R-parse
+    /// for packages whose imports actually get resolved. Parsing every
+    /// installed package's `NAMESPACE` eagerly would dominate the cost of
+    /// scanning a library with hundreds of packages, so we defer it.
```
DavisVaughan:
```suggestion
    /// scanning a library with hundreds of packages, so we defer it (#1265).
```

crates/oak_db/src/package.rs
```diff
@@ -0,0 +1,112 @@
...
+    ///
+    /// A missing or unparseable `NAMESPACE` yields an empty `Namespace`.
```
DavisVaughan, line 98: It's interesting to decide when to use `-> Namespace` vs `-> Result<Namespace>`

I generally agree with this approach for this one though

crates/oak_db/src/package.rs
```diff
@@ -0,0 +1,112 @@
...
+            // case and stays quiet. A file that exists but can't be read is
+            // logged so the failure isn't silently read as "no namespace".
+            Err(err) if err.kind() == io::ErrorKind::NotFound => return Namespace::default(),
+            Err(err) => {
+                log::error!("Failed to read `{namespace_path}`: {err:?}");
+                return Namespace::default();
+            },
+        };
+
+        Namespace::parse(&text).log_err().unwrap_or_default()
```
DavisVaughan: Move this up into `Ok(text)` to make it a little cleaner?

crates/oak_db/src/package.rs
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    /// The package's `Version:`, parsed lazily from `DESCRIPTION`. `None`
+    /// when the file is missing or has no version.
+    ///
+    /// A narrow query over [`Package::description`]: editing `DESCRIPTION`
+    /// without changing `Version:` backdates here, so downstream isn't
+    /// disturbed.
```
DavisVaughan, line 135: nice

crates/oak_db/src/package.rs
DavisVaughan: Oh actually i see `description()` uses `Option<Description>`. Should we be consistent across these two methods?

crates/oak_db/src/package.rs
```diff
@@ -0,0 +1,112 @@
...
+            // A missing `DESCRIPTION` is the normal "gone after a rescan" case
+            // and stays quiet. A file that exists but can't be read is logged
+            // rather than silently treated as absent.
+            Err(err) if err.kind() == io::ErrorKind::NotFound => return None,
+            Err(err) => {
+                log::error!("Failed to read `{path}`: {err:?}");
+                return None;
+            },
+        };
+        Description::parse(&text).log_err()
```
DavisVaughan: Move into `ok(text)` branch?

crates/oak_db/src/inputs.rs
DavisVaughan: Actually it seems nothing currently uses this

crates/oak_scan/src/packages.rs
```diff
@@ -44,32 +48,40 @@ pub(crate) struct PackageEntry {
...
 /// and `scripts`; callers fill those separately.
 ///
-/// The library scanner uses this directly: installed packages have no `.R`
-/// sources under `R/` (it holds the lazy-load db), so their `files` come from
-/// a cache later, not from a directory scan. The workspace scanner wraps this
-/// in [`read_workspace_package`], which discovers sources by walking the tree.
+/// `NAMESPACE` is only stat'd here, never read or parsed. The lazy
+/// [`oak_db::Package::namespace`] query does the parse. `DESCRIPTION` is read
+/// and parsed because the walk needs the `Package:` name (the workspace
+/// directory name isn't authoritative) and the `Collate:` order to sort
```
DavisVaughan: maybe "isn't authoritative, unlike an installed package's folder name"

because i thought you were actually talking about an installed package's folder name here, and i got confused

DavisVaughan (approved):

crates/oak_db/src/inputs.rs
lionel-: It should get refreshed by the scanner on the next file event, which is the ty model.

Claude says we could use this to force a reread of the query next time: https://github.com/salsa-rs/salsa/blob/86e69d7eacdbc4ee87e183a147af02ee5d932ca1/src/database.rs#L76-L80

crates/oak_db/src/inputs.rs
lionel-: Not needed for description rn

crates/oak_db/src/package.rs
lionel-: But those installed packages will not be queried if they are not part of the dependency graph of the workspace.

crates/oak_db/src/inputs.rs
lionel-: Ok let's try out the approach of reporting an untracked query if the stat failed. Untested and would be very unlikely in real usage though.

crates/oak_db/src/package.rs
lionel-: I've moved the input to package.rs. I'd keep package_resolve.rs separate for symmetry with file_resolve.rs though.

crates/oak_db/src/package.rs
lionel-: When we port completions to salsa, we should make sure completing e.g. `library()` doesn't depend on package details besides the name. If we need more than the name, then we may want LRU.

crates/oak_db/src/package.rs
lionel-: Since reading `DESCRIPTION` is important to get a version, and since versions are involved in what sources we get from oak_sources, I think I prefer not to use the empty file approach for that one.

crates/oak_db/src/inputs.rs
DavisVaughan: Yea I'm not super worried about it, more just curious about how you can express this in salsa, so report-untracked-read is neat to learn about

PR #1274, opened 2026-06-17
Separate editor state (`OpenFile`) from analysis state (`File`)

lionel-:
Branched from https://github.com/posit-dev/ark/pull/1271

We were closer than I thought to a cleaner LSP layer file type, so I went ahead and cleaned that up.

- Rename `ArkFile` to `OpenFile`. Now shrunk to editor/transport state (`inner: File`, `version`, `config`, `wire_url`), dropped the per-file `encoding` field which can be accessed via `LspConfig`.
- Analysis functions now take the bare salsa `File` instead of `&ArkFile`. Encoding is read from `state.config` at the handler boundary.
- Add `state.file(uri)` for handlers needing only the `File`; keep `state.open_file(uri)` for the two that need `wire_url`/`config` (code action, indent)
- `indent_edit` and `roxygen_documentation` now return domain types in tree-sitter coordinates (`IndentEdit`, `RoxygenEdit`). LSP conversion happens in the handler.


crates/ark/src/lsp/open_file.rs
```diff
@@ -0,0 +1,188 @@
...
+    pub(crate) inner: File,
+    pub(crate) version: Option<i32>,
+    pub(crate) config: DocumentConfig,
+    pub(crate) wire_url: Url,
+}
+
+/// Free functions over `LineIndex` + `PositionEncoding`, so anything holding
+/// those two (a `File` plus its `db`, or a `DocumentContext`) can convert
+/// without each carrying its own copy of the logic.
+pub(crate) fn lsp_position_from_tree_sitter_point(
```
DavisVaughan, line 81: Yea, just having free functions feels like the right way to do this to me

crates/ark/src/lsp/code_action.rs
```diff
@@ -25,15 +27,41 @@ pub(crate) struct CodeActions {
     response: lsp_types::CodeActionResponse,
 }
 
+/// Collect the code actions for `range` and assemble them into the LSP response.
 pub(crate) fn code_actions(
     db: &dyn ArkDb,
-    file: &ArkFile,
+    file: &OpenFile,
```
DavisVaughan: I guess if we wanted to make this `File` then it would be pretty complicated

We'd need to return a `CodeActionResponse`-ish type that we own that doesn't need `file.wire_url` or `file.version` but encodes everything else, then at the boundary we'd add in that extra info from the `OpenFile`

crates/ark/src/lsp/open_file.rs
```diff
@@ -0,0 +1,188 @@
...
+use crate::lsp::config::DocumentConfig;
+
+/// An editor-managed buffer: the salsa `File` plus the editor and transport
+/// state that should not cross into the analysis layer.
+#[derive(Clone, Debug)]
+pub(crate) struct OpenFile {
+    pub(crate) inner: File,
+    pub(crate) version: Option<i32>,
+    pub(crate) config: DocumentConfig,
+    pub(crate) wire_url: Url,
```
DavisVaughan: Mild preference for not making the fields `pub`, and instead providing getters for all of them. Most importantly you'd only provide _setters_ for `version` and `config`, making it clear that when you call `open_file_mut()`, those are the only 2 things you are allowed to change.

crates/ark/src/lsp/state.rs
```diff
@@ -82,10 +82,10 @@ impl WorldState {
         }
     }
 
-    pub(crate) fn ark_file_mut(&mut self, uri: &Url) -> anyhow::Result<&mut ArkFile> {
+    pub(crate) fn open_file_mut(&mut self, uri: &Url) -> anyhow::Result<&mut OpenFile> {
```
DavisVaughan, line 84: `did_change()` is the only other place we directly call `self.open_files.get_mut()`. Can we make that use `open_file_mut()` instead?

crates/ark/src/lsp/state.rs
```diff
@@ -107,17 +107,26 @@ impl WorldState {
...
-        let Some(ark_file) = self.open_files.get(&key) else {
+        let Some(open_file) = self.open_files.get(&key) else {
             return Err(anyhow!("Can't find document for URI {uri}"));
         };
-        Ok(ark_file.clone())
+        Ok(open_file.clone())
+    }
+
+    /// The salsa [`File`] handle for an open document.
+    pub(crate) fn file(&self, uri: &Url) -> anyhow::Result<File> {
```
DavisVaughan: Feels a bit duplicative to me over something like `open_file(uri).file()`

crates/ark/src/lsp/state.rs
```diff
@@ -128,33 +137,32 @@ impl WorldState {
         let path = file.path(&self.db);
         self.open_files
             .get(path)
-            .map(|ark_file| ark_file.url.clone())
+            .map(|open_file| open_file.wire_url.clone())
             .unwrap_or_else(|| path.to_url())
```
DavisVaughan, line 127: I keep questioning whether we need `wire_url()` at all, but I guess the idea is that the `db` might hand you a `File` and you don't have any idea if that is an "open file" or not from the `File` alone

crates/ark/src/lsp/state.rs
```diff
@@ -128,33 +137,32 @@ impl WorldState {
...
+        self.open_files.insert(key, open_file);
     }
 }
 
 pub(crate) fn open_file_uris(state: &WorldState) -> Vec<Url> {
     state
         .open_files
         .values()
-        .map(|doc| doc.url.clone())
+        .map(|doc| doc.wire_url.clone())
```
DavisVaughan:
```suggestion
pub(crate) fn open_file_wire_urls(state: &WorldState) -> Vec<Url> {
    state
        .open_files
        .values()
        .map(|doc| doc.wire_url.clone())
```

Probably worth the preciseness since these URLs are so troublesome to get right

crates/ark/src/lsp/state_handlers.rs
```diff
@@ -269,8 +269,11 @@ pub(crate) fn did_change(
     }
 
     // Fold the edits into the new buffer text and push it into `oak`
-    let new_contents =
-        apply_content_changes(file.contents(&state.db), &params.content_changes, encoding);
+    let new_contents = apply_content_changes(
+        file.inner.source_text(&state.db).as_str(),
```
DavisVaughan: Something in me would feel better about `file.file().source_text()`

Maybe I don't love that `.inner` is directly exposed? I think that's probably it.

Kind of like I want `OpenFile` to be more transparent, reexposing `source_text()` so you can just directly do `file.source_text(db)` when `file` is an `OpenFile`

crates/ark/src/lsp/code_action.rs
```diff
@@ -25,15 +27,41 @@ pub(crate) struct CodeActions {
...
     capabilities: &Capabilities,
 ) -> lsp_types::CodeActionResponse {
     let mut actions = CodeActions::new();
 
-    roxygen_documentation(db, file, &mut actions, range, capabilities);
+    // Our code actions return literal `CodeAction`s, so bail if the client
+    // can't accept them.
+    if !capabilities.code_action_literal_support() {
+        return actions.into_response();
+    }
```
DavisVaughan: I think the `if !capabilities.code_action_literal_support() {` check belongs with `roxygen_documentation` because that check is specific to the type of code action returned. Not all of them return literals, but the roxygen one does. It's tied to that.

crates/ark/src/lsp/code_action.rs
```diff
@@ -25,15 +27,41 @@ pub(crate) struct CodeActions {
...
+                file.wire_url.clone(),
+                file.version,
+                vec![text_edit],
+                capabilities,
+            );
+            actions.add_action(code_action(
+                "Generate a roxygen template".to_string(),
+                lsp_types::CodeActionKind::EMPTY,
+                workspace_edit,
+            ));
```
DavisVaughan: It might be nice to have `roxygen::to_code_action(edit: RoxygenEdit, otherfields) -> Option<CodeAction>` to encapsulate this logic in `roxygen.rs` and clean this up a little bit

When we have lots of other code actions, I think it will look a bit messy to inline all of the logic that converts each individual code action to its corresponding LSP type here.

i.e. the previous code was purposefully written with the expectation that we'd accumulate several code actions here, and I wanted it to be easy to understand them all

crates/ark/src/lsp/code_action/roxygen.rs
```diff
@@ -90,29 +90,19 @@ pub(crate) fn roxygen_documentation(
...
         let parameter_name = child.child_by_field_name("name")?;
-        let parameter_name = parameter_name.node_to_string(file.contents(db)).ok()?;
+        let parameter_name = parameter_name
+            .node_to_string(file.source_text(db).as_str())
+            .ok()?;
         parameter_names.push(parameter_name);
     }
 
     let indent_size = position.column;
     let documentation = documentation_builder(parameter_names, indent_size);
```
DavisVaughan, line 129: I think we lost an important `We insert the documentation string ` comment

crates/ark/src/lsp/code_action/roxygen.rs
```diff
@@ -188,24 +181,20 @@ mod tests {
...
         let capabilities = Capabilities::default()
             .with_code_action_literal_support(true)
             .with_workspace_edit_document_changes(true);
 
         let (text, point, offset) = roxygen_point_and_offset_from_cursor(text);
-        let (db, file) = test_ark_file(&text);
+        let (db, file) = test_open_file(&text);
 
-        roxygen_documentation(
+        let mut actions = code_actions(
```
DavisVaughan: You can't remove this. This literally doesn't even test the `roxygen_documentation()` function at all anymore 😢

DavisVaughan: Same below

DavisVaughan (approved):
A few important notes, but seems nice

crates/ark/src/lsp/code_action.rs
lionel-: yep that would be the cleaner way

crates/ark/src/lsp/open_file.rs
lionel-: > Most importantly you'd only provide setters for version and config, making it clear that when you call open_file_mut(), those are the only 2 things you are allowed to change.

That's mildly better, doing that

crates/ark/src/lsp/state.rs
lionel-: ok, which leads to returning a `&` from `open_file()`, which is actually the right thing to do now that it's the long-term API. (We returned an owned value for `ArkFile` so they can cross `r_task()`)

crates/ark/src/lsp/state.rs
lionel-: yep

crates/ark/src/lsp/code_action/roxygen.rs
lionel-: I think this dispatches to the roxygen code action right?

crates/ark/src/lsp/state.rs
DavisVaughan: Yes strong agree!

PR #1277, opened 2026-06-18
Move main LSP loop off Tokio to an owned thread

lionel-:
Branched from https://github.com/posit-dev/ark/pull/1274

Fixes a source of deadlocks when the main loop blocks on a Salsa write until it gains exclusive access, but background tasks hold a DB handle and are scheduled by Tokio to run on the same worker thread as the main loop. We could wrap writes in `block_in_place()` instead, which instructs Tokio to move all pending tasks to a different worker, but moving the main loop off tokio altogether is really the safer move.

From the documenting comment:

```
        // Since the main loop owns the Salsa DB and writes to it, we run on its
        // own thread instead of a Tokio worker. Salsa writes are potentially
        // blocking until the writer gains exclusive access. If background tasks
        // holding clones of the DB are stuck on the same thread as the main
        // loop, the LSP deadlocks. This can be avoided by wrapping writes in
        // `block_in_place()` but the safer structure is to have it run on an OS
        // thread that we're in control of.
```

Fixes a bunch of issues observed in frontend tests: https://github.com/posit-dev/positron/actions/runs/27755066106?pr=14254

We still use tokio for the auxiliary loop and the blocking worker pool, for the time being.

### DIFF
--- a/crates/ark/src/lsp/main_loop.rs
+++ b/crates/ark/src/lsp/main_loop.rs
@@ -26,10 +26,11 @@ use oak_scan::ScanRequest;
 use oak_scan::ScanScheduler;
 use oak_semantic::library::Library;
 use stdext::result::ResultExt;
+use stdext::spawn;
 use tokio::runtime::Handle;
-use tokio::runtime::RuntimeFlavor;
 use tokio::sync::mpsc;
 use tokio::sync::mpsc::unbounded_channel as tokio_unbounded_channel;
+use tokio::sync::oneshot;
 use tokio::task::JoinHandle;
 use tower_lsp::jsonrpc;
 use tower_lsp::lsp_types;
@@ -159,6 +160,21 @@ pub(crate) struct GlobalState {
     events_rx: TokioUnboundedReceiver<Event>,
 }
 
+/// Owns the running LSP loops. Dropping it shuts them down.
+///
+/// - The auxiliary loop is a runtime task, aborted when `_aux_loop` drops.
+/// - The main loop runs on its own thread. Dropping `_main_shutdown_tx` closes
+///   the channel the loop selects on, so it breaks, drops the owned
+///   `GlobalState`, and the thread exits. We hold the thread's handle but don't
+///   join it on drop (it winds down on its own), so dropping never blocks the
+///   caller.
+#[derive(Debug)]
+pub(crate) struct LoopHandles {
+    _main_loop: std::thread::JoinHandle<()>,
+    _main_shutdown_tx: oneshot::Sender<()>,
+    _aux_loop: tokio::task::JoinSet<()>,
+}
+
 /// Non-cloneable, per-session state mutated only by exclusive handlers.
 /// Sits alongside [`WorldState`] (which is cloneable for snapshot
 /// handlers); state that can't be cloned lives here instead.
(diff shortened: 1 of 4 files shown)


crates/ark/src/lsp/main_loop.rs
```diff
@@ -265,37 +283,66 @@ impl GlobalState {
...
-        // global transmission channel.
-        let aux = AuxiliaryState::new(self.client.clone());
-        set.spawn(async move { aux.start().await });
-
-        // Spawn main loop
-        set.spawn(async move { self.main_loop().await });
+    /// The returned [`LoopHandles`] owns everything the loops need. Drop it to
+    /// shut the loops down and release the owned state.
+    pub(crate) fn start(self) -> LoopHandles {
+        let mut aux = tokio::task::JoinSet::<()>::new();
```
DavisVaughan, line 291: Do you really need a `tokio::task::JoinSet` anymore for 1 task?

crates/ark/src/lsp/main_loop.rs
```diff
@@ -265,37 +283,66 @@ impl GlobalState {
...
+                    }
+                }
             }
         }
     }
 
+    /// Pull the next event off the channel. Only the test pump uses this; the
+    /// running loop selects on the channel directly so it can also watch for
+    /// shutdown.
+    #[cfg(test)]
```
DavisVaughan, line 345: i would just remove this

crates/ark/src/lsp/main_loop.rs
```diff
@@ -325,10 +372,16 @@ impl GlobalState {
         // handlers that mutate those still refresh explicitly.
         let old_revision = salsa::plumbing::current_revision(&self.world.db);
 
+        lsp::log_info!("Processing next event");
```
DavisVaughan: This one doesn't feel that useful? We log at the beginning of every event type

crates/ark/src/lsp/main_loop.rs
```diff
@@ -325,10 +372,16 @@ impl GlobalState {
...
+
         match event {
             Event::Lsp(msg) => match msg {
                 LspMessage::Notification(notif) => {
                     lsp::log_info!("{notif:#?}");
+                    lsp::log_info!(
+                        "Entering notification handler with {n} outstanding Salsa db holds",
+                        n = self.world.db.outstanding_holds()
+                    );
```
DavisVaughan, line 382: Why do we want to know this on every iteration? Why do we want to know this at all? It feels like itll be expected for the db to have outstanding holds in background threads (i.e. its working on diagnostics when we get a did-change update)

crates/ark/src/lsp/main_loop.rs
```diff
@@ -930,7 +993,11 @@ static DIAGNOSTICS_QUEUE: LazyLock<tokio::sync::mpsc::UnboundedSender<RefreshDia
 /// newer one and publish last, leaving diagnostics computed from the older
 /// console scopes or config until the next refresh.
 async fn process_diagnostics_queue(mut rx: mpsc::UnboundedReceiver<RefreshDiagnosticsTask>) {
-    while let Some(task) = rx.recv().await {
+    loop {
+        let Some(task) = rx.recv().await else {
+            lsp::log_warn!("process_diagnostics_queue: channel closed, task exiting");
+            return;
+        };
```
DavisVaughan: I feel like the previous `while` look was pretty nice?

You'd just put the `lsp::log_warn!("process_diagnostics_queue: channel closed, task exiting");` part after the while loop exits, i think?

crates/ark/src/lsp/main_loop.rs
```diff
@@ -950,6 +1015,9 @@ fn process_diagnostics_batch(batch: Vec<RefreshDiagnosticsTask>) {
         .map(|task| (task.file.wire_url.clone(), task))
         .collect();
 
+    tracing::trace!("Processing {n} diagnostic tasks", n = batch.len());
```
DavisVaughan, line 1013: Since we don't say `Generating diagnostics` or `Finished diagnostics` via `tracing::trace!` (we only do that via `lsp::log_info!`), i feel like this isn't actually that useful on the kernel side (i know its not new from you, but maybe we can just remove it)

How do we decide where to log?

DavisVaughan (approved):

crates/ark/src/lsp/main_loop.rs
lionel-: This is not every iteration, this is only when we're about to write to the DB.

This is meant to help debugging hangs from user logs, we can remove this later once proven stable.

crates/ark/src/lsp/main_loop.rs
lionel-: Let's keep it as a token of the nicer loop structure

crates/ark/src/lsp/main_loop.rs
lionel-: That's a nice way of cancelling on abort. Dropping a task handle doesn't cancel it, so we'd need a custom Drop method to call the handle's `abort()` method.

crates/ark/src/lsp/main_loop.rs
lionel-: Good point

crates/ark/src/lsp/main_loop.rs
lionel-: I decided to keep it there because diagnostics code logs at kernel level, so we get a section marker.

In general, the current split is that eventing and dispatch belong to the LSP, handling belongs to the kernel. This is a remnant from `r_task()`-heavy handling and we have nicer logging facilities on the kernel side (context, profiling).
But: when the LSP becomes standalone, everything moves to the LSP.

crates/ark/src/lsp/main_loop.rs
DavisVaughan: > Dropping a task handle doesn't cancel it

oh i was not aware of that

PR #1281, opened 2026-06-22
Extract out `Writing Tests` to document nextest processes

DavisVaughan:
We don't actually say in AGENTS.md that nextest means we don't have to worry about test hygiene, so lets give this a try? Might also give us a place to document other test specific conventions

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -69,30 +69,14 @@ just test -p ark
 
 ### Writing Tests
 
-Because we use `nextest` to run tests, each individual test runs in its own separate process. This means that tests cannot interact with each other, and cross-test hygiene is not a concern.
+- Do not worry about cross-test hygiene. We use `nextest` to run tests, so each individual test runs in its own separate process and cannot affect other tests.
 
-Prefer simple assertion macros without custom error messages:
+- Prefer simple assertion macros without custom error messages:
+    - Use `assert_eq!(actual, expected);` instead of `assert_eq!(actual, expected, "custom message");`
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -67,6 +67,33 @@ just test <test_name>
 just test -p ark
 ```
 
+### Writing Tests
+
+Because we use `nextest` to run tests, each individual test runs in its own separate process. This means that tests cannot interact with each other, and cross-test hygiene is not a concern.
+
+Prefer simple assertion macros without custom error messages:
+
+```rust
```
lionel-: I'm a bit worried about overloading our agents file with test styling instructions. So I'd prefer making this section terser. It's normal for past instructions to get increasingly ignored as the context gets longer anyway, but we pay the cost of the base agents prompt on every exchange (much less when cached but it's easy for the cache to get flushed).

If styling is super important the best way is to extract in a skill for the _reviewing_ agent, because then the instructions will be fresh and the focus of the agent.

lionel- (approved):
Approving but I'd make a bit terser for reasons explained inline.

AGENTS.md
DavisVaughan: This was pulled almost directly from a separate section where it was previously just a bullet point. The code block seemed easier to understand?

Can you please suggest an alternative that you are happy with?

lionel-: I prefer the old version that is terser. This new direction goes toward a style-guide layout that is heavier and might encourage verbosity, which doesn't fit an agents file in my mind.

PR #1283, opened 2026-06-23
Simplify ownership of the list of comms

lionel-:
Progress towards https://github.com/posit-dev/ark/issues/689 and https://github.com/posit-dev/ark/issues/691
Follow-up to https://github.com/posit-dev/ark/pull/1161 (see `console_comm.rs` changes)

See https://github.com/posit-dev/ark/pull/1075 for context: we're moving comms out of their own threads to run on the R thread, to simplify the concurrency structure of Ark and remove the risk of concurrent R evaluations which are UB (although these are now also tackled under a different angle, see https://github.com/posit-dev/ark/pull/1222).

In #1161 I made progress towards phasing out unsafe mutation on Console state (see https://github.com/posit-dev/ark/issues/1145) by using interior mutability in Console methods. To manage comm state in particular, I went for a take/move approach that removes a comm from the map of comms owned by the Console, and passes the moved comm by value to its handlers. This allowed data explorer comm handlers to manipulate the comm map and insert new explorer comms (when an inline explorer gets promoted to a full explorer).

Moving the comm was awkward with the UI comm though, which needed to be pinned down because its handlers might cause Console events that themselves require the UI comm (e.g. to send a message to the frontend). For this reason the UI comm used a different dispatch strategy with a shared borrow, that required ad hoc infrastructure.

The Help comm is also accessible from Console methods and is going to need a similar setup, so I thought I'd go ahead and try to simplify this:

- All comms now live in the map, including the UI comm. But they are now wrapped in an `Rc`.

- The dispatcher now holds an `Rc` clone of the comm rather than an exclusively owned value. The comm stays in the Console-owned map during dispatch, which is how they remain accessible for reentrancy while a handler is running. Thanks to the Rc wrapping and cloning, the comm map remains fully owned by the Console during dispatch and can be manipulated while a handler is running, allowing data explorers to clone/promote themselves.

  Note that the handler's `&mut` comes from interior mutability of the comm handler inside the `ConsoleComm`. So the one invariant we need to guarantee is that we never dispatch a comm's handler reentrantly.

- The UI comm still gets some ad hoc handling because of its "pinned" nature (there can only be one UI comm at any time, so opening a new one drops the old comm if any).

### DIFF
--- a/crates/ark/src/console/console_comm.rs
+++ b/crates/ark/src/console/console_comm.rs
@@ -4,6 +4,8 @@
 // Copyright (C) 2026 Posit Software, PBC. All rights reserved.
 //
 
+use std::rc::Rc;
+
 use amalthea::comm::comm_channel::CommMsg;
 use amalthea::comm::event::CommEvent;
 use amalthea::socket::comm::CommInitiator;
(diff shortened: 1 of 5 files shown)


crates/ark/src/console/console_comm.rs
```diff
@@ -125,98 +116,67 @@ impl Console {
...
-            self.with_comm_mut(&id, |comm| {
-                comm.handler.get_mut().handle_environment(event, &comm.ctx);
-            });
+        // Snapshot the `Rc`s so the `comms` borrow is dropped before we run any
+        // handler (a handler may reenter `self.comms`).
+        let comms: Vec<Rc<ConsoleComm>> = self.comms.borrow().values().map(Rc::clone).collect();
+        for comm in comms {
+            comm.handler
+                .borrow_mut()
+                .handle_environment(event, &comm.ctx);
```
DavisVaughan, line 143: So, to be clear, a handler may reenter `self.comms` but it CANNOT call `comm.handler.borrow_mut()` reentrantly on itself again, right? Otherwise that's a panic at debug time, and we log at release time

Would be nice if we can avoid reentrant issues entirely, but I get it

crates/ark/src/console/console_comm.rs
```diff
@@ -125,98 +116,67 @@ impl Console {
...
         };
-        f(&mut comm);
-        self.comms.borrow_mut().insert(comm.comm_id.clone(), comm);
-    }
-
-    fn take_comm(&self, comm_id: &str) -> Option<ConsoleComm> {
-        self.comms.borrow_mut().remove(comm_id)
+        log::info!("Replacing an existing UI comm.");
+        old.handler.borrow_mut().handle_close(&old.ctx);
     }
```
DavisVaughan: These don't feel quite right to me

In particular

```
log::info!("Replacing an existing UI comm.");
```

that doesn't really align with what happens. you actually remove it here and set it to `None` but don't replace it here. you replace it in the caller.

i kind of felt like we don't actually need this helper. like it feels like in `comm_open_frontend` we should do `remove_comm(ui_comm_id)` (or maybe `comm_handle_close`, so it does `handle_close`) followed up with setting the `ui_comm_id` to the new one.

DavisVaughan (approved):

crates/ark/src/console/console_comm.rs
lionel-: Good points

PR #1284, opened 2026-06-23
Run the Help comm on the R thread

lionel-:
Branched from #1283 
Progress towards https://github.com/posit-dev/ark/issues/689 and https://github.com/posit-dev/ark/issues/691

See https://github.com/posit-dev/ark/pull/1075 for context: we're moving comms out of their own threads to run on the R thread, to simplify the concurrency structure of Ark and remove the risk of concurrent R evaluations which are UB (although these are now also tackled under a different angle, see https://github.com/posit-dev/ark/pull/1222).

The help comm no longer has its own thread and event loop. Instead it implements `CommHandler` and its handlers get run on the R thread.

- Like the UI comm, it's a pinned comm that is accessible via methods on Console, so it uses the patterns established in #1283.

- `browseURL()` emits a Help event to the frontend. Now that the message is emitted from the R thread, instead of the old comm thread, the message is guaranteed to be emitted within the busy/idle window of the execute-request of the `browseURL()` call. This sort of simplification of message ordering is exactly what enables accurate and stable protocol-level tests.

- New kernel-level integration test for the Help comm (progress towards https://github.com/posit-dev/ark/issues/1074).


crates/ark/tests/integration/help.rs
```diff
@@ -152,3 +150,52 @@ fn test_custom_help_handlers() {
...
+    assert_eq!(
+        msg.data.get("method").and_then(|v| v.as_str()),
+        Some("show_help")
+    );
+    assert_eq!(msg.data["params"]["kind"], "url");
+    let content = msg.data["params"]["content"].as_str().unwrap();
+    assert!(content.starts_with("http://127.0.0.1:"));
+    assert!(content.contains("plot"));
+
+    frontend.recv_iopub_idle();
```
DavisVaughan, line 199: yay for busy/idle ordering stability!

crates/ark/src/console.rs
```diff
@@ -241,19 +241,20 @@ pub struct Console {
     tasks_idle_any_rx: Receiver<QueuedRTask>,
     pending_futures: HashMap<Uuid, (BoxFuture<'static, ()>, RTaskStartInfo, Option<String>)>,
 
-    /// Comm ID of the UI comm, if connected. The comm itself lives in `comms`
-    /// like any other.
+    /// Comm IDs of pinned comms, if connected.
```
DavisVaughan: is "pinned" the right word? they just seem kind of "special" to me

crates/ark/src/console/console_comm.rs
```diff
@@ -148,37 +154,50 @@ impl Console {
...
+        log::info!("Replacing an existing {role} comm.");
         old.handler.borrow_mut().handle_close(&old.ctx);
     }
 
+    fn close_ui_comm(&self) {
+        self.close_comm(&self.ui_comm_id, "UI");
+    }
+    fn close_help_comm(&self) {
+        self.close_comm(&self.help_comm_id, "help");
+    }
```
DavisVaughan: See https://github.com/posit-dev/ark/pull/1283#discussion_r3521055482 where I wonder if we really need these at all vs just `comm_handle_close`

crates/ark/src/console/console_comm.rs
```diff
@@ -148,37 +154,50 @@ impl Console {
...
         let comm = self.comms.borrow_mut().remove(comm_id)?;
 
         let mut ui_comm_id = self.ui_comm_id.borrow_mut();
         if ui_comm_id.as_deref() == Some(comm_id) {
             *ui_comm_id = None;
         }
+        let mut help_comm_id = self.help_comm_id.borrow_mut();
+        if help_comm_id.as_deref() == Some(comm_id) {
+            *help_comm_id = None;
+        }
```
DavisVaughan: Random thought that maybe you should only `borrow_mut()` if you've already confirmed that the `comm_id` is the same by using a more lax `borrow()`

crates/ark/src/console.rs
```diff
@@ -241,19 +241,20 @@ pub struct Console {
...
     last_error: Option<Exception>,
 
-    /// Channel to communicate with the Help thread
-    help_event_tx: Option<Sender<HelpEvent>>,
-    /// R help port
+    /// R help server port, used to recognize help URLs.
     help_port: Option<u16>,
 
+    /// Our help proxy server port, used to rewrite help URLs sent to the frontend.
+    help_proxy_port: Option<u16>,
```
DavisVaughan:
```suggestion
    help_r_port: Option<u16>,

    /// Our help proxy server port, used to rewrite help URLs sent to the frontend.
    help_proxy_port: Option<u16>,
```

In all of the functions we have `r_port` and `proxy_port`, so to me it makes sense that we store them here as the same name, just prefixed with `help_`

crates/ark/src/console/console_integration.rs
```diff
@@ -43,21 +43,24 @@ impl Console {
...
 impl Console {
-    pub(crate) fn set_help_fields(&mut self, help_event_tx: Sender<HelpEvent>, help_port: u16) {
-        self.help_event_tx = Some(help_event_tx);
-        self.help_port = Some(help_port);
+    pub(crate) fn set_help_ports(&mut self, r_port: u16, proxy_port: u16) {
+        self.help_port = Some(r_port);
+        self.help_proxy_port = Some(proxy_port);
     }
 
+    /// Translate a help event and send it on the help comm's outgoing channel.
```
DavisVaughan: Maybe dont need these docs since `RHelp::handle_event` has the same docs and does the real translation/sending?

crates/ark/src/console/console_integration.rs
```diff
@@ -43,21 +43,24 @@ impl Console {
...
     pub(crate) fn send_help_event(&self, event: HelpEvent) -> anyhow::Result<()> {
-        let Some(ref tx) = self.help_event_tx else {
-            return Err(anyhow!("No help channel available to handle help event. Is the help comm open? Event {event:?}."));
+        let (Some(r_port), Some(proxy_port)) = (self.help_port, self.help_proxy_port) else {
```
DavisVaughan: Maybe we should store them on `Console` as an all or nothing pair of `help_ports: Option<(u16, u16)>`?

DavisVaughan: Kind of like storing an all or nothing `help_state: Option<HelpState>` but without the struct

crates/ark/src/help/r_help.rs
```diff
@@ -11,77 +11,35 @@ use amalthea::comm::help_comm::HelpBackendRequest;
...
-        });
+pub const HELP_COMM_NAME: &str = "positron.help";
 
-        // Return the channel for sending help events to the help thread
-        Ok(help_event_tx)
-    }
+/// The R Help handler (together with the help proxy) provides the server side
+/// of Positron's Help panel.
+#[derive(Debug)]
+pub struct RHelp;
```
DavisVaughan, line 49: Something something "purposefully stateless"?

crates/ark/src/help/r_help.rs
```diff
@@ -310,6 +205,39 @@ impl RHelp {
...
+            },
+        };
+
+        Console::get_mut().set_help_ports(r_port, proxy_port);
+    }
+
+    fn handle_msg(&mut self, msg: CommMsg, ctx: &CommHandlerContext) {
+        let this = &*self;
+        handle_rpc_request(&ctx.outgoing_tx, HELP_COMM_NAME, msg, |req| {
+            this.handle_rpc(req)
```
DavisVaughan: ah, gotta downcast `&mut self` to `&self` to be able to pass it in the closure?

crates/ark/src/help/r_help.rs
```diff
@@ -235,38 +131,37 @@ impl RHelp {
...
+                }
             },
-        }?;
+        };
 
         Ok(found)
     }
 
-    // Must be called in a `r_task` context.
+    // Must be called on the R thread.
```
DavisVaughan: You could say this for a lot of methods in here. Maybe this moves to `handle_open` and `handle_msg`?

DavisVaughan (approved):

crates/ark/src/console.rs
lionel-: The idea of "pin" here is that you can't have more than one, so if a new one comes in you replace it in the slot. Agreed it's not the best term. I replaced "pinned" by "singleton".

crates/ark/src/console/console_integration.rs
lionel-: Now `Option<HelpPorts>`

crates/ark/src/help/r_help.rs
lionel-: hmm nope I think this is a leftover from a previous iteration

PR #1285, opened 2026-06-23
Shut down help proxy when help comm drops

lionel-:
Branched from #1284 

While working on that PR I noticed that we never shut down the proxy server when the Help comm is tore down. If we start a new one, we'd pile up a new proxy.

This is theoretical because Help is one of the comms that the frontend reuses on refresh/reload (unlike the UI comm). But it still seems nice to handle the shutdown path properly. Covered by a new test.


thomasp85 (approved):
This LGTM - it feels like a pretty clean fix to a non-issue (but issue none-the-less)

PR #1287, opened 2026-06-23
(Sources-2) `oak_db::workspace_dependencies()`

DavisVaughan:
Branched from #1286

Two new things here:

- `File::used_packages()` is a tracked query that returns the `Name`s of all packages used by this file
    - `library()` and `require()` are included (reusing the existing `attached_packages()`)
    - `::` and `:::` are included (using the new `namespaced_accesses()` from #1286)
    - The result is _sorted and unique_, maximizing the chance of backdating after small file edits (a reorder of `library()` or `::` calls should not affect this query result, unlike `attached_packages()`)

- `oak_db::workspace_dependencies()` is a tracked query that returns a `Vec<Package>` of every installed package used by this workspace.
    - It collects from `File::used_packages()` on each `workspace_file()`
    - It collects from `DESCRIPTION` if you have a workspace package open. In particular, it collects `Imports` and `Depends`. I have omitted `Suggests` because typically they are referenced by `::` already if they are used, so they probably aren't worth it to also include here?

Two notes about `oak_db::workspace_dependencies()`:
- Any _not installed_ package returned by `File::used_packages()` is thrown out. The user can't actually use this, and we won't have a package version to look up anyways.
- If you have a workspace package open, it is also removed from the list (the workspace package itself isn't really a workspace dependency). We assume that you already have package sources for that in your local `R/` folder.

`oak_db::workspace_dependencies()` will be used in a future PR to detect the set of changed packages after each revision change.


crates/oak_db/src/file.rs
```diff
@@ -207,6 +208,47 @@ impl File {
...
 
+    /// Package names from `::` / `:::` accesses in this file
+    ///
+    /// Packages are sorted by name and are unique. This maximizes the ability to
+    /// backdate after small file edits.
+    #[salsa::tracked(returns(ref))]
+    fn namespace_accessed_packages(self, db: &dyn Db) -> Vec<Name<'_>> {
+        // Likely that there are many `::` accesses for the same package within a single
+        // file, so it's useful to build as a BTreeSet that automatically handles sorting
+        // and deduplicating for us, rather than building a long `Vec` of duplicated
```
lionel-: I would expect cache-friendly vector dedup to be faster (as well as simpler) than insertion into a complex data structure of indirections?

crates/oak_db/src/file.rs
```diff
@@ -207,6 +208,47 @@ impl File {
             .collect()
     }
 
+    /// Package names from `::` / `:::` accesses in this file
+    ///
+    /// Packages are sorted by name and are unique. This maximizes the ability to
+    /// backdate after small file edits.
+    #[salsa::tracked(returns(ref))]
+    fn namespace_accessed_packages(self, db: &dyn Db) -> Vec<Name<'_>> {
```
lionel-, line 215: The past participles (namespace_accessed_packages, namespaced_accesses) are getting out of hands 😅 

The former is consistent with `used_packages()` and `attached_packages()` though, so simply renaming the latter as suggested in the other PR would fix it for me.

crates/oak_db/src/workspace.rs
```diff
@@ -0,0 +1,65 @@
...
+    // We take `Depends` and `Imports`, but not `Suggests`, as `Suggests` will
+    // instead be referenced by `::` elsewhere as required
+    for root in db.workspace_roots().roots(db).iter() {
+        for &package in root.packages(db) {
+            if let Some(description) = package.description(db) {
+                names.extend(description.depends.iter().map(String::as_str));
+                names.extend(description.imports.iter().map(String::as_str));
+            }
+        }
+    }
```
lionel-: Should these loops be extracted as individual queries so they can independently backdate? Maybe `workspace_package_dependencies()` and `workspace_script_dependencies()`?

I'm also wondering if we should extract a new `Workspace` singleton to collect these queries.

crates/oak_db/src/workspace.rs
```diff
@@ -0,0 +1,65 @@
...
+
+    // Now look up the actual `Package` by name, throwing out ones that aren't actually
+    // installed, and packages that are actively being worked on in the workspace already
+    let mut packages = Vec::with_capacity(names.len());
+
+    for name in names {
+        let Some(package) = db.package_by_name(name) else {
+            continue;
+        };
+        let Some(root) = db.root_by_package(package) else {
```
lionel-, line 91: We probably should make these two Salsa queries.

I think we started with `file_by_path()`, which can't be a query because `FilePath` is not a Salsa key, but that reason doesn't apply to package-by-name (if we make it take a `Name` instead of `&str`) and root-by-package.

lionel- (approved):

crates/oak_db/src/file.rs
DavisVaughan: Resolved by https://github.com/posit-dev/ark/pull/1286/changes/15fb5baf619cfad5d279b3c2db25f1742c655318

crates/oak_db/src/workspace.rs
DavisVaughan: Opened https://github.com/posit-dev/ark/issues/1293

PR #1288, opened 2026-06-23
(Sources-3) `oak_scan::set_package_sources()`

DavisVaughan:
Branched from https://github.com/posit-dev/ark/pull/1287

This PR teaches `oak_scan` how to update an installed `Package`'s `files` and `scripts` from an _external on disk directory_.

The idea is that some external source provider will return a flat `directory` of R files that correspond to the source files for a `Package`. The new `oak_scan::set_package_sources()` helper iterates over that `directory`, collecting the R files (but not reading them!), sorting and splitting them in accordance with `collate`, and stashes them in `files` and `scripts` of the `Package`.

> [!CAUTION]
> This PR has an important side effect. Since `package.files()` now live in a totally separate directory outside any `LiveRoot`, the `File::package` backpointer is now a load bearing feature.
>
> Consider `File::root()`. It switches on `self.package()`. If there is no package backpointer, it would go through either:
> - `root_by_file()`, which looks inside `LiveRoot::Workspace` and `LiveRoot::Library` root paths for the file's `path`, neither of which would hold it
> - `root_by_path()`, which is for orphaned files, and also uses paths contained within `WorkspaceRoots`
>
> So we need to instead detect `self.package()` and go through `root_by_package()`, which doesn't do a `path` check, it just detects that the `Package` backpointer is contained within a `LiveRoot` (by equality, not by `path` containment). We already do all this, it is just important to point it out now.
>
> `File::imports()` is similar

### DIFF
--- a/crates/oak_scan/src/packages.rs
+++ b/crates/oak_scan/src/packages.rs
@@ -210,6 +210,46 @@ fn read_workspace_package(package_dir: &Path) -> Option<PackageEntry> {
     Some(package)
 }
 
+/// Walk `directory`, collecting `.R` files
+///
+/// `.R` files are split into either `files` or `scripts` based on `collation`,
+/// matching [read_workspace_package()].
+///
+/// The `directory` is not walked recursively. We expect that this is a flat directory of
+/// `.R` files.
+pub(crate) fn read_package_sources(
+    directory: &Path,
+    collation: Option<&[String]>,
+) -> (Vec<FileEntry>, Vec<FileEntry>) {
+    let Ok(entries) = fs::read_dir(directory) else {
+        log::warn!("Cannot read sources directory: {}", directory.display());
+        return (Vec::new(), Vec::new());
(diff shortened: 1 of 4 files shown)


crates/oak_scan/src/packages.rs
```diff
@@ -210,6 +210,46 @@ fn read_workspace_package(package_dir: &Path) -> Option<PackageEntry> {
     Some(package)
 }
 
+/// Walk `directory`, collecting `.R` files
+///
+/// `.R` files are split into either `files` or `scripts` based on `collation`,
+/// matching [read_workspace_package()].
+///
+/// The `directory` is not walked recursively. We expect that this is a flat directory of
+/// `.R` files.
```
lionel-, line 219: Not important to support right now or ever, but in principle a package might install .R files in `inst` and source them at run time?

lionel- (approved):

PR #1289, opened 2026-06-23
(Sources-4) Implement `Event::SourceCompleted` for installed package sources

DavisVaughan:
Branched from https://github.com/posit-dev/ark/pull/1288

This PR sets up the `main_loop.rs` handling for package sources.

Here is the set up:

- The database changes via a `revision` bump
- We `dispatch()` a series of `SourceRequest`s via `lsp::spawn_blocking()`. One `SourceRequest` is made per package currently used by the workspace (removing any packages we've already requested sources for).
- On the tokio task, an implementer of `trait SourceProvider` is called to `provide()` sources for a package. On success, it sends the `directory` of the sources back to the main loop via a new `Event::SourceCompleted` event.
    - Note that this all happens on a task, not blocking the main loop while we are creating these sources.
- The main loop picks this `Event` up and calls `db.set_package_sources(package, &directory)` for that package, which actually loads the files from the directory.
- The next `revision` of the `db` can now use these `package.files()` in LSP requests

I purposefully have not hooked up a live production level implementer of `trait SourceProvider`. Instead we have a `TestProvider` that is used to verify that all of this infrastructure is working, but for production this is all set to `None` and no `SourceRequest`s are ever sent out, so it is safe to merge.

If we agree on the shape of all of this, then in the next PRs I can hook up the real source provider.

### DIFF
--- a/crates/ark/src/lsp/sources.rs
+++ b/crates/ark/src/lsp/sources.rs
@@ -11,12 +11,12 @@ use stdext::result::ResultExt;
 use crate::lsp::main_loop::Event;
 use crate::lsp::main_loop::TokioUnboundedSender;
 
-/// Provide R source files for an installed package
+/// Handle source requests to provide R source files for an installed package
 ///
 /// Implementations live outside of oak, oak is only in charge of ingesting the
 /// returned directory.
-pub(crate) trait SourceProvider: Send + Sync {
-    fn provide(&self, request: &SourceRequest) -> SourceResponse;
+pub(crate) trait SourceHandler: Send + Sync {
+    fn handle(&self, request: &SourceRequest) -> SourceResponse;
 }
 
 #[derive(Debug, Clone)]
@@ -26,7 +26,7 @@ pub(crate) struct SourceRequest {
     library_path: PathBuf,
 }
 
-// TODO!: Remove when we have a production `SourceProvider`
+// TODO!: Remove when we have a production `SourceHandler`
 #[cfg_attr(not(test), expect(dead_code))]
 #[derive(Debug)]
 pub(crate) enum SourceResponse {
@@ -49,21 +49,21 @@ enum SourceState {
 }
 
 pub(crate) struct SourceManager {
-    // TODO!: Remove the `Option<>` when we implement a production `SourceProvider`
-    provider: Option<Arc<dyn SourceProvider>>,
+    // TODO!: Remove the `Option<>` when we implement a production `SourceHandler`
+    handler: Option<Arc<dyn SourceHandler>>,
     state: HashMap<Package, SourceState>,
 }
 
 impl SourceManager {
-    pub(crate) fn new(provider: Option<Arc<dyn SourceProvider>>) -> Self {
+    pub(crate) fn new(handler: Option<Arc<dyn SourceHandler>>) -> Self {
         Self {
-            provider,
+            handler,
             state: HashMap::new(),
         }
     }
 
     pub(crate) fn dispatch(&mut self, db: &dyn Db, events_tx: &TokioUnboundedSender<Event>) {
-        let Some(provider) = &self.provider else {
+        let Some(handler) = &self.handler else {
             return;
         };
 
@@ -82,14 +82,14 @@ impl SourceManager {
                 continue;
             };
 
-            let provider = Arc::clone(provider);
+            let handler = Arc::clone(handler);
             let tx = events_tx.clone();
 
             // Set to `Pending` after all possible early exits
             self.state.insert(package, SourceState::Pending);
 
             crate::lsp::spawn_blocking(move || {
-                let response = provider.provide(&request);
+                let response = handler.handle(&request);
 
                 tx.send(Event::SourceCompleted(SourceCompleted {
                     package,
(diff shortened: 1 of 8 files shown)


crates/ark/src/lsp/sources.rs
```diff
@@ -0,0 +1,190 @@
...
+
+use crate::lsp::main_loop::Event;
+use crate::lsp::main_loop::TokioUnboundedSender;
+
+/// Provide R source files for an installed package
+///
+/// Implementations live outside of oak, oak is only in charge of ingesting the
+/// returned directory.
+pub(crate) trait SourceProvider: Send + Sync {
+    fn provide(&self, request: &SourceRequest) -> SourceResponse;
```
lionel-: I think this would speak more to me if this was called `SourceHandler` and `handle()`.

It goes well with the request argument and response return value, and fits with all the other places in Ark and Oak where "handler" indicates a dispatch/indirection abstraction (e.g. `CommHandler`, `ShellHandler` to take examples from Ark).

crates/ark/src/lsp/sources.rs
```diff
@@ -0,0 +1,190 @@
...
+}
+
+enum SourceState {
+    Pending,
+    Success,
+    Failed,
+    Retry,
+}
+
+pub(crate) struct SourceManager {
```
lionel-: I'm a bit allergic to "manager" naming because of how java-y it feels. How about naming it after the state this wraps?

Maybe `SourceStates`, since it holds a collection of `SourceState`?

Or `SourceTracker` if we want a role-oriented name would feel better to me, but `SourceStates` seems the most accurate option to me, because thinking in terms of data and procedures makes it easier for me to understand a system than thinking in terms of roles.

crates/ark/src/lsp/sources.rs
```diff
@@ -0,0 +1,190 @@
...
+
+impl SourceManager {
+    pub(crate) fn new(provider: Option<Arc<dyn SourceProvider>>) -> Self {
+        Self {
+            provider,
+            state: HashMap::new(),
+        }
+    }
+
+    pub(crate) fn dispatch(&mut self, db: &dyn Db, events_tx: &TokioUnboundedSender<Event>) {
```
lionel-: How about `refresh()`? Would read well in the main loop too, next to diagnostics refresh.

crates/ark/src/lsp/sources.rs
```diff
@@ -0,0 +1,190 @@
...
+pub(crate) struct SourceCompleted {
+    pub(crate) package: Package,
+    pub(crate) response: SourceResponse,
+}
+
+enum SourceState {
+    Pending,
+    Success,
+    Failed,
+    Retry,
```
lionel-: Not sure what this will be hooked to, but with the current setup:

- If the main loop is quiet, a retry may be stuck for a long time, which may delay diagnostics that could be useful even if the user isn't typing or interacting with the LSP. This suggests we should associate the retry with a timer instead?
 
- If the main loop is hot, this retries every time. I'm assuming retry is I/O-related, which suggests we should back off?

So in other words, retrying should probably be tied to a timer with backoff rather than to Salsa's revision counter (which is still the right place for refreshing newly known packages).

lionel- (approved):
Looks great!

Just a comment about what to do for retries, but it's not blocking.

crates/ark/src/lsp/sources.rs
DavisVaughan: I actually went with `SourceScheduler`, which aligns with `ScanScheduler`

```
    /// Coordinator for asynchronous workspace scans. Mutated only from
    /// main-loop handlers. Must be out of [`WorldState`] because the scheduler
    /// is not clonable.
    pub(crate) oak_scheduler: ScanScheduler,

    /// Scheduler of [crate::lsp::sources::SourceRequest]s. Scheduling and source
    /// consumption all happen from the main loop.
    pub(crate) source_scheduler: SourceScheduler,
```

And feels kind of right because it has a `.schedule()` method to kick of `SourceRequest`s and can `.finish()` them on the backend.

crates/ark/src/lsp/sources.rs
DavisVaughan: See https://github.com/posit-dev/ark/pull/1289#discussion_r3476930342

crates/ark/src/lsp/sources.rs
DavisVaughan: Actually, now that I'm thinking about this again, I think we can make it simpler for now until we get more experience with it (and have the full shape of it in front of us)

Since the downloading of sources happens off the main thread, I think we can afford to handle any network retries over there. So we can probably remove `Retry`.

And really `Pending`, `Success`, and `Failed` all say the same thing "we've seen this package before, please don't ask for it again".

So this all reduces down to `seen: HashSet<Package>`

---

This would mean that, for example, if you turn off your wifi and open Positron and we try and download sources for dplyr then it will get marked as `seen` but we won't be able to download sources for it. If you turn on your wifi within the same LSP session, then we won't ever ask for sources again. But this situation might end up being really rare and weird, especially since most of the time we will already have cached sources for dplyr from a previous session where a download was successful, so even if you're offline, we've already got them locally.

DavisVaughan: Actually it does seem like we need `Pending` to ensure that tests can wait for in flight requests to finish, but the idea is still roughly the same

crates/ark/src/lsp/sources.rs
lionel-: oh yep that's nice to have the symmetry here

PR #1298, opened 2026-06-29
(1) Implement generic `oak_cache`

DavisVaughan:
Part of #1234

Part of a series of PRs to get the sources mechanism integrated with oak. Most of the PRs revolve around restructuring the cache in a way that I think makes a lot more sense (particularly if we want it to be usable elsewhere).

We end up with:
- `oak_cache` - Generic `Cache` mechanism
- `oak_srcref` - Home to `SrcrefCache`, requires R, extracts the `srcref` attribute
- `oak_source` - Home to `SourceCache`, which knows how to download CRAN R package tarballs and R version tarballs

And `oak_sources` gets removed at the end.

When I worked on the caching stuff earlier, I struggled to reconcile with the fact that different caching sources return different things. We have 3 things we can return:
- srcrefs, which is just the reconstructed `R/` folder
- CRAN R package tarballs, which is a single R package in its entirety
- CRAN R version tarballs, which is a single R version in its entirety

Rather than trying to squish these together into a single cache mechanism, which required us to drop useful things from the CRAN downloads to align with the more minimal result we get from the srcref extractor, we now keep them separate in separate caches, so we end up with:

- `oak_srcref::SrcrefCache` feeding to `.cache/oak/srcref/`
- `oak_source::SourceCache::cran` feeding to `.cache/oak/source/cran/`
- `oak_source::SourceCache::r` feeding to `.cache/oak/source/r/`

Each cache folder now has a consistent structure that also maximizes information retention. i.e. in the CRAN download we keep the entire R package source, including things like `man/`, `src/`, and any `inst/` folder you might have included. I imagine this could be useful in the future (navigating to C functions?).

I'm also hoping that `oak_source` can be useful on its own without requiring the heavier R dependency that `oak_srcref` has.

The end result of this series of 5 PRs is a production level `SourceHandler` that is integrated in the LSP that pulls sources using `oak_srcref` and `oak_source` when you are in a debug build. Release builds still don't have it turned on yet. And it isn't being utilized in any LSP features yet. I'll turn them on one by one after merging this series.

### DIFF
--- a/crates/oak_cache/Cargo.toml
+++ b/crates/oak_cache/Cargo.toml
@@ -1,0 +1,19 @@
+[package]
+name = "oak_cache"
+version = "0.1.0"
+authors.workspace = true
+edition.workspace = true
+rust-version.workspace = true
+license.workspace = true
+
+[dependencies]
+anyhow.workspace = true
+etcetera.workspace = true
+log.workspace = true
+oak_fs.workspace = true
+
+[dev-dependencies]
+tempfile.workspace = true
+
+[lints]
+workspace = true
--- a/crates/oak_cache/src/lib.rs
+++ b/crates/oak_cache/src/lib.rs
@@ -1,0 +1,436 @@
+//! A generic, content-agnostic on disk directory cache
+//!
+//! A [`Cache`] is rooted at a single subfolder of the shared cache directory and
+//! holds a fixed number of entries (an LRU `capacity`). Each entry is a directory
+//! keyed by a caller-supplied string and filled by a caller-supplied `populate`
+//! closure. The cache knows nothing about what lives inside an entry. The lock,
+//! completion-sentinel, last-access, and LRU-eviction machinery all live here so
+//! callers don't reimplement it.
+
+use std::path::Path;
+use std::path::PathBuf;
+use std::time::SystemTime;
+
+use oak_fs::file_lock::FileLock;
+use oak_fs::file_lock::Filesystem;
+
+/// Name of the root lock file and the per-key lock file.
+const LOCK_FILENAME: &str = ".lock";
+
+/// Name of the completion sentinel, written last in each cache entry.
+///
+/// An entry without it is a crashed partial and is removed by [`Cache::clean`].
+///
+/// Its mtime doubles as the entry's last-access time. [`Cache::get`] bumps it on every
+/// retrieval (this is an atomic operation so we allow it even though we'll only hold a
+/// shared lock on this key's folder) and [`Cache::clean`] evicts the entries with the
+/// oldest mtime first.
+const COMPLETE_FILENAME: &str = ".complete";
+
+/// A generic on disk directory cache
+///
+/// # On disk layout
+///
+/// The cache is rooted at one subfolder of `<cache_dir>/oak/`. Each entry is a
+/// directory keyed by the caller:
+///
+/// ```text
+/// <cache_dir>/oak/<root>/
+///     .lock
+///     <key>/
+///         .lock
+///         .complete
+///         <content>
+/// ```
+///
+/// `.complete` is written last and is the completion sentinel.
+///
+/// # Locking
+///
+/// The root `.lock` can be held shared or exclusive:
+///
+/// - **Shared** is held for the lifetime of this `Cache`. While held, entries can be read
+///   and new entries appended, but nothing can be deleted. Multiple sessions can hold it
+///   simultaneously. It keeps any [`PathBuf`] handed out by [`Cache::get`] valid for the
+///   life of the cache.
+///
+/// - **Exclusive** is only attempted once, at [`Cache::new`], to run [`Cache::clean`].
+///   It is skipped if another session already holds the shared lock, in which case we
+///   just try again next time.
+///
+/// Each entry also has its own `<key>/.lock`, taken exclusively in [`Cache::insert`] so
+/// two sessions never populate the same key at once.
+///
+/// This mirrors cargo's model, which faces the same multi-process cache challenge.
+#[derive(Debug)]
+pub struct Cache {
+    /// On disk root of this cache's subfolder.
+    root: Filesystem,
+
+    /// Shared lock on the root `.lock`, held for the life of this `Cache`.
+    ///
+    /// Blocks any other process from taking the root exclusive lock (the only thing
+    /// that deletes entries), so any [`PathBuf`] we hand out stays valid as long as
+    /// this `Cache` lives.
+    root_lock: FileLock,
+}
+
+impl Cache {
+    /// Open `<cache_dir>/oak/<root>/`, creating it if needed.
+    ///
+    /// Runs a best-effort [`Cache::clean`] under the exclusive root lock (skipped if
+    /// another session holds the shared lock), then holds the shared root lock for the
+    /// life of the returned `Cache` so handed-out paths stay valid.
+    pub fn new(root: &str, capacity: usize) -> anyhow::Result<Self> {
+        Self::new_in(cache_dir()?.join(root), capacity)
+    }
+
+    /// Like [`Cache::new`], but rooted at an explicit `root` rather than a subfolder of
+    /// the shared cache directory. Only useful for testing against a temp directory.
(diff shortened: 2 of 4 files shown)


crates/oak_cache/src/lib.rs
```diff
@@ -0,0 +1,436 @@
...
+
+/// Name of the completion sentinel, written last in each cache entry.
+///
+/// An entry without it is a crashed partial and is removed by [`Cache::clean`].
+///
+/// Its mtime doubles as the entry's last-access time. [`Cache::get`] bumps it on every
+/// retrieval (this is an atomic operation so we allow it even though we'll only hold a
+/// shared lock on this key's folder) and [`Cache::clean`] evicts the entries with the
+/// oldest mtime first.
+const COMPLETE_FILENAME: &str = ".complete";
```
DavisVaughan, line 28: Perhaps the most interesting change is that we've switched the "cleanup" mechanism to a simpler LRU idea, kind of like salsa.

(Previously it would try and track the original libpath the request came from, and would cleanup if the package no longer existed in that libpath. It was too complicated. Plus, `oak_source` doesn't actually even take a libpath as an input anymore!)

We keep a max of `N` entries in the cache, where the mtime of the `.complete` completion sentinel file is our way of detecting "last use time" to order them.

We "touch" `.complete` on every `get()` of that package to keep it updated. Even though we only hold a shared lock on the folder, this one modification is fine because it is atomic, and there is a guarantee that no one can delete the file out from under us.

We keep at most:
- 200 srcrefs
- 200 R package sources
- 5 R versions

crates/oak_cache/src/lib.rs
```diff
@@ -0,0 +1,436 @@
+//! A generic, content-agnostic on disk directory cache
+//!
+//! A [`Cache`] is rooted at a single subfolder of the shared cache directory and
+//! holds a fixed number of entries (an LRU `capacity`). Each entry is a directory
```
lionel-, line 4: MRU capacity? Just seems like the LRU eviction is the result of caching only MRU items. Not important though.

crates/oak_cache/src/lib.rs
```diff
@@ -0,0 +1,436 @@
+//! A generic, content-agnostic on disk directory cache
```
lionel-, line 1: Should we try and mention in each lib.rs whether the crate is:

- Internal
- Experimental (aspirationally public)
- Public

For instance reading this I'm not sure if the crate is mostly for internal organisation or/and because you think it would be useful in a larger context.

lionel-: oh this seems to indicate "internal":

```rs
    Ok(etcetera::choose_base_strategy()?.cache_dir().join("oak"))
```

crates/oak_cache/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+authors.workspace = true
+edition.workspace = true
+rust-version.workspace = true
+license.workspace = true
+
+[dependencies]
+anyhow.workspace = true
+etcetera.workspace = true
+log.workspace = true
+oak_fs.workspace = true
```
lionel-, line 13: Should `FileLock` be a module in `oak_cache` rather than an entire crate? (I would lean yes to avoid over-granularity)

crates/oak_cache/src/lib.rs
lionel-: In my mind the cache would be GC'd based on time of access alone, without an upper bound on the number of items retained. Or the capacity bound could be kept as a defensive measure and set to a reasonably large number, so that we'd mostly rely on access time for GC.

crates/oak_cache/src/lib.rs
```diff
@@ -0,0 +1,436 @@
...
+        bump_last_access(&dir);
+        Some(dir)
+    }
+
+    /// Returns the entry directory for `key`, populating it if absent.
+    ///
+    /// - Takes the per-key exclusive lock
+    /// - Double-checks completion (another session may have populated it while we waited
+    ///   for the exclusive lock)
+    /// - Wipes any crashed partial results
```
lionel-, line 127: On `Err`?

crates/oak_cache/src/lib.rs
```diff
@@ -0,0 +1,436 @@
...
+    root_lock: FileLock,
+}
+
+impl Cache {
+    /// Open `<cache_dir>/oak/<root>/`, creating it if needed.
+    ///
+    /// Runs a best-effort [`Cache::clean`] under the exclusive root lock (skipped if
+    /// another session holds the shared lock), then holds the shared root lock for the
+    /// life of the returned `Cache` so handed-out paths stay valid.
+    pub fn new(root: &str, capacity: usize) -> anyhow::Result<Self> {
```
lionel-, line 84: Although there's `TempFile::new()` as precedent, I find it weird that `Cache::new()` has the (potential) side effect of creating the whole cache on disk.

How about `Cache::open()`?

crates/oak_cache/src/lib.rs
lionel-: Nope, on `clean()` when the cache is opened next and the `.complete` sentinel is missing.

Can we clean right away instead? On `Err` or `Ok(false)`.

lionel- (approved):
Nice API

crates/oak_cache/src/lib.rs
DavisVaughan: Hmm you know it started as this, but i think i got all the way up to `OakSourceHandler` and thought this looked odd

```r
impl OakSourceHandler {
    /// Build the handler, opening both caches against the shared on disk cache
    pub(crate) fn new(r: PathBuf) -> anyhow::Result<Self> {
        Ok(Self {
            srcref: SrcrefCache::open(r)?,
            source: SourceCache::open()?,
        })
    }

    /// Build the handler with both caches rooted under `root`, so tests don't touch the
    /// real on disk cache
    #[cfg(test)]
    pub(crate) fn new_in(root: &Path, r: PathBuf) -> anyhow::Result<Self> {
        Ok(Self {
            srcref: SrcrefCache::open_in(root.join("srcref"), r)?,
            source: SourceCache::open_in(root.join("source"))?,
        })
    }
}
```

compared with what i have now

```rust
impl OakSourceHandler {
    /// Build the handler, opening both caches against the shared on disk cache
    pub(crate) fn new(r: PathBuf) -> anyhow::Result<Self> {
        Ok(Self {
            srcref: SrcrefCache::new(r)?,
            source: SourceCache::new()?,
        })
    }

    /// Build the handler with both caches rooted under `root`, so tests don't touch the
    /// real on disk cache
    #[cfg(test)]
    pub(crate) fn new_in(root: &Path, r: PathBuf) -> anyhow::Result<Self> {
        Ok(Self {
            srcref: SrcrefCache::new_in(root.join("srcref"), r)?,
            source: SourceCache::new_in(root.join("source"))?,
        })
    }
}
```

but maybe thats not so bad.

ill mark it as a post-merge follow up PR

crates/oak_cache/src/lib.rs
DavisVaughan: Marked the age based GC-ing idea as something to follow up on once we determine an appropriate age

crates/oak_cache/src/lib.rs
DavisVaughan: I don't think we can _fully_ clean up. I think we can delete stray files in the folder but not the whole folder, which is why I just didn't bother with it.

When we hold a shared-lock on, say, `/oak/srcref/v1/.lock`, we can _create_ folders in there without race conditions (`std::fs::create_dir_all()` ensures this) but I don't think we can _delete_ folders as someone else might be trying to create it at the same time.

We have to way for a `clean()` when we hold the exclusive-lock on `/oak/srcref/v1/.lock` instead.

But I can go ahead and try and remove any stray files to clean up as much as we can at this point.

crates/oak_cache/Cargo.toml
DavisVaughan: Okay I'll do this and duplicate `set_readonly` in both oak_srcref/oak_source

PR #1299, opened 2026-06-29
(2) Extract out `oak_source` for CRAN downloads of R packages and R versions

DavisVaughan:
Part of #1234 
Branched from #1298 

Similar to the old cache model for CRAN sources, so not a ton to review. The main structural differences are:

- For R packages, we have `get_cran()` and `insert_cran()` 
- For R versions, we have `get_r()` and `insert_r()`

Where both of these now download and keep _the whole thing_ rather than trying to extract out only `R/`, which I think is going to pay off in the long run.

### DIFF
--- a/crates/oak_source/src/lib.rs
+++ b/crates/oak_source/src/lib.rs
@@ -1,0 +1,164 @@
+//! Download and cache whole R package and R version source trees
+//!
+//! [`SourceCache`] is a thin wrapper over two [`oak_cache::Cache`] instances pointing
+//! into:
+//!
+//! - `source/v1/cran/{name}_{version}/`, a full unpacked CRAN package tarball
+//! - `source/v1/r/{version}/`, a full unpacked R version source tarball
+//!
+//! Both keep the entire downloaded tree, unpacked eagerly with files marked read only.
+//! Each method returns the cached folder. Navigation is left to the caller, who knows the
+//! layout.
+
+mod cran;
+mod download;
+mod extract;
+mod r;
+
+use std::path::PathBuf;
+
+use oak_cache::Cache;
+
+/// Cache version
+const CACHE_VERSION: &str = "v1";
+
+/// LRU capacity for cached CRAN package source trees
+const CRAN_CAPACITY: usize = 200;
+
+/// LRU capacity for cached R version source trees, kept small because each is large
+const R_CAPACITY: usize = 5;
+
+/// Downloads and caches whole package / R version source trees
+///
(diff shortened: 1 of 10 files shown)


crates/oak_source/src/r.rs
```diff
@@ -0,0 +1,32 @@
...
+///
+/// Returns `Ok(false)` if the tarball isn't on CRAN (e.g. a development R version), which
+/// we treat as "source unavailable" rather than an error.
+pub(crate) fn populate(version: &str, dir: &Path) -> anyhow::Result<bool> {
+    let major = version
+        .split('.')
+        .next()
+        .ok_or_else(|| anyhow::anyhow!("Invalid R version for source download: {version}"))?;
+
+    let suffix = format!("src/base/R-{major}/R-{version}.tar.gz");
```
DavisVaughan, line 23: Currently we really do download and extract the entire R version source folder.

In retrospect, I think this is probably way too much. Its around 100mb unpacked on my machine.

I think in a follow up PR I'll do a careful analysis to see what we can drop. I imagine we can probably get this down to like 30mb.

I would want to keep most of `src/`, which has the R and C sources in it. But even in `src/library/base/` we can probably drop the 9mb `po/` folder of translations. So it'll take a little bit of analysis.

lionel-: Could we do the idea of bundling R 4.2 sources in Oak, with diffs for each subsequent versions? If all this data is zipped, the size overhead should be okay.

crates/oak_source/src/lib.rs
```diff
@@ -0,0 +1,162 @@
...
+
+use std::path::PathBuf;
+
+use oak_cache::Cache;
+
+/// Cache version
+const CACHE_VERSION: &str = "v1";
+
+/// LRU capacity for cached CRAN package source trees
+const CRAN_CAPACITY: usize = 200;
```
lionel-, line 26: Cf my comment in last PR about using time-based eviction instead. I would expect 200 to cause churn for users of packages with large dep trees from different domains.

Instead I'd set capacity to like 1000 so that it mostly has a defensive role, then evict unaccessed packages after e.g. 6 months (user-configurable).

lionel- (approved):

crates/oak_source/src/r.rs
DavisVaughan: Wrote this down so we can decide to do something smarter in a follow up PR

crates/oak_source/src/lib.rs
DavisVaughan: Will definitely add this after merging the stack

PR #1300, opened 2026-06-29
(3) Extract out `oak_srcref` for srcref rebuilding

DavisVaughan:
Part of #1234
Branched from #1299 

Not much to see here, exact same srcref mechanism as before. Just extracted into its own crate.

### DIFF
--- a/crates/oak_db/src/package.rs
+++ b/crates/oak_db/src/package.rs
@@ -140,6 +140,18 @@ impl Package {
             .map(|description| description.version.clone())
     }
 
+    /// The package's `Built:`, parsed lazily from `DESCRIPTION`. `None` when
+    /// the file is missing or has no `Built:` field (only installed packages have one).
+    ///
+    /// Build timestamps change on every install, so backdating probably isn't that
+    /// important here, but having the field easily accessible is nice.
+    #[salsa::tracked(returns(ref))]
+    pub fn built(self, db: &dyn Db) -> Option<String> {
+        self.description(db)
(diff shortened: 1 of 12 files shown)


crates/oak_db/src/package.rs
```diff
@@ -140,6 +140,18 @@ impl Package {
             .map(|description| description.version.clone())
     }
 
+    /// The package's `Built:`, parsed lazily from `DESCRIPTION`. `None` when
+    /// the file is missing or has no `Built:` field (only installed packages have one).
+    ///
+    /// Build timestamps change on every install, so backdating probably isn't that
+    /// important here, but having the field easily accessible is nice.
+    #[salsa::tracked(returns(ref))]
```
lionel-: Maybe untrack the method then? `description()` is already tracked.

lionel- (approved):
My (strong) expectation would be to have oak_source and oak_srcref as modules of a single crate, since they are really the same domain and same purpose, just different data acquisition approaches.

Because of the different sources (network vs disk) they have different deps, but the consumer depends on both anyway so the transitive deps are the same.

crates/oak_db/src/package.rs
DavisVaughan: Oh I forgot I can just not track it 😆

DavisVaughan:
We talked and decided to keep oak_source and oak_srcref separate in case a completely separate tool wants to use oak_source with as few deps as possible

PR #1301, opened 2026-06-29
(4) Wire up the production `SourceHandler` in the LSP

DavisVaughan:
Part of #1234
Branched from #1300 

This PR is most of the point of this series

The main point is that the source downloading is still off by default in release builds, but is on in debug builds (including unit and integration tests).


crates/ark/src/lsp/main_loop.rs
```diff
@@ -603,6 +609,33 @@ impl GlobalState {
     }
 }
 
+/// Build the LSP's [`SourceHandler`], or `None` to disable source fetching
+fn source_handler(r_home: &Path) -> Option<Arc<dyn SourceHandler>> {
+    if !cfg!(debug_assertions) {
+        // TODO!: Remove this to activate in release builds as well.
+        // Currently only active in debug builds (including unit and integration tests).
+        return None;
```
lionel-, line 617: Since we're now at the start of the release cycle I think we can remove this, so that daily build users get exposed to this new feature?

crates/ark/src/lsp/sources.rs
```diff
@@ -6,29 +6,113 @@ use std::sync::Arc;
...
+        let version = request.version();
+
+        // Base packages only exist in the R version source tarball, served from one
+        // download shared across all of them
+        if matches!(request.priority(), Some(Priority::Base)) {
+            return self
+                .source
+                .get_r(version)
+                .or_else(|| self.source.insert_r(version))
+                .map(|root| SourceResponse::Success(r_dir_for_base(&root, name)))
```
lionel-, line 69: Nicely ergonomic

lionel- (approved):
🎉

crates/ark/src/lsp/main_loop.rs
DavisVaughan: 👍  will do in a follow up PR

PR #1304, opened 2026-06-29
(6) Goto definition tests for workspace package `import()` and `importFrom()`

DavisVaughan:
Branched from #1302 

With some extracted support infrastructure: `DescriptionWriter`, `NamespaceWriter`, and the generic `TestSourceHandler`

---

Even without any additional changes, just hooking up the production `SourceHandler` has enabled some features we didn't have existing "live" tests for.

One of those is goto def on usage of `foo()` in your workspace package, when `foo()` comes from a NAMESPACE `import()` or `importFrom()`

### DIFF
--- a/crates/ark/src/lsp/tests/goto_definition.rs
+++ b/crates/ark/src/lsp/tests/goto_definition.rs
@@ -1,14 +1,32 @@
+use std::collections::HashMap;
+use std::sync::Arc;
+
 use aether_path::FilePath;
 use assert_matches::assert_matches;
+use oak_db::Db;
+use oak_db::OakDatabase;
+use oak_scan::DbScan;
+use oak_semantic::library::Library;
 use tower_lsp::lsp_types;
 use tower_lsp::lsp_types::GotoDefinitionParams;
 use tower_lsp::lsp_types::GotoDefinitionResponse;
 use url::Url;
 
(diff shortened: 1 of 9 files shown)


crates/ark/src/lsp/tests/utils/description_writer.rs
```diff
@@ -0,0 +1,45 @@
+use std::path::Path;
+
+/// Assembles a `DESCRIPTION` file from parts
```
lionel-, line 3: Good idea

crates/ark/src/lsp/tests/goto_definition.rs
```diff
@@ -214,3 +232,135 @@ fn test_sourced_file_with_sequential_redef_offers_runtime_winner() {
...
+    // The workspace folder is itself the package that imports `foo`.
+    let workspace = tempfile::tempdir().unwrap();
+    DescriptionWriter::new()
+        .package("mypackage")
+        .version("0.0.0")
+        .imports(&["foopkg"])
+        .write(workspace.path());
+    NamespaceWriter::new()
+        .import("foopkg")
+        .write(workspace.path());
```
lionel-, line 279: Nice

lionel- (approved):

PR #1305, opened 2026-06-29
(7) Allow `Package::resolve()` to skip the namespace export check for `base`

### LINKED ISSUE #1303: Mock `Package::namespace()` result for `base`
It's the one package that doesn't have a `NAMESPACE` file!

Right now that means that if you cmd+click on `environment()` with the sources support, then nothing happens even though we've downloaded the R sources and have access to base R's `R/` folder and can in theory navigate to `environment <- function() ...`

A special path in `Package::namespace()` is probably the right place to mock it? I don't _think_ modifying the on disk cache would be the correct place for this tweak.

DavisVaughan:
Branched from #1303 
Closes https://github.com/posit-dev/ark/issues/1303

I took a look at this and this seems to be the one place we use `namespace().exports` right now. Everything works as is if we just skip the gating check for `base` in particular. This works because _every_ top level assignment in base is an export by construction, so we don't really need to check this anyways.

If we find ourself getting burned by this in other places, we can reconsider, but right now this feels way cleaner than trying to reconstruct a fake namespace for base. It seems like we simply don't need it!

### DIFF
--- a/crates/oak_db/src/tests/package_resolve.rs
+++ b/crates/oak_db/src/tests/package_resolve.rs
@@ -273,3 +273,23 @@ fn test_same_name_defined_in_multiple_files_returns_each() {
     assert!(target_files.contains(&files[0]));
     assert!(target_files.contains(&files[1]));
 }
+
+#[test]
+fn test_base_exports_every_top_level_binding() {
+    // `base` doesn't have a `NAMESPACE`, so there are no namespace exports for it.
+    // Instead, `PackageVisibility::Exported` lookups on `base` skip the export check and
+    // fall straight through to looking for a top-level binding anywhere in the `base`
+    // source files.
+    //
+    // Note that this won't find primitives! These have no top-level binding.
+    let mut db = TestDb::new();
+    let (pkg, files) = setup_package(&mut db, "base", &[], &[(
+        "workspace/base/R/a.R",
+        "foo <- function() 1\n",
+    )]);
+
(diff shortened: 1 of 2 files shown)


lionel- (approved):

crates/oak_db/src/tests/package_resolve.rs
```diff
@@ -273,3 +273,23 @@ fn test_same_name_defined_in_multiple_files_returns_each() {
...
+fn test_base_exports_every_top_level_binding() {
+    // `base` doesn't have a `NAMESPACE`, so there are no namespace exports for it.
+    // Instead, `PackageVisibility::Exported` lookups on `base` skip the export check and
+    // fall straight through to looking for a top-level binding anywhere in the `base`
+    // source files.
+    //
+    // Note that this won't find primitives! These have no top-level binding.
+    let mut db = TestDb::new();
+    let (pkg, files) = setup_package(&mut db, "base", &[], &[(
+        "workspace/base/R/a.R",
```
lionel-, line 287: Interesting that you could add the R sources in the workspace to get LSP analysis against dev base packages.

DavisVaughan: Yea and I suppose the example should probably have really been using an "installed" `Package` version of `base`, but the `setup_package()` workspace package infra was right there already, and it mostly accomplishes the same thing, so I just went with that

PR #1306, opened 2026-06-30
Finish off `WorldState::root` and `WorldState::library` removal! 

DavisVaughan:
Cleans up a loooooooot of outdated things now that Salsa is pretty well integrated and fully featured

Support:
- `oak_db::Package` now has a `depends()` query (`DESCRIPTION` `Depends:`) and an `index()` query (for `INDEX`). This also requires an `index_revision` field on `Package`, which I thought we could avoid, but the 2nd commit shows that we cannot.

Tweaks:
- Diagnostics now look at `file.package()` for namespace imports instead of the old `SourceRoot`
- Diagnostics now use the new `package.index()` in `insert_package_exports()` to layer in the documented dataset symbols (that was our stopgap to avoid false positives)
- roxygen2 completions were tweaked again to directly use the `OakDatabase` rather than the old `oak_semantic::Library`
- `temp_palmerpenguin()` moved to `oak_package_metadata` since `oak_semantic::package.rs` is gone now. It is still useful for a test.

Removed:
- `WorldState` drops `root` and `library` fields, yay!
- `oak_semantic` drops `library.rs` and `package.rs`, both of which were "old style" features that we now access through `OakDatabase` directly


crates/ark/src/lsp/completions/sources/unique/comment.rs
```diff
@@ -68,11 +70,27 @@ fn completions_from_comment(
...
+        .and_then(|path| path.parent())
+    else {
+        return Ok(Some(completions));
+    };
+
+    // When you're actively working on roxygen2 itself, pull the tags from `inst/`
+    let tags = match root.kind(db) {
+        RootKind::Workspace => dir.join("inst").join("roxygen2-tags.yml"),
+        RootKind::Library => dir.join("roxygen2-tags.yml"),
+    };
```
DavisVaughan, line 93: A neat little change

crates/ark/src/lsp/state.rs
```diff
@@ -57,13 +55,6 @@ pub(crate) struct WorldState {
     /// Currently installed packages
     pub(crate) installed_packages: Vec<String>,
 
-    /// The root of the source tree (e.g., a package).
-    pub(crate) root: Option<SourceRoot>,
-
-    /// Map of package name to package metadata and package sources for installed
-    /// libraries. Lazily populated.
-    pub(crate) library: Library,
```
DavisVaughan, line 65: Woo!

crates/oak_db/src/package.rs
```diff
@@ -140,6 +148,16 @@ impl Package {
...
 
+    /// The package's `Depends:`, parsed lazily from `DESCRIPTION`. `None` when the file
+    /// is missing. Narrow query over [`Package::description`], same backdating story as
+    /// [`Package::version`].
+    #[salsa::tracked(returns(ref))]
+    pub fn depends(self, db: &dyn Db) -> Option<Vec<String>> {
+        self.description(db)
+            .as_ref()
+            .map(|description| description.depends.clone())
+    }
```
DavisVaughan, line 181: New query

crates/oak_db/src/package.rs
```diff
@@ -181,4 +199,43 @@ impl Package {
...
     }
+
+    /// The package's parsed `INDEX`
+    ///
+    /// Returns `None` for workspace packages, which never have an `INDEX`.
+    ///
+    /// Returns an empty `Index` for installed packages with missing or unreadable
+    /// indexes.
+    #[salsa::tracked(returns(ref))]
+    pub fn index(self, db: &dyn Db) -> Option<Index> {
```
DavisVaughan, line 232: New query + `index_revision` field

crates/ark/src/lsp/completions/sources/unique/comment.rs
lionel-: yep nice

crates/ark/src/lsp/diagnostics.rs
```diff
@@ -964,20 +955,31 @@ fn insert_package_exports(
...
-    let Some(package) = context.library.get(package_name) else {
+) -> anyhow::Result<()> {
+    let Some(package) = context.db.package_by_name(package_name) else {
         return Err(anyhow::anyhow!(
             "Can't get exports from package {package_name} because it is not installed."
         ));
     };
 
+    // Start from explicit `NAMESPACE` exports
+    let mut exports = package.namespace(context.db).exports.clone().into_vec();
```
lionel-, line 966: This should be a `Package::exports()` query mirroring `File::exports()`. I thought I had implemented it already, but can't find it (maybe it was abandoned at some point).

There will be some complex resolution in the future (e.g. for regex `exportPattern`), so we should call a single method everywhere exports are needed.

It'd return a vector of names that you can then resolve with top-level `Package::resolve()`.

crates/oak_db/src/package.rs
```diff
@@ -45,6 +46,13 @@ pub struct Package {
     /// time, only stat it, so installed packages the user never imports cost
     /// nothing beyond the stat.
     pub namespace_revision: FileRevision,
+    /// Mtime of the package's `INDEX` file. [`FileRevision::zero`] when it can't be
```
lionel-, line 50: Can just refer to previous comment instead of repeating

lionel- (approved):
How did I forget to do this very satisfying consolidation/cleanup!

PR #1309, opened 2026-07-01
Run Evaluate requests with timeout

lionel-:
Addresses https://github.com/posit-dev/positron/issues/14481

Adds `with_timeout()` to run R code with a timeout and use that in the Evaluate handlers (in notebook and console paths). This prevents an inflooping (or long-running) expression in the Watch Pane from freezing the kernel.

* `with_timeout()` runs a closure in a context where interrupts and polled events are restored. This allows R to check for interrupts, and allows polled events to check for a timeout.

* We check for a timeout from polled events to avoid the complexity of checking time in a side thread. When a timeout is detected, we record that fact in a thread-local variable and signal an R interrupt.

* The interrupt is either caught by `with_timeout()`'s `try_catch()`, or by an inner one running in the closure. Either way, `with_timeout()` returns whether the expression timed out to its caller.

This setup allows inner code to recover from the interrupt, which is not ideal. However this is much simpler than trying to propagate the interrupt with a Rust panic, since the latter approach would require making sure the panic does not cross a C stack which would be UB and likely to corrupt R's state. The simpler approach of propagating with Rust errors could in principle lead to surprising behaviour, but is simpler and safer to implement.

Regarding the safety of longjumping from polled events with an R interrupt, this should be sufficiently safe because:

- This is equivalent to an R-level error and should be caught the same way by proper safeguards like `try_catch()`.
- This can only happen from within `with_timeout()`, which is a controlled environment that expects the interrupt to fire any time R is called.

We should still assume it's possible for some Rust destructors to get bypassed by the interrupt if the calling context is not super vigilant about wrapping in `try_catch()`. However the alternative of having Ark freeze with runaway evals is much worse.

This is Unix-only until we figure out https://github.com/posit-dev/ark/pull/1222

#### New Features

- N/A

#### Bug Fixes

- Fixed hangs with the Evaluate pane of the debugger (https://github.com/posit-dev/positron/issues/14481). When an expression takes a long time to evaluate (or even infloops), it is now cancelled after 1 second.

### DIFF
--- a/crates/ark/src/console/console_repl.rs
+++ b/crates/ark/src/console/console_repl.rs
@@ -2570,6 +2570,10 @@ impl Console {
         if let Some(text) = self.debug_filter.check_timeout() {
             self.emit_stdout(text);
         }
+
+        // Interrupt any in-flight evaluation (e.g. a Watch pane expression) that
+        // has outlived its timeout.
+        crate::timeout::check_timeout();
     }
 
     pub(super) fn eval_env(&self) -> RObject {
--- a/crates/ark/src/dap/dap_jupyter_handler.rs
+++ b/crates/ark/src/dap/dap_jupyter_handler.rs
@@ -154,9 +154,13 @@ impl DapJupyterHandler {
 
         let outcome = match outcome {
             Ok(outcome) => outcome,
-            // A longjump that escaped `DapHandler::evaluate()`'s own `try_catch`
-            // (e.g. while building the response), caught by the try-idle sandbox.
-            Err(err) => return Ok(self.error_response(seq, "evaluate", &err.to_string())),
+            Err(err) => {
+                // `DapHandler::evaluate()` doesn't touch R outside of `with_timeout()`,
+                // which already converts any escaping error into `EvaluateOutcome::Error`
+                // (see `DapHandler::evaluate()`). So this shouldn't be reachable.
+                log::error!("Jupyter DAP: Unexpected error from `try_idle_task`: {err:?}");
+                return Ok(self.error_response(seq, "evaluate", &err.to_string()));
+            },
         };
 
         match outcome {
--- a/crates/ark/src/dap/dap_jupyter_handler.rs
+++ b/crates/ark/src/dap/dap_jupyter_handler.rs
@@ -32,12 +32,12 @@ use stdext::result::ResultExt;
 use crate::dap::dap_notebook;
 use crate::dap::dap_server::DapConsoleEvent;
 use crate::dap::dap_server::DapHandler;
+use crate::dap::dap_server::EvaluateOutcome;
 use crate::dap::dap_state::Breakpoint;
 use crate::dap::dap_state::BreakpointState;
 use crate::dap::dap_state::Dap;
 use crate::dap::dap_state::THREAD_ID;
 use crate::r_task;
-use crate::r_task::TryIdleOutcome;
 use crate::request::RRequest;
 
 pub struct DapJupyterHandler {
@@ -141,30 +141,35 @@ impl DapJupyterHandler {
 
         // This runs only if R is idle. Even though we stopped at some point,
         // causing the Evaluate request, R could be busy evaluating the next
-        // expression already. A runaway expression is interrupted after a
-        // timeout and reported as `TimedOut`.
-        let result = r_task::try_idle_task(move |capture| {
+        // expression already, in which case `try_idle_task()` returns `None`.
+        // `DapHandler::evaluate()` runs the eval under a timeout so a runaway
+        // expression is interrupted and doesn't freeze the kernel.
+        let outcome = r_task::try_idle_task(move |capture| {
             DapHandler::evaluate(&state, &expression, frame_id, capture)
         });
 
-        match result {
-            TryIdleOutcome::Busy => Ok(self.error_response(seq, "evaluate", "R is busy")),
-            TryIdleOutcome::TimedOut => {
-                Ok(self.error_response(seq, "evaluate", "Evaluation timed out"))
-            },
-            // The task raised an R error, caught by the sandbox on the R thread.
-            TryIdleOutcome::Ran(Err(err)) => {
-                Ok(self.error_response(seq, "evaluate", &format!("{err}")))
-            },
-            TryIdleOutcome::Ran(Ok(Ok(body))) => {
-                let ResponseBody::Evaluate(eval) = body else {
-                    return Err(anyhow::anyhow!("Unexpected response body from evaluate"));
-                };
+        let Some(outcome) = outcome else {
+            return Ok(self.error_response(seq, "evaluate", "R is busy"));
+        };
+
+        let outcome = match outcome {
+            Ok(outcome) => outcome,
+            // A longjump that escaped `DapHandler::evaluate()`'s own `try_catch`
+            // (e.g. while building the response), caught by the try-idle sandbox.
+            Err(err) => return Ok(self.error_response(seq, "evaluate", &err.to_string())),
+        };
+
+        match outcome {
+            EvaluateOutcome::Ok(ResponseBody::Evaluate(eval)) => {
                 Ok(self.success_response(seq, "evaluate", serde_json::to_value(eval)?))
             },
-            TryIdleOutcome::Ran(Ok(Err(err))) => {
-                Ok(self.error_response(seq, "evaluate", &format!("{err}")))
+            EvaluateOutcome::Ok(_) => {
+                Err(anyhow::anyhow!("Unexpected response body from evaluate"))
             },
+            EvaluateOutcome::TimedOut => {
+                Ok(self.error_response(seq, "evaluate", "Evaluation timed out"))
+            },
+            EvaluateOutcome::Error(err) => Ok(self.error_response(seq, "evaluate", &err)),
         }
     }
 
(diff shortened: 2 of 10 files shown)


crates/ark/src/console/console_repl.rs
```diff
@@ -2509,6 +2509,9 @@ impl Console {
             return;
         }
 
+        // Interrupt any in-flight evaluations that have outlived their timeout
+        crate::timeout::check_timeout();
```
DavisVaughan: But if we do #1222, where does this go?

#1222 removed `polled_events()` entirely.

In the current state of that PR, we are left with a `process_events()` hook for the `debug_filter` to still be checked occasionally (https://github.com/posit-dev/ark/pull/1222#discussion_r3267489082) but as you discovered in https://github.com/posit-dev/ark/pull/1222#discussion_r3310467187 we shouldn't actually take over `ptr_R_ProcessEvents` because other packages like Quartz may use it instead.

So we are going to have to do something else there. I really wasn't thrilled with the idea of setting `Callback` (i.e. `R_ProcessEvents`) on Windows and `R_PolledEvents` on Unix as mentioned in https://github.com/posit-dev/ark/pull/1222#pullrequestreview-4370255430. I remember sending this slack message about it

```
i would really rather not mix
* setting `R_ProcessEvents` on Windows via `CallBack`
* setting `R_PolledEvents` on Unix

that sounds like something that is likely to just confuse me greatly in the future, and is part of what im trying to get away from in this pr. it is very hard to keep what these do (and when they are called) straight in my head

since:
* we cant use `R_PolledEvents` on windows (doesnt exist)
* we cant use `R_ProcessEvents` on mac (Quartz needs it)

and because we only have 1 thing we need to regularly check, the `debug_filter` thing, i vote we get out of the business of setting `R_PolledEvents` and `R_ProcessEvents` at all and instead do one of two things

* the side thread that checks `debug_filter` and eat the 2mb cost
* drop caring about `debug_filter` at all for this nice use case. it isn't worth this amount of trouble IMO.

IMO not setting `R_PolledEvents` and `R_ProcessEvents` would be a huge win for us in terms of complexity and never having to think about this again
```

With the main conclusion there being that it would be nice to not use EITHER `R_PolledEvents` or `R_ProcessEvents`, to try to avoid all of the funny business with those, and instead maybe use a "watcher" thread of some kind.

The watcher thread could probably handle `debug_filter` flushing, but I don't think it could directly do `crate::timeout::check_timeout()` since that pokes an R API via `crate::signals::set_interrupts_pending(true)`.

I'll also add that I'm not sure that calling this at `R_PolledEvents` / `R_ProcessEvents` time really reliably saves us. It still relies on _someone_ calling these events handlers, and I'm pretty sure that arbitrary R code running in an `Evaluate` could just never check those, right? But maybe if that is the case, it means they wouldn't be able to respect an interrupt request either, so that's okay?

So I'm not really sure what the right thing to do is! But I would hate to add this and box ourselves out from being able to finish off #1222 when we were soooooo close to getting rid of interrupt tasks, so I think I'd like us to have a plan before merging this.

crates/ark/src/dap/dap_jupyter_handler.rs
```diff
@@ -138,24 +139,37 @@ impl DapJupyterHandler {
...
-                };
+        let Some(outcome) = outcome else {
+            return Ok(self.error_response(seq, "evaluate", "R is busy"));
+        };
+
+        let outcome = match outcome {
+            Ok(outcome) => outcome,
+            // A longjump that escaped `DapHandler::evaluate()`'s own `try_catch`
+            // (e.g. while building the response), caught by the try-idle sandbox.
+            Err(err) => return Ok(self.error_response(seq, "evaluate", &err.to_string())),
```
DavisVaughan: It looks to me like `DapHandler::evaluate` cannot actually jump.

Instead `DapHandler::evaluate` has this exact kind of comment about `Dap::evaluate`, so I would have thought that this Err case would have already been handled fully by `DapHandler::evaluate`

crates/ark/src/dap/dap_server.rs
```diff
@@ -706,18 +717,31 @@ impl DapHandler {
...
+
+        if timed_out {
+            return EvaluateOutcome::TimedOut;
+        }
+
+        let variable = match res {
+            Ok(Ok(variable)) => variable,
+            Ok(Err(err)) => return EvaluateOutcome::Error(format!("{err}")),
+            // An R longjump that escaped `Dap::evaluate()`'s own `try_catch`.
+            Err(err) => return EvaluateOutcome::Error(err.to_string()),
```
DavisVaughan, line 739: This is the comment i was referring too, this seems like where the longjmp is actually handled?

crates/ark/src/r_task.rs
```diff
@@ -152,7 +148,10 @@ where
         },
     }
 
-    done_rx.recv().log_err()?;
+    if done_rx.recv().is_err() {
```
DavisVaughan, line 147: is an Err normal behavior here?

crates/ark/src/dap/dap_server.rs
```diff
@@ -1007,13 +1036,11 @@ impl<R: Read, W: Write> DapServer<R, W> {
...
-        log::trace!("DAP: Spawning idle task for evaluate");
         r_task::spawn(RTask::send_idle_any_prompt(async move |mut capture| {
-            log::trace!("DAP: Idle task started for evaluate");
-
             let rsp = match DapHandler::evaluate(&state, &expression, frame_id, &mut capture) {
-                Ok(body) => req.success(body),
-                Err(err) => req.error(&format!("Error: {err}")),
+                EvaluateOutcome::Ok(body) => req.success(body),
+                EvaluateOutcome::TimedOut => req.error("Evaluation timed out"),
+                EvaluateOutcome::Error(err) => req.error(&format!("Error: {err}")),
```
DavisVaughan, line 1043: Yea like here you are relying on `DapHandler::evaluate` to fully handle that longjmp case for you, I think, and translate it into a `EvaluateOutcome::Error` already.

So that other reference in dap_jupyter_handler.rs to longjmp does seem not quite right? I think?

crates/ark/src/timeout.rs
```diff
@@ -0,0 +1,132 @@
...
+//! and the deadline is never checked. Fixed by https://github.com/posit-dev/ark/pull/1222.
+
+use std::cell::Cell;
+use std::time::Duration;
+use std::time::Instant;
+
+use harp::raii::RLocalInterruptsSuspended;
+
+/// How long an evaluation in `with_timeout()` may run before we interrupt it.
+pub(crate) const EVAL_TIMEOUT: Duration = Duration::from_secs(1);
```
DavisVaughan: That's pretty darn aggressive for slow windows machines! 😬

crates/ark/src/timeout.rs
```diff
@@ -0,0 +1,132 @@
...
+fn try_catch_with_timeout<F, T>(f: F) -> harp::Result<T>
+where
+    F: FnOnce() -> T,
+{
+    let _interrupts = RLocalInterruptsSuspended::new(false);
+    // Re-enable Ark's polled-events handler so `check_timeout()` runs during `f`.
+    let _polled = harp::raii::RLocal::new(
+        Some(crate::console::r_polled_events as unsafe extern "C-unwind" fn()),
+        unsafe { libr::R_PolledEvents },
+    );
```
DavisVaughan: hmm you kind of want the opposite of RLocalPolledEventsSuspended, but that's a harp thing...

DavisVaughan (commented):
Mostly want to talk through the ultimate plan with #1222 before committing to this, because relying on being able to poke an R API from polled-events / process-events might box us in a bit.

crates/ark/src/timeout.rs
lionel-: hmm maybe we should bump that on CI to avoid flakes? The main thing I want to avoid is 10 Watch Pane expressions hanging the console for 30sec

crates/ark/src/r_task.rs
lionel-: Nope probably a main thread panic

lionel-:
Rebased on #1222 and cancellation tests now pass on Windows!

crates/ark/src/timeout.rs
```diff
@@ -0,0 +1,139 @@
...
+/// propagate the interrupt with a Rust panic, but that needs to be carefully
+/// engineered. Be aware that because of the approach taken here, it's possible
+/// for code inside `f` to try and recover from Rust errors
+/// (`harp::Error::TopLevelExecError`) caused by the cancellation interrupt.
+///
+/// Returns `f`'s value wrapped in the `try_catch()` result and a boolean
+/// indicating whether the timeout fired.
+///
+/// Must run on the R thread.
+pub(crate) fn with_timeout<F, T>(timeout: Duration, f: F) -> (harp::Result<T>, bool)
```
DavisVaughan: If you hit a timeout, is it _ever_ valid to look at `harp::Result<T>`? 

My intuition is no, and that we should make that impossible via the type system, like

```rust
enum TimeoutResult<T> {
  Ok(harp::Result<T>),
  Timeout
}
```

crates/ark/src/timeout.rs
```diff
@@ -0,0 +1,139 @@
+//
+// timeout.rs
+//
+// Copyright (C) 2026 Posit Software, PBC. All rights reserved.
+//
+//
+
+//! A timeout that breaks runaway R-thread work with an interrupt.
```
DavisVaughan, line 8:
```suggestion
//! A timeout that breaks runaway R-thread work with an interrupt
```

Random nit: I have explicit Claude.md instructions not to add a period after 1 liner code comments, which IIUC is our usual convention. I don't care that much, but I seem to notice it every time I see them 😆

DavisVaughan (approved):

PR #1313, opened 2026-07-03
Don't update Salsa inputs unnecessarily

lionel-:

Flagged by Fable. There's no backdating for Salsa inputs! So we have to conditionally set them to prevent invalidation.

Unfortunately there's no easy way to gate the setters. We could wrap the Salsa inputs in a newtype but that would have ergonomy costs.

Since new code that writes to the database should be fairly rare from now on, we can just be careful about this when it comes up.


DavisVaughan:
I'm remembering that I read this in https://hackmd.io/5ddaZLKYSVC_YTTD3SzRDw?view#Input-Structs

Screenshot 2026-07-03 at 8 31 16 AM

DavisVaughan (approved):

PR #1314, opened 2026-07-03
Make it impossible for background threads to mutate Salsa inputs

lionel-:

This approach is taken from Rust-Analyzer where a snapshot is guaranteed to never write to the DB. Having this type-enforced safety is nice because writing to the DB from outside the main loop could deadlock.

- New `WorldSnapshot` type that's the read-only counterpart to `WorldState`. It holds a clone of the Salsa data wrapped in `Analysis`, as well as clones of other world state fields.

- New `Analysis` type that holds the `OakDatabase` in a private field, and lends it only as a shared reference. This is what prevents writes.

### DIFF
--- a/crates/ark/src/lsp/db.rs
+++ b/crates/ark/src/lsp/db.rs
@@ -17,6 +17,37 @@ pub(crate) trait ArkDb: oak_db::Db {}
 #[salsa::db]
 impl ArkDb for OakDatabase {}
 
+/// Read-only handle to an `OakDatabase` that acts as a read-only snapshot.
+///
+/// An `Analysis` snapshot holds the DB clone in a private field and lends a
+/// shared ref via `read()`. This prevents a read-only background workers
+/// from reaching DB setters and deadock with the main loop. This is oak's
+/// version of rust-analyzer's `Analysis` facade.
+#[derive(Clone, Debug)]
+pub(crate) struct Analysis {
+    db: OakDatabase,
+}
+
+impl Analysis {
+    pub(crate) fn new(db: OakDatabase) -> Self {
+        Self { db }
+    }
+
+    /// The database as a shared `&dyn ArkDb`. Only lends `&`, so a reader can
+    /// query but can't reach a setter through it.
+    pub(crate) fn read(&self) -> &dyn ArkDb {
+        &self.db
+    }
+
+    /// The database's salsa cancellation token. Read-side only: it observes and
+    /// arms cancellation, it doesn't mutate any input, so it's safe to expose
+    /// on the read-only facade. Only cancellation tests arm it by hand.
+    #[cfg(test)]
+    pub(crate) fn cancellation_token(&self) -> salsa::CancellationToken {
+        salsa::Database::cancellation_token(&self.db)
+    }
+}
+
 /// Extension trait that adds ark-side query methods to `oak_db::File`.
 pub(crate) trait FileArkExt {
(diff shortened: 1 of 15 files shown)


crates/ark/src/lsp/state.rs
```diff
@@ -8,13 +8,21 @@ use oak_semantic::library::Library;
...
 /// The world state, i.e. all the inputs necessary for analysing or refactoring
 /// code. This is a pure value. There is no interior mutability in this data
 /// structure. It can be cloned and safely sent to other threads.
+///
+/// The main loop owns and mutates this. Background readers get a
+/// [`WorldSnapshot`] instead, which holds a read-only [`Analysis`] in place of
+/// the writable `OakDatabase`. This prevents background threads from reaching a
+/// Salsa input setter. See [`Self::diagnostics_snapshot`] and
+/// [`Self::snapshot`]. This split mirrors rust-analyzer's `GlobalState`  and
+/// `GlobalStateSnapshot`.
```
DavisVaughan:
```suggestion
/// The main loop owns and mutates this. Background readers get a
/// [`WorldStateSnapshot`] instead, which holds a read-only [`OakDatabaseSnapshot`] in place of
/// the writable `OakDatabase`. This prevents background threads from reaching a
/// Salsa input setter. See [`Self::diagnostics_snapshot`] and
/// [`Self::snapshot`]. This split mirrors rust-analyzer's `GlobalState`  and
/// `GlobalStateSnapshot`.
```

I like this idea a lot, but I find the names to be very confusing in the long term.

Like this used to make sense

```rust
state: WorldState,
```

but no longer does to me as

```rust
state: WorldSnapshot,
```

I would prefer

```rust
state: WorldStateSnapshot,
```

---

And `Analysis` doesn't really mean much to me at all, I would definitely like to see that tied to the `OakDatabase` name in some way. I'd be fine with `OakDatabaseSnapshot` or `ReadOnlyOakDatabase`. 

Like seeing `pub(crate) db: Analysis,` just confuses me naming wise :/

crates/ark/src/lsp/db.rs
```diff
@@ -17,6 +17,37 @@ pub(crate) trait ArkDb: oak_db::Db {}
...
+/// Read-only handle to an `OakDatabase` that acts as a read-only snapshot.
+///
+/// An `Analysis` snapshot holds the DB clone in a private field and lends a
+/// shared ref via `read()`. This prevents a read-only background workers
+/// from reaching DB setters and deadock with the main loop. This is oak's
+/// version of rust-analyzer's `Analysis` facade.
+#[derive(Clone, Debug)]
+pub(crate) struct Analysis {
+    db: OakDatabase,
+}
```
DavisVaughan: I have a gut feeling this is too complex for us

What if we just had

```rust
pub(crate) struct WorldStateSnapshot {
    // Private field. We only give out `&dyn ArkDb` handles on threads to enforce that it is read-only.
    db: OakDatabase
}

impl WorldStateSnapshot {
    /// Read-only access to the `OakDatabase`. Important that we return as `&dyn ArkDb`
    /// to prevent the thread from simply `.clone()`ing to get its own copy that it can call
    /// setters with.
    pub(crate) db(&self) -> &dyn ArkDb {
        &self.db
    }
}
```

Your little test only `cancellation_token()` could live on `WorldStateSnapshot`, I think.

This setup makes a lot of sense to me! And avoids a lot of the layers that seem to make naming this thing hard.

DavisVaughan (approved):
I really really like the idea a lot, but the naming is throwing me off rn

PR #1315, opened 2026-07-03
Bump Salsa revision manually for non-Salsa updates

lionel-:
Branched from https://github.com/posit-dev/ark/pull/1314

Removes a FIXME that I left for non-Salsa inputs, like console inputs and diagnostics config. These updates manually triggered diagnostics out-of-band, which had the issue that ongoing diagnostics did not get cancelled by a Salsa bump. The concurrent diagnostics could publish out of order, resulting in stale results.

It's actually easy to bump the revision manually, which both cancels in flight diagnostics and ensures ordering. That's because the synthetic write:

- Causes Salsa to trigger cancellation and park the caller until all readers are dropped
- Refreshes diagnostics automatically via the revision check at the end of the main loop.

### DIFF
--- a/crates/ark/src/lsp/main_loop.rs
+++ b/crates/ark/src/lsp/main_loop.rs
@@ -384,8 +384,9 @@ impl GlobalState {
         // resolved definitions), so any handler that writes to oak invalidates
         // them. Rather than have each write site remember to refresh, we watch
         // the oak revision across the whole tick: if a handler advanced it,
-        // refresh centrally. Config and console state live outside oak, so
-        // handlers that mutate those still refresh explicitly.
+        // refresh centrally. Config and console state live outside oak, so the
+        // handlers that mutate those advance the revision synthetically (see
+        // `WorldState::bump_revision`) to route through this same path.
         let old_revision = salsa::plumbing::current_revision(&self.world.db);
 
         match event {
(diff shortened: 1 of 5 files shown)


crates/ark/src/lsp/main_loop.rs
```diff
@@ -1031,15 +1032,12 @@ static DIAGNOSTICS_QUEUE: LazyLock<tokio::sync::mpsc::UnboundedSender<RefreshDia
...
-/// FIXME: Batches triggered without an oak write (console inputs, diagnostics
-/// config) have no such barrier. An older pass can run concurrently with the
-/// newer one and publish last, leaving diagnostics computed from the older
-/// console scopes or config until the next refresh.
+/// Batches can't publish out of order. Each pass holds a db snapshot, and a
+/// salsa write blocks until all snapshots drop, so by the time the write
+/// completes and the newer batch is enqueued, any older pass has either unwound
+/// with `Cancelled` or already produced its result. Changes to state that lives
+/// outside oak (console inputs, diagnostics config) get the same barrier by
+/// advancing the revision synthetically (see `WorldState::bump_revision`).
```
DavisVaughan:
```suggestion
/// advancing the revision synthetically (see [`WorldState::bump_revision`]).
```

if you want to linkify the doc comments

crates/ark/src/lsp/state_handlers.rs
```diff
@@ -506,10 +506,11 @@ async fn update_config(
         }
     }
 
-    // Refresh diagnostics if the configuration changed
+    // Refresh diagnostics if the configuration changed. The config lives
+    // outside Oak so we bump the revision manually.
     if state.config.diagnostics != diagnostics_config {
         tracing::info!("Refreshing diagnostics after configuration changed");
-        lsp::main_loop::diagnostics_refresh_all(state);
+        state.bump_revision();
```
DavisVaughan, line 485:
```suggestion
    // Refresh diagnostics if the configuration changed. The config lives
    // outside Oak so we bump the revision manually, causing diagnostics
    // to refresh on the next tick.
if state.config.diagnostics != diagnostics_config {
tracing::info!("Bumping salsa revision after configuration changed");
        state.bump_revision();
```

idk, feels more precise to me this way?

crates/ark/src/lsp/state_handlers.rs
```diff
@@ -526,8 +527,9 @@ pub(crate) fn did_change_console_inputs(
     // We currently rely on global console scopes for diagnostics, in particular
     // during package development in conjunction with `devtools::load_all()`.
     // Ideally diagnostics would not rely on these though, and we wouldn't need
-    // to refresh from here.
-    lsp::diagnostics_refresh_all(state);
+    // to refresh from here. The scopes live outside oak so we bump the revision
+    // manually.
+    state.bump_revision();
```
DavisVaughan, line 504:
```suggestion
    // to refresh from here. The scopes live outside oak so we bump the revision
    // manually, causing diagnostics to get refreshed on the next tick.
    state.bump_revision();
```

DavisVaughan (approved):

PR #1316, opened 2026-07-03
Add session mode section in `ark --help`

### LINKED ISSUE #1311: Expand `--session-mode` documentation
Hi,

With the default `--session-mode console` errors come through on IoPub like this:

```
ename = "",
evalue = "Error in `dplyr::select()`:\n! Can't select columns that don't exist.",
traceback = { 
  " ▆",
  "1. ├─dplyr::select(dplyr::starwars, bla)",
  "2. ├─dplyr:::select.data.frame(dplyr::starwars, bla)",
  // More lines
}
```

But with `--session-mode notebook` you get the `evalue` prepended to the `traceback`:

```
ename = "",
evalue = "Error in `dplyr::select()`:\n! Can't select columns that don't exist.",
traceback = { 
  "Error in `dplyr::select()`:\n! Can't select columns that don't exist.",
  " ▆",
  "1. ├─dplyr::select(dplyr::starwars, bla)",
  "2. ├─dplyr:::select.data.frame(dplyr::starwars, bla)",
  // More lines
}
```

This follows the format expected by [JupyterLab](https://github.com/jupyterlab/jupyterlab/blob/4a19f8fff4049a9536e21fa2980fe7e5ee592d01/packages/rendermime/src/outputmodel.ts#L304), which doesn't display the `ename` or `evalue` if the `traceback` is supplied.

Thankfully `ark --install` writes `--session-mode notebook` into the kernelspec, but I still managed to trip up on this and it took me a while to figure out what I was doing wrong! 

Many thanks :)

lionel-:
Closes #1311

### DIFF
--- a/crates/ark/src/main.rs
+++ b/crates/ark/src/main.rs
@@ -42,7 +42,9 @@ Available options:
 -- arg1 arg2 ...             Set the argument list to pass to R; defaults to
                              --interactive
 --startup-file FILE          An R file to run on session startup
---session-mode MODE          The mode in which the session is running (console, notebook, background)
+--session-mode MODE          The mode in which the session is running: "console" (the
+                             default), "notebook", or "background". See the
+                             "Session modes" section.
 --no-capture-streams         Do not capture stdout/stderr from R
 --default-repos              Set the default repositories to use, by name:
                              "rstudio" ('cran.rstudio.com', the default), or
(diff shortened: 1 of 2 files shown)


crates/ark/src/main.rs
```diff
@@ -72,6 +74,28 @@ Available options:
...
+  all.
+
+- Autoprint of intermediate expressions. When a single request contains several
+  top-level expressions (e.g. `a` and `b` in `a; b; c`), "console" mode streams
+  each intermediate result to the frontend, matching R behaviour. "notebook"
+  mode emits only the last expression's result, matching how a notebook cell is
+  typically evaluated by Jupyter kernels.
+
+Note that `ark --install` sets the generated kernelspec to run Ark in notebook
+mode.
```
DavisVaughan: I think CLI help should be relatively quiet, reporting just the options available with a brief description about them. If we started adding these extra docs about each option, our `--help` output would get way too long imo.

This feels like it should be in Ark website documentation somewhere. For lack of a better place, maybe a new `Ark CLI` section of https://github.com/posit-dev/ark/blob/main/doc/configuration.md

I feel like this is especially true here, with these extremely niche details that are better left to more in depth docs, rather than the first thing a user sees when requesting help with `--help`

DavisVaughan (approved):
Fine, but I do think we should not put the extended docs in the CLI output itself

PR #1319, opened 2026-07-06
Fix blank Data Explorer when viewing objects piped with `%>%`

### LINKED ISSUE #7385: Pipe dataframe to View() does not work for tidyverse pipes
#### Positron and OS details:

Positron Version: 2025.04.0 (user setup) build 250
Code - OSS Version: 1.98.0
Commit: 627b3d723ee0b3a5bd86731a75902d419cafe22d
Date: 2025-04-03T16:50:39.209Z
Electron: 34.2.0
Chromium: 132.0.6834.196
Node.js: 20.18.2
V8: 13.2.152.36-electron.0
OS: Windows_NT x64 10.0.26100

#### Interpreter details:

R 4.4.2

## Describe the issue:

Minor bug: when piping a dataframe to View() with the tidyverse/magrittr `%>%` pipe, the positron dataview pops up but shows no data. When using the R native `|>` pipe, the dataview does correctly display the dataframe. 

## Steps to reproduce the issue:

1. Load tidyverse and any dataframe 
2. `foo %>% View()` (Does not work)
3. `foo |> View()` (Does work)

## Expected or desired behavior:
Dataviewer should show the dataframe with the `%>%` pipe as well

### LINKED ISSUE #8167: data explorer not rendering any data when piped from select()
Type: <b>Bug</b>

when piping a select statement and passing to the data explorer, nothing loads

VS Code version: Positron 1.100.0 (Universal) (e7e5d5590adc6bf7b760f837051cd2f57ea9e3dc, 2025-05-29T20:44:08.369Z)
OS version: Darwin arm64 24.5.0
Modes:

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Apple M1 (8 x 2400)|
|GPU Status|2d_canvas: enabled<br>canvas_oop_rasterization: enabled_on<br>direct_rendering_display_compositor: disabled_off_ok<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>opengl: enabled_on<br>rasterization: enabled<br>raw_draw: disabled_off_ok<br>skia_graphite: disabled_off<br>video_decode: enabled<br>video_encode: enabled<br>webgl: enabled<br>webgl2: enabled<br>webgpu: enabled<br>webnn: disabled_off|
|Load (avg)|2, 2, 2|
|Memory (System)|16.00GB (0.79GB free)|
|Process Argv||
|Screen Reader|no|
|VM|0%|
</details><details><summary>Extensions (11)</summary>

Extension|Author (truncated)|Version
---|---|---
ruff|cha|2025.24.0
vscode-pull-request-github|Git|0.110.0
pyright|ms-|1.1.402
debugpy|ms-|2025.6.0
jupyter|ms-|2025.4.1
jupyter-keymap|ms-|1.1.2
vscode-jupyter-cell-tags|ms-|0.1.9
vscode-jupyter-slideshow|ms-|0.1.6
air-vscode|Pos|0.14.0
publisher|pos|1.16.1
quarto|qua|1.123.0

</details>

juliasilge:
- Closes posit-dev/positron#7385 
- Closes posit-dev/positron#8167 

This PR fixes a blank Data Explorer when viewing a data frame through the magrittr pipe, e.g. `ggplot2::diamonds %>% View()`. Currently the tab opens, shows a loading spinner, and then goes blank because the kernel closes the comm before any data is sent. We originally thought this was something to do with renv, but that seems like it was incorrect. The base pipe (`ggplot2::diamonds |> View()`) was always unaffected.

## Why did this happen?

`View()` registers the Data Explorer against a *variable binding* (a name plus an environment) so the grid can live-update when the underlying object changes. On each top-level execution, `RDataExplorer::update()` re-resolves that binding with `Rf_findVarInFrame()` and compares the result against the object it is currently showing.

The magrittr pipe binds the piped object to `.` inside its private pipe environment, and it stores `.` as a promise rather than a plain value. So the re-resolution returned the promise object (a `PROMSXP`), not the data frame. That promise pointer never equals the forced data frame we are viewing, so `update()` treated it as a replacement value, failed to compute a table shape for it, and closed the comm. The base pipe works because `diamonds |> View()` is pure syntax that rewrites to `View(diamonds)`, and `diamonds` in the global environment is a plain value.

## What changes in this PR

When re-resolving the binding, force the value if it is a promise before comparing and computing its shape. For the magrittr case the promise is already forced, so this is a cached read with no side effects, and the resolved value now equals the data frame we are viewing, so the comm stays open. The change preserves existing behavior for genuine top-level variables: removing a watched variable (`rm(x)`) still resolves to an unbound value and correctly closes the comm.

This seems like a pretty innocuous fix to me, but if you all think we should _not_ take it just for magrittr pipe users, I will not be super offended.

#### New Features

- N/A

#### Bug Fixes

- Fixed the Data Explorer being blank when viewing objects piped with `%>%` (posit-dev/positron#7385 )

### DIFF
--- a/crates/ark/src/data_explorer/r_data_explorer.rs
+++ b/crates/ark/src/data_explorer/r_data_explorer.rs
@@ -72,7 +72,8 @@ use harp::r_symbol;
 use harp::table_kind;
 use harp::tbl_get_column;
 use harp::utils::r_is_promise;
-use harp::utils::r_promise_force;
+use harp::utils::r_promise_is_forced;
+use harp::utils::r_promise_value;
 use harp::vector::CharacterVector;
 use harp::vector::Vector;
 use harp::ColumnNames;
@@ -248,15 +249,22 @@ impl RDataExplorer {
             Rf_findVarInFrame(env, sym)
         };
 
-        // A binding can hold a promise rather than a plain value. Most
-        // importantly, the magrittr pipe `df %>% View()` binds the object to
-        // `.` as a promise inside the pipe's environment. Force the promise so
-        // we track and view the underlying value rather than the promise
-        // object, which has no data frame shape and would otherwise make
-        // `get_shape()` fail and close the comm.
+        // A binding can hold a promise rather than a plain value. The magrittr
+        // pipe `df %>% View()` binds the object to `.` as a promise inside the
+        // pipe's environment, and `View()` has already forced that promise to
+        // obtain the value it displays. Here we only read the value of an
+        // already-forced promise; we never force one ourselves. Forcing runs
+        // arbitrary R code that could emit warnings or output, force other
+        // promises, or mutate bindings and options, and such side effects must
+        // never originate from a comm update. If the binding currently holds an
+        // unforced promise, we can't read its value without side effects, so we
+        // leave the current view in place until the binding resolves.
         // See https://github.com/posit-dev/positron/issues/7385.
         let new = if r_is_promise(new) {
-            r_promise_force(new).map_or(new, |value| value.sexp)
+            if !r_promise_is_forced(new) {
+                return Ok(true);
+            }
+            r_promise_value(new)
         } else {
             new
         };
--- a/crates/ark/src/data_explorer/r_data_explorer.rs
+++ b/crates/ark/src/data_explorer/r_data_explorer.rs
@@ -71,6 +71,8 @@ use harp::object::RObject;
 use harp::r_symbol;
 use harp::table_kind;
 use harp::tbl_get_column;
+use harp::utils::r_is_promise;
+use harp::utils::r_promise_force;
 use harp::vector::CharacterVector;
 use harp::vector::Vector;
 use harp::ColumnNames;
@@ -244,6 +246,19 @@ impl RDataExplorer {
         let new = unsafe {
             let sym = r_symbol!(binding.name);
             Rf_findVarInFrame(env, sym)
+        };
+
+        // A binding can hold a promise rather than a plain value. Most
+        // importantly, the magrittr pipe `df %>% View()` binds the object to
+        // `.` as a promise inside the pipe's environment. Force the promise so
+        // we track and view the underlying value rather than the promise
+        // object, which has no data frame shape and would otherwise make
+        // `get_shape()` fail and close the comm.
+        // See https://github.com/posit-dev/positron/issues/7385.
+        let new = if r_is_promise(new) {
+            r_promise_force(new).map_or(new, |value| value.sexp)
+        } else {
+            new
         };
 
         let changed = if new == self.table.get().sexp {
--- a/crates/ark/tests/integration/data_explorer.rs
+++ b/crates/ark/tests/integration/data_explorer.rs
@@ -143,51 +143,6 @@ fn open_data_explorer_from_expression(expr: &str, bind: Option<&str>) -> anyhow:
     })
 }
 
-// Open a data explorer whose binding tracks `.` bound to a *promise* in a
-// private environment. This mirrors how the magrittr pipe `df %>% View()`
-// binds the piped object inside its pipe environment, which used to cause the
-// comm to close on the next environment change (posit-dev/positron#7385).
-fn open_data_explorer_with_promise_binding() -> TestSetup {
-    let (iopub_tx, iopub_rx) = bounded::<IOPubMessage>(10);
-
-    let comm_id = uuid::Uuid::new_v4().to_string();
-    let outgoing_tx = CommOutgoingTx::new(comm_id, iopub_tx);
-    let (comm_event_tx, _) = bounded::<CommEvent>(10);
-    let ctx = CommHandlerContext::new(outgoing_tx, comm_event_tx);
-
-    let inner = r_task(|| {
-        let env = harp::parse_eval_global(
-            "local({
-                e <- new.env()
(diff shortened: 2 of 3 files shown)


crates/ark/src/data_explorer/r_data_explorer.rs
```diff
@@ -246,6 +248,19 @@ impl RDataExplorer {
...
 
+        // A binding can hold a promise rather than a plain value. Most
+        // importantly, the magrittr pipe `df %>% View()` binds the object to
+        // `.` as a promise inside the pipe's environment. Force the promise so
+        // we track and view the underlying value rather than the promise
+        // object, which has no data frame shape and would otherwise make
+        // `get_shape()` fail and close the comm.
+        // See https://github.com/posit-dev/positron/issues/7385.
+        let new = if r_is_promise(new) {
+            r_promise_force(new).map_or(new, |value| value.sexp)
```
lionel-: Should at least use `r_promise_force_with_rollback()` to avoid `"restarting interrupted promise evaluation"` warnings when the promise fails.

But a comm is not the right place to be forcing a promise as this can have all sorts of side effects. The promise might force other promises which won't be rolled back, might cause warnings, might emit output, change bindings in arbitrary environments, modify options, etc. Note that completions only force lazyload promises, and leaves all other promises intact. These lazyload promises have well known behaviour.

Looking at the use case (piping into `View()`), it seems fine to force the magrittr input, it's just that this should not happen here.

crates/ark/tests/integration/data_explorer.rs
```diff
@@ -1276,6 +1321,45 @@ fn test_search_filters() {
...
 
+// Regression test for https://github.com/posit-dev/positron/issues/7385.
+//
+// The magrittr pipe `df %>% View()` binds the piped object to `.` as a promise
+// inside magrittr's pipe environment. When the data explorer re-resolves the
+// binding on an environment change, it must force that promise instead of
+// treating the promise object (which has no data frame shape) as a replacement
+// value and closing the comm.
+#[test]
+fn test_promise_binding_does_not_close() {
```
lionel-: These legacy tests are slated to be migrated to the new protocol-level backend (https://github.com/posit-dev/ark/issues/1074).

Should now use a proper integration test with a dummy frontend: https://github.com/posit-dev/ark/blob/main/crates/ark/tests/integration/data_explorer_integration.rs

crates/ark/src/data_explorer/r_data_explorer.rs
lionel-: IIUC, the promise should already be forced by `View()` right? Then I'd prefer to do something like:

```
if prom_is_forced(x)
then prom_force(x)
```

lionel-: Would be nice to test for the edge case of an object watched by a data explorer going from forced to delayed, and check that the promise is not forced by the explorer even though it could watch forced values just before.

crates/ark/tests/integration/data_explorer.rs
```diff
@@ -143,6 +143,51 @@ fn open_data_explorer_from_expression(expr: &str, bind: Option<&str>) -> anyhow:
     })
 }
 
+// Open a data explorer whose binding tracks `.` bound to a *promise* in a
+// private environment. This mirrors how the magrittr pipe `df %>% View()`
+// binds the piped object inside its pipe environment, which used to cause the
+// comm to close on the next environment change (posit-dev/positron#7385).
+fn open_data_explorer_with_promise_binding() -> TestSetup {
```
lionel-: Should not add more helpers for legacy tests.

lionel- (commented):
I think this should work out in principle, but the fix and testing approach need a different shape.

crates/ark/src/data_explorer/r_data_explorer.rs
juliasilge: Thank you, I understand! I updated so the re-resolution now reads the value of an already-forced promise via `r_promise_value()` instead of forcing. If the binding holds an *unforced* promise, we `return Ok(true)` early and leave the current view in place rather than forcing it. That also means no `r_promise_force_with_rollback()` is needed, since we never force from here.

juliasilge:
I dropped the legacy helper and test and moved the tests to the protocol-level backend using the dummy frontend:

- `test_magrittr_pipe_data_explorer_stays_open`: drives the real `df %>% View()`, triggers an environment change, and asserts the comm stays open and still serves the data frame.
- `test_data_explorer_does_not_force_delayed_binding`: I added your suggested edge case. The watched binding transitions from a plain value to an unforced `delayedAssign` promise that flips a flag when evaluated. We assert the flag stays `FALSE` (never forced) and the explorer keeps showing the last resolved value.

I confirmed both tests fail with the exact `comm_close` on the unfixed code.

lionel- (approved):
Perfect!

lionel-:
Here are the release notes template from https://github.com/posit-dev/ark/pull/1318 in case you'd like to try https://github.com/posit-dev/positron/pull/14676 for bumping Ark:

```md
### Positron Release Notes

<!--
  These notes are typically filled by the Positron team. If you are an external
  contributor, you may leave the `N/A` bullets in place.
-->

#### New Features

- N/A

#### Bug Fixes

- N/A
```

juliasilge:
I updated the PR body with the release notes! ✅

PR #1329, opened 2026-07-09
Always list base packages in `workspace_dependencies()`

DavisVaughan:
I noticed that if you open purrr to https://github.com/tidyverse/purrr/blob/481e829f297fd4315b386518215157f361475ad0/R/adverb-safely.R#L42 and just try to goto definition on `force()`, then nothing happens.

That's because `force()` is a base function, but base isn't listed in Imports or Depends so it doesn't get picked up in `workspace_dependencies()` and we never try and set its package `R/` sources.

Similar story if you open an arbitrary workspace of scripts where you use any package from the default R `search()` path without first calling `library()` or `::` on that package somewhere in the file.

I think the right thing to do is to just ensure that the default `search()` packages are always listed in `workspace_dependencies()`. This is definitely right for workspace scripts. For workspace packages _technically_ an individual package file would only need {base}, but if you add any extra scripts into the package, like `inst/data.R`, then we'd require the rest of the base packages anyways, so I think it is fine to just default to all of the default `search()` packages for workspace packages too. It doesn't affect diagnostics there, just the set of packages we try and download sources for, so over approximation is pretty harmless.

### DIFF
--- a/crates/oak_db/src/workspace.rs
+++ b/crates/oak_db/src/workspace.rs
@@ -17,10 +17,13 @@ pub fn workspace_dependencies(db: &dyn Db) -> Vec<Package> {
     // shouldn't require re-querying `DESCRIPTION` dependencies)
     let from_files = workspace_file_dependencies(db);
     let from_packages = workspace_package_dependencies(db);
+    let from_search = default_search_package_packages(db);
 
-    let mut packages: Vec<Package> = Vec::with_capacity(from_files.len() + from_packages.len());
+    let mut packages: Vec<Package> =
+        Vec::with_capacity(from_files.len() + from_packages.len() + from_search.len());
     packages.extend(from_files);
     packages.extend(from_packages);
+    packages.extend(from_search);
 
     packages.sort_by_cached_key(|package| package.name(db));
     packages.dedup_by_key(|package| package.name(db));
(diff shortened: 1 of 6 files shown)


crates/oak_db/src/workspace.rs
```diff
@@ -77,6 +80,23 @@ fn workspace_package_dependencies(db: &dyn Db) -> Vec<Package> {
...
+///
+/// - Workspace package files technically only depend on {base}, but if your package has
+///   any arbitrary scripts in it then we are going need the rest of the base packages
+///   anyways, and returning them all here just ensures that their sources are available,
+///   it doesn't affect package diagnostics, so it's okay to over approximate for
+///   simplicity.
+///
+/// These are effectively static, so we don't need to sort them by name at this point
+#[salsa::tracked(returns(ref))]
+fn default_search_package_packages(db: &dyn Db) -> Vec<Package> {
```
lionel-:
```suggestion
fn default_search_packages(db: &dyn Db) -> Vec<Package> {
```
?

crates/oak_db/src/workspace.rs
```diff
@@ -77,6 +80,23 @@ fn workspace_package_dependencies(db: &dyn Db) -> Vec<Package> {
...
+///   mentioned via a `library()` / `::` call.
+///
+/// - Workspace package files technically only depend on {base}, but if your package has
+///   any arbitrary scripts in it then we are going need the rest of the base packages
+///   anyways, and returning them all here just ensures that their sources are available,
+///   it doesn't affect package diagnostics, so it's okay to over approximate for
+///   simplicity.
+///
+/// These are effectively static, so we don't need to sort them by name at this point
+#[salsa::tracked(returns(ref))]
```
lionel-, line 95: Technically we might want to retrieve `R_DEFAULT_PACKAGES` / `getOption("defaultPackages")` from the side car session: https://cran.r-project.org/doc/manuals/r-release/R-admin.html#Default-packages-1

Totally fine to ignore it for now, but maybe mention in a comment

lionel- (approved):
> For workspace packages technically an individual package file would only need {base}, but if you add any extra scripts into the package, like inst/data.R, then we'd require the rest of the base packages anyways, so I think it is fine to just default to all of the default search() packages for workspace packages too. It doesn't affect diagnostics there, just the set of packages we try and download sources for, so over approximation is pretty harmless.

For scripts in packages, I agree. For files in `R/`, we should only provide base by default. Is that still the case (or was the case to begin with? I don't remember)?

This _will_ affect diagnostics later on so I would rather get this right now.

PR #1332, opened 2026-07-09
Add `goto_definition()` support for `::` and `:::`

DavisVaughan:
This is the official version of the POC `feature/sources-demo`, with full test coverage and correct `NamespaceVisibility` handling


crates/oak_db/src/package_resolve.rs
```diff
@@ -1,20 +1,9 @@
...
-///
-/// Mirrors R's `::` vs `:::` distinction. `Exported` requires `name` to
-/// appear in the package's NAMESPACE `export()` directives. `Internal`
-/// returns any top-level binding the package's files define, exported
-/// or not.
-#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
-pub enum PackageVisibility {
-    Exported,
-    Internal,
-}
```
DavisVaughan, line 16: This file and `identifier.rs` need to share a single `NamespaceVisibility` type

I renamed it from `PackageVisibility` to `NamespaceVisibility` because it feels like `identifier.rs` is a better owner of this type, and there we have `Identifier::NamespaceAccess` as the struct that owns the `NamespaceVisibility` field, so that name felt better.

crates/oak_ide/src/find_references.rs
```diff
@@ -198,23 +200,24 @@ fn find_member_references(db: &dyn Db, file: File, name: &str, kind: MemberKind)
...
     primary: File,
-    namespace: Name<'db>,
+    package: Name<'db>,
     name: Name<'db>,
     include_declaration: bool,
 ) -> Vec<FileRange> {
-    if let Some(package) = db.package_by_name(namespace.text(db).as_str()) {
-        let defs = package.resolve(db, name, PackageVisibility::Internal);
+    if let Some(package) = db.package_by_name(package.text(db).as_str()) {
+        let defs = package.resolve(db, name, NamespaceVisibility::Internal);
```
DavisVaughan, line 208: Is this hardcoding of `Internal` really correct? We could pass through a known `visibility` from above if we wanted to. I see the doc comment, but it doesn't make sense to me.

crates/oak_ide/src/goto_definition.rs
```diff
@@ -19,20 +23,43 @@ use crate::NavigationTarget;
...
 pub fn goto_definition(db: &dyn Db, file: File, offset: TextSize) -> Vec<NavigationTarget> {
-    let Some(Identifier::Variable { range, .. }) = Identifier::classify(db, file, offset) else {
-        return Vec::new();
+    let defs = match Identifier::classify(db, file, offset) {
+        Some(Identifier::Variable { range, .. }) => goto_local_variable(db, file, range),
+        Some(Identifier::NamespaceAccess {
+            package,
+            name,
+            visibility,
+        }) => goto_namespace_access(db, package, name, visibility),
```
DavisVaughan, line 32: The crux of this PR

crates/oak_db/src/identifier.rs
```diff
@@ -31,13 +31,11 @@ pub enum Identifier<'db> {
...
-    /// the same wherever the cursor sits. `part` records which half it was on.
+    /// the same wherever the cursor sits.
     NamespaceAccess {
-        namespace: Name<'db>,
+        package: Name<'db>,
         name: Name<'db>,
-        part: NamespacePart,
-        operator_range: TextRange,
-        name_range: TextRange,
+        visibility: NamespaceVisibility,
```
DavisVaughan, line 38: I have simplified this type for now since nothing actually used `part`, `operator_range`, or `name_range`, and it kind of felt like overkill to keep them around, and it felt like a strange set of things to keep. It seemed reasonable to me that we could add them back as required

lionel- (approved):
LGTM!

PR #1333, opened 2026-07-14
Add build version info to kernel info reply

jmcphers:
This change adds a couple more Positron-specific fields to the `kernel_info_reply`, with the goal of making it possible for Positron to know exactly what version of ark it's dealing with, without making additional RPCs. 

Needed for https://github.com/posit-dev/positron/issues/14407.

### DIFF
--- a/crates/ark/src/shell.rs
+++ b/crates/ark/src/shell.rs
@@ -139,7 +139,7 @@ impl ShellHandler for Shell {
                 input_prompt: kernel_info.input_prompt.clone(),
                 continuation_prompt: kernel_info.continuation_prompt.clone(),
                 build_version: Some(String::from(crate::BUILD_VERSION)),
-                commit: Some(String::from(env!("BUILD_GIT_HASH"))),
+                commit: Some(String::from(crate::BUILD_GIT_HASH)),
             }),
         };
         let mut supported_features = vec![String::from("debugger")];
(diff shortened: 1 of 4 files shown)


crates/ark/src/shell.rs
```diff
@@ -138,6 +138,8 @@ impl ShellHandler for Shell {
             positron: Some(LanguageInfoPositron {
                 input_prompt: kernel_info.input_prompt.clone(),
                 continuation_prompt: kernel_info.continuation_prompt.clone(),
+                build_version: Some(String::from(crate::BUILD_VERSION)),
+                commit: Some(String::from(env!("BUILD_GIT_HASH"))),
```
lionel-: For consistency we might want to do the same for git commit as you did for build version with Davis in https://github.com/posit-dev/ark/pull/1188

lionel- (approved):

crates/ark/src/shell.rs
DavisVaughan: If you're asking for `crate::BUILD_GIT_HASH` to be a constant alongside `crate::BUILD_VERSION`, then yea I'd be in favor of that

DavisVaughan (approved):

crates/ark/src/shell.rs
jmcphers: thanks, that's a good idea! https://github.com/posit-dev/ark/pull/1333/commits/b3e9c0df945e5773d342b24d1ec6e8716fae693b

Issue #1338, opened 2026-07-16
Add support for NSE functions and semantic effects


lionel-:
Most important effect to support: evaluation in a nested scope (e.g. `test_that()`). Definitions in a nested scope shouldn't escape.

Other effects to support soon:
- Attach (`library()`, `require()`)
- Source
- Assign (S7's `:=`, magrittr's `%<>%`, `base::delayedAssign()`, etc)
- Quote (`quote()`, `bquote()`), including unquoted holes with e.g. `.(arg)` or `!!arg`

Effects for later:
- Eval (to be done a bit later because this requires data flow to be useful)
- ExprCapture (`substitute()`)
- EnvCapture (`new.env()`, `parent.frame()`, `environment()`)

Not straightforward because these effects can interact with each other, e.g. a Source may hide effects, a nested NSE scope may cause an effect to be masked, or on the contrary may prevent an effect to be masked, etc. Main idea: resolve effect linearly. Any effect in a lazy scope that contradicts a linear effect is an ambiguous program that should be linted.

All these effects should be expressible with either:

- Local declarations with `declare()`, which requires a mini language spec for describing these effects

- Custom Rust handlers vendored in Oak. Many functions have complex effects depending on arguments and trying to express these with a DSL is a lost cause. We'll provide support for argument matching, static resolution of values, etc.

PR #1339, opened 2026-07-16
Scaffold NSE support

lionel-:

This PR adds preliminary support for NSE functions that create a local scope, like `local()` or `test_that()`. Data-masking functions like `with()` are also supported, although only for the effect of opening a scope, not for the effect of injecting a data frame in the scope.

The semantic index is created in two passes per scan unit to resolve a chicken and egg problem: to recognise an NSE function while we're building the index, we need to resolve the function callee: is it locally bound, perhaps to a function with NSE annotations? If it's locally unbound, does it come from an imported file?

The two passes are structured by scan units, which are either the file scope, or a lazy scope. Eager scopes are just traversed linearly without creating a new unit.

We already had a similar pre-scan to collect bound symbols for lazy scopes. This first pass now has more responsibilities. The two passes design is forward-looking and takes into account that file imports from `source()` (and `library()` in a future PR) calls may themselves be masked by assignments, which may or may not have local effects if they happen within a nested NSE scope.

NSE functions are parameterised along two axes:

- `NseScope`: Either `Current` (`evalq()`, `rlang::on_load()`) or `Nested` (`local()`, `test_that()`).

- `NseTiming`: Either `Eager` (`local()`, `test_that()`) or `Lazy` (`rlang::on_load()`, `shiny::observe()`)

Each combination gets distinct handling:

- Current + Eager: Same as regular evaluation. The reason this combination exists is that in the future we'll support the `envir` argument of `evalq()`, at which point it will stop falling into the Current bucket depending on its arguments.

- Nested + Eager: Pushes a scope. Scanned inline during the eager linear scan. Uses of enclosing names get a point-in-time snapshot, unlike the accumulated union a lazy scope gets. That's the main difference with lazy scopes like functions.

- Current + Lazy: It's a special scan unit whose definitions are deferred, just like `<<-`

  ```r
  x <- 1
  on_load({ x <- 2 })   # routed to file scope, deferred
  x                     # sees {1, 2}: deferred def didn't shadow `x <- 1`
  ```

- Nested + Lazy: Pushes a scope, special scan unit. Uses refer to an enclosing snapshot, exactly like `function` scopes.

The linear scanner decides whether a function is NSE by resolving symbols eagerly. This is always correct in an _eager_ context, but is only a guess in a _lazy_ context, like inside a function or Shiny reactive. If a binding comes in later and contradicts that guess, we can't order these two states of the world.

```r
f <- function() local({ x <- 1 })
local <- identity
```

The position that we take from now on is that this is an ambiguous program that should be linted. We record the ambiguity in `LazyShadowAmbiguity { name, range }` and log it later on. In a future PR, we'll surface in diagnostics.

That's a general principle that we'll apply for all effects, for instance this will produce a lint:

```r
function() source("path.R")
source <- identity
```

Imported effects are now handled by new ImportsResolver methods. The PR also introduces an effect registry where we'll start documenting effects of known functions from base and other important packages (rlang, shiny, etc).

#### New Features

- Improved support for symbol navigation and symbol renaming within well known NSE functions like `test_that()`.

#### Bug Fixes

- N/A

### DIFF
--- a/crates/oak_semantic/src/effects.rs
+++ b/crates/oak_semantic/src/effects.rs
@@ -1,0 +1,33 @@
+use crate::semantic_index::NseScope;
+use crate::semantic_index::NseTiming;
+
+/// Effects of a resolved function.
+///
+/// Currently only records NSE effects. In the future this will include other
+/// effects such as `attach` (for e.g. `library()`) and `assign` (for the
+/// eponymous function).
+#[derive(Debug, Clone, Copy, Default)]
+pub struct Effects {
+    pub nse: Option<NseAnnotation>,
+}
+
+impl Effects {
+    pub fn nse(nse: NseAnnotation) -> Self {
+        Self { nse: Some(nse) }
+    }
+}
+
+/// Annotation describing how an NSE function's arguments create scopes.
+#[derive(Debug, Clone, Copy)]
+pub struct NseAnnotation {
+    pub arguments: &'static [NseArgument],
+}
+
+/// A single argument that creates an NSE scope.
+#[derive(Debug)]
+pub struct NseArgument {
+    pub name: &'static str,
+    pub position: usize,
+    pub scope: NseScope,
+    pub timing: NseTiming,
+}
--- a/crates/oak_semantic/src/effects.rs
+++ b/crates/oak_semantic/src/effects.rs
@@ -8,24 +8,24 @@ use crate::semantic_index::NseTiming;
 /// eponymous function).
 #[derive(Debug, Clone, Copy, Default)]
 pub struct Effects {
-    pub nse: Option<NseAnnotation>,
+    pub nse: Option<ArgumentsAnnotation>,
 }
 
 impl Effects {
-    pub fn nse(nse: NseAnnotation) -> Self {
+    pub fn nse(nse: ArgumentsAnnotation) -> Self {
         Self { nse: Some(nse) }
     }
 }
 
 /// Annotation describing how an NSE function's arguments create scopes.
 #[derive(Debug, Clone, Copy)]
-pub struct NseAnnotation {
-    pub arguments: &'static [NseArgument],
+pub struct ArgumentsAnnotation {
+    pub arguments: &'static [Argument],
 }
 
 /// A single argument that creates an NSE scope.
 #[derive(Debug)]
-pub struct NseArgument {
+pub struct Argument {
     pub name: &'static str,
     pub position: usize,
     pub scope: NseScope,
--- a/crates/oak_semantic/src/resolver.rs
+++ b/crates/oak_semantic/src/resolver.rs
@@ -1,5 +1,8 @@
 use url::Url;
 
+use crate::effects::Effects;
+use crate::effects_registry;
+
 /// The result of resolving a `source()` call. Returned by
 /// [`ImportsResolver::resolve_source`].
 #[derive(Clone)]
@@ -28,14 +31,15 @@ pub struct SourceResolution {
 ///   for isolated indexing (CLI tools, unit tests).
 /// - `oak_db::SalsaImportsResolver`: salsa-backed lookup against the source graph.
 ///
-/// The trait grows along two axes as new analyses land:
+/// The trait has three queries:
 ///
 /// - [`resolve_source`](ImportsResolver::resolve_source) is the bulk
 ///   query, "enumerate every name this `source("path")` brings in," used
 ///   to inject `DefinitionKind::Import` entries at each source() offset.
-/// - A future `resolve_name(scope, name)` is the point query, "find
-///   the import that resolves this specific name in this scope," used by
-///   NSE call-site analysis.
+/// - [`resolve_effects`](ImportsResolver::resolve_effects) resolves a bare
+///   callee against imports, e.g. the search path, and returns known effects.
+/// - [`resolve_qualified_effects`](ImportsResolver::resolve_qualified_effects)
+///   resolves the effects of a `pkg::fn` (or `:::) callee against a named package.
 pub trait ImportsResolver {
     /// Resolve a `source("path")` call to the target file's exported names
     /// and transitive `library()` attachments. The path is the literal
@@ -43,6 +47,25 @@ pub trait ImportsResolver {
     /// anchoring it (workspace root, calling file's directory, ...).
     /// Returns `None` when the target can't be located.
     fn resolve_source(&mut self, path: &str) -> Option<SourceResolution>;
+
+    /// Resolve a bare callee `name` to its effects. The builder state is passed
+    /// in because the resolver can't query our own semantic index without
+    /// creating a cycle:
+    ///
+    /// - `attached`: packages attached at this point, in flow order.
+    /// - `lazy`: whether the callee sits in a lazy context like a function.
+    fn resolve_effects(&mut self, name: &str, attached: &[String], lazy: bool) -> Option<Effects> {
+        let _ = (name, attached, lazy);
+        None
+    }
+
+    /// Resolve a namespace-qualified callee `pkg::fn` (or equivalently with
+    /// `:::`) to its effects.
+    fn resolve_qualified_effects(&mut self, package: &str, name: &str) -> Option<Effects> {
+        effects_registry::lookup(package, name)
+            .copied()
+            .map(Effects::nse)
+    }
 }
 
 /// Resolver that returns nothing. The builder skips all cross-file
--- a/crates/oak_semantic/src/semantic_index.rs
+++ b/crates/oak_semantic/src/semantic_index.rs
@@ -335,8 +335,7 @@ impl SemanticIndex {
         let local = bindings.definitions().iter().map(move |&d| (scope_id, d));
 
         let enclosing = if bindings.may_be_unbound() {
-            let symbol_id = self.uses(scope_id)[use_id].symbol();
-            self.enclosing_bindings(scope_id, symbol_id)
+            self.enclosing_bindings(scope_id, use_id)
         } else {
             None
         };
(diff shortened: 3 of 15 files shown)


crates/oak_semantic/src/resolver.rs
```diff
@@ -28,21 +31,41 @@ pub struct SourceResolution {
...
+    }
+
+    /// Resolve a namespace-qualified callee `pkg::fn` (or equivalently with
+    /// `:::`) to its effects.
+    fn resolve_qualified_effects(&mut self, package: &str, name: &str) -> Option<Effects> {
+        effects_registry::lookup(package, name)
+            .copied()
+            .map(Effects::nse)
+    }
 }
```
DavisVaughan, line 69: Do you think this one really needs to be a trait method?

It feels like this is the definitive implementation for this function, and it doesn't use `self` at all.

Like, could it just be a "free" function that queries this external effect registry?

crates/oak_semantic/src/semantic_index.rs
```diff
@@ -764,6 +805,15 @@ pub enum NamespaceAccessKind {
     Internal,
 }
 
+/// A diagnostic surfaced while building the semantic index, for downstream
+/// consumers to turn into user-facing diagnostics.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum SemanticDiagnostic {
+    /// An NSE call recognized in a lazy context whose binding is overwritten
+    /// later on (in subsequent parent code or in another lazy context).
+    LazyShadowAmbiguity { name: String, range: TextRange },
```
DavisVaughan: I feel like I'm expecting to see 2 `range`s here!

- One for the initial NSE call in the lazy context
- One for the overwrite location

That way you can give a nice full circle diagnostic that tells you which binding was overwritten and where the overwrite happened. What do you think?

crates/oak_semantic/src/semantic_index.rs
```diff
@@ -405,6 +412,40 @@ pub enum ScopeKind {
...
+    Nested,
+}
+
+/// Whether an NSE scope evaluates eagerly (at the call site) or lazily
+/// (at an unknown later time).
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum NseTiming {
+    Eager,
+    Lazy,
+}
```
DavisVaughan, line 443: Maybe a little more description of Eager vs Lazy here? Like the `e.g.` examples from above in `NseScope` helped me understand that feature.

crates/oak_semantic/src/semantic_index.rs
```diff
@@ -405,6 +412,40 @@ pub enum ScopeKind {
...
+/// Where definitions in an NSE scope land.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum NseScope {
+    /// Definitions go to the current (parent) environment.
+    /// e.g. `rlang::on_load()`
+    Current,
+    /// Definitions go to a nested environment.
+    /// e.g. `local()`, `test_that()`, `with()`
+    Nested,
+}
```
DavisVaughan, line 430: Isn't there a 3rd possibility? Like `Parent` maybe? I'm thinking of some environment that is passed through multiple layers before finally being used. This is a bad example but kind of demonstrates the idea:

```r
x <- "foo"

fn <- function(env = caller_env()) {
  fn2(env)
}

fn2 <- function(env = caller_env()) {
  x <- "bar"

  # `env` here isn't really Current, it's multiple layers up
  eval(quote(x), envir = env)
}

fn()
#> [1] "foo"
```

crates/oak_semantic/src/semantic_index.rs
```diff
@@ -405,6 +412,40 @@ pub enum ScopeKind {
...
+impl ScopeKind {
+    /// Whether free variables in this scope resolve against the union of all
+    /// enclosing definitions (lazy) or against a point-in-time snapshot at the
+    /// call site (eager). `Function` bodies run at an unknown later time, so
+    /// they're always lazy.
+    pub fn is_lazy(self) -> bool {
+        match self {
+            ScopeKind::File => false,
+            ScopeKind::Function => true,
+            ScopeKind::Nse(_, laziness) => laziness == NseTiming::Lazy,
```
DavisVaughan:
```suggestion
            ScopeKind::Nse(_, timing) => timing == NseTiming::Lazy,
```

crates/oak_semantic/src/effects.rs
```diff
@@ -0,0 +1,33 @@
+use crate::semantic_index::NseScope;
+use crate::semantic_index::NseTiming;
+
+/// Effects of a resolved function.
+///
+/// Currently only records NSE effects. In the future this will include other
+/// effects such as `attach` (for e.g. `library()`) and `assign` (for the
+/// eponymous function).
```
DavisVaughan, line 8: eponymous lmao, what a word!

crates/oak_semantic/src/effects.rs
```diff
@@ -0,0 +1,33 @@
...
+impl Effects {
+    pub fn nse(nse: ArgumentsAnnotation) -> Self {
+        Self { nse: Some(nse) }
+    }
+}
+
+/// Annotation describing how an NSE function's arguments create scopes.
+#[derive(Debug, Clone, Copy)]
+pub struct ArgumentsAnnotation {
+    pub arguments: &'static [Argument],
```
DavisVaughan, line 23: Neat `entry!` macro to get a static lifetime

crates/oak_semantic/src/effects.rs
```diff
@@ -0,0 +1,33 @@
...
+/// eponymous function).
+#[derive(Debug, Clone, Copy, Default)]
+pub struct Effects {
+    pub nse: Option<ArgumentsAnnotation>,
+}
+
+impl Effects {
+    pub fn nse(nse: ArgumentsAnnotation) -> Self {
+        Self { nse: Some(nse) }
+    }
```
DavisVaughan, line 17: Nit: 

I have some gut feeling that `pub nse` here should really be `pub arguments` or something?

Like when use use `Effect::nse` it seems like its because you're wrapping argument info.


Or maybe it's that a function can (eventually) contribute 3 ish kinds of effects:
- NSE
- Attach
- Assign

And it just so happens to be that the only way it can contribute an NSE effect is via these arguments? That's fine then, I guess.

crates/oak_semantic/src/use_def_map.rs
```diff
@@ -304,7 +306,11 @@ pub(crate) struct UseDefMapBuilder {
     // Currently used for `<<-` extra definitions in ancestor scopes.
     deferred_defs: Vec<(SymbolId, DefinitionId)>,
     enclosing_snapshots: IndexVec<EnclosingSnapshotId, Bindings>,
-    snapshot_watchers: FxHashMap<SymbolId, SmallVec<[EnclosingSnapshotId; 1]>>,
+    // Snapshots subscribed to every definition of a symbol (lazy snapshots).
+    def_watchers: FxHashMap<SymbolId, SmallVec<[EnclosingSnapshotId; 1]>>,
+    // Snapshots subscribed only to deferred (`<<-`) definitions of a symbol
+    // (eager snapshots).
+    deferred_def_watchers: FxHashMap<SymbolId, SmallVec<[EnclosingSnapshotId; 1]>>,
```
DavisVaughan: I find it confusing that "lazy snapshots" go with `def_watchers`

But "eager snapshots" go with _deferred_ _def_watchers

Like the eager+deferred throws me off, is it just me?

crates/oak_semantic/src/use_def_map.rs
```diff
@@ -458,23 +483,68 @@ impl UseDefMapBuilder {
...
+    /// snapshot. One snapshot is shared by every use of the symbol in the body,
+    /// so a use before the `<<-` picks it up too. That is a known over-approximation.
+    ///
+    /// ```r
+    /// x <- 1
+    /// local({
+    ///     x        # {1, 2}: Should be {1} but shares one snapshot with the use below
+    ///     x <<- 2  # deferred def, folded into the snapshot
+    ///     x        # {1, 2}
+    /// })
```
DavisVaughan, line 507: And what happens when you start incorporating the `envir` argument?

crates/oak_semantic/src/use_def_map.rs
```diff
@@ -458,23 +483,68 @@ impl UseDefMapBuilder {
...
+    /// eager body changes the binding while the body runs, and uses later in
+    /// the body must see that change. A plain `<-` after the eager call on the
+    /// other hand runs once the body has finished, so it stays out of the
+    /// snapshot. One snapshot is shared by every use of the symbol in the body,
+    /// so a use before the `<<-` picks it up too. That is a known over-approximation.
+    ///
+    /// ```r
+    /// x <- 1
+    /// local({
+    ///     x        # {1, 2}: Should be {1} but shares one snapshot with the use below
```
DavisVaughan: "Should be" makes me think we have a bug, is it?

DavisVaughan (commented):
I've looked at everything except for `builder_nse.rs`, which I'll do on Monday, but here are some starting comments

crates/oak_semantic/src/resolver.rs
lionel-: The registry is a crutch, but ideally the effects should come from annotations, and then we do need the imports resolver.

The method will probably evolve over time though. Some thoughts:

- Effects parsing belongs in oak_semantic, so the oak_db impl would have to call oak_semantic to parse decls
- Maybe instead we resolve to a definition and parse the decls from there. In this case the trait method is no longer about effects at all.

crates/oak_semantic/src/semantic_index.rs
lionel-: I'm not sure yet how this will play out exactly, but this struct will change completely when we introduce environment capture. This will require some thinking.

crates/oak_semantic/src/effects.rs
lionel-: Not the AI's choice in this case ;)

crates/oak_semantic/src/effects.rs
lionel-: Yep this becomes `arguments` in an ulterior PR. In addition to `Nse` (which will become `EvalQ`), an argument might be `Quote` or `Eval`.

crates/oak_semantic/src/use_def_map.rs
lionel-: More like a known limitation, but I've just refactored things to lift it.

crates/oak_semantic/src/use_def_map.rs
lionel-: It'll keep working as is when the environment is nested or current. What happens in the other cases depends on what the environment resolves to (unknown, data-masked, etc).

PR #1351, opened 2026-07-16
Filter same-version rebuilds from `pkg_outdated`

juliasilge:
This addresses part of posit-dev/positron#14294, where the packages pane surfaces a large number of packages as having updates available when in fact they don't.

`.ps.rpc.pkg_outdated()` was returning everything from `utils::old.packages()`. That function flags a package as outdated when the repository version is strictly newer *or* when the version is identical but the repository copy has a newer build or publication date (see the `needs.install` closure inside `utils:::old.packages`). Right after a new R minor is installed, CRAN's freshly rebuilt binaries carry newer `Built` dates than the user's carried-over packages, so hundreds of same-version packages get flagged even though no newer version exists.

The pane's update indicator means "a newer version is available," so this filters `old.packages()` down to packages whose repository version is strictly greater than the installed version, dropping same-version rebuilds.

### Changes

- Split the filtering and formatting out of `.ps.rpc.pkg_outdated()` into a pure `pkg_outdated_result()` helper so it can be tested without querying live repositories.
- Keep only packages where `ReposVer > Installed` (compared with `package_version()`, so `10.0.0` correctly beats `9.0.0`).

### Testing

Added integration tests in `crates/ark/tests/integration/packages_pane.rs` driving `pkg_outdated_result()` with fabricated `old.packages()` matrices: a genuine upgrade is kept, a same-version rebuild is dropped, a mixed list keeps only real upgrades in order, versions compare numerically rather than as strings, and `NULL` / all-rebuild inputs return an empty list.

Also verified manually that `old.packages()` really does flag a same-version rebuild (via injected `instPkgs`/`available` matrices with a newer `Built` date) and that the filter drops it while keeping a genuine upgrade.

#### New Features

- N/A

#### Bug Fixes

- Fixed the packages pane reporting available updates for packages that have only been rebuilt (same version) rather than genuinely updated (posit-dev/positron#14294).

### DIFF
--- a/crates/ark/src/modules/positron/packages_pane.R
+++ b/crates/ark/src/modules/positron/packages_pane.R
@@ -192,8 +192,30 @@
 # cannot guarantee.
 #' @export
 .ps.rpc.pkg_outdated <- function() {
-    outdated <- utils::old.packages()
+    pkg_outdated_result(utils::old.packages())
+}
+
+# Turn an `old.packages()` matrix into the pane's outdated-package list. Split
+# out from the RPC so the filtering and formatting can be tested without
+# querying live repositories.
+#
+# We keep only packages whose repository version is strictly greater than the
+# installed version. old.packages() also flags packages whose repository copy
+# has the *same* version but a newer build or publication date (see the
+# `needs.install` closure inside utils::old.packages). Right after a new R
+# minor is installed, CRAN's freshly rebuilt binaries carry newer Built dates
+# than the user's carried-over packages, so hundreds of same-version packages
+# get flagged even though no newer version exists. The pane's update indicator
+# means "a newer version is available", so a same-version rebuild is not an
+# update we should surface.
+pkg_outdated_result <- function(outdated) {
     if (is.null(outdated) || nrow(outdated) == 0) {
+        return(list())
+    }
+    newer <- package_version(outdated[, "ReposVer"]) >
+        package_version(outdated[, "Installed"])
+    outdated <- outdated[newer, , drop = FALSE]
+    if (nrow(outdated) == 0) {
         return(list())
     }
     unname(Map(
--- a/crates/ark/src/modules/positron/packages_pane.R
+++ b/crates/ark/src/modules/positron/packages_pane.R
@@ -198,20 +198,14 @@
 # Turn an `old.packages()` matrix into the pane's outdated-package list. Split
 # out from the RPC so the filtering and formatting can be tested without
 # querying live repositories.
-#
-# We keep only packages whose repository version is strictly greater than the
-# installed version. old.packages() also flags packages whose repository copy
-# has the *same* version but a newer build or publication date (see the
-# `needs.install` closure inside utils::old.packages). Right after a new R
-# minor is installed, CRAN's freshly rebuilt binaries carry newer Built dates
-# than the user's carried-over packages, so hundreds of same-version packages
-# get flagged even though no newer version exists. The pane's update indicator
-# means "a newer version is available", so a same-version rebuild is not an
-# update we should surface.
 pkg_outdated_result <- function(outdated) {
     if (is.null(outdated) || nrow(outdated) == 0) {
         return(list())
     }
+    # Since R 4.6, `old.packages()` also reports equal-version packages with
+    # newer `Built` or `Published` timestamps, which flags hundreds of packages
+    # right after an R minor upgrade rebuilds CRAN's binaries. The pane promises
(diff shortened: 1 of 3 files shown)


juliasilge:
@bricestacey, I believe Jenny is OOO now and we'd like to get this fix in before we branch for 2026.08. Can you take a look and review, at least based on what your understanding is? Thanks!

juliasilge:
@lionel-, I am having a bit of trouble getting this PR reviewed. 😅 We do want to get this fix in for this milestone, so could you take a look at it?

We believe there may be further work we can do here for packages with compiled code that actually _should_ be marked outdated in some situations, but this is a first step to improve what people are seeing.

crates/ark/src/modules/positron/packages_pane.R
```diff
@@ -192,10 +192,32 @@
...
+#
+# We keep only packages whose repository version is strictly greater than the
+# installed version. old.packages() also flags packages whose repository copy
+# has the *same* version but a newer build or publication date (see the
+# `needs.install` closure inside utils::old.packages). Right after a new R
+# minor is installed, CRAN's freshly rebuilt binaries carry newer Built dates
+# than the user's carried-over packages, so hundreds of same-version packages
+# get flagged even though no newer version exists. The pane's update indicator
+# means "a newer version is available", so a same-version rebuild is not an
+# update we should surface.
```
lionel-: Does this simpler comment (to move below near the check) retain all useful information?

```r
# Since R 4.6, `old.packages()` also reports equal-version packages with newer
# `Built` or `Published` timestamps. The pane promises a newer package version,
# so exclude those entries.
```

crates/ark/tests/integration/packages_pane.rs
```diff
@@ -0,0 +1,107 @@
+//
+// packages_pane.rs
```
lionel-, line 2: I'd love an integration test that exercises the format produced by the R version on CI.

Claude thinks this is feasible. If you have time, could you take a look?

````md
A real integration test is cheap

No network, no `R CMD INSTALL`, no compilation. `installed.packages()` only needs `<lib>/<pkg>/Meta/package.rds` holding a list with `$DESCRIPTION` and `$Built$R` (`utils:::.readPkgDesc`), and `available.packages()` only needs a DCF `PACKAGES` file under `<repo>/src/contrib`. Both are a few `saveRDS()`/`cat()` calls, and the run is instant — the whole thing is much cheaper than the current six r_task processes.

```r
lib <- tempfile("lib"); repo <- tempfile("repo")
dir.create(file.path(repo, "src", "contrib"), recursive = TRUE)

install_fake <- function(pkg, version, built) {
    dir.create(file.path(lib, pkg, "Meta"), recursive = TRUE)
    saveRDS(
        list(
            DESCRIPTION = c(Package = pkg, Version = version, Built = built),
            Built = list(R = package_version(sub("^R ([0-9.]+).*", "\\1", built)))
        ),
        file.path(lib, pkg, "Meta", "package.rds")
    )
}

install_fake("rebuild", "1.5.0", "R 4.5.2; ; 2026-01-01 00:00:00 UTC; unix")
install_fake("upgrade", "1.0.0", "R 4.6.0; ; 2026-05-01 00:00:00 UTC; unix")
# PACKAGES: rebuild 1.5.0 with a newer `Built:` date, upgrade 2.0.0

options(
    repos = c(CRAN = paste0("file:///", normalizePath(repo, winslash = "/"))),
    pkgType = "source",
    # `Built` is only standard for binary repos, so ask for it explicitly and
    # the fixture stays portable instead of faking a mac.binary layout.
    available_packages_fields = "Built"
)
.libPaths(c(lib, .libPaths()))
.ps.rpc.pkg_outdated()
```
````

lionel- (approved):
LG although a real integration test would be great

crates/ark/tests/integration/packages_pane.rs
juliasilge: Done! Thanks for that idea, which works pretty much as written, with just some adjustments for older R.
 
The one thing worth a look is the version gating. On R < 4.6 `old.packages()` never flags the same-version rebuild, so "the rebuild is absent from the result" would pass vacuously on most of our matrix. The tests now assert that `old.packages()` flags it exactly when `getRversion() >= "4.6.0"`, so the fixture fails loudly if it ever stops reproducing the 4.6 behavior instead of going quiet.

Verified passing against R 4.2.1, 4.4.2, and 4.6.0 locally, and confirmed both tests fail if I relax the filter to `>=`.

PR #1362, opened 2026-07-28
Switch to `tower_lsp_server`

### LINKED ISSUE #1361: LSP transport panics ("unreachable code") when the client disconnects abruptly, killing the LSP for the session
## Summary

When the LSP client (Positron's extension host) disconnects abruptly, ark's tower-lsp transport hits `Connection reset by peer` while writing a response, then **panics** with `internal error: entered unreachable code` in `tower-lsp`'s `transport.rs:167`. This tears down the LSP for the rest of the kernel session. A recoverable client disconnect (which is routine when the extension host is recycled, e.g. on a remote/Workbench reconnect) becomes a hard, unrecoverable LSP failure. R language features (completion, hover, signature help, document symbols, folding ranges, **statement range for Cmd+Enter**, etc.) stay broken until the whole kernel is restarted.

Found while analyzing logs from https://rstudioide.zendesk.com/agent/tickets/131110. 

## Observed behavior

From a user's kernel log (Positron 2026.06.1 build 6, remote Posit Workbench session). The extension host was being recycled; the moment its LSP client connection dropped, ark logged:

```
2026-07-23T12:33:15.777016Z ERROR  failed to encode message: failed to encode response: Connection reset by peer (os error 104)
    at .../tower-lsp-ad3208de34ae5bea/49ef549/src/transport.rs:123

2026-07-23T12:33:15.779684Z ERROR  Panic! In file '.../tower-lsp-ad3208de34ae5bea/49ef549/src/transport.rs' at line 167: internal error: entered unreachable code
Backtrace:
   0: <unknown>
   ... (frames unresolved)
  14: <unknown>
    at crates/ark/src/main.rs:415
```

Immediately before and after, the main loop floods with:

```
WARN  LSP channel is closed, redirecting messages to Jupyter kernel
    at crates/ark/src/lsp/main_loop.rs (log() fallback)
```

i.e. the LSP's auxiliary channel is already dead, so all subsequent LSP log messages fall back to the kernel log. The LSP does not recover for the remainder of the session.

## Sequence of events

1. The extension host (LSP client) is torn down / reconnected. On a remote or Posit Workbench session this is routine -- the ext host is recycled on reconnect, and each teardown severs the local socket to ark's LSP.
2. ark is mid-write on a response; the socket write fails with `ECONNRESET` ("Connection reset by peer", os error 104). This is logged by tower-lsp's transport at `transport.rs:123`.
3. tower-lsp's transport then reaches `transport.rs:167`, which is `unreachable!()`, and panics.
4. The panic is caught/logged via `crates/ark/src/main.rs:415`, but the LSP transport is now gone; the main loop falls back to redirecting log messages to the kernel (`main_loop.rs`).
5. All LSP features are dead until the kernel is restarted.

## Root cause

This is a `tower-lsp` transport bug: on an abrupt peer disconnect during response encoding/writing, the transport codec ends up in a state it treats as `unreachable!()` instead of returning a clean EOF/shutdown. Ark is pinned to a `tower-lsp` fork (`tower-lsp-ad3208de34ae5bea` @ `49ef549`), so the `unreachable!()` at `transport.rs:167` is in our fork and is fixable there. A client disconnect must be handled as an ordinary end-of-stream, not a panic.

## Impact

- A routine, recoverable event (client/ext-host disconnect) is turned into an unrecoverable LSP outage.
- Especially bad on remote / Posit Workbench deployments, where the extension host reconnects/recycles periodically, so this can fire repeatedly.
- User-visible result: R language features silently stop working -- notably **statement range**, so Cmd+Enter (run selection/statement from a script) hangs or does nothing -- with no obvious recovery short of restarting the R session.

## Suggested fix

- In the pinned `tower-lsp` fork, replace the `unreachable!()` at `transport.rs:167` with graceful handling: treat a write/encode error caused by peer disconnect (`ECONNRESET`, broken pipe, EOF) as a normal transport shutdown rather than a panic. Return/propagate a clean shutdown so the LSP service loop can exit (or, ideally, allow a reconnect) without taking the process path through `unreachable!()`.
- Defensively, ensure a panic on the LSP transport thread cannot leave the kernel with a permanently dead LSP: either allow the LSP main loop to restart its transport on the next client connection, or surface a clear error so Positron can restart just the LSP.
- Add a regression test that drops the client socket mid-response and asserts the server shuts the transport down cleanly (no panic).

## Environment

- Positron 2026.06.1 (build 6), remote Posit Workbench session (Workbench 2026.06.0+242.pro13), R 4.5.1.
- ark `tower-lsp` fork: `tower-lsp-ad3208de34ae5bea` @ `49ef549`.
- Panic logged via `crates/ark/src/main.rs:415`; fallback logging at `crates/ark/src/lsp/main_loop.rs`.

## Notes

Seen alongside a separate, unrelated performance problem in the same session (ark LSP periodically stalling 30-90s in a stall-then-drain pattern). That is being investigated separately; this issue is specifically about the panic-on-disconnect turning a recoverable event into a dead LSP.

lionel-:
Closes https://github.com/posit-dev/ark/issues/1361

Test suite passed at https://github.com/posit-dev/positron/pull/15174

Because `tower_lsp` doesn't handle `exit` notifications and is unmaintained, we've pinned it to a fork that includes a patch for `exit`: https://github.com/lionel-/tower-lsp/tree/bugfix/patches. However it turns out this patch is buggy in case of unexpected disconnections.

To fix both the exit handling and the unexpected disconnection issue, this PR switches us to `tower_lsp_server`, a maintained community fork of `tower_lsp` (https://github.com/tower-lsp-community/tower-lsp-server/). It's used by Biome, which indicates reliability.

#### tower_lsp_server and the disconnection fix

The first two commits add integration tests for shutdown and disconnection. The failing test output with our current version of tower-lsp matches the panics documented in the linked issue:

```
thread 'lsp::test_lsp_survives_abrupt_disconnect' (1793302) panicked at crates/ark/tests/integration/lsp.rs:56:9:
`ark-lsp` thread panicked on abrupt disconnect: panicked at /Users/lionel/.cargo/git/checkouts/tower-lsp-ad3208de34ae5bea/49ef549/src/transport.rs:167:41:
internal error: entered unreachable code
test lsp::test_lsp_survives_abrupt_disconnect ... FAILED
```

The third commit is the mechanical switch to the community fork, and it fixes the failing test.

The third commit also copies `aether_lsp_utils` from the Air repo, because the latter uses a different incompatible `lsp_types` crate (from tower-lsp). Ideally we'd move it to posit-dev/aether and update it to use the community fork, then point both Ark and Air to it, but this can wait until later.

#### tower_lsp_server and the `Uri` type

The fourth commit streamlines the Uri/FilePath boundary. The main change from switching to `tower_lsp_server` is that LSP URLs are represented by `ls_types::Uri` which wraps https://docs.rs/fluent-uri/latest/fluent_uri. This
type has different properties than `url::Url`, which we use for Jupyter and `tower_lsp` used to use for LSP. `Uri` implements RFC 3986, which is recommended by the LSP protocol, whereas `Url` implements WHATWG. The main
differences are:

- `Url` does some normalisation, whereas `Uri` stores bytes as is.
- `Url` accepts `| [ ] ^` characters, `Uri` rejects them.

Given that `Uri` doesn't do any normalisation, it's a better representation of wire URLs which we were already tracking for sending back URLs to the frontend exactly as we got them in requests. It's a good thing that we had to switch to follow `tower_lsp_server`.

Here are the principles that guided this refactor (these already existed, but we are reinforcing them):

- Identity is governed by our normalised `FilePath`, not by any kind of URLs. For non-virtual paths, this wraps `AbsPathBuf`.
- Wire values are stored byte for byte. `Uri` now plays that role.
- `Uri` and `FilePath` are converted at the LSP boundary.
- File system queries take `AbsPathBuf` (unwrapped from `FilePath`).


crates/aether_lsp_utils/src/proto/to_proto.rs
```diff
@@ -0,0 +1,122 @@
+// Utilites for converting internal types to LSP types
```
jmcphers, line 1: Here and elsewhere: do these new files need Posit copyright headers?

jmcphers (approved):
Nice to get onto a more maintained version of this crate!

crates/aether_lsp_utils/src/proto/to_proto.rs
lionel-: I'd say no because Oak and Aether crates will soon move to their own repo, where we'll follow the Rust community convention of headerless files, which we already follow in Air.

PR #1370, opened 2026-07-30
Add source fetching settings

lionel-:
Follow up to #1369 for https://github.com/posit-dev/positron/issues/15211.

- Disable source fetching on CI by default. No need to impact the network and cause LSP flickering as new information comes in on CI.

- Add a user-facing LSP setting `oak.sourceFetching.enabled`, and a corresponding env var `OAK_SOURCE_FETCHING_ENABLED`. The env-var can be used on CI to opt back in the source fetching mechanism. The env-var always has precedence over the LSP setting, which follows ruff/ty.

- Start an Oak documentation page that describes the source fetching and cache and how to configure it.

#### New Features

- N/A

#### Bug Fixes

- N/A

### DIFF
--- a/crates/ark/src/lsp/config.rs
+++ b/crates/ark/src/lsp/config.rs
@@ -37,11 +37,11 @@ pub static GLOBAL_SETTINGS: &[Setting<LspConfig>] = &[
         },
     },
     Setting {
-        key: "oak.enableSourceFetching",
+        key: "oak.sourceFetching.enabled",
         set: |cfg, v| {
-            cfg.oak.enable_source_fetching = v
+            cfg.oak.source_fetching_enabled = v
                 .as_bool()
-                .unwrap_or_else(|| OakConfig::default().enable_source_fetching)
+                .unwrap_or_else(|| OakConfig::default().source_fetching_enabled)
         },
     },
     Setting {
@@ -64,7 +64,7 @@ pub static GLOBAL_SETTINGS: &[Setting<LspConfig>] = &[
 
 /// Source fetching is enabled by default except on CI.
 /// This env-var opts a CI job back into source fetching.
-pub(crate) const OAK_ENABLE_SOURCE_FETCHING_ENV_VAR: &str = "OAK_ENABLE_SOURCE_FETCHING";
+pub(crate) const OAK_SOURCE_FETCHING_ENABLED_ENV_VAR: &str = "OAK_SOURCE_FETCHING_ENABLED";
 
 pub struct EnvOverride<T> {
     pub name: &'static str,
@@ -74,8 +74,8 @@ pub struct EnvOverride<T> {
 /// Environment variables that override an LSP setting. Only listed here if
 /// there is a corresponding client setting that needs resolution.
 pub static ENV_OVERRIDES: &[EnvOverride<LspConfig>] = &[EnvOverride {
-    name: OAK_ENABLE_SOURCE_FETCHING_ENV_VAR,
-    set: |cfg, on| cfg.oak.enable_source_fetching = on,
+    name: OAK_SOURCE_FETCHING_ENABLED_ENV_VAR,
+    set: |cfg, on| cfg.oak.source_fetching_enabled = on,
 }];
 
 /// Resolve [`ENV_OVERRIDES`] into `config`.
@@ -159,13 +159,13 @@ pub struct OakConfig {
     /// Whether to recover R sources for package dependencies. Sources are
     /// needed for full analysis, but the LSP degrades gracefully if it doesn't
     /// have them.
-    pub enable_source_fetching: bool,
+    pub source_fetching_enabled: bool,
 }
 
 impl Default for OakConfig {
     fn default() -> Self {
         Self {
-            enable_source_fetching: true,
+            source_fetching_enabled: true,
         }
     }
 }
--- a/crates/ark/src/lsp/config.rs
+++ b/crates/ark/src/lsp/config.rs
@@ -62,6 +62,27 @@ pub static GLOBAL_SETTINGS: &[Setting<LspConfig>] = &[
     },
 ];
 
+/// Resolve global settings from the client's `initializationOptions`.
+///
+/// Keys are the dotted paths from [`GLOBAL_SETTINGS`], looked up through nested
+/// objects, so `oak.sourceFetching.enabled` reads `{oak: {sourceFetching:
+/// {enabled: ...}}}`.
+pub(crate) fn apply_initialization_options(config: &mut LspConfig, options: &Value) {
+    for setting in GLOBAL_SETTINGS {
+        if let Some(value) = nested_setting(options, setting.key) {
+            (setting.set)(config, value.clone());
+        }
+    }
+}
+
+fn nested_setting<'options>(options: &'options Value, key: &str) -> Option<&'options Value> {
+    let mut value = options;
+    for segment in key.split('.') {
+        value = value.get(segment)?;
+    }
+    Some(value)
+}
+
 /// Source fetching is enabled by default except on CI.
 /// This env-var opts a CI job back into source fetching.
 pub(crate) const OAK_SOURCE_FETCHING_ENABLED_ENV_VAR: &str = "OAK_SOURCE_FETCHING_ENABLED";
--- a/crates/ark/src/lsp/config.rs
+++ b/crates/ark/src/lsp/config.rs
@@ -37,7 +37,7 @@ pub static GLOBAL_SETTINGS: &[Setting<LspConfig>] = &[
         },
     },
     Setting {
-        key: "oak.sourceFetching.enabled",
+        key: OAK_SOURCE_FETCHING_ENABLED_SETTING,
         set: |cfg, v| {
             cfg.oak.source_fetching_enabled = v
                 .as_bool()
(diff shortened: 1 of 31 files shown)


lionel-:
Positron tests running at https://github.com/posit-dev/positron/pull/15226

lionel-:
Consider rust-analyzer's nested style: `oak.sourceFetching.enable`

crates/ark/src/lsp/config.rs
```diff
@@ -27,6 +28,14 @@ pub static GLOBAL_SETTINGS: &[Setting<LspConfig>] = &[
                 .unwrap_or_else(|| DiagnosticsConfig::default().enable)
         },
     },
+    Setting {
+        key: "oak.sourceFetching.enable",
```
juliasilge: In terms of user-facing settings, the norm is "enabled" and not "enable":

Image

Will this eventually get contributed to Positron?

doc/configuration-oak.md
```diff
@@ -0,0 +1,67 @@
...
+
+1. Environment variables.
+2. LSP settings from the client.
+3. Defaults.
+
+Environment variables read `1` and `true` as on, `0` and `false` as off, case-insensitively. Anything else, including the empty string, counts as unset and falls through to the setting.
+
+# LSP settings
+
+## `oak.enableSourceFetching`
```
juliasilge: I think this needs to be updated.

crates/ark/src/lsp/state.rs
```diff
@@ -75,10 +75,15 @@ pub(crate) struct Workspace {
...
-        Self {
+        let mut state = Self {
             db,
             ..Default::default()
-        }
+        };
+        // Resolved here too, not just in `update_config()`. This way an
+        // override holds over the window before the client's first
+        // `didChangeConfiguration`.
+        crate::lsp::config::apply_env_overrides(&mut state.config);
```
juliasilge: Resolving the env var here means `WorldState::new()` now reads the process environment, and three existing tests build their world this way while expecting the default (on). So setting `OAK_ENABLE_SOURCE_FETCHING=0` in your shell breaks the suite:

```
$ OAK_ENABLE_SOURCE_FETCHING=0 just test -p ark sources
FAIL  lsp::tests::sources::test_failed_source_is_not_retried
FAIL  lsp::tests::sources::test_source_pipeline_ingests_package_sources
FAIL  lsp::tests::sources::test_source_pipeline_ingests_real_srcref_sources
```

This is a bit of a trap, I think. The documentation advertises the variable, so a developer who exports it to save bandwidth locally then gets three unrelated-looking failures. Could those three tests pin `world.config.oak.enable_source_fetching = true` after `WorldState::new()`, the same way the new test pins `false`? That also makes each test's expectation self-evident rather than inherited from the default.

(`CI=true just test -p ark sources` passes cleanly, since these tests inject a handler directly and never reach `source_handler()`.)

crates/ark/src/lsp/config.rs
juliasilge: Also, the environment variable is `OAK_ENABLE_SOURCE_FETCHING` which _is_ user-facing and inverts the order relative to the setting. Renaming is free right now and awkward once it ships, so it seems worth it to me. (ruff and ty do keep this pairing aligned, and the PR description cites them for the precedence rule, so there may be an argument for matching them here too.)

crates/ark/src/lsp/sources.rs
```diff
@@ -164,9 +165,18 @@ impl SourceScheduler {
...
         db: &dyn Db,
+        config: &OakConfig,
         pool: &IoPool,
         events_tx: &TokioUnboundedSender<Event>,
     ) {
+        // Checked per call rather than at construction, because the client can
+        // flip this mid-session. Turning it back on picks up every package
+        // we've seen so far. Already resolved against the environment by
+        // `apply_env_overrides()`.
+        if !config.enable_source_fetching {
```
juliasilge: This gate is correct, but I think it runs too late to deliver what the setting is meant to do. With `"oak.sourceFetching.enable": false` in `settings.json`, reloading the window and opening no R file at all still produces:

```
[Info] Fetching sources for package base, 1 pending
[Info] Fetching sources for package datasets, 2 pending
[Info] Fetching sources for package grDevices, 3 pending
[Info] Fetching sources for package graphics, 4 pending
[Info] Fetching sources for package methods, 5 pending
[Info] Fetching sources for package stats, 6 pending
[Info] Fetching sources for package utils, 7 pending
```

Those are all `Priority::Base`, so they take the branch at `sources.rs:67` and call `insert_r()`, which downloads the `r-source.tar.zst` release archive on a cold cache. The setting is off and we're still hitting the network at startup, which is the specific thing it's meant to prevent, if I am understanding this correctly.

The cause is that `update_config()` has exactly one caller, `did_change_configuration` at `state_handlers.rs:397`. Nothing pulls the client's settings during `initialize` or `initialized`. But `initialize` dispatches the workspace scan right away (`state_handlers.rs:101-108`), so the first `schedule()` runs while `config.oak` is still `OakConfig::default()`, i.e. `enable_source_fetching: true`. The client's `false` arrives afterwards, if it arrives at all.

It compounds, because anything fetched in that window is marked `Finished` and then skipped forever by the `contains_key` check just below. That skip is a bare `continue` with no log line, so the package silently never appears again. I hit this while testing the mid-session toggle documented in `configuration-oak.md`. I believe tibble was fetched during startup, and flipping the setting off and back on never re-dispatched it, with nothing in the output channel explaining why. That took a while to work out from the outside.

Worth noting the env var doesn't have this problem because `WorldState::new()` resolves it eagerly; the env var is protected against it while the LSP setting isn't. That asymmetry isn't in the docs, which present the two as equivalent apart from precedence.

Would it work to call `update_config()` from `handle_initialized()`, before or alongside the initial scan dispatch, so the client's settings are in hand for the first `schedule()`? Alternatively, hold off source scheduling until the first config response lands? As it is, I don't think `false` in `settings.json` prevents the startup fetch.

juliasilge (changes requested):
Thank you so much for working on this, and the doc page is a really nice addition! I did some manual testing in Positron and I think the environment variable works exactly as advertised and the LSP setting works mid-session too (flipping it back on re-dispatches the packages Oak already knows about). What doesn't work is the setting at _startup_. Given your timing constraints, what do you think about one of these two options:

1. Ship the env var only for now. Drop the `GLOBAL_SETTINGS` entry and the matching section of `configuration-oak.md`, and we get the CI benefit right away with nothing user-facing that half-works. We can open a follow-up issue for the setting.

2. If you have time, calling `update_config()` from `handle_initialized()` may be all it takes, though I haven't tried it and there may be a reason it isn't there already.

crates/ark/src/lsp/config.rs
lionel-: oops, we've used `enable` for a long time in ark/positron-r: https://github.com/posit-dev/ark/blob/db054ebabb074e58af0b4ddcbfdb9c995415c0b3/crates/ark/src/lsp/config.rs#L23

I was going for consistency with rust-analyzer. I don't mind that much, so renamed to `enabled`. The existing variable will be deprecated once Oak gains a corresponding switch.

crates/ark/src/lsp/state.rs
lionel-: Fair point

juliasilge:
Thanks for the rename and the `initializationOptions` handling! I think this is better. The three tests I flagged do pass now with the pin.

One thing to flag before this lands is that the env var trap I mentioned has moved rather than gone. The suite still fails if a developer has the documented variable exported, in both directions, on the two new tests:

```
$ OAK_SOURCE_FETCHING_ENABLED=0 just test -p ark sources
FAIL  lsp::tests::sources::test_initialize_without_options_fetches
  left: []
 right: ["donor"]

$ OAK_SOURCE_FETCHING_ENABLED=1 just test -p ark sources
FAIL  lsp::tests::sources::test_initialization_options_gate_the_first_dispatch
```

(Baseline with the variable unset is clean, 55 passed.)

If I'm reading this right, the cause is that pinning the world is no longer sufficient once a test goes through `initialize()`. `test_initialize_without_options_fetches` builds its world with `world_with_source_fetching(db)`, but `initialize()` then calls `apply_env_overrides()` at `state_handlers.rs:103` and re-resolves the variable over the pin, so the test's setup gets flipped back. `test_initialization_options_gate_the_first_dispatch` has the mirror problem where it passes `enabled: false` in the options and the env var overrides it to on.

That's correct behaviour for the handler, so I think the fix belongs in the tests. Can we have those two control the variable themselves rather than inheriting the ambient environment, the way `test_env_var_overrides_the_setting` already does? Maybe also worth a note in `tests/utils/mod.rs` too, since `world_with_source_fetching()` reads as if it makes a test env-proof and it only does so for tests that never reach `initialize()`.

Not blocking if you don't have time, but there's still no log line when the setting disables fetching. `schedule()` returns early silently, and the one "disabled" message is the CI branch in `main_loop.rs`, which is `log::info!` to the kernel log rather than `lsp::log_info!` to the LSP output channel. So in the output channel, "correctly disabled" and "nothing was dispatched for some other reason" look identical. I had to move `~/.cache/oak` aside to tell the two apart while testing this.

lionel-:
@juliasilge That PR is quite tricky!

I've comprehensively revised the approach:
- Delay source fetching to `initialized` response. This way clients don't have to correctly configure init options, the initial setting fetch for clients that support pull-based configuration is sufficient.

- New layered abstractions for settings, to merge overrides into defaults more consistently.

- Logging (setting enabled/disabled, source fetching via network) should now work as expected.

- CI should work with either value of the setting override.

- Improved test coverage quite a bit.

juliasilge (approved):
This is working well for me now, in all the permutations of settings and such. Thank you! 🙌

PR #1384, opened 2026-07-31
Add support for `targets::tar_source()` and `sourceDir()`

lionel-:
Addresses https://github.com/posit-dev/positron/issues/14790

Source effects can now target a directory instead of a single file, which covers the two idioms for loading a folder of scripts: `targets::tar_source()` and the `sourceDir()` example from `?source`.

```r
# _targets.R
library(targets)
tar_source()                 # defaults to files = "R"
tar_target(x, my_helper())   # goto-def lands in R/my_helper.R
```

The resolver expands the directory into one target per script, in `list.files()` order. Load order within the call is respected: `code/b.R` sees `code/a.R` but not the other way around. Combined with the previous PR this also means the sourced files see each other, through the sourcing file's scope, without relying on the `R/` auto-collation of the first PR.

The PR also adds ad hoc support for `sourceDir()` with a total hack: any calls to `sourceDir()` with a resolvable string causes a `Source` effect. This is to handle the documented pattern in `?source` that people are copying around: https://github.com/search?q=sourceDir+language%3AR&type=code. Ideally we'd infer the effects from the `list.files()` and `source()` calls. One day.

#### New Features

- Added support for directory sourcing with `targets::tar_source()` and a `sourceDir()` call as documented in `?source` (https://github.com/posit-dev/positron/issues/14790).

#### Bug Fixes

- N/A

### DIFF
--- a/crates/ark/src/lsp/diagnostics.rs
+++ b/crates/ark/src/lsp/diagnostics.rs
@@ -995,7 +995,7 @@ fn handle_package_attach_call(node: Node, context: &mut DiagnosticContext) -> an
     let package_name = package_node.get_identifier_or_string_text(context.contents())?;
     let attach_pos = node.end_position();
 
-    insert_package_exports(package_name, attach_pos, context)?;
+    insert_package_exports(package_name, attach_pos, context);
 
     // Also attach packages from `Depends` field, if any
     if let Some(package_names) = context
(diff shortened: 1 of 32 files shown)


lionel-:
Edit: See https://github.com/posit-dev/ark/pull/1388#issuecomment-5170395436

PR #1386, opened 2026-08-03
Respect a lone `fig-width` or `fig-height` in plot sizing metadata

seeM:
Addresses posit-dev/positron#15259, identified in https://github.com/posit-dev/positron/pull/15234#issuecomment-5143761675.

Plot sizing from cell options previously required both `fig-width` and `fig-height`: a lone dimension was ignored and the plot fell back to `output_width_px` sizing. Either option may now be set alone, matching `quarto render` and the Python kernel; the missing dimension falls back to Quarto's default figure size (7 x 5 inches).

### Screenshots

image

image

### QA notes

In a Quarto document with inline output enabled, run a cell containing only `#| fig-width: 2` and `plot(1:10)`. The plot renders 2 inches wide and 5 inches tall. `#| fig-height: 2` alone gives 7 x 2 inches.

### Known issues

* The images are 2x larger than expected because of https://github.com/posit-dev/positron/issues/15261


juliasilge (approved):
This is working well for me!

Image

PR #1387, opened 2026-08-03
Report logical plot size in `display_data` metadata

seeM:
Addresses posit-dev/positron#15261.

## Summary

Ark renders notebook plots at `logical size × pixel ratio` physical pixels. Previously, its `display_data` and `update_display_data` messages did not report the intended logical dimensions, so frontends could display a 2× render at twice its requested width and height.

This PR adds standard Jupyter PNG metadata containing the logical `width` and `height`:

```json
{
  "image/png": {
    "width": "<logical width>",
    "height": "<logical height>"
  }
}
```

The PNG retains its higher physical resolution, while frontends—including Positron's Quarto inline-output renderer—display it at the requested logical size.

## Screenshots

**Before**

Plot displayed at twice its intended size

**After**

Plot displayed at its intended logical size

## QA

On a HiDPI display, enable inline output in a Quarto document and run:

```r
#| fig-width: 4
#| fig-height: 3
plot(1:10)
```

The plot should:

- display at 4 × 3 inches rather than 8 × 6 inches;
- remain sharp at the display's native pixel ratio.

### DIFF
--- a/crates/ark/src/plots/graphics_device.rs
+++ b/crates/ark/src/plots/graphics_device.rs
@@ -678,12 +678,10 @@ impl DeviceContext {
         let ctx = self.capture_execution_context();
         self.store_plot_context(id, &ctx);
 
-        let data = unwrap!(self.create_display_data_plot(id, &ctx), Err(error) => {
+        let (data, metadata) = unwrap!(self.create_display_data_plot(id, &ctx), Err(error) => {
             log::error!("Failed to create plot due to: {error}.");
             return;
         });
-
-        let metadata = json!({});
 
         // For `DisplayData`, the `transient` slot is a simple `Value`,
         // but we can use the `TransientValue` required by `UpdateDisplayData`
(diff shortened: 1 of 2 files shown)


crates/ark/src/plots/graphics_device.rs
```diff
@@ -844,7 +844,14 @@ impl DeviceContext {
         let mut map = serde_json::Map::new();
         map.insert("image/png".to_string(), serde_json::to_value(data)?);
 
-        Ok(serde_json::Value::Object(map))
+        let metadata = json!({
+            "image/png": {
```
lionel-: This is partly preexisting, but it now occurs to me we should not hard-code the format, and instead pick it up from the settings just above.

lionel-: e.g. with `get_mime_type()`

```rust
let mime_type = Self::get_mime_type(&settings.format);

let data = json!({ mime_type: data });

let metadata = json!({
    mime_type: {
        "width": settings.size.width,
        "height": settings.size.height,
    }
});

Ok((data, metadata))
```

lionel- (approved):
@thomasp85 might be interested in taking a look too

crates/ark/src/plots/graphics_device.rs
seeM: Good idea! Added in 8b9e742

PR #1391, opened 2026-08-05
Keep the plot origin when a render is held past source()

### LINKED ISSUE #1334: Plot origin lost on the source() / run-file path (no code_location fallback)
Running an R file with `source()` (Positron's "run file" / `r.sourceCurrentFile`) intermittently produces a plot with no origin attribution, so Positron never shows the source-file link on the plot.

Found while chasing a flaky Positron e2e test (`plot-file-attribution.test.ts › R - Plot origin shows source file after run file command`). It fails as *flaky* (passes on retry), which is the signature of a timing race rather than a hard bug.

### Flake rates

Last 14 days on `main`, 121 runs each ([see metrics](https://connect.posit.it/e2e-test-insights/?tab=test_health&branch=main&search=plot-file-attribution)):

| Test | Path | Flaky |
|---|---|---|
| R - run file command | `source()` | **21%** |
| R - line execution | line-by-line | 1% |
| Python - line execution | line-by-line | 0% |
| Python - run file command | `python.execInConsole` | 5% |

The R `source()` path is 21% flaky vs. 1% for R line-execution. The line-execution path attaches a `code_location` that the `source()` path lacks, so it isn't exposed to this race.

### What the R Kernel log shows

The `source()` execute request arrives with `positron: None` (no `code_location`):

```
ExecuteRequest { code: "source(\"/tmp/.../plot-attribution-test.R\")", ... positron: None }
```

The line-by-line variant carries `code_location: Some(...)`. Both plots then go through the same deferred path: `hook_holdflush` -> `process_changes` -> `store_plot_context` -> `take_pending_origin`.

### Why the `source()` path is exposed

`capture_plot_origin` checks the source-context stack first, then falls back to `ctx.code_location`. Line execution always has `code_location`, so even if the stack timing is off, origin still resolves. The `source()` path has no such fallback: origin depends entirely on the source-context stack (pushed/popped by the `source()` hook in `hooks_source.R` via `defer()`) being populated at capture time.

When the eager `pending_origin` snapshot in `hook_mode` is skipped (it's gated on `!old_has_changes && is_drawing`), `take_pending_origin` re-derives origin at `process_changes` time. If `defer()` has already popped the stack by then, origin comes back `None` for good, with no signal to the frontend.

### Expected vs. actual

- Expected: a plot from `source("file.R")` is attributed to `file.R` every time.
- Actual: attribution is empty ~21% of the time on the `source()` path.

### Reproduction

No reliable one-shot repro - it's a timing race, reproducing intermittently under load.

Manual (occasional):

1. Start a fresh R console in Positron.
2. Open a `.R` file containing just `plot(1:10)`.
3. Run the whole file ("Source" / `r.sourceCurrentFile`).
4. Check the Plots pane. Usually the origin link appears; occasionally it's missing.

Via the Positron e2e test: `tests/plots/plot-file-attribution.test.ts` › "R - Plot origin shows source file after run file command". Fails ~21% in CI (Linux x86_64 electron), which runs parallel workers under load. 

Existing ark coverage: `test_plot_source_context_stacking` and `test_plot_from_source_dynamic` in `crates/ark/tests/integration/plots.rs` exercise `source()` + origin and pass deterministically. The `DummyArkFrontend` harness is single-threaded, so the `defer()` pop never races the deferred `process_changes` - which is why the existing suite doesn't catch this.

### Fix direction (per Claude)

Make source-path capture independent of the pop timing, e.g. capture origin eagerly when the source context is pushed, hold the pop until pending plots' `process_changes` runs, or drop the `!old_has_changes` gate while a source context is active.

### Notes

- I couldn't confirm the exact stack state from logs alone - the push/pop and origin-capture functions have no trace logging, so the empty-stack moment is inferred from the surrounding sequence. Trace logging there (or a repro under load) would pin down the precise sub-mechanism.
- The Python run-file case is also elevated (5% flaky vs. 0% for Python line-execution). Python doesn't use `source()`, so this is likely a separate, milder origin-attribution gap on the Python side rather than the same bug - worth a look independently.

midleman:
Fixes #1334.

### Summary

A plot created inside `source()` could lose its origin, leaving the Plots pane with no source file and no way to recover it. 

- `hook_mode()` only captured the origin on the `has_changes` false->true edge, so a new page starting while earlier changes were pending never re-captured after `new_positron_page()` cleared the snapshot
- `graphics_on_did_execute_request()` cleared the snapshot at the end of every request, so a render held into a later request lost it
- Now captures whenever no snapshot is pending, and drops the end-of-request clear as redundant with the new-page clear
- Adds `test_plot_origin_survives_hold_across_source`, which fails on `main` inside the assertion and passes with the fix

### E2E Triage Diagnosis

<details>
<summary>🟢 <b>High confidence</b> -- Plot origin is lost on the source() path when the origin snapshot is dropped before process_changes consumes it, leaving the fallback to run after source() popped the context stack</summary>

- **Test:** [Plot File Attribution > R - Plot origin shows source file after run file command](https://connect.posit.it/e2e-test-insights/?tab=test_health&repo=positron&test=Plot%20File%20Attribution%20%3E%20R%20-%20Plot%20origin%20shows%20source%20file%20after%20run%20file%20command%7C%7C%7Ctest%2Fe2e%2Ftests%2Fplots%2Fplot-file-attribution.test.ts)
- **Targeted failure:** Error: expect(locator).toBeVisible() failed
- **Signal:** Observed via file probes in a RED ark integration test: push_source_context -> EAGER CAPTURE succeeds -> new_positron_page clear wipes it -> hook_mode skips re-capture (old_has_changes=true) -> end-of-request clear in graphics_on_did_execute_request drops it again -> pop_source_context -> take_pending_origin FALLBACK -> None. The end-of-request clear is the decisive path for a render held across requests. Plot renders and the metadata RPC succeeds, but .plot-origin-file never enters the DOM. Local e2e did NOT reproduce (10/10 green on macOS, origin consumed inside the same request), so the CI route is predicted, not observed.
- **Frequency:** 32/137 runs (23.4%) on main, ubuntu/electron
- **Hypothesis:** race

</details>

### DIFF
--- a/crates/ark/src/console/console_graphics.rs
+++ b/crates/ark/src/console/console_graphics.rs
@@ -39,6 +39,5 @@ impl Console {
         let dc = Rc::clone(self.device_context());
         dc.process_changes(self);
         dc.clear_execution_context();
-        dc.clear_pending_origin();
     }
 }
--- a/crates/ark/src/plots/graphics_device.rs
+++ b/crates/ark/src/plots/graphics_device.rs
@@ -277,7 +277,7 @@ impl DeviceContext {
     }
 
     /// Clear any unconsumed pending origin.
-    pub(crate) fn clear_pending_origin(&self) {
+    fn clear_pending_origin(&self) {
         self.pending_origin.replace(None);
     }
 
@@ -338,10 +338,13 @@ impl DeviceContext {
         let old_has_changes = self.has_changes.get();
         self.has_changes.replace(old_has_changes || is_drawing);
 
-        // Eagerly capture the plot origin when drawing first starts for this
-        // change set. The source context stack may be popped before
-        // `process_changes()` runs, so we snapshot it now while it's available.
-        if !old_has_changes && is_drawing {
+        // Eagerly capture the plot origin while the source context stack is still
+        // available, since `process_changes()` may not run until after `source()`
+        // popped it. Capture whenever we have no snapshot rather than only on the
+        // `has_changes` false->true edge, which never comes back around for a new
+        // page started while earlier changes are still pending (e.g. `dev.hold()`).
+        let needs_origin = self.pending_origin.borrow().is_none();
+        if is_drawing && needs_origin {
             let ctx = self.capture_execution_context();
             let origin = self.capture_plot_origin(&ctx);
             self.set_pending_origin(origin);
--- a/crates/ark/src/plots/graphics_device.rs
+++ b/crates/ark/src/plots/graphics_device.rs
@@ -187,10 +187,10 @@ pub(crate) struct DeviceContext {
     /// provides the file attribution even though the execute_request came from the console.
     source_context_stack: RefCell<Vec<String>>,
 
-    /// The plot origin captured eagerly when drawing starts (i.e. when `has_changes`
-    /// transitions from false to true). This is necessary because the source context
-    /// stack may be popped before `process_changes()` runs (e.g. `source()` completes
-    /// before the execute request finishes), so we snapshot the origin at drawing time.
+    /// The plot origin for the current page, captured eagerly whenever drawing
+    /// starts and no snapshot is pending. This is necessary because the source
+    /// context stack may be popped before `process_changes()` runs (e.g. `source()`
+    /// completes before the execute request finishes).
     pending_origin: RefCell<Option<Option<PlotOrigin>>>,
 }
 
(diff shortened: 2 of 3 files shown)


lionel- (approved):
I've pushed an update to make sure a stale plot origin isn't mistakenly applied to a new plot, with a test.

However AFAIU I don't think this will fix https://github.com/posit-dev/ark/issues/1334. All the plot hooks and callbacks are called deterministically on a single thread, and the `dev.hold()` in `plot()` (the E2E test calls `plot()`) is flushed on exit: https://github.com/r-devel/r-svn/blob/0017c9e54f90bdc4a10dd1f41a7afdf7cac8ac0e/src/library/graphics/R/plot.R#L79. 

Still a good fix to have, even though it only fixes an unlikely case.

PR #1392, opened 2026-08-19
Render intrinsic size plots at correct dpi

jmcphers:
## Intent

Fixes plots rendered at intrinsic size looking pixelated and saving at low resolution when a Quarto figure size (`fig-width`/`fig-height`) is in play. 

Addresses posit-dev/positron#15026.

## Approach

A Quarto figure size is a physical size in inches. When the plot pane asks for the plot at its intrinsic size, we were converting those inches to pixels at the screen's base DPI (72 on Linux, 96 on macOS) and only scaling further by the display's pixel ratio. On an ordinary (non-HiDPI) display that left a 4x2 inch plot at just 288x144 pixels, so it looked blurry when the pane stretched it to fit, and it saved at that same low resolution. Quarto renders the same plot at 768x384.

The pixel ratio is really two things rolled into one: how dense the display is, and how sharp we render. For a physical-inch plot that the pane will scale up, tying sharpness to the display alone isn't enough.

So when rendering an inch-based intrinsic size, we now raise the effective pixel ratio to reach a target resolution of 192 DPI, matching Quarto's default (its 96 DPI figures at 2x retina). This is a floor: displays that already exceed it are left alone, and pixel-based intrinsic sizes (such as Matplotlib's) keep their own resolution. As a result a 4x2 inch plot now renders at least 768x384 on every OS.

image

A small companion change rounds device dimensions to whole pixels, since the new floor can produce a fractional pixel ratio.

## Testing

Adds an integration test that creates a Quarto-sized plot and sends the same intrinsic render request the plot pane sends when "intrinsic" is selected, then checks the resulting PNG is 768x384. It fails before this change and passes after.

#### New Features

- N/A

#### Bug Fixes

- Fixes plots rendered at intrinsic size looking pixelated and saving at low resolution when a Quarto figure size (`fig-width`/`fig-height`) is in play (#15026).

### DIFF
--- a/crates/ark/src/modules/positron/graphics.R
+++ b/crates/ark/src/modules/positron/graphics.R
@@ -376,10 +376,11 @@ finalize_device_arguments <- function(format, width, height, pixel_ratio) {
         # values are upscaled by `pixel_ratio`.
         #
         # `res` is nominal resolution specified in pixels-per-inch (ppi).
+        # Round to whole pixels, since `pixel_ratio` can be fractional.
         return(list(
             res = default_resolution_in_pixels_per_inch() * pixel_ratio,
-            width = width * pixel_ratio,
-            height = height * pixel_ratio
+            width = round(width * pixel_ratio),
+            height = round(height * pixel_ratio)
         ))
     }
 
--- a/crates/ark/src/plots/graphics_device.rs
+++ b/crates/ark/src/plots/graphics_device.rs
@@ -481,6 +481,7 @@ impl DeviceContext {
             PlotBackendRequest::Render(plot_meta) => {
                 log::trace!("PlotBackendRequest::Render");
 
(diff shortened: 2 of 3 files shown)


thomasp85 (approved):
This generally LGTM

One question is whether a user setting a `dpi` option explicitly should have this honoured, even if it's below the target resolution? Since dpi drives how absolute sizing (font-size, line width, point size) is relative to relative sizes, I expect a user setting this value wants the output to honor it so that they can expect the same relative look as they get when rendering the markdown

jmcphers:
> One question is whether a user setting a dpi option explicitly should have this honoured

Seems reasonable! Can you open an issue for that and name the specific DPI option(s) you were thinking of? (not sure if you're thinking R level, Ark level, Quarto level, etc.)

PR #1394, opened 2026-08-30
Migrate diagnostics setting to `positron.r.diagnostics.enabled`

juliasilge:
- Addresses posit-dev/positron#13354
- Related to https://github.com/posit-dev/positron/pull/15814

This PR migrate the R diagnostics setting from `positron.r.diagnostics.enable` to `positron.r.diagnostics.enabled`. We'll continue reading and registering the legacy setting for a good long while so existing configurations can still disable diagnostics. This applies the compatibility behavior to both initialization options and live `workspace/configuration` updates.

## Testing

- Added coverage for the new and legacy settings in initialization options.
- Added coverage for configuration updates where the legacy setting disables the new default.

#### New Features

- N/A

#### Bug Fixes

- Updated the R diagnostics setting to use `positron.r.diagnostics.enabled` while continuing to honor the legacy setting.

### DIFF
--- a/crates/ark/src/lsp/config.rs
+++ b/crates/ark/src/lsp/config.rs
@@ -61,7 +61,7 @@ global_settings! {
     OAK_DIAGNOSTICS_EXPERIMENTAL_ENABLED_SETTING
         => diagnostics_experimental: diagnostics.experimental,
     OAK_SOURCE_FETCHING_ENABLED_SETTING => source_fetching_enabled: oak.source_fetching_enabled,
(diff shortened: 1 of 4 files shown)


thomasp85 (approved):
Generally LGTM - at the risk of asking for premature optimisation, might it be better to create a general solution to renaming config entries rather than litter legacy fallbacks around the resolve path that we then need to gradually deprecate? Not a blocker but more an idea to think about

juliasilge:
That's a great idea for when we've got a few more settings like that! Managing a bunch of migration manually will not be sustainable in the long run.

PR #1400, opened 2026-09-02
Strip carriage returns when splitting lines

### LINKED ISSUE #1399: `strings::lines` never strips carriage returns, so CRLF input breaks input boundaries
`strings::lines` in `crates/ark/src/strings.rs` is meant to split text into lines and drop a trailing `\r` so CRLF input is handled. The `\r` is never stripped: the guard that gates it can never be true, so every line of CRLF input keeps its carriage return.

```rust
pub fn lines(text: &str) -> impl DoubleEndedIterator<Item = &str> {
    text.split('\n').map(|line| {
        let Some(line) = line.strip_suffix('\n') else {
            return line;
        };
        let Some(line) = line.strip_suffix('\r') else {
            return line;
        };
        line
    })
}
```

After `split('\n')`, no piece ends in `\n`, so `strip_suffix('\n')` always returns `None` and the closure returns early. The `\r` strip below it is unreachable.

### Impact

`input_boundaries` (behind the `positron/inputBoundaries` LSP request) feeds these lines to R's parser as a character vector. A carriage return that lands outside a string literal is an invalid token to the parser, so complete R code is misclassified. In the worst case a multi-line quoted string followed by another line is reported as incomplete, which is what caused posit-dev/positron#15734 (the code was pasted into the Console but never executed on Windows/CRLF files).

### Reproduction

Call `input_boundaries` on the same code with CRLF endings:

```
text1 <- 'First line \r\n second line'\r\ntext2 <- ''
```

Expected: two complete boundaries (one per statement), the same result as the LF version. Actual: no complete boundaries; the input is classified incomplete/invalid because the retained `\r` characters break the parse.

### Fix

Strip the trailing `\r` unconditionally after splitting on `\n`, e.g. drop the `strip_suffix('\n')` guard and just `strip_suffix('\r')` (or split with something that already handles CRLF). The existing test at `strings.rs` only exercises LF input; add a CRLF case so this can't regress.

### Notes

Positron also normalizes line endings before it calls this request now (posit-dev/positron#15734), so the immediate user-facing bug is fixed there. This is still worth fixing in Ark: any other caller of `strings::lines`, or a direct `positron/inputBoundaries` request with CRLF text, hits the same mis-parse.

jmcphers:
`strings::lines` is supposed to split text into lines and drop a trailing carriage return so CRLF input is handled, but the carriage return was never stripped. The code first checked for a trailing newline and returned early when it wasn't found. Since the function splits on newlines, no piece ever ends in a newline, so that check always returned early and the carriage return stripping below it was never reached.

The effect shows up in input boundary detection. Those lines are handed to R's parser as a character vector, and a carriage return outside a string literal is an invalid token. Complete R code then gets misclassified, and in some cases a multi-line quoted string followed by another line is reported as incomplete. On Windows with CRLF files this meant pasted code landed in the Console but never ran.

The fix drops the dead newline check and strips the carriage return directly. A new test covers CRLF input so this does not regress.

Positron already normalizes line endings before making this request, so the user-facing bug is handled there, but any other caller or a direct request with CRLF text would still hit the same mis-parse.

Fixes #1399.

### DIFF
--- a/crates/ark/src/strings.rs
+++ b/crates/ark/src/strings.rs
@@ -12,15 +12,8 @@
 /// Returns a `DoubleEndedIterator`, which is the same as the
 /// one returned by `split()` in this particular case.
 pub fn lines(text: &str) -> impl DoubleEndedIterator<Item = &str> {
-    text.split('\n').map(|line| {
-        let Some(line) = line.strip_suffix('\n') else {
-            return line;
-        };
-        let Some(line) = line.strip_suffix('\r') else {
-            return line;
-        };
-        line
-    })
+    text.split('\n')
+        .map(|line| line.strip_suffix('\r').unwrap_or(line))
 }
 
 #[cfg(test)]
@@ -31,5 +24,11 @@ mod tests {
     fn test_lines() {
         let lines: Vec<&str> = lines("foo\n\n\nbar\n\n").collect();
         assert_eq!(lines, vec!["foo", "", "", "bar", "", ""])
+    }
+
+    #[test]
+    fn test_lines_crlf() {
+        let lines: Vec<&str> = lines("foo\r\n\r\nbar\r\n").collect();
+        assert_eq!(lines, vec!["foo", "", "bar", ""])
     }
 }


thomasp85 (approved):
LGTM

PR #1403, opened 2026-09-08
Follow-up for attached packages fix

lionel-:
Follow-up for #1393:

1. Extend test coverage for cycles
2. Mention file loader in user-surfaced diagnostics
3. Remove the implicit `R/` folder collation load. This addresses https://github.com/posit-dev/positron/issues/15665.

I decided to go for 3 because of two undesirable consequences when the implicit collation conflicts with an explicit `source()` call written by the user:

- The recovery handler reports empty results and makes language analysis incorrect.
- When experimental diagnostics are enabled, the user gets an incorrect and unactionable lint:

Screenshot 2026-09-08 at 14 39 59

I think that pretty much kills the approach. Potentially we could disable the implicit folder loader for files that have an explicit source, but that behaviour would we hard to understand and predict for users. So I think we're better off modelling explicit source effects, like we did with the Shiny and testthat file loaders, or with `targets::tar_source()`.

Here is an example of the improved lint when an explicit loader conflicts with a user Source effect. Compared to the problematic lint above, this one is both correct and actionable:

Screenshot 2026-09-08 at 14 49 59

### Positron Release Notes

Let me know if you'd like a release note item here. It seemed to me mentioning this change would more likely confuse users.

#### New Features

- N/A

#### Bug Fixes

- N/A


crates/oak_db/src/diagnostic.rs
```diff
@@ -199,14 +208,27 @@ fn lower_uninstalled_package(package: &str, range: TextRange) -> Diagnostic {
...
-/// Every file in the cycle gets its own diagnostic.
-fn lower_source_cycle() -> Diagnostic {
+/// Anchor at the file start because cycle records carry no source range.
+///
+/// Report every participant. Recovery rebuilds without cross-file resolution,
+/// so we can't identify the exact source call or effect that formed the cycle.
+fn lower_source_cycle(loader: Option<LoaderInfo>) -> Diagnostic {
+    let cause = match loader {
+        Some(LoaderInfo { name, loads }) => {
+            format!("{name} already loads {loads}, so a `source()` call between them is redundant.")
```
juliasilge, line 218: This sentence tells the reader to remove a `source()` call that a Shiny app actually needs to work, right? An alternative would be to just state the fact and stop:

```rust
format!("{name} already loads {loads}. A `source()` call into a file that the loader loads later creates this cycle.")
```

Let me know if I am misunderstanding!

juliasilge: Also related to the message here, `loader(db, file)` reads the diagnosed file only and IIUC nothing makes sure that the other files in the cycle have the same loader. The message still makes a claim about the relationship between them: "a `source()` call between them is redundant".

`SemanticDiagnostic::SourceCycle` is a unit variant, so the participants are not available here. A gate on the shared loader would be pretty hard and needs the participant list in the variant. I think the cheaper, better option is to keep this sentence more descriptive.

juliasilge (approved):
This is working well for me in various workspace configurations!

Image

I have one requested change before you merge, to update one of the user facing messages; I think it will be wrong/confusing to folks as is.

crates/oak_db/src/diagnostic.rs
lionel-: That makes sense! I made it:

```
This Shiny app already loads its `global.R` and `R/` files through `shiny::loadSupport()`.
A `source()` call into a file that the loader also loads creates this cycle.
```

PR #1404, opened 2026-09-10
Support bundled ipykernel for reticulate sessions

t-kalinowski:
Python (reticulate) sessions can fail to start when the project's Python environment lacks `ipykernel`, even when the user's code doesn't depend on it. Users see an error about `positron_ipykernel`, as reported in [this comment](https://github.com/posit-dev/positron/issues/6675#issuecomment-5604984461).

Positron already ships `ipykernel`. This adds the Ark support needed to make that bundled copy available to `reticulate`. Ark reports which Python version and architecture are running inside R, then accepts the locations of the matching packages before starting the kernel.

A companion Positron PR https://github.com/posit-dev/positron/pull/16045 supplies those locations. Existing callers can omit them and continue using packages installed in their Python environment.


crates/ark/src/modules/positron/reticulate.R
```diff
@@ -106,6 +114,12 @@
     # and passing the communication files that Positron Jupyter's Adapter sets up.
     tryCatch(
         {
+            # PYTHONPATH cannot change an already initialized interpreter. Apply
+            # the launcher's paths directly, preserving their precedence.
+            sys <- reticulate::import("sys", convert = FALSE)
```
lionel-, line 119: reticulate is wild to watch in action

lionel- (approved):

PR #1405, opened 2026-09-11
Consistently catch LSP panics

lionel-:
This PR revisits how LSP panics are caught to:

- Catch panics produced in notification handlers and other parts of the LSP service. Previously only request handlers were properly guarded. Addresses https://github.com/posit-dev/positron/issues/15666

- Catch panics on the I/O and analysis task pools without aborting the process. Addresses https://github.com/posit-dev/positron/issues/15667

Special care is taken for:
- Salsa cancellation panics, which should not be reported as crash but simply advance the LSP loop or close the background task.
- Panics and cancellations across an `r_task()`. This is mainly defensive, but `r_task()` is used in legacy parts of the LSP so we better be safe.

Unlike panics in request handlers, panics in notification handlers are unrecoverable. See these screencasts, the first causes a notification panic and shuts down the LSP in an orderly fashion after notifying the user:

https://github.com/user-attachments/assets/3f11768c-1c0d-4bdc-a310-6ae316af22ba

The second is a request panic that only notifies the user and keeps the LSP in working order:

https://github.com/user-attachments/assets/93e90374-c0d2-4900-8021-99334ad6d984

To avoid spamming the user with notifications, panics are only reported once per handler. They keep being logged in the background so user reports are complete.

When a panic occurs in a background task (e.g. diagnostics), a notification is sent once. Further panics are logged only.

https://github.com/user-attachments/assets/a1695e2b-b261-4b16-8f10-ae48f3227142

#### New Features

- N/A

#### Bug Fixes

- N/A

### DIFF
--- a/crates/ark/src/console.rs
+++ b/crates/ark/src/console.rs
@@ -126,7 +126,6 @@ pub(crate) use console_debug::FrameSource;
 use console_error::stack_overflow_occurred;
 use console_filter::strip_step_lines;
 use console_filter::ConsoleFilter;
-pub use console_repl::catching_panics;
 pub(crate) use console_repl::console_inputs;
 pub(crate) use console_repl::r_busy;
 pub(crate) use console_repl::r_interrupt_events;
--- a/crates/ark/src/console/console_repl.rs
+++ b/crates/ark/src/console/console_repl.rs
@@ -731,7 +731,10 @@ impl Console {
     pub fn with<T>(f: impl FnOnce(&Console) -> anyhow::Result<T>) -> anyhow::Result<T> {
         match panic::catch_unwind(Recovery::ReleaseOnly, || f(Console::get())) {
             Ok(result) => result,
-            Err(msg) => Err(anyhow!("Panic in Console callback: {msg}")),
+            Err(payload) => {
+                let message = panic::message(&payload);
+                Err(anyhow!("Panic in Console callback: {message}"))
+            },
         }
     }
 
--- a/crates/ark/src/console/console_repl.rs
+++ b/crates/ark/src/console/console_repl.rs
@@ -13,12 +13,13 @@
 use std::path::Path;
 use std::rc::Rc;
 
-use stdext::panic_message;
 use stdext::DebugRefCell;
 
 use super::*;
 use crate::dap::dap_notebook;
 use crate::data_explorer::r_data_explorer::POSITRON_DATA_EXPLORER_MIME;
+use crate::panic;
+use crate::panic::Recovery;
 use crate::r_task::QueuedRTask;
 use crate::r_task::RTask;
 use crate::r_task::TryIdleTask;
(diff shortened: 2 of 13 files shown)


juliasilge:
I want to make sure I understand what's going on here. The old code caught unconditionally, right? A panic with no recovering caller was caught on the R thread, ferried back, and re-raised on the calling thread? The hook then saw no boundary and aborted. The new code aborts at the original panic site instead. Both paths abort, so is there a case where the difference is visible?

I ask because the condition removes the guard that the old comment called out: "a Rust unwind crossing those C frames is UB". Can we keep `catch_unwind` unconditional and decide whether to abort at the re-raise site? That would keep the guarantee from the old comment, and AFACT it gives the same observable behavior.

lionel-:
@juliasilge Assuming this is about `r_task()`, before I implemented this catch site in https://github.com/posit-dev/ark/commit/4de75ff79923b9ebd83596a303aa2361925887a1, we were not catching anything. In that case the panic hook runs its full course with specific dispositions for the R thread (backtrace capture) and _aborts the process_. There is no unwinding or crossing of C frames. The catching and rethrow was specifically implemented for Salsa cancellations.

I'd say the new behaviour of checking if there is a recovery boundary is marginally better because of the immediate process exit in case of panic. I also like that this unifies `r_task()` and `Console::with()` recovery mechanisms.

Are you concerned about other kinds of `resume_unwind()` throwing (same mechanism as Salsa cancellations) which does not invoke the panic hook? I'd says those are UB (from our own policy standpoint, not the language) to throw on the R thread. The Salsa situation is a temporary stopgap. If we wanted to do better here, we'd have to consider other boundaries like `try_catch()` and that seems daunting. For instance I've stayed clear of Rust stack unwinding in https://github.com/posit-dev/ark/pull/1309 for this reason.

lionel-:
Worth noting that in parallel to that commit for catching salsa cancellations we've started catching LSP panics. And since the legacy LSP is the main consumer of r_task (now that comms run directly on the R threads) it should be unexpected to go through that immediately aborting path, since we are now consistently installing recovery boundaries.

Also since resume_unwind doesn't reinvoke the panic hook, re throwing panics unconditionally would result in a greater risk of a thread silently dying. Although we've been heading in the opposite direction of keeping the process alive in case of panics, we should arguably only recover in controlled situations rather than let the process in a partially working state.

juliasilge (approved):
Ah, thank you! I was not thinking about this quite correctly. I overlooked that the panic hook runs before unwinding, so an ordinary panic in the old code would abort on the R thread before the unconditional `catch_unwind()` could ferry it back.

The conditional catch makes sense given that distinction, and I see why rethrowing without a known recovery boundary risks silently losing the calling thread. Do you think it may be worth documenting the remaining invariant? IIUC, it's that an `r_task()` without a recovering caller must not originate a hookless unwind such as `resume_unwind()`, since that could cross the `r_sandbox()` C frames.

But this looks good to me!

thomasp85:
Not going to contest the review by Julia. One observation though which may be due to my lack of understanding of the life cycle.

LSP_HAS_CRASHED/BACKGROUND_PANIC_REPORTED is set once and never reset, so an LSP restart (e.g. reloading the window) will inherit the crashed state. You have to restart ark to recover. If this is intended or not a concern then all good

lionel-:
@thomasp85 I thought of both of these flags as anti-spam measures:

- The global crash flag would prevent the client from reconnecting repeatedly
- The background panic report flag would prevent a repeatedly failing handler from spamming the user

However after taking another look, I think both of these were a bit off.

For the global flag, it turns out the client doesn't repeatedly reconnect. I think because it only sees a disconnection, not a process crash. Instead, what you see when switching R sessions (which causes the LSPs to cycle), is repeated crash notifications:


https://github.com/user-attachments/assets/c4dd6a2e-d165-4684-9433-8dddea6ccba9



So it's the worst of both worlds. We don't guard against repeated reconnections, and we prevent the LSP from restarting even though the cause of the crash might not be a repeating event.

I've changed that flag to only disable LSP features while the LSP is tearing down. So if the LSP is repeatedly crashing down, you'll still see repeated notifications as the LSPs cycle, but if the crash was transient, they will work as expected. Here is what you now see when the crash is _not_ transient:

https://github.com/user-attachments/assets/c795eb01-a16d-47e2-abf3-581553e7d725

Regarding the background panic flag, it seems fine to reset it when the LSP reconnects. That would already prevent most of the spamming. Now done:

https://github.com/user-attachments/assets/6a3f777f-6e24-4a2c-949f-311c5c96c92f

And a nice outcome is that now both of these flags are now local state of an LSP session, not global flags.

Thanks for pointing this out!

PR #1406, opened 2026-09-11
Enforce panic catching in the Oak LSP

lionel-:
Follow-up to #1405.

This PR enforces good panic catching conventions for Salsa accesses.

- All Salsa DB access now go through dedicated accessors that check whether a "panic boundary" (i.e. a catch site) is active on the stack. This ensures that new code/handlers in the Oak LSP are properly guarded against panics.

- A clippy rule disallows bare panic catching. Instead the codebase must now use `crate::panic::catch_unwind()` which declares a panic boundary.

- Add boundaries around Tokio tasks and processes.

#### New Features

- N/A

#### Bug Fixes

- N/A


crates/ark/src/debug.rs
```diff
@@ -164,6 +164,7 @@ pub fn capture_console_output(cb: impl FnOnce()) -> *const ffi::c_char {
 
     // We protect from panics to correctly restore `captured_output`'s state.
     // The panic is resumed right after.
+    #[expect(clippy::disallowed_methods)]
```
thomasp85, line 167: Maybe a comment on this to clarify why we allow a raw catch_unwind here with no boundary? (if I understand the new architecture correctly)

thomasp85 (approved):
Overall LGTM. Do we have any way of ensuring that future tokyo processes are correctly handled so errors in subprocesses can't bring the whole thing down?

PR #1408, opened 2026-09-12
Pass through console document `id` from rstudioapi editor context

juliasilge:
- Addresses posit-dev/positron#15571
- Related to https://github.com/posit-dev/positron/pull/16063

This set of PRs makes `rstudioapi` context aware of the console. The two PRs can merge in either order, because R reads the raw JSON reply instead of the generated struct.

This change adds an optional `id` field to the editor context from `.rs.api.getSourceEditorContext()`. `rstudioapi::getActiveDocumentContext()` calls this function, so it gets the new field too. When the context is the console, the frontend sets `id` to `"#console"`. Editor contexts do not yet include this field.

#### New Features

- `rstudioapi::getActiveDocumentContext()` and `getSourceEditorContext()` now return a document `id`. This is `"#console"` when the context is the console (posit-dev/positron#15571).

#### Bug Fixes

- N/A

### DIFF
--- a/crates/amalthea/src/comm/ui_comm.rs
+++ b/crates/amalthea/src/comm/ui_comm.rs
@@ -435,6 +435,7 @@ pub enum UiBackendRequest {
  */
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
 #[serde(tag = "method", content = "result")]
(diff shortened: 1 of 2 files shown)


jmcphers (approved):
LGTM!

PR #1416, opened 2026-09-21
pin x86 Homebrew installer to maintain support

isabelizimm:
Homebrew/install's HEAD now refuses to run under Rosetta/x86_64 emulation on Apple Silicon, breaking the x86 cross-compiling step for macOS x64 releases. Pin to the last commit before Intel/Rosetta support was dropped.

Related to failure in: https://github.com/posit-dev/positron-builds/actions/runs/35573145309/job/106249465641#step:5:11

#### New Features

- N/A

#### Bug Fixes

- N/A

PR #1417, opened 2026-09-21
ci: drop unused x86 Homebrew install from macOS release build

testlabauto:
The x64 cross-compile step in `release-macos.yml` installed an x86 Homebrew into `/usr/local` and put its `bin` first on `PATH` so pkg-config could find an x86 libsodium for zeromq. Neither is actually used: `zmq-sys` builds libzmq from vendored sources via `zeromq-src` and calls `with_libsodium(None)`, so no libsodium or pkg-config lookup happens and nothing under `/usr/local` is consulted during the build. The last passing x64 build log (July) confirms no `/usr/local` usage after the install step.

Homebrew 7.0 (released 2026-09-13) made the installer refuse to run under Rosetta on Apple Silicon, which turned this no-op step into a hard failure. This removes the step instead of pinning an old Homebrew (alternative to #1416).

Testing: `release-macos.yml` is not exercised by PR CI, so it was run via `workflow_dispatch` on this branch; link in comments.

Note: the Positron daily build fails in the same way from a copy of this step in `posit-dev/positron-ark`'s `build-ark-macos.yml`, which needs the same change (see https://github.com/posit-dev/positron-builds/actions/runs/35573145309).

#### New Features

- N/A

#### Bug Fixes

- N/A

🤖 Generated with [Claude Code](https://claude.com/claude-code)


testlabauto:
Test run of `release-macos.yml` on this branch via `workflow_dispatch`: https://github.com/posit-dev/ark/actions/runs/35613337684

testlabauto:
Test run passed: all four builds (arm64/x64 × release/debug) and both universal binaries succeeded without the Homebrew step. Verified the downloaded x64 artifact is `Mach-O 64-bit executable x86_64` and the universal artifact contains both `x86_64` and `arm64` slices.

https://github.com/posit-dev/ark/actions/runs/35613337684

jmcphers (approved):

